PR #71, opened 2020-12-22
Interactions: brush

### LINKED ISSUE #5: Brushing.
Brush a rectangular region to select data.

Fil:
Fixes #5.

TODO

- [x] Update the plot’s value on interaction, and emit an *input* event
- [x] Set the value on the figure when the svg is wrapped (#717)
- [x] Support faceting (either with unioned or exclusive brushes)  (#717)
- [x] Allow brush styling (stroke, strokeWidth, strokeOpacity, fill, fillOpacity, _etc._)?
- [x] Decide what happens if multiple marks expose selections: the last one wins (mostly).
- [x] ~~Disable the standard mark channels (*e.g.*, href, title)?~~ No, just don’t do that.
- [x] Set the plot’s initial value when there’s no initial selection
- [x] Improve band scales (select when the band intersects the brush)
- [x] ~~Support scale insets (use scale ranges rather than chart dimensions)?~~ No, it’s fine as is.
- [x] A more explicit cue to indicate that a mark exposes a selection (currently *node*.selection)
- [x] ~~Set the plot’s initial value when there is an initial selection (what about faceting)?~~ Not yet.
- [x] Tests (minimal…)
- [x] Documentation

Previous tests of brush, brushX and brushY: https://observablehq.com/d/2044e02a89cc688a


nami868:
@Fil My demo report error `TypeError: Plot.brush is not a function`. How do i fix it? Thanks..
image

Fil:
@nami868 are you using a build that includes this code? Plot.brush is not part of the official distribution, it only lives on this development branch. See https://observablehq.com/@fil/plot-early-bird

nami868:
> @nami868 are you using a build that includes this code? Plot.brush is not part of the official distribution, it only lives on this development branch. See https://observablehq.com/@fil/plot-early-bird

thanks

Fil:
superseded by #669

mbostock:
I think this is close. I want to try resurrecting this branch, making a few tweaks, and thinking about the ownerSVGElement problem. We can keep thinking about the selection channel in #669 but we may be able to land brushing (and perhaps pointing? or lasso?) ahead of that.

mbostock:
Idle musing: I wonder if marks could emit _events_?

Also: When brushing with faceting, I could imagine a mode where you can have multiple brushes active, and the selection is the union across facets.

mbostock:
Okay, I think this is ready to go @Fil. It’s missing the ability to set the initial selection, but I think this is still a fine starting point.

mbostock (approved):
LG! Fil, since you opened this, can you just comment on whether you approve?

Fil:
Yes! 🚀

mishatsvelik:
Regarding faceting, how about supporting an option which synchronises the brush area across all facets? Why is this useful? For example lets say we have plot.line faceted and we use brush.x to select points of comparison between the right-most point in the brush area vs the left most point let's say these are dates in a time-series

mishatsvelik:
Also just realised it does not seem to work with time scale

Fil:
rebooting this with https://observablehq.com/@observablehq/plot-brush-interaction-dev

PR #253, opened 2021-03-18
className option for plot, facets, axes and marks

Fil:
supersedes #126

very ugly test https://observablehq.com/@fil/plot-classname-253

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -15,7 +15,8 @@ export class AxisX {
     labelAnchor,
     labelOffset,
     line,
-    tickRotate
+    tickRotate,
+    className
   } = {}) {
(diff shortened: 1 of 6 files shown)


Fil:
https://observablehq.com/@mcmcclur/a-plot-selection-hack by @mcmcclur also wants class names, though it's not totally clear to me if this pull-request corresponds to Mark's use-case.

Fil:
if there is interest, we should redo this as a common style (https://github.com/observablehq/plot/pull/490)

Fil:
@RLesser There wasn't, I've opened https://github.com/observablehq/plot/issues/1002

PR #289, opened 2021-03-31
offset = {center, normalize}

mbostock:
These are for consistency with Vega-Lite:

* offset = center is an alias for offset = silhouette
* offset = normalize is an alias for offset = expand

Although, I thought maybe offset = proportion should be an alias for offset = expand, given that we support a proportion reducer? Though it’s not exactly the same thing. We could have multiple aliases, I suppose, to help people with familiarity with other libraries.


mbostock:
I’m keeping the old options not to violate parsimony but to maintain backwards compatibility.

I also renamed anchor = center to middle, since that better matches textAnchor. Note that labelAnchor for axes supports center, but that’s used in conjunction with left and right or top and bottom.

PR #292, opened 2021-04-01
Render mark

### LINKED ISSUE #182: Function marks?
Should we allow marks to be specified as simple render functions (i.e., with no channels, no properties, and no initialization)?

Fil:
I've made two versions of this (in two consecutive commits):

- 1) the function mark:  `marks: [ () => svg'<text>function mark' ]`
- 2) the render mark:  `marks: [ { render: () => svg'<text>render mark' } ]`

The second solution is imo much better since you can discover it by inspecting a mark. (It might be hard to figure out what initialize does without reading the documentation, but the render function is immediate and simple.)

closes #182

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -15,11 +15,12 @@ export function plot(options = {}) {
   }
 
   // Flatten any nested marks.
-  const marks = options.marks === undefined ? [] : options.marks.flat(Infinity)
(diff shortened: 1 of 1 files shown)


mbostock:
Closing in favor of #498.

Issue #236, opened 2021-03-11
radius legend?


Fil:
This might be a bit difficult to do (do we want superposed circles? how to chose the reference values? even by hand it's difficult…), and it would in any case need inferRadialRange (#235) to be feasible.

related #23

Fil:
superseded by https://github.com/observablehq/plot/pull/432

arky:
+1 I think this would be a good enhancement for plot.

SPattanaik:
>> see #665<https://nam02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fobservablehq%2Fplot%2Fpull%2F665&data=05%7C01%7CSumanta.Pattanaik%40ucf.edu%7C6c949d73ee0449328bf308da38d30ae2%7Cbb932f15ef3842ba91fcf3c59d5dd1f1%7C0%7C0%7C637884776311152352%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=dGETsaOuaQ4zvX1ouZNtVctKLTspiYHSCDChFg3OCmM%3D&reserved=0>
It will be a good addition.

Sumanta Pattanaik
________________________________

Fil:
Another tentative here https://observablehq.com/@observablehq/plot-us-bubble-map#radiusLegend

jonhelfman:
this is Vega-Lite style:
![image](https://github.com/observablehq/plot/assets/17505060/39e4225f-d63f-4b17-8db6-cbe0534c5023)

mbostock:
Maybe we should just start with a simple swatches-style legend? It’s better than nothing (and what Vega-Lite provides).

Issue #23, opened 2020-11-03
Color legend.


mbostock:
When using a color legend (e.g., #2 #19), should we automatically include a nice [legend](https://observablehq.com/@d3/color-legend) (probably yes)?

PR #299, opened 2021-04-05
Separate legends (including main axes) from marks.

Fil:
This allows to add a global color or radius legend to the plot as in https://observablehq.com/@data-workflows/plot-radiuslegend (ref #23 and #236).

Legends have the same signature as marks, but they receive only scales (index and channels are empty).

An issue is that the axes are below the marks, but I would tend to put the color and radius legends above them.
(I kept "unshift" to maintain the exact same order as in the test plots)

plot with two legends

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -15,6 +15,7 @@ export function plot(options = {}) {
   }
 
   // Flatten any nested marks.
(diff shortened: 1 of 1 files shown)


Fil:
see also #425

mbostock:
Related #1197.

PR #301, opened 2021-04-05
avoid illegible ticks when the size of the domain exceeds x times the number of ticks

### LINKED ISSUE #74: Hide tick labels when an ordinal x or y has a high-cardinality domain.
Per https://github.com/observablehq/plot/pull/75#issuecomment-751497435 we want to hide the axis if it has much too many ticks.

Fil:
TODO:
- [ ] update with **tickSpacing**
- [ ] revise with #1790 

http://localhost:8008/?test=mobyDickFaceted has approximately 8 times the number of autoticks (26 letters vs 3.3 autoticks), and it's acceptable because it's single letters (there's even a bit of space left).

Capture d’écran 2021-04-05 à 18 47 50

With a threshold at 10, and setting the height at the ultimate point before this kicks in, it's really become illegible. 
threshold10

With a threshold at 8, it's just barely begun to be illegible. 
threshold8

With a threshold at 5, it's packed, but still legible. 
threshold5

closes #74

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -64,8 +64,10 @@ export class AxisX {
       labelOffset,
       line,
       name,
+      ticks,
       tickRotate
     } = this;
+    if (!x.interpolate && x.domain().length > 8 * ticks) return; // avoid a huge ordinal domain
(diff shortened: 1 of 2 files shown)


Fil:
needs an update with the new **tickSpacing** option

Issue #175, opened 2021-02-25
Should *any* constant property be expressible, somehow, as a data-driven channel if desired?


Fil:
Mike's comment:
> My working assumption is that Marks currently get to decide whether a property is a constant or a channel (or either, e.g., with maybeColor), so if a given Mark doesn’t expose what you want, we either need to change the Mark definition to make it more flexible, or extend the Mark to create a new one. I do think it ends up being simpler if a Mark implementation can declare something as a constant property, rather than needing to support every property as possibly being data-driven. Also, even for constant properties, you can make things somewhat data-driven by declaring multiple marks (as in the Marimekko example) and possibly filtering the data.

Fil:
This question comes up often, and it's not easy to remember which of the properties are constants (basically you discover by trying and failing).

Maybe marks should accept constants (without checking that they may be field names), but *when their typeof is function*, use the function. It would need to be passed the datum.

mbostock:
I don’t think we should do anything immediately here, and this issue has been open for a while, so I am going to close so we can focus on more actionable items. We can always revisit this issue if we keep being frustrated by certain options only being channels, but I think switching to everything potentially being a channel will add complexity and burden to mark implementations that we should avoid if possible.

tobiassjosten:
Would it be an option to pick some properties and make those into channels? I'm thinking specifically of `dy` and `dx` for the text plot, which I right now would have loved to offset differently depending on the data.

Fil:
Yes, see the work in progress #909

PR #322, opened 2021-04-13
strokeWidth as a channel

### LINKED ISSUE #311: strokeWidth channel on links
would be useful to draw a graph (I thought we already had an issue for this?)

Fil:
closes #311

build : https://observablehq.com/@fil/plot-strokewidth-322

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,7 +476,7 @@ All marks support the following optional channels:
 * **strokeOpacity** - a stroke opacity; bound to the *opacity* scale
 * **title** - a tooltip (a string of text, possibly with newlines)
 
-The **fill**, **fillOpacity**, **stroke**, and **strokeOpacity** options can be specified as either channels or constants. When the fill or stroke is specified as a function or array, it is interpreted as a channel; when the fill or stroke is specified as a string, it is interpreted as a constant if a valid CSS color and otherwise it is interpreted as a column name for a channel. Similarly when the fill or stroke opacity is specified as a number, it is interpreted as a constant; otherwise it is interpeted as a channel. When the radius is specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel.
+The **fill**, **fillOpacity**, **stroke**, **strokeWidth**, and **strokeOpacity** options can be specified as either channels or constants. When the fill or stroke is specified as a function or array, it is interpreted as a channel; when the fill or stroke is specified as a string, it is interpreted as a constant if a valid CSS color and otherwise it is interpreted as a column name for a channel. Similarly when the fill or stroke opacity or the stroke width is specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel. When the radius is specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel.
 
 The rectangular marks ([bar](#bar), [cell](#cell), and [rect](#rect)) support insets and rounded corner constant options:
 
--- a/README.md
+++ b/README.md
@@ -474,6 +474,7 @@ All marks support the following optional channels:
 * **fillOpacity** - a fill opacity; bound to the *opacity* scale
 * **stroke** - a stroke color; bound to the *color* scale
 * **strokeOpacity** - a stroke opacity; bound to the *opacity* scale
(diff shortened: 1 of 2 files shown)


Fil:
Using this for a Sankey diagram, I'll need to set the width with a scale: identity. (Doesn't mean a scale isn't going to be useful in general.)

ryoqun:
@Fil @mbostock hi, is this going to be merged soon? :)

Also, I want tick's strokeWidth to be a channel as well.

Fil:
See #175 and #380

mbostock (changes requested):
I support not having a scale for strokeWidth (at least for now), but it does feel arbitrary to only expose this property on the link mark. I would prefer to make this a standard channel on all marks, like we do with strokeOpacity. Even if that channel often goes unused, it’s still meaningful, and should be supported for consistency/symmetry. I might take a pass at this now.

ryoqun:
> I support not having a scale for strokeWidth (at least for now), but it does feel arbitrary to only expose this property on the link mark. I would prefer to make this a standard channel on all marks, like we do with strokeOpacity. Even if that channel often goes unused, it’s still meaningful, and should be supported for consistency/symmetry. I might take a pass at this now.

hey, I've always been fan of d3.js. thanks for moving this forward. :)

README.md
```diff
@@ -459,7 +459,6 @@ All marks support the following style options:
 * **fill** - fill color
 * **fillOpacity** - fill opacity (a number between 0 and 1)
 * **stroke** - stroke color
-* **strokeWidth** - stroke width (in pixels)
```
mbostock, line 462: This should be kept, right? The options that are supported as channels are also listed as constant options.

ryoqun:
hooray! thanks for working hard on this. so, strokeWidth for all marks? cool. :)

fyi, I'm creating a nice chart like this with plot (simply put, we're debugging any performance issues for distributed computing node's consensus formulation):

image

And I want to use `strokeWidth` to embed extra scalar values into marks, instead of radius for visual conciseness and proper visual comparison (radius/area isn't cognitively easy than just width for visual measuring).

Fil:
wow! great example, I'd be curious to see it in action if you can share it! If you want to test this branch I've updated the build at https://observablehq.com/@fil/plot-strokewidth-322 (see also https://observablehq.com/@fil/plot-early-bird)

ryoqun:
>  I'd be curious to see it in action if you can share it!

yeah, I'd like to share but it's still in bad shape to share currently... needs some preparation to be published to the public..

> If you want to test this branch I've updated the build at https://observablehq.com/@fil/plot-strokewidth-322 (see also https://observablehq.com/@fil/plot-early-bird)

seems to work nicely! (I applied strokeWidth to Plot.dot and Plot.tickX. I can't wait official new version release with it. :)

image

README.md
```diff
@@ -476,7 +476,7 @@ All marks support the following optional channels:
 * **strokeOpacity** - a stroke opacity; bound to the *opacity* scale
```
mbostock, line 476: Add strokeWidth here?

mbostock (approved):

Fil:
example for CHANGELOG:
strokeWidth-channel

Plot.dotX(d3.range(41), { strokeWidth: (d) => (1 + d) / 15 }).plot()

enjalot:
> Can do, but I'm not sure it's worth it. Links tend to come with value=1…10 (co-occurences as in the miserables.json example), or 0…1, or sometimes from distance matrices 0…infty, and I feel it's always more "graphic tweaking" than really using it as an encoding to be shared with another mark.

I didn't think about the "shared with another mark" aspect, but from the perspective of "Plot figures out how to map data space to graphic space for me and I just want to tweak the pixels" having a scale here is what I expect. It's similar to how I would think about the `r` channel in dots, where I usually want to impose some maximum size. That could be considered a graphic tweak but it's a pretty meaningful one.
I suppose for the same reason I also want a scale for fontSize, because without a scale I have to go outside of Plot to calculate the extent of the data and do the math that Plot would do for me in the other channels, like opacity

PR #333, opened 2021-04-20
Renames

Fil:
* window transform: _shift_ to _anchor_ (old options: _leading/centered/trailing_ ; new options: _start/center/end_)
* stack offset: _silhouette_ to _center_

question:
* stack offset: should "expand" be renamed "normalize"? ; cf #289 (which offers them as aliases)

### DIFF
--- a/test/transforms/windowMean-test.js
+++ b/test/transforms/windowMean-test.js
@@ -34,16 +34,15 @@ it("movingAverage treats null as NaN", () => {
   assert.deepStrictEqual(m3.x.transform(), [, 1, NaN, NaN, NaN, 1, 1, 1, NaN, NaN, NaN, 1,, ]);
 });
 
-it("movingAverage respects shift", () => {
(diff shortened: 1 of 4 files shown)


test/transforms/windowMean-test.js
```diff
@@ -34,15 +34,16 @@ it("movingAverage treats null as NaN", () => {
   assert.deepStrictEqual(m3.x.transform(), [, 1, NaN, NaN, NaN, 1, 1, 1, NaN, NaN, NaN, 1,, ]);
 });
 
-it("movingAverage respects shift", () => {
+
```
Fil, line 37:
```suggestion
```

Fil:
@mbostock would be good to decide on this before the next release.

I would vote for all renamings, including expand => normalize, and not having aliases, but I don't have a strong preference.

mbostock:
I’m going to merge this into #289 and then ensure backwards compatibility with aliases.

PR #334, opened 2021-04-20
sort bins

### LINKED ISSUE #297: Sort bins and groups?
The bin and group transforms don’t support an *order* and *reverse* option like the stack transform does, but I think we need it. For example, if you use the binMid transform together with a dot mark, you usually want to draw the smallest dots on top so they don’t get obscured.

Fil:
The proposed syntax specifies *sort* in the reduce/outputs object, enabling sorting on a different reducer.

~~~js
      Plot.dot(data, Plot.bin({
        r: "count",
        sort: "count",
        reverse: true
      }, {
        x: "body_mass_g",
        y: "culmen_length_mm",
        stroke: "species",
        fill: "species"
      }))
~~~
 

closes #297


mbostock:
This is very similar to what I propose for filter! #495

mbostock (approved):

mbostock (changes requested):
Oops, I just realized there’s an issue here. This **sort** option allows a corresponding *sort* input so that you can, for example, sort by sum of weights. However, the *sort* input channel is retained in *options* and thus is *also* applied as a basic sort transform prior to the bin and group transform. That seems confusing: either you intended the *sort* option to be the basic sort transform, or you intended the *sort* option to be the input to the sort reducer, but not both. We should try to avoid this ambiguity.

Another issue is that sorting is a little different than most reducers: you don’t strictly want to compute an output value; instead you want to define a total order. And hence it’d be nice if you could specify a comparator to the **sort** output in the same way you can to the basic transform.

I’ll try to work on this tomorrow.

mbostock:
I’m not sure about the comparator issue, but as far as the input channel vs. basic transform ambiguity goes with the sort option, I’m thinking we can resolve that by changing how we interpret the sort option based on the presence of be sort output. If a sort output is specified, then the sort option represents the input channel; but otherwise the sort option is a basic transform that is applied before the bin or group. (And you can always make this explicit using Plot.sort if you like.) This approach could be extended to the filter option in the other PR, too.

mbostock:
I am going to merge this into the filter option PR #495 so we can do both simultaneously.

PR #349, opened 2021-05-01
dot default sort: descending R

Fil:
it is the most common use case; deactivates as soon as the sort <del>or reverse</del> option is defined

related: #334


mbostock:
The problem here is that `options.sort === undefined && options.reverse === undefined` is not a great test for whether the data is already sorted, since you could have a transform in place which sorts the data (*e.g.*, Plot.sort, but all transforms support the basic sort transform).

Another way of implementing this might be

```js
function dot(data, {x, y, r, ...options} = {}) {
  ([x, y] = maybeTuple(x, y));
  options = {...options, x, y, r};
  if (r != null) options = reverse(sort(options));
  return new Dot(data, options);
}
```

but again this doesn’t solve the problem of not knowing whether it’s safe to apply the default sort by descending *r*.

Here are a few alternatives:

1. Do nothing; always require explicit sort if desired.
2. Always sort by descending radius; do not allow custom sort orders.
3. This PR; handles the common case, but may require `{...transform({…}), sort: null}` to disable.
4. This PR, but also only sort if `options.transform === undefined`.
4. Make a new dedicated option apart from *sort* and *reverse* that controls this additional sort behavior (“sorted”?).
5. Make a new convenience constructor for sorted dots (as shown above).
6. Change the dot constructors to sort, and offer a convenience constructor for unsorted dots.
7. Bring back the *z* channel and default it to *r* (undoing part of #345)?

None of these feel like great options, so I’m probably inclined to do nothing until I feel more confident.

mbostock:
With the sort option for bin #334 we can write:

```js
Plot.dot(data, Plot.bin({
  r: "count",
  sort: "count",
  reverse: true
}, {
  x: "body_mass_g",
  y: "culmen_length_mm",
  stroke: "species",
  fill: "species",
  fillOpacity: 0.5
}))
```

This is one more line of code that the sortedMark option, but it avoids introducing a new concept (assuming we still want the sort option for bins and groups, but I think we do).

Fil:
OK since we already have two options, let's punt on introducing a third one :)

I just want to update penguinSexMassCulmenSpecies to be sorted, as an example.

PR #352, opened 2021-05-01
Default to ylgnbu instead of turbo.

mbostock:
Feels like a safer default, though still a little worried about contrast on white.


Fil:
just to mention that we've also discussed batlow, and need more research

mbostock:
Going to close this until we do more research.

PR #368, opened 2021-05-08
Add diverging-log et al.

### LINKED ISSUE #367: Additional diverging scale types.
We should support additional scale types:

* diverging-log
* diverging-pow
* diverging-sqrt
* diverging-symlog

mbostock:
Fixes #367.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ The normal scale types — *linear*, *sqrt*, *pow*, *log*, *symlog*, and *ordina
 * *sequential* - equivalent to *linear*
 * *cyclical* - equivalent to *linear*, but defaults to the *rainbow* scheme
 * *diverging* - like *linear*, but with a pivot; defaults to the *rdbu* scheme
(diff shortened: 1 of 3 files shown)


Fil (changes requested):
diverging-log fails if the _first_ value is negative

```
Plot.plot({
  marks: [Plot.cellX([-1, 0.1, 0.5, 1, 2, 10], { fill: d => d, x: d => d })],
  color: { type: "diverging-log", scheme: "rdbu" }
})
```
fails.


```
Plot.plot({
  marks: [Plot.cellX([0.1, 0.5, 1, 2, 10, -1], { fill: d => d, x: d => d })],
  color: { type: "diverging-log", scheme: "rdbu" }
})
```
works.

I think it should use the sign of the pivot instead of the sign logic that we have in type:"log".

mbostock:
Using the pivot makes a lot of sense. 👍

Fil (approved):

PR #402, opened 2021-05-14
include margins in the label translate

### LINKED ISSUE #375: marginRight in y facet does not translate label of x
Observed when I applying facet y on a dot plot:

When I change marginRight  within the facet y object, 
it updates the dot plot and the x axis  with the ticks (as expected) but does not translate the label of x to the new end of the axis

image

(see the age label in: https://observablehq.com/d/1d9e39bd4f83a69a)

Fil:
fixes #375

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -43,8 +43,8 @@ export class AxisX {
       marginLeft,
       facetMarginTop,
       facetMarginBottom,
-      labelMarginLeft = 0,
-      labelMarginRight = 0
+      labelMarginLeft = 20,
+      labelMarginRight = 20
     }
   ) {
     const {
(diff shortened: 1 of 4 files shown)


Fil:
rebased

mbostock:
A little explainer would be nice here… I’m finding it hard to follow what the logic was vs. what it should be. I feel like the current behavior was intentional (put the _x_-axis label as far right as possible to give it space); now I’m trying to decide what the new behavior should be, if the current one is not desirable.

Fil:
The idea is to position the axis label relative to the "tip of the axis", instead of having it float to the far end. In the case of the OP, that gives:

### before 

Capture d’écran 2021-09-07 à 07 47 22


### after

Capture d’écran 2021-09-07 à 07 47 57

mbostock:
But what is the default label margin of 20?

Fil:
That was to match (most of) the existing plots.

>

mbostock:
Because normally there isn’t a second axis? It’s matching the 20 from here, I presume.

https://github.com/observablehq/plot/blob/07cf90baf428ade198d1cac5ec3ab3f531804995/src/plot.js#L118-L121

Fil:
I see now that this branch works only in a few cases, and is far from working in general.

I've made an interactive here https://observablehq.com/d/79d3d39812dd3441 ; it seems to me that the behavior of the two first charts is desirable, *i.e.* that the label should be tied to the tip of the axis. But it would need to work in the four faceting possibilities (no faceting, x, y, x+y), for labelAnchor=left, and also for the vertical axis (which I haven't touched here either).

Maybe an alternative would be to have a configurable label.**shift**? (label.**offset** is for the orthogonal direction). I'm closing this for the time being.

mbostock:
Nice notebook. That helps! 👍 It’s definitely tricky and I expect we do want some tweaks to the default behavior as well as likely new options.

PR #403, opened 2021-05-14
ignore null transform

### LINKED ISSUE #357: A null scale transform should be equivalent to undefined.
This currently errors, but it shouldn’t, I think:

```js
Plot.plot({
  y: {
    transform: null
  },
  marks: [
    Plot.areaY(sftemp, {x: "date", y1: "low", y2: "high"})
  ]
})
```

Fil:
fixes #357

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -35,7 +35,7 @@ export function plot(options = {}) {
       if (scale !== undefined) {
         const scaled = scaleChannels.get(scale);
         const {percent, transform = percent ? x => x * 100 : undefined} = options[scale] || {};
-        if (transform !== undefined) channel.value = Array.from(channel.value, transform);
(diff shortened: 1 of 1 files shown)


src/plot.js
```diff
@@ -35,7 +35,7 @@ export function plot(options = {}) {
       if (scale !== undefined) {
         const scaled = scaleChannels.get(scale);
         const {percent, transform = percent ? x => x * 100 : undefined} = options[scale] || {};
-        if (transform !== undefined) channel.value = Array.from(channel.value, transform);
+        if (typeof transform === "function") channel.value = Array.from(channel.value, transform);
```
mbostock, line 38: I didn’t mean we should silently ignore a invalid argument, like if you pass a transform object instead of a function. Instead I meant we should consider null and undefined to be equivalent here. I’ll make the change and merge.

PR #405, opened 2021-05-14
throw when facet is missing data

### LINKED ISSUE #371: Better error message for unspecified facet data
A plot call like this

```js
Plot.lineY(data, {
  x: 'year',
  y: 'employ_n'
}).plot({ facet: { x: 'major_occupation', y: 'minor_occupation' } })
```

Currently gives the error `TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`. What this actually means is that the `facet` specification wants to be passed a `data` array:

```js
Plot.lineY(data, {
  x: 'year',
  y: 'employ_n'
}).plot({ facet: { data: data, x: 'major_occupation', y: 'minor_occupation' } })
```

Perhaps the error could be more clear on this point, e.g. something along the lines of `Error: Faceting requires 'data' to be specified in the plot facet options.`

Fil:
closes #371

### DIFF
--- a/src/facet.js
+++ b/src/facet.js
@@ -10,7 +10,7 @@ export function facets(data, {x, y, ...options}, marks) {
 
 class Facet extends Mark {
   constructor(data, {x, y, ...options} = {}, marks = []) {
-    if (data === undefined) throw "Missing facet data";
+    if (data === undefined) throw new Error("missing facet data");
     super(
       data,
       [
--- a/src/facet.js
+++ b/src/facet.js
@@ -10,6 +10,7 @@ export function facets(data, {x, y, ...options}, marks) {
 
 class Facet extends Mark {
   constructor(data, {x, y, ...options} = {}, marks = []) {
+    if (data === undefined) throw "Missing facet data";
     super(
       data,
       [


src/facet.js
```diff
@@ -10,6 +10,7 @@ export function facets(data, {x, y, ...options}, marks) {
 
 class Facet extends Mark {
   constructor(data, {x, y, ...options} = {}, marks = []) {
+    if (data === undefined) throw "Missing facet data";
```
mbostock, line 13:
```suggestion
    if (data === undefined) throw Error("missing facet data");
```

mbostock:
I’m not sure this is the right fix. My intent was that if you don’t specify the facet data, it’s equivalent to not having faceting—not that your Plot now throws an error. I think there’s a different error at play here perhaps?

Fil:
The bug is that Plot crashes when there is a facet with no (iterable) data attribute. Both approaches (error or ignore) can be valid fixes for that bug.

Now, a common difficulty I've experienced and seen people struggle with is that when you focus on faceting and you are working on one dataset, you will naturally write  `facet: {x: "key"}` and forget to add `data`. A second difficulty is that when you start from an example that is written `facet: {data, x: "key"}` and yo want to substitute your own dataset, you will tend to replace `data` by the new dataset, and do `facet: {penguins, x: "key"}`.

With the "ignore" proposal, these two do nothing and leave you wondering why. With the "error" proposal they give you a reason. "error" has my preference.

Writing this I realize that we could error if data is undefined, and ignore if it is null or false (ie explicitely disabled).

(Currently, to disable faceting  temporarily I rename `facet` to `_facet`.)

mbostock (approved):

PR #406, opened 2021-05-17
add a {line: true} option to axes

Fil:
handy for materializing the vertical axis of a bar chart with an ordinal x
note that in the case of a continuous *X*, tickX([*extent min*]) is often preferable, since it creates the support line for all x/y facets

see https://observablehq.com/@fil/axis-support-406

suggested by @ee2dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ Plot automatically generates axes for position scales. You can configure these a
 * *scale*.**tickFormat** - how to format tick values as a string (a function or format specifier)
 * *scale*.**tickRotate** - whether to rotate tick labels (an angle in degrees clockwise; default 0)
 * *scale*.**grid** - if true, draw grid lines across the plot for each tick
-* *scale*.**axisLine** - if true, draw the axis line
+* *scale*.**line** - if true, draw the axis line
 * *scale*.**label** - a string to label the axis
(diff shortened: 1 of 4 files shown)


enjalot:
this makes me want to have a `{ frame: true }` option that adds a frame (to the bottom of the marks?). For most facet use cases a default frame is probably desired and good enough.
this wouldn't replace using the frame mark to create a customized frame, just make it easier to have one when exploring.

mbostock:
This seems like a good option to have, but I’m struggling with the name “support”—that doesn’t resonate with me. D3 calls this the “domain path”. Maybe **axisLine** would be okay? Should we allow you to style this line separately from the grid? It does seem related to the “frame” concept, too, so maybe **frame** would be okay as an option.

Fil:
Is "support" a frenchism? Maybe a better translation for what I had in mind is “bearing” or “beam”.

Anyway, in this PR the axis line would be added (or "not removed" :)) on the materialized axes, not on all facets. 

By contrast, Plot.ruleX([0]) and Plot.frame are present on all facets (more precisely "active" facets #332).

mbostock:
Needs a better name than “support” but otherwise I’m onboard with the option.

Fil:
axisLine is fine, I think?

mbostock (approved):

Fil:
axis-line

(example for the CHANGELOG)

Issue #408, opened 2021-05-18
Dark mode


Fil:
Dark mode almost works: the use of currentColor for stroke and fill is solving 90% of the problem.

In some cases, though, it's not ideal:
- in cars-parcoords we use white to detach the text against the background. Using "fill: currentColor; filter: invert(1)" would work.
- in metroUnemploymentRidgeline the areas are filled with #ccc, which could be replaced by "fill: currentColor; filter: invert(20%);"
- and more.

Adding support for the filter attribute would help a lot.

See also https://github.com/observablehq/plot/discussions/1461

Fil:
A smaller issue is that hard-coded `.plot background {#fff}`, which I had to override in bf7b99fbe9a77052008c38b5e50a390618c8fad9

reubano:
Reference [metricsgraphics](https://metricsgraphicsjs.org/examples.htm#rollovers)

mbostock:
We support dark mode in our documentation and I even tweeted about it:

https://twitter.com/mbostock/status/1653069773645938689

Maybe we can close this issue? Maybe the last remaining issues are the default white background #1434 and the default white marker stroke #1463 but both issues can be worked around with a stylesheet.

Fil:
At some point we'll have to consolidate the documentation, but in terms of support I think we have all we need.

PR #409, opened 2021-05-18
imageFilter (support for the CSS filter attribute)

Fil:
Helps with dark mode (#408)


src/style.js
```diff
@@ -14,6 +15,7 @@ export function Style(mark, {
   strokeDasharray,
   mixBlendMode
 } = {}) {
+  mark.colorFilter = string(colorFilter);
```
mbostock, line 18:
```suggestion
  mark.colorFilter = impliedString(colorFilter, "none");
```

mbostock (commented):
I’m going to punt on this. I’m unsure about the name “colorFilter”. It’s unfortunate that we can’t use *filter* because it conflicts with the basic transform. We could, possibly, interpret the *filter* as CSS filter if it’s a valid CSS filter function string, similar to how we validate colors. But that’s a little fancy…

https://drafts.fxtf.org/filter-effects/#filter-functions

mbostock:
I think we should do this, but we should call it **imageFilter** instead of **colorFilter**.

Fil:
rebased as imageFilter.

mbostock (commented):
Looks good, but you can add a tiny test too?

mbostock (approved):

PR #410, opened 2021-05-18
for string values, the "proportion" reduce uses counts instead of sums

### LINKED ISSUE #399: reduce: proportion for counts
As @yurivish remarks in https://observablehq.com/d/0d9e55233e524500 the "proportion" reducer doesn't get an automatic label if we don't pass _x_, and crashes if we pass it a string channel _x_ (since the sum of strings is 0).

Fil:
the type detection is consistent (done on the whole set), so mixes of nulls and numbers will not give a weight of 1 to nulls. 

however it's not done once, since I don't know how to cache the result — but in most cases it will return immediately.

fixes #399

test build at https://observablehq.com/d/1dd4f1c3e491b6d6

maybe it would be good to add a few test cases mixing nulls, undefined, numbers, strings.


Fil:
@yurivish if you want to review

src/transforms/group.js
```diff
@@ -201,10 +202,16 @@ const reduceCount = {
   }
 };
 
+function sumOrCount(I, V) {
+  return typeof V.find(v => v != null) === "number"
```
mbostock, line 206: This could lead to inconsistent behavior across series (and facets). Would it be possible to do this earlier, on the first non-null value in the entire channel, as we do for other forms of type inference?

Fil: I think that's what I'm doing here: V is the entire channel. What I don't like is I'm doing the same for each series, without caching.

mbostock: Ah, thanks for the clarification.

src/transforms/group.js
```diff
@@ -213,8 +214,13 @@ const reduceDistinct = {
 
 const reduceSum = reduceAccessor(sum);
 
+function sumOrCount(I, V) {
+  if (!("type" in V)) V.type = typeof V.find(v => v != null) === "number";
```
mbostock, line 218: You’ll want a WeakMap here instead of mutating V. (I believe that V can sometimes be an externally-provided value, but even if it isn’t, we generally want to avoid mutating inputs.)

Or another simple approach that I believe will work here is single-value memoization: record the last-passed V value and the last-computed type; if the new V value is the same as the old one, return the previously-computed type.

Also, if we do this, it’ll persist across render calls even if the caller mutates the contained values. That might be surprising, but it’s probably rare enough to not worry about.

mbostock: Also, we’ll want an isNumeric helper here modeled after isTemporal and isOrdinal, and we should probably use for-of instead of array.find for consistency:

https://github.com/observablehq/plot/blob/8aafa7c41d08afdaaf8a800a612f9c243d6e9ce9/src/mark.js#L332-L338

Fil: yes! and memoization will be useful for all of isOrdinal, isTemporal, isNumeric, not just for isNumeric. In a78d7d4 I've centralized the content type detection.

mbostock (changes requested):
I think we should remove the memoization until we have evidence that suggests that it is a bottleneck.

mbostock (changes requested):
I’d prefer to leave the isOrdinal and isTemporal as-is, and just introduce a new isNumeric that follows the similar pattern.

src/transforms/group.js
```diff
@@ -164,6 +164,7 @@ export function maybeSubgroup(outputs, Z, F, S) {
 
 function reduceFunction(f) {
   return {
+    label: f.label,
```
mbostock: This change seems unrelated (and untested), so I’m going to revert this part. Elsewhere we don’t promote the reducer name to the axis label so I think it probably isn’t appropriate to promote it here. (For example, when you use the “sum” reducer, we don’t use or want “sum” as the axis label; we want the name of the column being summed.)

test/plots/software-versions.js
```diff
@@ -0,0 +1,31 @@
...
+export default async function() {
+  const data = await d3.csv("data/software-versions.csv");
+  const options = Plot.stackX(
+    Plot.groupZ(
+      {
+        x: "proportion", // uses sumOrCount in the patched version
+        text: "first"
+      },
+      {
+        x: "key", // this is only strings
```
mbostock: This isn’t strictly needed here because it doesn’t affect the composition of the groups (which here for groupZ is determined only by the _fill_ channel), i.e. removing this key doesn’t affect the chart (but with it removed, we’re no longer testing the feature added by this PR). It’d be nice to have a more explicit test.

mbostock:
I’m wondering if this PR makes sense? If there’s a value input to the _proportion_ reducer, it means we want a weighted proportion. But if those values are ordinal, there’s no way to know what the appropriate weights are to compute the weighted proportion. We could ignore the input value, but that seems undesirable to silently ignore it? Perhaps it would be better to throw an error in this case since the idiomatic fix is to remove the unnecessary `x: "key"` from the group transform options.

mbostock:
Closing as argued in https://github.com/observablehq/plot/issues/399#issuecomment-890596083 but please continue the discussion if you are not convinced. (Also, it might still be worth adding this test, since it now works in main.)

Issue #170, opened 2021-02-24
Expose more Plot internals to facilitate building new mark types.


mbostock:
In an upcoming Plot release, we should document how to implement custom mark types and expose the necessary internals to make this easier. Ideally, all of Plot’s built-in mark types are “unprivileged” in the sense that they rely exclusively on stuff that’s available publicly. That way we can feel good about the barrier to custom mark types being low.

I’m thinking we expose [options.js](https://github.com/observablehq/plot/blob/main/src/options.js) mostly as-is, but perhaps adopt a naming convention that delineates these implementation helpers from Plot’s more commonly-needed plotting API. For example, maybe it’s renderDirectStyles, renderIndirectStyles, renderTransform, renderChannelStyles, renderOffset instead of applyDirectStyles, applyIndirectStyles, applyTransform, applyChannelStyles, offset. Or at least we do a pass on naming before exposing publicly, and have a separate section in the README (or even a different README) for custom marks.

Fil:
Closing, this feels 99% done.

PR #411, opened 2021-05-19
Expose Plot.column and Plot.transform 

Fil:
helpers to create custom transforms (#170)

An example uses Plot.transform to create a uniform sample transform (diamond-carat-sampling)
An example of Plot.column is available at https://observablehq.com/@fil/plot-splom


Fil:
Rewritten; this time I've left out the renaming of maybeLazyChannel into column, and basic into transform, since it only creates merge conflicts. But when (if) we merge, it would be good to rename these functions internally.

mbostock:
Looks like the documentation needs updating here, since Plot.transform now takes (*options*, *transform*) rather than (...*transforms*)?

mbostock:
Also, do we want to expose maybeLazyChannel here, or lazyChannel? My inclination is that I’d prefer Plot.column to be an alias for the latter, so that a setter function is always returned (even if there is no *source* input).

Fil:
rebased with fixes to the documentation and adoption of lazyChannel instead of maybeLazyChannel.

mbostock (approved):

Issue #388, opened 2021-05-12
Sort a scale’s domain according to a channel, or other scale?


ericemc3:
It is common to have to sort marks according to another channel: vertical or horizontal bars sorted by decreasing size for example.

![image](https://user-images.githubusercontent.com/17407796/117989654-cafe3880-b33c-11eb-90d7-24918bdb9d02.png)
 
Vega-Lite offers a very simple syntax to do this: `sort("-y")`, to sort for example markBars according to channel `y`, and ggplot2 offers a similar syntax (`x = reorder(...)`).

With Plot, it's more complex, especially in the context of grouping, with domain specifications via `d3.groupSort()` that require a fine-grained knowledge of these advanced D3 functions, and I'd rather avoid this kind of dependency to D3.

I would therefore suggest a syntax like this, for a barY mark, for example:
```
x : {
  order : "-y"
}
```
or
```
x : {
  order : {
    channel: "y",
    orientation: "desc"
  }
}
```
And, more advanced (for instance sorting facets according to the median of each group) :
```
fy: {
  order: "-median(y)"
}
```

PS: Plot is a great library, very intellectually stimulating, and the 21 beautifully written articles in the https://observablehq.com/collection/@observablehq/plot collection should be read by all data scientists!

mbostock:
We did a lot of previous exploration on this topic; some of it is captured in #32 #106 #108. I’m supportive of continuing this effort, but I don’t think it’s easy. I have two fairly strong opinions here. First, it should be explicit and opt-in: you have to declare which mark channel is driving the sort order. Second, we should avoid a string DSL like `"-y"` or `"median(y)"` because we want this to be extensible with JavaScript.

josesho:
As a newcomer to Observable Plot (longtime user of ggplot2 and matplotlib/seaborn, light d3 user), I found it a little counter-intuitive that Plot auto-sorts the order of the categories in a (simple/basic) barchart, especially if in prior data-munging one has already sorted it a certain way. 

Is it possible to either turn off this auto-sort by default?

Great work by the way, I'm really looking forward to when you guys bake-in basic interactivity!

Fil:
Thanks for pushing the question again.

[...]

PR #414, opened 2021-05-23
sort groups

Fil:
Don't sort the groups in `Plot.group()`; the only case where it makes a difference is when the groups are stacked, and the stacking order can be made explicit (see test: penguins-species-island).

Ordinal scales now accept a sort option, which can be:
- `"ascending"`: default, sorts in ascending order with undefined first
- false, null (or `input`): sort in input order
- `"count"`: sort by the number of times each ordinal value is present in the marks.

Note: when using Plot.barY the "count" is equal to 1 for each bar—it is *not* the number of times each value is present in the data. However `{sort: "count"}` works well to sort facets, or an ordinal scale for a Plot.dot chart. See the discussion in #388 for more details.

The build and a few examples are available at https://observablehq.com/d/d797c5a9572385bd


Fil:
90% obsolete after #442 ; but not sorting groups is probably still a good idea

Fil:
Note that #332 also takes care of facet sorting (by count, by input, or by value).

mbostock:
There’s a lot of overlap here with the new *sort* and *filter* options for bin and group #495. But related to #435, we don’t pass the key(s) to the group reducer, so it’s not currently possible to implement the default natural ordering of groups and hence there’s a separate internal sorting that happens prior to the user-specified *sort* reducer. But I think we can make it work…

mbostock:
Removing the sort here also affects the bin transform, since the bin transform groups internally, too (e.g., on *z*). So this is extra tricky: maybe the sort reducer would also need to be passed the group keys *in addition to* the bin extents.

mbostock:
Going to close this as covered by #495, but I recognize there are still things to do (mostly covered by #435).

PR #415, opened 2021-05-24
quantile and threshold color scales

### LINKED ISSUE #372: Threshold scales
E.g., color: {type: "threshold", domain: [30, 50], range: ["red", "green", "blue"]}.

### LINKED ISSUE #373: Quantile scales
E.g., color: {type: "quantile", scheme: "blues", quantiles: 7}.

Fil:
`color: {type: "threshold", domain: [30, 50], range: ["red", "green", "blue"]}`
closes #372

`color: {type: "quantile", scheme: "blues", quantiles: 7}`
closes #373

*TODO*
- [ ] see bug below with ordinal schemes

(still needs a few test plots)


Fil:
we'll use the test plot in #419

Fil:
I've added a build in https://observablehq.com/@fil/quantile-threshold-shaperendering-415-419

Here's a bug though: *Threshold scales should accept categorical schemes.*

```js
Plot.dot(data, { x: "bill_length", y: "species", fill: "body_mass" }).plot({
  color: {
    type: "threshold",
    domain: [2000, 3000],
    scheme: "dark2"
  }
})
```

gives `TypeError: ordinalScheme(scheme === undefined ? "rdylbu" : scheme) is not a function. (In 'ordinalScheme(scheme === undefined ? "rdylbu" : scheme)({length: domain.length + 1})', 'ordinalScheme(scheme === undefined ? "rdylbu" : scheme)' is an instance of Array)`

mbostock:
For the diverging schemes (scheme11’s) you don’t get very good results with n = 2: you get the first two (red and gray in RdBu or red and yellow in RdYlBu).

Screen Shot 2021-08-02 at 3 23 20 PM

What you want is probably the outer two (red and blue).

Screen Shot 2021-08-02 at 3 23 28 PM

Maybe we can fix this in scheme11 for the n = 2 case.

https://github.com/observablehq/plot/blob/4286100d658ebef38f8d5fb4e0d213cfc0556685/src/scales/schemes.js#L153

mbostock:
The reason for the TypeError is that sometimes ordinalScheme returns an array (for the ordinal schemes) while other times it returns a function that will later be converted into an array when the domain is known:

https://github.com/observablehq/plot/blob/96d2f3f3d5318a80f9fe2d27ce6dcf721b9f8951/src/scales/ordinal.js#L16-L17

So we either need to make sure ordinalScheme always returns a function, or we need to handle a returned array when we call ordinalScheme. I can fix.

src/scales/quantitative.js
```diff
@@ -198,10 +104,34 @@ export function ScaleLog(key, channels, {base = 10, domain = inferLogDomain(chan
...
+  if (reverse) range = reverseof(range);
+  return ScaleQ(key, scaleQuantile(), [], {domain, range, ...options});
+}
+
 export function ScaleSymlog(key, channels, {constant = 1, ...options}) {
   return ScaleQ(key, scaleSymlog().constant(constant), channels, options);
 }
 
+export function ScaleThreshold(key, channels, {
+  domain = inferDomain(channels),
```
mbostock, line 124: A threshold scale’s domain represents the thresholds, so if we default to [min, max] here, we’ll have a 3-element range, but we’ll only use one of the three output colors (because all of the input values to the scale will necessarily be >=min and <=max). Probably a more sensible default would be [median] or [first quartile, median, third quartile]… or maybe we just have no default domain and require you to specify one, which might make the most sense for a threshold scale. If you want quantiles you can use a quantile scale.

mbostock: I decided that `domain = [0]` is a more reasonable default (better than crashing).

mbostock (approved):

PR #416, opened 2021-05-24
expose color schemes

Fil:
Maybe we should tighten the API a bit, for example when an ordinal scheme receives a domain's length it receives in fact the domain in most cases, but we wouldn't want people to think that it is always the domain, since in the case of quantile scales it is only a {length} object.

Also, I don't like the names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,31 @@ Plot.plot({
 })
 ```
 
+#### Extending schemes
+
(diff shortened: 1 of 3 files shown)


mbostock:
I don’t think we should expose mutators, but I could see something like Plot.withSchemes({…}) or Plot.configure({…}) that returns a function like Plot.plot but with some behavior overridden.

mbostock:
I’m going to close this for now since I think Plot.withSchemes(…) would be a fair amount of work, and I’m not sure how often people would want to do it over just using the existing range or interpolate option. But I’ve rebased this so if you do wish to pick it up, hopefully it will be ready to start tinkering. 🙏

Issue #417, opened 2021-05-25
Default bins too narrow using Plot.binX


mkfreeman:
The default binning in Plot can create rects with a width that is too small to see (making a user -- namely, me -- believe the code is broken). As an example (also see [this notebook](https://observablehq.com/d/2029a9da3400238d)):

<img width="663" alt="Screen Shot 2021-05-25 at 6 12 33 PM" src="https://user-images.githubusercontent.com/1373882/119575275-c8f89880-bd84-11eb-880a-2097d5352b24.png">

Compared to the same plot made using ggplot2:

<img width="779" alt="ggplot_example" src="https://user-images.githubusercontent.com/1373882/119576700-42918600-bd87-11eb-9388-9b5d8f9a3dd0.png">

mbostock:
Good find. I think we should consider not applying an inset (`{inset: 0}`) if there are a lot of bins:
![untitled-82](https://user-images.githubusercontent.com/230541/119577690-f555ea00-bd6f-11eb-90c9-4d67bc1c984d.png)

We could also cap the number of bins returned by the default d3.thresholdScott based on the width of the chart:
![untitled-81](https://user-images.githubusercontent.com/230541/119577686-f4bd5380-bd6f-11eb-91b2-0ef5f768b6a1.png)

Fil:
Doesn't this mean we have two issues?

a) in some cases the default binning strategy creates too many bins

this might be addressed at the level of d3.thresholdScott, not based on the chart's width; or on the level of Plot.bin, based on the chart's width?

b) insets can "reverse" a rect and make it invisible

the formula is ([in rect.js](https://github.com/observablehq/plot/blob/1077eef0cb31e19216ad015dcc95474add6fdf27/src/marks/rect.js#L79)):
`.attr("width", i => Math.max(0, Math.abs(X2[i] - X1[i]) - this.insetLeft - this.insetRight))`

here maybe the minimum width should be more than 0, perhaps 0.5? It would not eliminate all cases, in particular if you have a white stroke and the default fill-then-stroke [paint order](https://developer.mozilla.org/fr/docs/Web/CSS/paint-order), but it would mean that a mark however small is never "zero-width".

[...]

PR #421, opened 2021-05-29
cap the default number of bins at 200

Fil:
(see #417)

(200 is working for almost any chart size; 100 would be a fine choice too)

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -171,7 +171,10 @@ function maybeBin(options) {
   return bin;
 }
 
-function maybeThresholds(thresholds = thresholdScott) {
+function maybeThresholds(thresholds) {
+  if (thresholds === undefined) {
+    return (...args) => Math.min(200, thresholdScott(...args));
+  }
   if (typeof thresholds === "string") {
     switch (thresholds.toLowerCase()) {
       case "freedman-diaconis": return thresholdFreedmanDiaconis;
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -172,9 +172,7 @@ function maybeBin(options) {
 }
 
 function maybeThresholds(thresholds) {
-  if (thresholds === undefined) {
-    return (...args) => Math.min(200, thresholdScott(...args));
-  }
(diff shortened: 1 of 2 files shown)


src/transforms/bin.js
```diff
@@ -171,7 +171,10 @@ function maybeBin(options) {
   return bin;
 }
 
-function maybeThresholds(thresholds = thresholdScott) {
+function maybeThresholds(thresholds) {
+  if (thresholds === undefined) {
+    return (...args) => Math.min(200, thresholdScott(...args));
```
mbostock, line 176: Let’s have a `thresholdDefault` static function (a global, but local to this file) rather than creating a closure each time it is needed. Also, would be possible to pass arts explicitly through to thresholdScott since we know which arguments that function expects.

src/transforms/bin.js
```diff
@@ -172,9 +172,7 @@ function maybeBin(options) {
 }
 
 function maybeThresholds(thresholds) {
-  if (thresholds === undefined) {
-    return (...args) => Math.min(200, thresholdScott(...args));
-  }
+  if (thresholds === undefined) return thresholdDefault;
```
mbostock, line 175: Can now be a default argument again.

mbostock (changes requested):
Related to the _facet_ option #450, if we have a default implementation strategy for an option, there should be a way to specify it explicitly rather than requiring you to pass undefined. So, let’s call the min(200, _scott_) strategy _auto_, so that you can specify it explicitly?

I was separately wondering if the max of 200 should apply to any of the other threshold strategies, and if so, should it be a separate option, say _thresholdsMax_ which defaults to 200? Or are we doing this only because Scott’s normal reference rule is prone to too many bins? It’s nice, though not obvious, that using _scott_ instead of _auto_ will disable the cap.

Lastly, I was wondering if the 200 should be dependent on the dimensions of the chart. That might suggest having a separate _thresholdsMax_ option, too, which perhaps could somehow default to a value derived from the chart dimensions. Though these dimensions are not currently made available to transforms so it’d require further plumbing to make a smart default, and in any case, I think 200 is almost certainly fine for now.

PR #422, opened 2021-05-29
keep non-zero rects visible

Fil:
When drawing a rect with an inset leave a sliver of matter (here 0.1px), unless the original width or height is 0.

See discussion in #417


mbostock:
I’d like to avoid this special case for now. If we were to keep this, here’s what I’d think about:

* Should there be a hard limit (of 0.1? of 0.5?), or should there be a gradual taper as the insets are larger than the desired width of the mark?
* When the full inset cannot be applied, should the effective insets be reduced equally from both sides, rather than just one side?
* Should this technique apply to other marks too, at the least bars?

The default bin cap #421 was a cleaner solution for the motivating example here, though I fully recognize that there are other cases where rects can disappear. (We addressed another special case for collapsed domains in #470.) I’d rather keep it simple for now and revisit if this problem crops up again in other forms.

Issue #92, opened 2021-01-11
Plot title (and signature, sources).


Fil:
as part of legends

ericemc3:
According to the usual rules of legibility, a chart should include an informative title, coming first (above the chart), possibly a subtitle, obligatorily a source, at the bottom of the chart, more optionally a copyright (which can also be arranged vertically, e.g. bottom right).

![image](https://user-images.githubusercontent.com/17407796/118007833-ce99bb80-b34c-11eb-9bf8-780cdab39b77.png)

ggplot2 and vega-lite allow you to specify these essential elements, it would be very nice if Plot could as well.

The SVG export of the Plot should allow you to get the whole graphic, with titles and sources.

As vega-lite proposes, specifying a title as an array allows it to be spread over several lines.

The height specified for the graphic should not take these elements into account, but only the central body of the graphic.

breck7:
Just another vote in support of title and subtitle support :). I would take @martgnz 's demo (https://github.com/observablehq/plot/pull/423) right now as it looks good enough and would unblock me. This is the one thing making it hard to build *ready-to-share* charts in Plot.

At Our World in Data the titles and subtitles are key, and the charts would be shared thousands of times less without them.

![coronavirus-data-explorer](https://user-images.githubusercontent.com/74692/178867508-5f44e78b-bfed-4969-8d85-dc246a140723.png)

kentr:
Have [`<title>` and `aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) been considered?

Fil:
We haven't considered aria-labelledby for now, because it needs an associated element with a given id, which makes it a bit complicated to handle if we want to be able to have several charts on a single page (they're not supposed to share ids). But we'll definitely consider it if there is a clear use case. Not that this property is usually visible only to screen readers, so I seems to fall under a different category, and might need a separate issue?

mbostock:
I think maybe @kentr’s comment is regarding tooltips—which we already support via the **title** option; also note that we also already support **ariaLabel** and **ariaDescription** options. This issue is primarily regarding a visual title element, i.e., a heading above the chart.

kentr:
@mbostock @Fil 

My comment was regarding a title element for the whole SVG (not tooltips).

I envision markup like this:

```html
<div id="svg-title">Daily new confirmed COVID-19 deaths per million people</div>
  <svg aria-labelledby="svg-title" ...>
    <g>
      ...
    </g>
[...]

PR #423, opened 2021-05-30
Title and subtitle support

martgnz:
Adding tentative HTML title and subtitle support (live example https://observablehq.com/d/90ecbc9920d7456e). A SVG approach would be better for chart portability (e.g. exporting to Illustrator) but it would mean implementing text wrapping and some lookups for text positioning that could be potentially troublesome. 

For titles, we add a `h1` element, and for subtitles a `h2`. We wrap everything in a figure tag, extending the support for captions.

I wonder how minimal we want to go with the styling and which elements we'd want to use. A `h1` and `h2` make sense for an isolated element but it might be too much for a chart that will be inside a document.

imagen

Closes https://github.com/observablehq/plot/issues/92.


Fil:
I think the use of figcaption to describe the sources is incorrect, per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption:
> [The] Figure Caption element represents a caption or legend describing the rest of the contents of its parent &lt;figure> element.

We'd also need to place signature and sources (or maybe only sources), in a semantically correct way.

CatChenal:
The proposed solution is a great improvement, but as @martgnz noted, fixing the style to `h1` and `h2` is very limiting.
Perhaps adding a "header channel" with a list for _header-part-name_: {_part-text_, _part-style_} would work better?

```javascript
plot.plot({
  header: [ title: {"This is a title in Observable plot", "h3"},  
            subtitle: { "Relative frequency of letters in the English language", "h4"}
            ],
  marks: [plot.barY(data, { y: "frequency", x: "letter", fill: "steelblue" })]
})
```

Note: I am a newbie in both javascript and Observable notebooks, but going through the **Observable Overview** tutorial, I came to the lack of title problem when I wanted to indicate which country the data was filtered for in the plot title so that if I wanted to download a picture of that specific plot, the information would not be lost.  
I did try this:  
```javascript
label: "price_in_usd: " + country
```
... but that's really not a title (& neither is a caption)!

Fil:
superseded by #1761 
Thank you for the contribution!

PR #425, opened 2021-05-31
expose scales

Fil:
The scales are exposed on the figure through a .scales property, a function that lazily computes scale option objects that can be used to reproduce the same scales in a different plot.

The function takes a scale name as parameter, and returns a scale definition that matches the given scale *as instanciated by the plot's data*. (If no scale name is given, it returns all the scales.)

With a lot of unit tests.

A (very messy) notebook with the build is https://observablehq.com/@fil/expose-scales-425

TODO:
- [x] return the scales as options
- [x] expose Plot.scale() to create a D3 scale from options.
- [x] README
- [x] Make sure we can build all the things we want (color legends, radius legends), "inside" the plot as well as outside
- [x] play a bit with it, and once we set on good legends, bake them in and expose them as Plot.colorLegend(), Plot.colorSwatches() or Plot.radiusLegend(). (Easy to extend via a Plot definition.)

Capture d’écran 2021-05-31 à 17 31 46

Capture d’écran 2021-05-31 à 19 24 27


Fil:
Same concept (legend drawn with Plot) for a radius legend
Capture d’écran 2021-05-31 à 21 26 57

Fil:
Now we can add `color: {legend: function(color) { … }}` to embed the resulting legend in the figure; also works for the r scale (and others if necessary).

Fil:
For anyone following this, https://github.com/observablehq/plot/pull/425/commits/228565b251e26ba4ced4f47604e4b885fca7c8ed exposes Plot.scale(*options*), allowing to create a D3 scale from Plot options.

README.md
```diff
@@ -200,6 +200,25 @@ Plot.plot({
...
+color.range // ["red", "blue"]
+```
+
+And, to reuse the scale in another plot:
+
+```js
+const plot1 = Plot.plot(…);
+
+Plot.plot({
+  color: plot:.scales.color
```
mbostock, line 216:
```suggestion
  color: plot1.scales.color
```

src/scales.js
```diff
@@ -105,3 +116,22 @@ function inferScaleType(key, channels, {type, domain, range}) {
...
+}
+
+function exposeScale({scale, label}) {
+  return {
+    domain: scale.domain(),
+    range: scale.range(),
+    ...scale.interpolate && {interpolate: scale.interpolate()},
+    ...label !== undefined && {label},
+    ...scale.type && {type: scale.type},
+    ...scale.clamp && scale.clamp() && {clamp: true}
```
mbostock, line 135:
```suggestion
    ...scale.clamp && {clamp: scale.clamp()}
```

Fil: When do we want to have the {clamp: false} option exposed, since it's the default?

mbostock: I think the exposed scale should materialize the defaults, not leave them unspecified, unless you want to specifically distinguish between the case where the user specified `clamp: false` versus when they didn’t specify the clamp option and it defaulted to false.

Fil: OK. My design choice was because I was aiming for the terser option, i.e. return the minimal options object that defines the scale. I don't think we can address the difference between explicit and default false at this point, since the option has been "folded in" the scale.

mbostock: Philosophically speaking, it is good to be minimal on input, but unambiguous on output, and hence we want to materialize the defaults.

src/plot.js
```diff
@@ -143,3 +140,27 @@ function autoHeight({y, fy, fx}) {
...
+function wrap(svg, scaleDescriptors, {caption} = {}) {
+  const scales = exposeScales(scaleDescriptors);
+  const legends = [];
+  for (let key in scaleDescriptors) {
+    const {legend} = scaleDescriptors[key];
+    if (typeof legend === "function") {
+      const l = legend(scales[key]);
+      if (l instanceof Node) legends.push(l);
+    }
+  }
```
mbostock, line 154: I think we should remove this legend feature to a separate PR.

src/plot.js
```diff
@@ -143,3 +140,18 @@ function autoHeight({y, fy, fx}) {
...
+
+// Wrap the plot in a figure with a caption, if desired.
+function wrap(svg, scaleDescriptors, {caption} = {}) {
+  const scales = exposeScales(scaleDescriptors);
+  if (caption == null) {
+    return svg.scales = scales, svg;
+  }
+  const figure = create("figure");
+  figure.append(() => svg);
+  if (caption != null) {
```
mbostock, line 152: This is guaranteed to be true here.

src/plot.js
```diff
@@ -143,3 +140,18 @@ function autoHeight({y, fy, fx}) {
...
 }
+
+// Wrap the plot in a figure with a caption, if desired.
+function wrap(svg, scaleDescriptors, {caption} = {}) {
+  const scales = exposeScales(scaleDescriptors);
+  if (caption == null) {
+    return svg.scales = scales, svg;
+  }
+  const figure = create("figure");
+  figure.append(() => svg);
```
mbostock, line 151: I don’t see much value in using a D3 selection here (versus using the DOM API as we did before), since this is always just a single element.

src/plot.js
```diff
@@ -143,3 +140,18 @@ function autoHeight({y, fy, fx}) {
...
+  if (caption != null) {
+    figure.append("figcaption")
+    .append(() => caption instanceof Node ? caption : document.createTextNode(caption));
+  }
+  return Object.assign(figure.node(), {scales});
```
mbostock, line 156: Similarly we can just `figure.scales = scales`.

src/scales.js
```diff
@@ -105,3 +116,22 @@ function inferScaleType(key, channels, {type, domain, range}) {
 function asOrdinalType(key) {
   return registry.get(key) === position ? "point" : "ordinal";
 }
+
+// prepare scales for exposure through the plot's scales() function
+export function exposeScales(scaleDescriptors) {
```
mbostock, line 121: It’d probably be nice if *plot*.scales was implemented as a lazy getter so that we don’t need to materialize these copies in the vast majority of cases where they won’t be needed. That can be done by using Object.defineProperty to define a getter on the return plot element, but we’ll also want to make that getter replace itself with the materialized value if it’s called more than once so that we don’t create a new object each time.

src/scales.js
```diff
@@ -105,3 +116,22 @@ function inferScaleType(key, channels, {type, domain, range}) {
...
+  );
+}
+
+function exposeScale({scale, label}) {
+  return {
+    domain: scale.domain(),
+    range: scale.range(),
+    ...scale.interpolate && {interpolate: scale.interpolate()},
+    ...label !== undefined && {label},
+    ...scale.type && {type: scale.type},
```
mbostock, line 134:
```suggestion
    label,
    type: scale.type,
```

Can we always expose these? Seems fine for *label* to be undefined if missing, and I’m not sure when scale.type would be missing, if ever.

src/scales.js
```diff
@@ -57,22 +57,33 @@ function autoScaleRound(scale) {
...
+    case "utc": scale = ScaleUtc(key, channels, options); break;
+    case "time": scale = ScaleTime(key, channels, options); break;
+    case "point": scale = ScalePoint(key, channels, options); break;
+    case "band": scale = ScaleBand(key, channels, options); break;
+    case "identity": scale = registry.get(key) === position ? ScaleIdentity(key, channels, options) : undefined; break;
+    case undefined: break;
     default: throw new Error(`unknown scale type: ${options.type}`);
   }
+  if (scale) {
+    if (options.legend) scale.legend = options.legend;
```
mbostock, line 79:
```suggestion
```

src/scales.js
Fil: I implemented each of scales[key] as an independent getter with a cache. First time I used Object.defineProperty :)

Fil:
Color legends in #432 !

Fil:
I have to work again on this, since some options are missing (like exponent: 0.3 for type="exp" scales), and we want all options including those that are not consumed by the scales, but could be used by the legend (like "columns" for swatches).

mbostock:
I plan on looking at this next. Exposing (materialized) scales will be great for debugging, too. I’m leaning towards just exposing plot.scales to start, and not the plot.scale method which returns a D3 scale instance. But mostly that’s so I can focus on getting this landed sooner and minimizing new API. 🙂

Fil:
deferring to #484

Issue #397, opened 2021-05-14
style the figure element?


Fil:
The use case (at the bottom [of this notebook](https://observablehq.com/@ee2dev/plot-exploration-how-to-customize-the-calendar-visualizat)) is a Plot that we want to make 2800px wide, and which should scroll horizontally in Observable.

My suggestion was to create a figure parent by adding a caption, then setting  max-width + scroll overflow styles to the parent, like so:

~~~js
Object.assign(
Plot.plot({ 
  width: 2800, height: 700, // sets the viewBox of the .plot svg
  caption: "", // wraps the plot in a figure element
  style: {maxWidth: "2800px"}, // overrides max-width: 100%; of the .plot default css
  … rest of the plot definition
}),
  { style: `max-width: ${width}px; overflow: scroll` } // overrides the figure/max-width style of observable and adds scrolling
)
~~~

it looks like there should be a better way at least to set styles on the figure element

EE2dev:
I find the default font-size fairly small, but I can adjust it with ...style: { fontSize: 14} .
However, just the text elements of the SVG is rerendered. It would be nice to also style the text of the legend.

Fil:
"Dark mode" is another use case for this: https://talk.observablehq.com/t/plot-legend-fill-color-when-embedded/7494

Hvass-Labs:
I also have the problem where the font-size of the legend doesn't change. See this example:

![Legend has wrong font-size](https://github.com/observablehq/plot/assets/13588114/aa424024-5e5b-4499-b94e-d7aa66360977)

Which was generated using the following code:

    var data = [
        { Date: new Date('1962-01-02'), Value: 0.0406, Name: 'Foo' },
        { Date: new Date('1962-01-03'), Value: 0.0403, Name: 'Bar' },
        { Date: new Date('1962-01-04'), Value: 0.0399, Name: 'Foo' },
        { Date: new Date('1962-01-05'), Value: 0.0402, Name: 'Bar' },
        { Date: new Date('1962-01-06'), Value: 0.0402, Name: 'Foo' },
        { Date: new Date('1962-01-07'), Value: 0.0403, Name: 'Bar' },
        { Date: new Date('1962-01-08'), Value: 0.0403, Name: 'Foo' },
        { Date: new Date('1962-01-09'), Value: 0.0405, Name: 'Bar' },
        { Date: new Date('1962-01-10'), Value: 0.0407, Name: 'Foo' },
[...]

PR #432, opened 2021-06-10
color, opacity and radius legends

### LINKED ISSUE #23: Color legend.

### LINKED ISSUE #236: radius legend?

Fil:
sequel of #425

build at https://observablehq.com/d/c324a9a466e66329

todo:
- [x] fix API (1 options object)
- [x] examples
- [x] remove dependency on htl
- [x] fix the figure so it gets the same width as the plot
- [x] expose and document Plot.colorLegend
- [x] continuous date color legend

closes #23 
closes #236

related: #397 #92


mbostock:
The color legend should inherit the scale’s automatic label, no? E.g., here both the _x_-axis and color legend should be labeled with “random”, without you needing to specify a label explicitly.

```js
Plot.plot({
  color: {
    legend: true,
    width: 120,
    nice: true
  },
  marks: [
    Plot.dot({length: 100}, {
      x: Object.assign(() => Math.random(), {label: "random"}),
      fill: Object.assign(() => Math.random(), {label: "random"})
    })
  ]
})
```

src/legends/color.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+  return legend(scale(descriptor), {
+    title: descriptor.label,...options
+  });
+}
+
+function legend(color, {
+  title,
+  tickSize = 6,
+  width = 320,
```
mbostock, line 19:
```suggestion
  width = 240, 
```

320 feels a little wide as a default. Should probably also be capped at the width of the chart in the legend: true case.

src/scales.js
```diff
@@ -116,25 +119,34 @@ function asOrdinalType(key) {
...
+  const scales = {};
+  const legends = [];
   for (const key in scaleDescriptors) {
     let cache;
     Object.defineProperty(scales, key, {
       enumerable: true,
       get: () => cache = cache || exposeScale(scaleDescriptors[key])
     });
+    const options = scaleDescriptors[key].options;
+    let {legend} = options;
```
mbostock, line 132: This is scoping the legend options under a legend object, which is different from how we handle axis options (which are in the same namespace as scale options, i.e., not under an axis object). I wonder whether we should flatten the legend options as well; certainly the ticks and label options seem the same as an axis, so no reason to make them different.

src/scales.js
```diff
@@ -116,25 +119,34 @@ function asOrdinalType(key) {
   return registry.get(key) === position ? "point" : "ordinal";
 }
 
-export function exposeScales(figure, scaleDescriptors) {
-  const scales = figure.scales = {};
+export function exposeScales(scaleDescriptors) {
```
mbostock, line 122: The name “exposeScales” no longer seems accurate. Can we separate the two functions, keeping exposeScales as it was and having a separate method for computing the legends?

src/legends/swatches.js
```diff
@@ -0,0 +1,71 @@
...
+  max-width: calc(100% - ${+swatchWidth}px - 0.5em);
+}
+
+.${id}-swatch {
+  width: ${+swatchWidth}px;
+  height: ${+swatchHeight}px;
+  margin: 0 0.5em 0 0;
+}
+
+  </style>
```
mbostock, line 37: We can move all of this to style.css and then just use CSS variables for the variables (the swatch width, height, and color). Then we won’t need uid or inline styles (other than to define the CSS variables), either.

Fil:
> color legend should inherit the scale’s automatic label

Yes that's the intent. I swear it was working at some point :)

Thank you for the review! I should have marked this as a Draft because there's quite a lot of screws to tighten.

I think this shows that the "scale options" object (returned as plot.scales.color etc) should contain all the options that we passed and that were not consumed when building the scale.

Also, I want more unit tests—for example I realize that calling Plot.scale on a diverging scale will crash if we don't tell it that the key is "color".

Fil:
deferring to #484

Fil:
see https://github.com/observablehq/plot/pull/665

PR #437, opened 2021-06-17
pass the bins' extent (x1, x2, y1, y2) to the reducer

### LINKED ISSUE #435: Pass the bin extent, and group keys, to reducers?
A typical example would be to have the title of each bar show the date extent in https://observablehq.com/@fil/plot-stargazer-history

~~~js
    Plot.rectY(
      data,
      Plot.binX(
        {
          y: "count",
          title: (bin) => `${bin.x1}…${bin.x2}` // ⇐ something like this.
        },
        {
          x: "date",
          fill: "brown",
          thresholds: 100
        }
      )
    ),
~~~

Fil:
closes #435

not sure about the way I've coded "around" the take() function…


mbostock (commented):
Some possible TODO’s:

1. The reducers might want group keys, too (even when binning).
2. Avoid mutation and `"x1" in I`  test, somehow (only pass if reducer needs it?). 
3. Use this technique to implement the default sort-by-key for groups?

These issues are a little tricky so not sure when we can work on this.

src/transforms/bin.js
```diff
@@ -132,8 +133,8 @@ function binn(
       maybeSort(groupFacets, sort, reverse);
       return {data: groupData, facets: groupFacets};
     }),
-    ...BX1 ? {x1: BX1, x2: BX2, x: mid(BX1, BX2)} : {x},
-    ...BY1 ? {y1: BY1, y2: BY2, y: mid(BY1, BY2)} : {y},
+    ...BX1 && !hasOutput(outputs, "x") ? {x1: BX1, x2: BX2, x: mid(BX1, BX2)} : {x},
+    ...BY1 && !hasOutput(outputs, "y") ? {y1: BY1, y2: BY2, y: mid(BY1, BY2)} : {y},
```
mbostock, line 137: The idea here is that if you specified your own custom output for *x* (_e.g._, x: null), then we’ll suppress the *x1* and *x2* outputs, too.

src/transforms/group.js
```diff
@@ -159,8 +163,10 @@ export function maybeEvaluator(name, reduce, inputs) {
         context = reducer.reduce(I, V);
       }
     },
-    reduce(I) {
-      return reducer.reduce(I, V, context);
+    reduce(I, extent) {
+      return reducer.scope == null
+        ? reducer.reduce(I, V, extent)
+        : reducer.reduce(I, V, context, extent);
```
mbostock, line 169: Most reducers aren’t scoped, so it felt simpler to pass the extent directly as the third argument.

mbostock (approved):
I avoided the mutation (Object.assign). Instead, the bin extent is passed as an additional argument to the reducer.

I also added a feature to this PR which lets you suppress the built-in outputs of the bin transform. For example x: null in the outputs will suppress the *x*, *x1* and *x2* channels. This simplifies the binned stargazers example where want to compute the median across bins.

@Fil Can you review these changes and if it looks good, merge?

mbostock:
(I’m punting on the exposing the group keys to reducers, too. I think there’s a reasonable way to do that by extending the extent argument `{[gk]: …, x1, x2, y1, y2}`, but feels like something we can do in the future.)

Fil:
Merging this. Let’s explore "exposing the group keys to reducers" in  https://github.com/observablehq/plot/pull/541

PR #438, opened 2021-06-18
create a non-empty bar if the domain is collapsed

Fil:
(e.g. binning of a one-element data)

degenerate case example given by @severo in https://github.com/observablehq/plot/issues/417

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -225,6 +225,7 @@ function bincumset([bin], j, bins) {
 }
 
 function binfilter([{x0, x1}, set]) {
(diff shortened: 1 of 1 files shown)


mbostock (changes requested):
I would like to fix this issue, but I think we’ll need a different fix. One issue here is that if you have a log _x_-scale (which admittedly is a little weird when binning, but is theoretically valid), then adding the fixed offset to _x0_ = _x1_ could mean that your _x_-domain will cross zero. I think we may want to fix this at the rendering layer (i.e., marks) rather than by changing the semantics of the data transform.

Fil:
superseded by #470 (& #474)

PR #439, opened 2021-06-20
the bin transform should not consume {reverse}

Fil:
since it doesn't make use of it… instead pass it up so we can stack the resulting bins in the desired order.
solves https://stackoverflow.com/questions/68056843/in-observable-plot-how-to-sort-order-the-stack-from-a-bin-transform/68057660

build and test: https://observablehq.com/@fil/stack-bins-ordered-439


mbostock (commented):
My first thought was “shouldn’t the bin transform pass through all options that it doesn’t understand?” My second thought was “doesn’t it do that already?” And yes, it does; but the _reverse_ option is consumed by maybeTransform:

https://github.com/observablehq/plot/blob/068f8d8125a910ee1a528c8b299a79c27ca1a9a5/src/mark.js#L239

So, isn’t the problem here that we in effect want the _reverse_ transform to apply _after_ the bins have been created, rather than _before_ (where it has no effect)? I guess it’s okay to change the behavior for the bin transform specially, but it seems a little inconsistent with how the basic transforms work. What do you think?

Fil:
Same analysis — I didn't see a better way to fix this than by having the bin transform explicitly "save it for later", since it makes no sense to order the elements in the bins. But I agree it goes a little against the genericity of the transforms composition.

mbostock:
It raises the question: should the _sort_ transform then similarly apply after the bins have been constructed, rather than before?

Fil:
There's one case where *sort* must be done before stacking: when we compute the wiggle offset (which minimizes the total movement), the data should be sorted according to the stack's "location" dimension (*x* if we're using stackY).
https://github.com/observablehq/plot/pull/177#issuecomment-789062973 
However in that use case: 1. it doesn't matter if it's done before or after binning 2. it doesn't matter if the data are sorted or sorted reverse—they just have to be "aligned".

If we need to be explicit about where {reverse} is applied (or want to apply it several times with different targets), there's always the possibility of using a combination of spread (as in [this example](https://stackoverflow.com/questions/68056843/in-observable-plot-how-to-sort-order-the-stack-from-a-bin-transform/68057660)) and Plot.transform.

The full example would become something like:

~~~js
Plot.rectY(
  data,
  Plot.stackY({
    ...Plot.binX(
      { y: "count", fill: "first", title: (d) => JSON.stringify(d[0]) },
      Plot.transform({
        x: "body_mass",
        fill: "species",
        z: "species",
        // order: "value" // ***
        sort: "sex",
        reverse: true // applies on "sort", consumed by Plot.transform hence not passed up by Plot.bin
      })
    ),
    order: "value", // alternatively, could be defined in ***
    reverse: true // applies on "order"
  })
).plot()
~~~


Also related https://github.com/observablehq/plot/issues/297

mbostock (changes requested):
The main point I’m trying to make here is that _sort_ and _reverse_ should always travel together: you need both to do descending sort (at least, for convenience). Otherwise I’m okay making an exception for the bin transform to not consume these options, and instead pass them through (as here, by removing them from _options_ so they don’t get consumed by maybeTransform).

Should we make the same exception for the group transform?

mbostock:
This is what I meant

https://github.com/observablehq/plot/compare/fil/fix-bin-reverse...mbostock/fix-bin-sort

but thinking about it some more, I’m not sure it makes sense: if the bin transform is composed with the stack transform, then yes, the _reverse_ option will be consumed. But the _sort_ option will not be consumed—it’ll just be ignored—since that _sort_ option is only consumed by maybeTransform (which we’re talking about bypassing). So I think maybe the root of the confusion here is that there are two different _reverse_ options, one for the stack transform, and one for the basic transform:

https://github.com/observablehq/plot/blob/7e45462fc1e8026ba9daebbae4f7371da0ad6ceb/src/transforms/stack.js#L53

https://github.com/observablehq/plot/blob/7e45462fc1e8026ba9daebbae4f7371da0ad6ceb/src/mark.js#L230

Here we want to pass through the _reverse_ option to the stack transform, but not the _reverse_ option to the basic transform (which should travel together with the _sort_ option, unlike the stack transform, which uses an _order_ option).

So, I’m thinking that being explicit about who the _reverse_ option is intended for is actually desirable here. It does have a meaning (together with the _sort_ option) for the bin transform, albeit a subtle one affecting the order of the data within the bin. That would suggest doing `stackY({..binX(…), reverse: …})` if you do want to pass the _reverse_ option explicitly to the stack transform.

But we still don’t have a way to apply sort and reverse _after_ binning, because the basic transform always composes transforms in a fixed order:

https://github.com/observablehq/plot/blob/7e45462fc1e8026ba9daebbae4f7371da0ad6ceb/src/mark.js#L235-L237

This makes me think that maybe the basic transforms should be pulled out into functions, so you can say `binX(sort(order, options))` if you want to sort _before_ binning and `sort(order, binX(options))` if you want to sort _after_ binning. And similarly `filter(test, options)` to replace the _filter_ option and `reverse(options)` to replace the _reverse_ option. We should ideally maintain backwards-compatibility even if we deprecate the existing options, though.

mbostock:
Closing in favor of #494.

PR #441, opened 2021-06-23
create the test plot output if different from the expected value. 

Fil:
useful to compare changes

### DIFF
--- a/test/plot.js
+++ b/test/plot.js
@@ -40,7 +40,7 @@ import * as plots from "./plots/index.js";
         test.ok(actual === expected, `${name} must match snapshot`);
         if (actual !== expected) {
           const outfile = path.resolve("./test/output", path.basename(name, ".js") + "-changed." + ext);
-          fs.writeFile(outfile, actual, "utf8");
+          await fs.writeFile(outfile, actual, "utf8");
         }
       } finally {
         delete global.document;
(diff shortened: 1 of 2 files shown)


test/plot.js
```diff
@@ -38,6 +38,10 @@ import * as plots from "./plots/index.js";
         }
 
         test.ok(actual === expected, `${name} must match snapshot`);
+        if (actual !== expected) {
+          const outfile = path.resolve("./test/output", path.basename(name, ".js") + "-changed." + ext);
+          fs.writeFile(outfile, actual, "utf8");
```
mbostock, line 43: Probably want await here.

mbostock (changes requested):
If we do this, we should add the appropriate entry to the gitignore so these files don’t get accidentally committed to the repo.

test/plot.js
```diff
@@ -38,6 +38,10 @@ import * as plots from "./plots/index.js";
         }
 
         test.ok(actual === expected, `${name} must match snapshot`);
+        if (actual !== expected) {
+          const outfile = path.resolve("./test/output", path.basename(name, ".js") + "-changed." + ext);
```
mbostock, line 42: It might be nice to console.log here the outfile so that it’s obvious to folks where they can find the generated output.

mbostock (approved):

PR #444, opened 2021-06-29
center the silhouette and wiggle offsets

### LINKED ISSUE #424: Centering "silhouette" and "wiggle" stacks
After #285 the wiggle and silhouette offsets start at 0; when faceting, this results in asymmetric stacks.

The following transform makes them centered around 0:

~~~js
function centerStackX(options) {
  const transform = options.transform;
  options.transform = function(data, facets) {
    ({data, facets} = transform(data, facets));
    options.x1 = options.x1.transform();
    options.x2 = options.x2.transform();
    for (const I of facets) {
      const shift = (d3.min(I, i => options.x1[i]) - d3.max(I, i => options.x2[i])) / 2;
      for (const i of I) {
        options.x1[i] += shift;
        options.x2[i] += shift;
      }
    }
    return {data, facets}
  }  
  return options;
}
~~~

I'm thinking it should probably be the default (?)

cc @tophtucker

Fil:
Capture d’écran 2021-06-29 à 12 32 05

closes #424


mbostock:
I don’t like the negative values in the y-axis, though; with a streamgraph, the y-axis represents length rather than position. Can we avoid negative values while still centering across facets?

Fil:
OK!

src/transforms/stack.js
```diff
@@ -81,6 +81,9 @@ function stack(x, y = () => 1, ky, {offset, order, reverse, ...options} = {}) {
         }
         if (offset) offset(stacks, Y1, Y2, Z);
       }
+      if (offset === offsetWiggle || offset === offsetSilhouette) {
+        offsetCenterFacets(Y1, Y2, facets);
+      }
```
mbostock, line 86: It’s a bit of a bummer that we special-case the logic for these two offset implementations. Philosophically, none of the built-in methods (e.g., for offsets and orders when stacking, or reducers when binning or grouping) should have “privileged” abilities that go beyond what can be done with a passed-in function. The idea being as much as possible we want people to be able to pass in arbitrary functions that can do as much as the built-in ones.

So, I wonder if that would be possible here? One idea that comes to mind is a separate “offsetCenter” boolean option that defaults to true iff the *offset* option is wiggle or silhouette. But, maybe there’s another way of doing it.

Fil: The offset logic needs either:
a) each of the stacks, individually, for expand and silhouette
b) all the stacks in one facet, for wiggle
The center logic needs c) all the stacks in all the facets.

Since c > b > a, we can create a reference of all the stacks in all facets and pass them at once to the offset function. This makes writing an offset function a tiny bit more complex (you receive an array of arrays of stacks).

I'm favoring this over having another option.

src/transforms/stack.js
```diff
@@ -112,51 +114,59 @@ function extent(stack, Y2) {
...
+      const s1 = sum(Fi);
+      if (s1) y -= sum(Fi, (d, i) => (Df[i] / 2 + Cf1[i]) * d) / s1;
     }
-    const s1 = sum(Fi);
-    if (s1) y -= sum(Fi, (d, i) => (Df[i] / 2 + Cf1[i]) * d) / s1;
+    offsetZero(stacks, Y1, Y2);
   }
-  offsetZero(stacks, Y1, Y2);
+  offsetCenterFacets(Y1, Y2, allstacks.map(d => d.flat()));
```
mbostock, line 169: Can we avoid this flatMap by using the same signature? `offsetCenterFacets(allstacks, Y1, Y2)`

PR #447, opened 2021-06-30
default outputs for bins

### LINKED ISSUE #363: Defaults for bin counts
Filing this issue at Mike's suggestion in the d3 slack. I ran into some surprising behavior using `Plot.binX` without specifying any arguments (https://observablehq.com/d/cf5ba27c74172e3e) and Mike suggested adding a default to handle this case.

Right now this produces the expected plot:
```js
Plot.rectY(data, Plot.binX({ y: 'count' })).plot()
```

While this produces an unexpected (not empty, but also not correct) plot:
```js
Plot.rectY(data, Plot.binX()).plot()
```

One question would be what channels to default to `count` – in this case I wanted the `y` channel, but I'm not sure if that's always the right default.

Fil:
binX (y:count), binY (x: count), and bin (<del>r: count,</del> fill: count)
(the rationale for the 2d bin is that they can be used indifferently with Plot.rect or Plot.dot)

closes #363

Capture d’écran 2021-06-30 à 21 51 18
Capture d’écran 2021-06-30 à 23 31 21


src/transforms/bin.js
```diff
@@ -4,23 +4,23 @@ import {offset} from "../style.js";
...
+export function binY(outputs = {x: "count"}, {inset, insetTop, insetBottom, ...options} = {}) {
   let {x, y} = options;
   y = maybeBinValue(y, options, identity);
   ([insetTop, insetBottom] = maybeInset(inset, insetTop, insetBottom));
   return binn(null, y, x, null, outputs, {inset, insetTop, insetBottom, ...options});
 }
 
 // Group on {z, fill, stroke}, then bin on x and y.
-export function bin(outputs, {inset, insetTop, insetRight, insetBottom, insetLeft, ...options} = {}) {
+export function bin(outputs = {r: "count", fill: "count"}, {inset, insetTop, insetRight, insetBottom, insetLeft, ...options} = {}) {
```
mbostock, line 23:
```suggestion
export function bin(outputs = {fill: "count"}, {inset, insetTop, insetRight, insetBottom, insetLeft, ...options} = {}) {
```

Are you sure you wouldn’t prefer only providing a *fill* output by default, since it works with both rect and dot and avoids a redundant encoding in the case of dot? I’d prefer `{r: "count"}` for the dot case, but I don’t see how we’d infer how it will be consumed.

mbostock (commented):
I was wondering for a moment, if only one argument is specified, should it be interpreted as _outputs_ or _options_? But it’s already the case that we interpret it as _outputs_, and I think it makes sense to preserve that behavior, so, this seems good to me.

src/transforms/bin.js
Fil: the simpler the better

test/plots/random-bins-xy.js
```diff
@@ -0,0 +1,8 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export default async function() {
+  const random = d3.randomNormal.source(d3.randomLcg(42))(10, 3);
+  const data = Array.from({length: 500}, () => [random(), random()]);
+  return Plot.dot(data, Plot.bin()).plot();
```
mbostock, line 7: Plot.rect is a slightly better choice here, but this is a fine test.

mbostock (approved):
👍👍

PR #450, opened 2021-07-08
explicit facet option in a mark

Fil:
you can enable or disable faceting explicitly on a mark with the facet option, which accepts the following values:

* "auto" (default) - enable faceting if the mark's data is === to the facet data
* null (or *false*) - disable faceting for this mark
* "include" (or *true*) - enable faceting for this mark
* *"exclude"* - enable exclusion faceting for this mark (each facet receives all the data except those of the facet’s subset) 

See https://observablehq.com/@fil/plot-explicit-facet-option


src/mark.js
```diff
@@ -7,9 +7,10 @@ const TypedArray = Object.getPrototypeOf(Uint8Array);
 const objectToString = Object.prototype.toString;
 
 export class Mark {
-  constructor(data, channels = [], options = {}) {
+  constructor(data, channels = [], {facet, ...options} = {}) {
     const names = new Set();
     this.data = data;
+    this.facet = facet;
```
mbostock, line 13: We should do some validation and coercion here.

I also think we can unify the types a little more. Currently:

* `"exclude"` is a string,
* `null` and `undefined` are equivalent, and
* `true` and `false` are booleans.

If we want to be more analogous to the *scale*.axis option, perhaps:

* `"exclude"`
* `"include"`, and `true` is shorthand for `"include"`
* `undefined` is `"include"` if data is strict equal, and otherwise `null` to disable faceting
* `null` or any other falsey value is mapped to `null` to disable faceting

Any other truthy value, after being coerced to a string, would throw an error. We’d use the `keyword` helper:

https://github.com/observablehq/plot/blob/068f8d8125a910ee1a528c8b299a79c27ca1a9a5/src/mark.js#L126

src/facet.js
```diff
@@ -38,24 +38,29 @@ class Facet extends Mark {
     }
     for (let i = 0; i < this.marks.length; ++i) {
       const mark = this.marks[i];
-      const markFacets = mark.data === this.data ? facetsIndex : undefined;
-      const {index, channels} = mark.initialize(markFacets);
+      const markFacets = mark.facet == null ? mark.data === this.data ? facetsIndex : undefined
+        : mark.facet === "exclude" ? facetsIndex.map(facet => Uint32Array.from(difference(index, facet)))
+        : mark.facet === true ? facetsIndex
+        : undefined;
+      if (mark.facet && range(mark.data).length !== index.length) {
```
mbostock, line 45: We should use the same test we use below:

```suggestion
      if (markFacets && range(mark.data).length !== index.length) {
```

mbostock (changes requested):
This is great. I was thinking we should do something like this. My feedback is mostly on how to make the *facet* option more consistent with other Plot options.

src/facet.js
```diff
@@ -38,9 +38,10 @@ class Facet extends Mark {
     }
     for (let i = 0; i < this.marks.length; ++i) {
       const mark = this.marks[i];
-      const markFacets = mark.facet == null ? mark.data === this.data ? facetsIndex : undefined
-        : mark.facet === "exclude" ? facetsIndex.map(facet => Uint32Array.from(difference(index, facet)))
-        : mark.facet === true ? facetsIndex
+      const facet = mark.facet && keyword(mark.facet, "facet", ["include", "exclude"]);
```
mbostock, line 41: This needs to happen in the mark constructor (as I said previously), not here in initialize.

src/facet.js
```diff
@@ -38,9 +38,10 @@ class Facet extends Mark {
...
+      const markFacets = facet === undefined ? mark.data === this.data ? facetsIndex : undefined
+        : facet === "include" ? facetsIndex
+        : facet === "exclude" ? facetsIndex.map(f => Uint32Array.from(difference(index, f)))
```
mbostock, line 44: I think we probably shouldn’t bother with this optimization, but: this could be extracted out of the for loop using a lazily-constructed variable so that it can be shared across multiple excluded marks.

mbostock (changes requested):

src/facet.js
Fil: yes I understood your comment just after pushing the previous commit 8-)
I think it's worth adding an "auto" option value (rather than undefined?), which makes the code easier to read, and would maybe clarify the documentation

mbostock: Yea, I wasn’t sure whether we needed `"auto"` in addition to `undefined`, but sure, if you want.

src/facet.js
Fil: it's cheap so I added it too

src/mark.js
```diff
@@ -7,10 +7,11 @@ const TypedArray = Object.getPrototypeOf(Uint8Array);
 const objectToString = Object.prototype.toString;
 
 export class Mark {
-  constructor(data, channels = [], {facet, ...options} = {}) {
+  constructor(data, channels = [], {facet = "auto", ...options} = {}) {
     const names = new Set();
     this.data = data;
-    this.facet = facet;
+    this.facet = facet && keyword(facet, "facet", ["auto", "include", "exclude"]);
```
mbostock, line 13: Almost. 🙂 

```suggestion
    this.facet = facet ? keyword(facet, "facet", ["auto", "include", "exclude"]) : null;
```

src/facet.js
```diff
@@ -36,26 +36,33 @@ class Facet extends Mark {
...
-      const markFacets = mark.data === this.data ? facetsIndex : undefined;
-      const {index, channels} = mark.initialize(markFacets);
+      const {facet} = mark;
+      const markFacets = facet === "auto" ? mark.data === this.data ? facetsIndex : undefined
+        : facet === "include" ? facetsIndex
+        : facet === "exclude" ? facetsExclude || (facetsExclude = facetsIndex.map(f => Uint32Array.from(difference(index, f))))
+        : undefined;
+      if (markFacets && range(mark.data).length !== index.length) {
+        throw new Error("faceted mark data must match facet data length");
+      }
```
mbostock, line 49: `range(mark.data).length` is inefficient test in the common case where mark.data is already an array. Also, this test isn’t strictly related to the _facet_ option, and might be overly strict. Maybe we shouldn’t add this?

```suggestion
```

mbostock (changes requested):

src/facet.js
Fil: My idea here was to raise a flag before users tried to pass data they had reduced themselves and wondered why it didn't magically follow facets, but the test isn't fool-proof.
Shouldn't we have an all-purpose function for length? Or is it simply `arrayify(data).length`?

mbostock (approved):
Nice work!

zanarmstrong:
Wow!

PR #451, opened 2021-07-09
Use columnar data

### LINKED ISSUE #449: Arquero shorthand
Currently to use Arquero, you say:

```js
Plot.plot({
  marks: [
    Plot.tickX(table.indices(), {y: table.column("age"), x: table.column("percent")})
  ]
})
```

But with a little magic, we could support shorthand for Arquero, treating it like an array of objects:

```js
Plot.plot({
  marks: [
    Plot.tickX(table, {y: "age", x: "percent"})
  ]
})
```

Specifically:

1. If _data_ is an Arquero table, then data is _data_.indices().
2. If _data_ is an Arquero table, then field names are interpreted as _data_.column(_field_).

Fil:
optimizes accesses to arquero tables through *Table*.column(fieldname)

closes #449

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -54,25 +54,22 @@ export class Mark {
 
 // TODO Type coercion?
 function Channel(data, {scale, type, value}) {
-  return {
-    scale,
-    type,
-    value: valueof(data, value),
-    label: value ? value.label : undefined
-  };
+  const label = value ? value.label : undefined;
+  value = value._column !== undefined && isColumnar(data) ? data.column(value._column) : valueof(data, value);
+  return {scale, type, value, label};
 }
 
 // This allows transforms to behave equivalently to channels.
 export function valueof(data, value, type) {
   const array = type === undefined ? Array : type;
-  return typeof value === "string" ? array.from(data, field(value))
+  return typeof value === "string" ? isColumnar(data) ? data.column(value) : array.from(data, field(value))
     : typeof value === "function" ? array.from(data, value)
     : typeof value === "number" || value instanceof Date ? array.from(data, constant(value))
     : value && typeof value.transform === "function" ? arrayify(value.transform(data), type)
     : arrayify(value, type); // preserve undefined type
 }
 
-export const field = label => Object.assign(d => d[label], {label});
+export const field = label => Object.assign(d => d[label], {label, _column: label});
 export const indexOf = (d, i) => i;
 export const identity = {transform: d => d};
 export const zero = () => 0;
@@ -135,7 +132,7 @@ export function keyword(input, name, allowed) {
 // the specified data is null or undefined, returns the value as-is.
 export function arrayify(data, type) {
   return data == null ? data : (type === undefined
-    ? (data instanceof Array || data instanceof TypedArray) ? data : Array.from(data)
+    ? (data instanceof Array || data instanceof TypedArray || isColumnar(data)) ? data : Array.from(data)
     : (data instanceof type ? data : type.from(data)));
 }
 
@@ -182,7 +179,7 @@ export function titleGroup(L) {
 
 // Returns a Uint32Array with elements [0, 1, 2, … data.length - 1].
 export function range(data) {
-  return Uint32Array.from(data, indexOf);
+  return Uint32Array.from(data.length ? {length: data.length} : data, indexOf);
 }
 
 // Returns an array [values[index[0]], values[index[1]], …].
@@ -341,4 +338,8 @@ export function isTemporal(values) {
     if (value == null) continue;
     return value instanceof Date;
   }
+}
+
+function isColumnar(data) {
+  return typeof data.column === "function";
 }
--- a/test/columnar-test.js
+++ b/test/columnar-test.js
@@ -1,0 +1,28 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+import tape from "tape-await";
+
+tape("Plot uses the data.column accessor", test => {
+  const columnChecks = [];
+  const columnX = ["A", "B", "C", "D"];
+  const columnY = [1, 1, 2];
+  const data = {
+    length: 5,
+    column: function(field) {
+      columnChecks.push(field);
+      switch(field) {
+        case "x": return columnX;
+        case "y": return columnY;
+      } 
+    },
+    *[Symbol.iterator] () {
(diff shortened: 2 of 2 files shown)


Fil:
OK this doesn't fully work, in particular we can't use these columns for the title option. The reason is that Plot's code for the title option checks if the title channel L value for each index *i* is nonempty, by calling `nonempty(L[i])`. But arquero's columns do not have a getter for `L[i]`, only `L.get(i)`, so `L[i]` is always `undefined`.

There might be ways to fix this from Plot, but I think this is a case where we would prefer to patch arquero and add a getter to the column prototype?

Fil:
Here's a way to accomodate this case in Plot:

~~~patch
diff --git a/src/mark.js b/src/mark.js
index de42510..bf6ecdc 100644
--- a/src/mark.js
+++ b/src/mark.js
@@ -55,20 +55,29 @@ export class Mark {
 // TODO Type coercion?
 function Channel(data, {scale, type, value}) {
   const label = value ? value.label : undefined;
-  value = value._column !== undefined && isColumnar(data) ? data.column(value._column) : valueof(data, value);
+  value = valueof(data, value);
   return {scale, type, value, label};
 }

 // This allows transforms to behave equivalently to channels.
   const array = type === undefined ? Array : type;
-  return typeof value === "string" ? isColumnar(data) ? data.column(value) : array.from(data, field(value))
-    : typeof value === "function" ? array.from(data, value)
+  if (isColumnar(data) && typeof value === "string") value = field(value);
+  return typeof value === "string" ? array.from(data, field(value))
+    : typeof value === "function" ? column(data, value, array)
     : typeof value === "number" || value instanceof Date ? array.from(data, constant(value))
     : value && typeof value.transform === "function" ? arrayify(value.transform(data), type)
     : arrayify(value, type); // preserve undefined type
 }

+function column(data, value, array) {
+  if (value._column !== undefined && isColumnar(data)) {
+    value = data.column(value._column);
+    return "data" in value ? value.data : value;
+  }
+  return array.from(data, value);
+}
+
 export const field = label => Object.assign(d => d[label], {label, _column: label});
 export const indexOf = (d, i) => i;
 export const identity = {transform: d => d};
~~~

Another approach could be to create a Proxy if the column object has a get method, but it's starting to become quite a complication. (and I still feel this should be done in arquero.)

Fil:
ok it was not that much of a complication:
`return "get" in data ? new Proxy(data, {get: (_, i) => i in _ ? _[i] : _.get(i)}) : data;`

src/mark.js
```diff
@@ -54,25 +54,22 @@ export class Mark {
...
-  return typeof value === "string" ? array.from(data, field(value))
+  return typeof value === "string" ? isColumnar(data) ? column(data, value) : array.from(data, field(value))
     : typeof value === "function" ? array.from(data, value)
     : typeof value === "number" || value instanceof Date ? array.from(data, constant(value))
     : value && typeof value.transform === "function" ? arrayify(value.transform(data), type)
     : arrayify(value, type); // preserve undefined type
 }
 
-export const field = label => Object.assign(d => d[label], {label});
+export const field = label => Object.assign(d => d[label], {label, _column: label});
```
mbostock, line 72: What’s up with this `_column` property? I don’t think we want to introduce a hidden property.

Fil: The unit tests checked that channel.value was a function, so I wanted to respect that, and used this hidden property to note that it also could be a "data.column". But I'm cool with keeping channel.value as a string, it makes everything much simpler. It also breaks 61 unit tests ;-)

Fil: I think this is the right call. Will fix the tests later.

mbostock (changes requested):
Thanks for this, Fil. It’s a good effort, but it raises subtle concerns; I’m not yet comfortable landing this. Unfortunately I don’t have an actionable recommendation yet on how we should proceed, but I want to share my thoughts so far.

My main concern is that Arquero support is currently a leaky abstraction. It could have a downstream ripple effect because Arquero data goes through different code paths. And this will be a particular problem for custom marks and transforms.

For example, arrayify(*data*) no longer converts the Arquero Table instance into an array; arrayify can now return something that is *not* an array. Plot transforms are passed data, which means that they will now potentially receive an Arquero table instance, too, which could cause them to break. For example the stack transform expects there to be a *data*.length property, but Arquero only implements *table*.size.

https://github.com/observablehq/plot/blob/20d118c5ac44ff036cf617eaa3702f848170c99c/src/transforms/stack.js#L66

This problem is masked somewhat in this PR because `valueof` and `range` also support *data* being an Arquero Table, but we can’t control arbitrary code in transforms, so this makes Arquero support a leaky abstraction.

In the motivating issue #449, note that I wanted the data to be *table*.indices(), not the Arquero table itself. This means that the data can be a valid array. However, if the data is an index rather than the Arquero table, it’ll make it harder to define channels as functions since those functions will now be passed an integer index, not a row object.

I also wonder: if the data is an Arquero table, then could channels be defined as [table expressions](https://uwdata.github.io/arquero/api/expressions.html#table) to avoid instantiating row objects (i.e., for derived columns rather than for pre-existing named columns)?

I’m going to think about this some more.

Fil:
At least it helps clarify a few things.

- Maybe what we need from arquero (or could build as a wrapper around the arquero table format, with Proxy) is to expose a more "array-like" API: accessing data with [i], having a length, and possibly a column API that returns an object with the same "array-like" characteristics.
- The n = data.length line in the stack transform should probably be Y.length (independently of this particular use case).
- The removal of field() seems a good change, also independently of this use case.

Fil:
I've added details in the demo https://observablehq.com/d/eab191ed35920c7c to what needs to be "proxied" for an arquero table to have an API that is closer to an array. Still pretty much inconclusive at this point.

mbostock:
I believe that Proxy is fairly slow, so I wonder if this is still faster than the existing (iterative over row objects) approach. And also whether Arquero is a performance bottleneck in any case. I was also considering using *table*.array instead of *table*.column, but then we’re making a copy of the column which would be nice to avoid if possible.

Anyway, speaking of performance, one optimization I know we want to make is binning of many values (a few millions, say.) This currently uses bisection in D3 but I suspect we could do something faster (quantization) when the thresholds are uniformly-spaced. I’ll file an issue for that.

Fil:
superseded by #1324

PR #452, opened 2021-07-10
test arquero

Fil:
Testing arquero adds quite a few packages, but probably worth it.


package.json
```diff
@@ -48,6 +48,7 @@
     "tape-await": "^0.1.2"
   },
   "dependencies": {
+    "arquero": "^4.8.4",
```
mbostock, line 51: This should be a devDependency.

mbostock (changes requested):

mbostock (approved):

PR #453, opened 2021-07-12
Don’t promote string channel values to functions.

mbostock:
Instead, leave them as-is, and let valueof (and labelof) do the work. Broken out of #451.

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -19,9 +19,6 @@ export class Mark {
         if (optional) return false;
         throw new Error(`missing channel value: ${name}`);
       }
-      if (typeof value === "string") {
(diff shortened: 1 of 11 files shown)


Fil:
Yes. We still keep field() for shorthand? Is this what we want in d3 too? In that case should field check its argument x and return d=>d[x] only iff x is a string (or, not a function)?

mbostock:
I only kept it to avoid repeating `d => d[name]` twice. It’s not exposed externally, and we only call it given a string name, so it doesn’t need to check that the argument is a string.

Fil (approved):

Issue #283, opened 2021-03-30
Hyperlinks?


mbostock:
We need a way to make Plot marks linkable (meaning a + xlink:href).

PR #459, opened 2021-07-21
Create Image mark

RLesser:
As mentioned in #157, an image mark could be useful as an alternative for scenarios where it would represent a datapoint better than a text label.

Here's an example using NBA logos (included as a new snapshot test):
Screen Shot 2021-07-20 at 9 18 06 PM

### Changes in this Pull Request
- Added image mark
- Added NBA data file
- Added snapshot test
- Added unit test
 
### Open Questions
- `href` option
  - _Should `href` be a required option? If not, what's the default result?_
  - _Is `href` the best name? Maybe `src` instead?_
  - _Should it support other forms of image specifications, not just urls?_

- `size` option
  - _Should `size` just be `r`? Not sure if the sqrt scale makes sense here._
  - _Should it support a height and width instead? How would that work with image rendering?_
  - _What should the default `size` be?_

- Image mark generally
  - _When should `applyAttr` be used instead of `.attr()`? Do any attributes need to be switched?_

- `Unit test`
  - _Are there other aspects of this mark that need to be tested? This is just a pared down test set from the dot mark._

- `Snapshot test`
  - _This dataset is currently from an ESPN API, with ESPN image urls. Should it be switched to an open-source data source and image source?_
  - _Any ideas for other snapshot tests that might be useful / illustrative?_

### TODO
- [x]  Generate snapshot for snapshot test
- [x] Switch `size` to `r`
- [x] Generate branch bundle and example notebook
- [x] Change to non-copyrighted material
- [x] remove imageX and imageY
- [x] Update to 0.2.0
- [x] Add preserveAspectRatio and crossorigin properties


Fil:
Here are a few remarks and opinions:

It's a pity that both `svg:image` and `svg:a` use `href` to denote different things (first case: image source; second case: link); this means that if we want to use href across all marks to indicate a link, we can't use href here. But using `src` feels a bit wrong too. Maybe to respect standards would mean to use `href` here, and `link` for the generic "create a A:href wrapper" thing '(#283).

`applyAttr` calls selection.attr(…) if the value is defined (and not null). It should be used on conditionals such as `.call(applyAttr, "fill", F && (i => F[i]))` when F can be undefined.

Can we stay away from copyrighted material? (even if it can be considered fair use in this case). I'm sure we can find a good example with public domain illustrations.

I don't think we need imageX and imageY at all. dotX and dotY are useful for quick visualisation of lists of values, but with images we're already in something a bit more elaborate.

(Opinion): size should be `r` for consistency. Width and height seem problematic since I don't really know a use case where changing the aspect ratio doesn't result in awful visuals. There might be a need for cropping an image instead (if we have a bunch of images with different aspect ratios, cropping them to squares or circles often makes a nicer image).

If you want to share an example notebook I'll add a link and the build in the "plot early bird" notebook.

RLesser:
Made updates per your comments @Fil. The test folder now contains examples using presidential portraits (public domain from wikipedia) and polling data. 

I've created an example Observable that uses the NBA data, and will add the presidential examples there as well: https://observablehq.com/@rlesser/observable-plot-image-mark

Let me know if you have other changes you think would be good, or notice any issues in the notebook

Fil:
Thank you for all the changes. Do you think it would be worth to add the [preserveAspectRatio](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio) and crossorigin properties?

RLesser:
> @RLesser Is the https://observablehq.com/@rlesser/observable-plot-image-mark notebook unpublished? Would love to take a look

Thanks for the heads up @tophtucker! It should now be shared and viewable. 

> Do you think it would be worth to add the [preserveAspectRatio](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio) and crossorigin properties?

Yea that would make sense - I can add those.

Also @Fil, it looks like the build is failing, but I'm not entirely sure why. The issue is with the import `tape-await`, but that is already present right?

Fil:
we have changed our test framework to mocha—which means we need to rewrite all the unit tests (fortunately it's not too painful)

RLesser:
One issue I noticed with `preserveAspectRatio` - since we are currently setting `width` and `height` through `r`, users can only have their images be squares (if they set the second param to 'slice') or have them be the original shape (if they set the second param to 'meet' or leave it undefined).

Is it worth trying to allow for more customizability here? I see a few options:

1. Leave as is, with `r` defining the size and 'square' and 'original' being the only two shapes possible.
2. Add `height` and `width` fields in addition to `r`, which override `r` for more customized rectangles
3. Some other combination that allows for more customizability?

Let me know what you all think
cc: @Fil @tophtucker

RLesser:
Any thoughts on the above options @Fil @tophtucker?

Fil:
In a sense I have a hunch that this would be better as a Plot plugin so people could begin using it immediately, and we could then see various use cases. It's hard to imagine what people will really want to do once it's available :)

Re: the aspect ratio, I don't feel that there is a need for anything else than the original shape. Maybe we'll want to add a mask on the image, to make it square, or circular? But I feel that changing the aspect ratio almost inevitably results in ugly visuals.

mbostock:
I have created a branch to push additional changes in #599.

Looking at the example, I think we should go with _width_ and _height_ unscaled channels (or constants) rather than a scaled _r_ channel. To me the _r_ more strongly implies that it should be a circle with clipping, so that it could for example work together with the radius legend. But we could support both, too.

PR #469, opened 2021-07-31
text stroke channels

### LINKED ISSUE #467: Text marks do not support stroke and strokeOpacity as channels.
stroke and strokeOpacity are [documented](https://github.com/observablehq/plot/blob/main/README.md#marks) as standard mark channels, but they are not supported on text marks. (Related, rules also do not yet support strokeOpacity as channel #466, and in the near future, strokeWidth may become a standard mark channel #311.)

My inclination is that we should probably support these as channels on text marks, even if they will almost never be used; alternatively, we could explicitly call out in the documentation that these standard channels are not supported on text marks, but that would leave a gotcha for anyone who doesn’t exactingly read the full documentation.

mbostock:
Fixes #467.

(It’d be nice to consolidate this code somehow, but I don’t think that needs to block this PR.)


Fil (approved):

mbostock:
Thank you for adding a test! I’d like our examples to try to be representative of good visualizations and best practices (for some minimal definition of “good” and “best”); people will inevitably use them to learn and as starting points for exploration. I couldn’t think of a “good” use of the text stroke channel, so I removed it from the test, but maybe it’d be worth adding as a unit test?

PR #470, opened 2021-07-31
handle collapsed domains

mbostock:
Supersedes #438.
Ref. https://github.com/observablehq/plot/issues/417#issuecomment-849149359.

This adds special logic to marks with continuous position dimensions such that when the associated scale’s domain is collapsed (i.e., all values in the input domain map to the same value in the output range), the mark spans the full extent of the chart. For example, here is a plot with the collapsed _y_-domain [0, 0]; note that the bar and rule span the full _y_-extent of the chart. Previously this chart would be empty because the bar and rule would have zero height.

Screen Shot 2021-07-31 at 9 35 21 AM

```js
Plot.plot({
  marks: [
    Plot.barY({length: 1}, {x: ["foo"], y1: [0], y2: [0]}),
    Plot.ruleX(["foo"], {stroke: "red", y1: [0], y2: [0]})
  ]
})
```

In practice, this seems most likely to occur when binning values that all happen to be equal, but unlike #438 this does not change the semantics of the chart and thus is more robust and general.

The isCollapsed check could be memoized if performance is a concern in the future, given how rare it is for a scale to be collapsed. And also it’s slightly unfortunate that this logic must be applied by mark implementations rather than somehow generalizing to the scales. I think it might be possible to do this more generically, and faster, but this seems fine for now?

### DIFF
--- a/src/facet.js
+++ b/src/facet.js
@@ -72,7 +72,7 @@ class Facet extends Mark {
     }
     return {index, channels: [...channels, ...subchannels]};
   }
-  render(index, scales, channels, dimensions, axes) {
+  render(I, scales, channels, dimensions, axes) {
     const {marks, marksChannels, marksIndex, marksIndexByFacet} = this;
     const {fx, fy} = scales;
     const fyMargins = fy && {marginTop: 0, marginBottom: 0, height: fy.bandwidth()};
--- a/src/mark.js
+++ b/src/mark.js
@@ -326,18 +326,3 @@ export function isTemporal(values) {
     return value instanceof Date;
   }
 }
-
-// Certain marks have special behavior if a scale is collapsed, i.e. if the
-// domain is degenerate and represents only a single value such as [3, 3]; for
-// example, a rect will span the full extent of the chart along a collapsed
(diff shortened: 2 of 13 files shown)


Fil:
Works perfectly. But following your hint of doing this “more generically”, I offer an alternative in #474.

Notice that the singleValueBar example is identical if we omit y1 and y2:

```
Plot.plot({
    marks: [
      Plot.barY({length: 2}, {x: ["foo", "bar"]}),
      Plot.ruleX(["foo"], {stroke: "red"})
    ]
  });
```

Fil:
A serious issue I'm seeing only now is what happens when if we have only 1 bar with a 0-height (barY({y:0}))… it would probably be very surprising to see it shown at all? OTOH one can argue that it's "continuous" with barY({y: 𝜺}) with 𝜺 => 0?  Wanted to flag it anyway.

mbostock:
I’ve rebased this and I’m still in favor of this over #474. Do you have any objection, @Fil?

Fil:
Not a strong objection—it's one of these cases where we need an arbitrary resolution anyway.

Fil (approved):

PR #471, opened 2021-08-01
Shorthand for axis-aligned links

### LINKED ISSUE #428: Shorthand for horizontal and vertical links
> Plot.link(subdata, {x1: "low", x2: "high", y: "country" }),

could be a shorthand for 
> Plot.link(subdata, {x1: "low", x2: "high", y1: "country", y2: "country" }),

mbostock:
Fixes #428. I mirrored the approach from the area mark so that we don’t need to compute both channels when only one is needed.

### DIFF
--- a/src/marks/link.js
+++ b/src/marks/link.js
@@ -30,8 +30,8 @@ export class Link extends Mark {
       [
         {name: "x1", value: x1, scale: "x"},
         {name: "y1", value: y1, scale: "y"},
-        {name: "x2", value: x2, scale: "x"},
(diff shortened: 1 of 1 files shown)


Fil (approved):
we could test it by switching in test/plots/us-population-state-age-dots.js
~~~diff
-      Plot.ruleY(stateage, Plot.groupY({x1: "min", x2: "max"}, position)),
+      Plot.link(stateage, Plot.groupY({x1: "min", x2: "max"}, position)),
~~~

PR #474, opened 2021-08-02
a generic alternative for the collapsed scales

Fil:
apply min(range) to x1 and y1, and max(range) to x2 and y2.

It concentrates the derogatory code in the application of the scale, instead of having to test it in each mark.

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -340,3 +340,18 @@ export function isTemporal(values) {
     return value instanceof Date;
   }
 }
+
+// Certain marks have special behavior if a scale is collapsed, i.e. if the
+// domain is degenerate and represents only a single value such as [3, 3]; for
+// example, a rect will span the full extent of the chart along a collapsed
+// dimension (whereas a dot will simply be drawn in the center).
+export function isCollapsed(scale) {
+  const domain = scale.domain();
+  const value = scale(domain[0]);
+  for (let i = 1, n = domain.length; i < n; ++i) {
+    if (scale(domain[i]) - value) {
+      return false;
+    }
+  }
+  return true;
+}
--- a/src/mark.js
+++ b/src/mark.js
@@ -1,4 +1,4 @@
-import {color} from "d3";
+import {color, max, min} from "d3";
 import {ascendingDefined, nonempty} from "./defined.js";
 import {plot} from "./plot.js";
 
@@ -317,7 +317,20 @@ export function values(channels = [], scales) {
       if (scale !== undefined) {
         scale = scales[scale];
         if (scale !== undefined) {
-          value = Array.from(value, scale);
+          if (scale.isCollapsed) {
+            switch(name) {
+              case "x1":
+              case "y1":
+                value = Array.from(value).fill(min(scale.range())); break;
+              case "x2":
+              case "y2":
+                value = Array.from(value).fill(max(scale.range())); break;
+              default:
+                value = Array.from(value, scale); break;
+            }
+          } else {
+            value = Array.from(value, scale);
+          }
         }
       }
       values[name] = value;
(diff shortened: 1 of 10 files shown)


src/mark.js
```diff
@@ -317,7 +317,20 @@ export function values(channels = [], scales) {
       if (scale !== undefined) {
         scale = scales[scale];
         if (scale !== undefined) {
-          value = Array.from(value, scale);
+          if (scale.isCollapsed) {
+            switch(name) {
+              case "x1":
+              case "y1":
+                value = Array.from(value).fill(min(scale.range())); break;
```
mbostock, line 324: This is slightly different behavior than my PR this it will apply the scale’s inset. That might be desirable, though it’s different than the behavior of one-dimensional marks, and it’s also now dependent on the way the scale range is defined (which can be overridden). I assume it was done because the dimensions aren’t passed into this method, but it’s worth thinking about.

Also this requires materializing another array which is slow although not likely to be an issue. 😅

src/scales/quantitative.js
```diff
@@ -284,3 +285,18 @@ function inferLogDomain(channels) {
   }
   return [1, 10];
 }
+
+// Certain marks have special behavior if a scale is collapsed, i.e. if the
+// domain is degenerate and represents only a single value such as [3, 3]; for
+// example, a rect will span the full extent of the chart along a collapsed
+// dimension (whereas a dot will simply be drawn in the center).
+export function isCollapsed(scale) {
```
mbostock, line 293: You moved this into quantitative.js, but this behavior isn’t limited to quantitative scales (although that is where it most often applies). See the ordinalBar example, which should also be considered collapsed if the domain only contains a single value.

src/scales/quantitative.js
```diff
@@ -179,6 +179,7 @@ export function ScaleQ(key, scale, channels, {
 
   if (range !== undefined) scale.range(range);
   if (clamp) scale.clamp(clamp);
+  scale.isCollapsed = isCollapsed(scale);
```
mbostock, line 182: It’d be good to avoid the mutation here.

mbostock (changes requested):
Thanks for the PR and looking into consolidating the logic here!

I’ve not really articulated this as a principle before, but this is the first time we’re ascribing a special meaning to specific channels by name. Heretofore only scales had such special meanings, based on this registry:

https://github.com/observablehq/plot/blob/2f326d2e4f33a7cdac031b30a00d9dad595f53e4/src/scales/index.js#L18-L29

And eventually I thought we’d find a way to remove the specialness of scale names, although that seems less likely given the way Plot has grown, and certainly not urgent… though it could pave the way for marks to define new scales in the future.

So, I’d like to think about whether marks should declare their desire for this behavior explicitly. That is: whether to replace the channel value with an extremum in the case that the domain is collapsed rather than relying on the scale’s default behavior which is to use the midpoint of the range, rather than rely on a special behavior triggered by the channel name.

Overall, I think I’m inclined to proceed with my PR as it stands rather than trying to tackle the additional issues raised in this PR. Although mostly that’s because there’s a lot of other stuff I want to do. Always happy to discuss and appreciate the work you’ve put in here!

src/scales/quantitative.js
Fil: Can you clarify the ordinalBar example? I've tried several variations and all seem to give the same chart under both branches. For example `Plot.barY("A", {x: (d, i) => i, y: d => d}).plot()` gives a large rect (width x height).

Fil:
I wanted to see if there was a way to not complicate the marks' code for what is only a corner case. I attacked the scale, but maybe another approach would allow us to switch a mark with a collapsed dimension to a 1-d mark? 🤔

src/scales/quantitative.js
mbostock: The example is with a point scale:

```js
Plot.plot({
  y: {
    type: "point",
    grid: true
  },
  marks: [
    Plot.barY([0, 0, 0], {x: (d, i) => i, y2: d => d}),
    Plot.ruleY([0])
  ]
})
```

This is admittedly a pretty pointless plot, but it does demonstrate the issue: in this PR, the _y_-domain is not considered collapsed because it is not a quantitative scale, even though its domain only has a single value [0].

Screen Shot 2021-08-09 at 6 34 35 PM

Whereas in my PR you get:

Screen Shot 2021-08-09 at 6 34 47 PM

Fil:
Might revisit later. The "pros" of this branch were essentially in the separation of concerns (collapsed domains handled in the scale rather than in the marks, resulting in a code that seemed simpler to maintain). The "cons" are listed in the conversation :)

PR #475, opened 2021-08-02
reinstate the "save changed figure" functionality of tests

Fil:
(I think it disappeared after 13052a87 when assert replaced test.ok)


Fil:
not sure I need to bother you with such a tiny fix :)

mbostock (approved):

PR #480, opened 2021-08-03
map option for map[XY]

### LINKED ISSUE #479: Plot.normalizeX and Plot.mapX have different structure
I was confused because I tried to do Plot.normalizeX(*basis*, *options*), which is similar to Plot.mapX(*reducer*, *options*), but it didn’t work! Instead you need to do Plot.normalizeX({basis, ...options}). Maybe it’d be nice to support both forms just in case you don’t read the documentation as closely as you should.

mbostock:
Fixes #479.

My first take was to support Plot.normalizeX(*basis*, *options*) and Plot.normalizeY(*basis*, *options*). While that’s more consistent with Plot.mapX(*map*, *options*) and Plot.mapY(*map*, *options*), it’s not consistent with Plot.windowX(*options*) and Plot.windowY(*options*). So, I then decided it would be better to standardize on Plot.mapX(*options*) and Plot.mapY(*options*), and only use the two-argument form when you want to declare the output channels explicitly (e.g., Plot.map, Plot.group, and Plot.bin).

For backwards-compatibility we can continue to support the old form of Plot.mapX and Plot.mapY, but I consider it deprecated and removed it from the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1239,21 +1239,21 @@ Plot.map({y: "cumsum"}, {y: d3.randomNormal()})
 
 Groups on the first channel of *z*, *fill*, or *stroke*, if any, and then for each channel declared in the specified *outputs* object, applies the corresponding map method. Each channel in *outputs* must have a corresponding input channel in *options*.
 
-#### Plot.mapX(*map*, *options*)
+#### Plot.mapX(*options*)
 
 ```js
-Plot.mapX("cumsum", {x: d3.randomNormal()})
+Plot.mapX({map: "cumsum", x: d3.randomNormal()})
 ```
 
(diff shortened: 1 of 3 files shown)


Fil (approved):
Good! Since we're still in the 0.x, do we have to maintain any sort of backwards-compatibility? And if so, for how many versions?

mbostock:
We need to maintain backwards-compatibility for Observable until we ship version pinning. Otherwise notebooks will upgrade and break.

Also, I’ve realized that the pattern in this PR is inconsistent with the pattern I proposed for basic transforms in #472, so I want to think about how to resolve this discrepancy before landing.

mbostock:
Closing in favor of #494.

PR #484, opened 2021-08-05
legends and scales

### LINKED ISSUE #23: Color legend.

### LINKED ISSUE #236: radius legend?

Fil:
(rebased #432)

build at https://observablehq.com/@fil/plot-color-legend-484

closes #23
closes #236

related: #397 #92 #425

TODO:
- [ ] split "expose scales" from "legends"
- [ ] add legend for **threshold** and **quantile** scales
- [x] check polylinear color scales (#530)
  - merged locally
- [ ] check legend for all the diverging scales
- [ ] color legends with a date domain
- [ ] more tests
- [ ] document exhaustively


yurivish:
Is there a way to configure the positioning of the legend with respect to the plot, e.g. above or below?

Fil:
No, there is no such thing for the moment. We'll probably want some configurability but we don't want to invent a new layout engine :) What you can already do, is to create the legend separately (as shown in a few of the [examples](https://observablehq.com/d/c324a9a466e66329)). Alternatively, move it inside the figure element with d3.select?.

mbostock:
I wanted to work on landing exposing scales first before tackling legends. Ref. https://github.com/observablehq/plot/pull/425#issuecomment-892270954

Fil:
In the two other branches we had accumulated quite a bit of conflict with main, and I wanted to clear that. But feel free to remove anything from this PR (I think the legend part is well separated).

About the issue of exposing Plot.scale(options) or not, I think that there will be a way around in any case if we expose the d3 scale: something like `Plot.plot({ color: options }).scales.color.scale`.

mbostock:
My preference for scales would be to just expose the materialized options (type, domain, range, etc.) with the goal of being able to take a scale constructed from one chart and propagate it to another chart; I would prefer not to expose a D3 scale instance or a scale function. We removed most of the need for that in #280 by precomputing the scaled channels, though scales still have access to the scale functions… which is a little messy. In my ideal world the implementation of scales would be hidden so that we have more flexibility in changing the implementation.

Fil:
I'm good with this plan (including _not_ exposing the d3 scale); it still gives enough information to re-create the scale outside of Plot if necessary.

mbostock:
Superseded by #539.

PR #487, opened 2021-08-07
Raise an error when a reducer expects numeric input but is given non-numeric input

### LINKED ISSUE #473: When a reducer excepts numeric input but is given non-numeric input, it should raise an error.
Following up on https://github.com/observablehq/plot/issues/399#issuecomment-890596083. For example, if you pass ordinal data (strings) to a _sum_ reducer, Plot could detect that and raise a helpful error rather than producing NaN output.

Fil:
Add tests for the normalize transform
closes #473

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -340,3 +340,14 @@ export function isTemporal(values) {
     return value instanceof Date;
   }
 }
+
+function isNumeric(values) {
+  for (const value of values) {
+    if (value == null) continue;
+    return !isNaN(+value);
+  }
+}
+
+export function checkNumeric(S) {
+  if (!isNumeric(S)) throw new Error("the reducer expects numeric input");
+}
--- a/src/transforms/group.js
+++ b/src/transforms/group.js
@@ -1,6 +1,6 @@
 import {group as grouper, sort, sum, deviation, min, max, mean, median, mode, variance, InternSet} from "d3";
 import {firstof} from "../defined.js";
-import {valueof, maybeColor, maybeInput, maybeTransform, maybeTuple, maybeLazyChannel, lazyChannel, first, identity, take, labelof, range} from "../mark.js";
+import {valueof, maybeColor, maybeInput, maybeTransform, maybeTuple, maybeLazyChannel, lazyChannel, first, identity, take, labelof, range, checkNumeric} from "../mark.js";
 
 // Group on {z, fill, stroke}.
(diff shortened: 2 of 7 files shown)


src/mark.js
```diff
@@ -340,3 +340,14 @@ export function isTemporal(values) {
     return value instanceof Date;
   }
 }
+
+function isNumeric(values) {
+  for (const value of values) {
+    if (value == null) continue;
+    return !isNaN(+value);
+  }
+}
```
mbostock, line 349: This is a slightly different assertion than what we do elsewhere, e.g.:

https://github.com/observablehq/plot/blob/b1885ad65ff5351485260ec72acd12833969dbbb/src/scales.js#L104-L108

I think what we’re saying here is whether the values are “weakly” numeric, i.e., whether they can be treated as numbers (in this context). So renaming this to isWeaklyNumeric is probably good.

Also, I think we want to treat strict NaN as numeric in this context, since the goal here is to detect things that are _not_ numbers (such as strings), not to check that the numbers themselves are invalid. So, one possibility:

```suggestion
function isWeaklyNumeric(values) {
  for (const value of values) {
    if (value == null) continue;
    if (typeof value === "number") return true; // note: includes NaN!
    return !isNaN(+value);
  }
}
```

src/mark.js
```diff
@@ -340,3 +340,14 @@ export function isTemporal(values) {
...
+
+export function checkNumeric(S) {
+  if (!isNumeric(S)) throw new Error("the reducer expects numeric input");
+}
```
mbostock, line 353: This error message is too specific (referring to a reducer here). I’ll think about this…

src/transforms/group.js
```diff
@@ -170,9 +170,10 @@ function reduceFunction(f) {
   };
 }
 
-function reduceAccessor(f) {
+function reduceAccessor(f, check) {
```
mbostock, line 173: I think I’d prefer a separate reduceNumericAccessor here rather than an argument.

mbostock (changes requested):
I’ll chip in to help with these.

Fil:
related #751

Fil:
I've rewritten this PR against main and following your comments.

One thing to note is that we run the test on each call to the reducer. It will usually be fast (returning immediately), so there is no need to memoize, but there might be weird cases (large tables full of null values) where it's not.

Issue #174, opened 2021-02-25
Should more constant properties (e.g., textAnchor) be specifiable as channels?


Fil:

mbostock:
I don’t see us taking any immediate action on this for now, so I’m going to close.

PR #488, opened 2021-08-07
support dx, dy on all marks

### LINKED ISSUE #379: Support dx, dy on all marks?
Sometimes for presentation purposes you want e.g. a tickX to be off by x pixels

Fil:
On all marks except text, dx, dy are rendered as a transform(translate) or (x,y) property, possibly with the 0.5px offset on high-density screens. On text marks, the dx and dy properties are used.

closes #379

build : https://observablehq.com/@fil/plot-dx-dy-488 ; also available in https://observablehq.com/@fil/plot-early-bird

### DIFF
--- a/src/marks/rule.js
+++ b/src/marks/rule.js
@@ -2,7 +2,7 @@ import {create} from "d3";
 import {filter} from "../defined.js";
 import {Mark, identity, number} from "../mark.js";
 import {isCollapsed} from "../scales.js";
-import {applyDirectStyles, applyIndirectStyles, applyTransform, applyChannelStyles} from "../style.js";
+import {applyDirectStyles, applyIndirectStyles, applyTransform, applyChannelStyles, offset} from "../style.js";
 
 const defaults = {
   fill: null,
@@ -39,7 +39,7 @@ export class RuleX extends Mark {
     const index = filter(I, X, Y1, Y2);
     return create("svg:g")
         .call(applyIndirectStyles, this)
-        .call(applyTransform, X && x, null, 0.5, 0)
+        .call(applyTransform, X && x, null, offset, 0)
         .call(g => g.selectAll("line")
           .data(index)
           .join("line")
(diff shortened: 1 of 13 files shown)


mbostock:
What’s the use case? Feels a bit niche.

Fil:
I often find I want to translate a mark by a few pixels, and bummed that only Plot.text offers this feature. This would be convenient to add a rule at the side of a bar, make shades, annotations / overlays (manual legends), [separation lines](https://observablehq.com/d/dca184f528bc6847) between categories, etc.

src/marks/rule.js
```diff
@@ -103,9 +103,10 @@ export class RuleY extends Mark {
     {width, height, marginTop, marginRight, marginLeft, marginBottom}
   ) {
     const index = filter(I, Y, X1, X2);
+    const {dx, dy} = this;
     return create("svg:g")
         .call(applyIndirectStyles, this)
-        .call(applyTransform, null, Y && y, 0, 0.5)
+        .call(applyTransform, null, Y && y, dx, offset + dy)
```
mbostock, line 109: Maybe in cases like this it would be preferable to initialize {dx = 0, dy = offset}, rather than adding offset to dy?

Fil: No, it's a separate concern. If you want a shape to be 1px to the left of another, they should both be moved by .5px if that helps make them sharper.

Fil:
> if this becomes a channel like the new behavior of `strokeWidth`, it would be really cool as well.

Ref.: https://github.com/observablehq/plot/issues/174

ryoqun:
https://github.com/observablehq/plot/pull/322#issuecomment-895146363:

> wow! great example, I'd be curious to see it in action if you can share it! If you want to test this branch I've updated the build at https://observablehq.com/@fil/plot-strokewidth-322 (see also https://observablehq.com/@fil/plot-early-bird)

@Fil hi, I've finally made my plot application public. (not on observablehq.com because i wanted full screen..)

still, buggy, slow and ugly code. :)

I think you can click around even if you don't understand: https://codepen.io/ryoqun/full/JjNbeyE

fyi, we're using it like this: https://github.com/solana-labs/solana/issues/18308#issuecomment-901262267

lastly, so, I want to plot multiple circles or rects per y/z with dx/dy shift if possible.  :)

mbostock (approved):
Looks good! 👍

PR #489, opened 2021-08-07
An option to include empty bins

Fil:
When using line or area on binned data, it can be necessary to set {skip: false} in order to reduce to count=0 values.

(Not fully sure that this is the correct solution, or the correct naming.)

build https://observablehq.com/@fil/missing-data-in-lines-489

(let me know if you think the principle sounds correct—currently this is breaking quite a few tests, I'll fix that later.)


README.md
```diff
@@ -1032,6 +1032,7 @@ To control how the quantitative dimensions *x* and *y* are divided into bins, th
 * **thresholds** - the threshold values; see below
 * **domain** - values outside the domain will be omitted
 * **cumulative** - if positive, each bin will contain all lesser bins
+* **skip** - skip empty bins (default: true)
```
mbostock, line 1035: How about the name **empty**, where true means to include empty bins, defaulting to false?

src/transforms/bin.js
```diff
@@ -124,12 +124,13 @@ function binn(
...
 }
 
-function maybeBinValue(value, {cumulative, domain, thresholds} = {}, defaultValue) {
+function maybeBinValue(value, {cumulative, domain, thresholds, skip = true} = {}, defaultValue) {
   value = {...maybeValue(value)};
   if (value.domain === undefined) value.domain = domain;
   if (value.cumulative === undefined) value.cumulative = cumulative;
   if (value.thresholds === undefined) value.thresholds = thresholds;
   if (value.value === undefined) value.value = defaultValue;
+  value.skip = skip;
```
mbostock, line 133:
```suggestion
  value.skip = !!skip;
```

mbostock (changes requested):
Good idea!

mbostock:
Another thought is whether the bin transform should not filter by default, since we are trading off a performance optimization for a change in semantics. I think ideally we’d be able to tell whether the bins were consumed by a mark that requires continuous semantics? Anyway, this is a fine change to start.

mbostock (approved):
Switched to the **empty** option as suggested. If this looks good to you, please merge?

One minor thing: I noticed that this code doesn’t allow you to set the _empty_ option for an individual dimension, as we do for _domain_, _cumulative_, and _thresholds_. It would be possible to support this by adding the following:

```js
if (value.empty === undefined) value.empty = empty;
value.empty = !!value.empty;
```

However, I don’t think it’s meaningful to support this on a per-dimension basis, because the _empty_ option applies to the constructed bins across dimensions. So even though the code separately looks at `bx.empty` and `by.empty`, these should always have the same value, and even if they have different values, the transform will only return non-empty values if both `bx.empty` and `by.empty` are truthy.

Another tiny observation is that `b[0]` will be undefined when the bins are empty; or in other words, the _z_, _fill_, and _stroke_ channels if any will be undefined for empty bins. But that is exactly what we want.

And lastly I wonder if the rect mark should be smart enough to not create rect elements if the computed width or height is nonpositive? That might make rendering faster even when rendering empty bins. But I don’t feel the need to implementation that optimization right now. 🙂

Fil:
I want to add a test plot before I merge, but this works well in https://observablehq.com/@fil/missing-data-in-lines-489

Fil:
OK now I want to walk this back, and replace it with a different strategy based on #490.

mbostock:
Can you give me a hint as to what you’re thinking? I don’t see the obvious connection between this and #490.

Fil:
The idea is that all the default reducers should return _null_ when given an empty array.

Currently these reducers are not called for empty bins, but if they were called, sum, length and proportion would return 0. 

However if we change them to return return null, and if they are used to inform a "filtered channel" (like any of stroke, fill, fillOpacity etc, or y for barY etc):
* a simple mark (say, a bar) is filtered out
* a grouped mark (line) knows how to stop (instead of interpolating)
* we don't need the empty: true option—it's still possible to impute missing values by passing a custom reducer (e.g. (bin) => bin.length) if you want to draw an empty bin with a value of 0.

"filtered channels" are consolidated in #490, which made me think there was a direct connection, but in fact it's orthogonal.

PR #490, opened 2021-08-08
Common style channels

mbostock:
Related #322, this is a sketch at how we might consolidate the definition of “standard” channels in Plot. This idea is to define them in one place so that we can add a new standard channel to all marks with less risk of drift between marks.

~~This only currently works for the Area mark. In addition to porting the other marks to use this system, we need to address that some marks have different defaults (for example Line has a default stroke, whereas Area has a default fill), and that some marks have different structure (for example Area and Line produce a single path element per series and thus group styles, whereas Dot and Bar product distinct elements for each data point).~~ I’ve done most of this now.

TODO

- [x] areas
- [x] bars
- [x] cells
- [x] dots (quirk: default fill)
- [x] frames (quirk: does not support channels)
- [x] lines
- [x] links
- [x] rects
- [x] rules
- [x] texts
- [x] ticks

If you like this idea I’d be happy to continue work and/or work on it together.

### DIFF
--- a/src/marks/area.js
+++ b/src/marks/area.js
@@ -2,8 +2,13 @@ import {area as shapeArea, create, group} from "d3";
 import {Curve} from "../curve.js";
 import {defined} from "../defined.js";
 import {Mark, indexOf, maybeZ} from "../mark.js";
-import {applyDirectStyles, applyIndirectStyles, applyTransform, applyGroupedChannelStyles, primaryFill} from "../style.js";
+import {applyDirectStyles, applyIndirectStyles, applyTransform, applyGroupedChannelStyles} from "../style.js";
 import {maybeStackX, maybeStackY} from "../transforms/stack.js";
+
+const defaults = {
+  strokeWidth: 1,
+  strokeMiterlimit: 1
+};
 
 export class Area extends Mark {
   constructor(data, options = {}) {
@@ -18,7 +23,7 @@ export class Area extends Mark {
         {name: "z", value: maybeZ(options), optional: true}
       ],
       options,
-      primaryFill
+      defaults
     );
     this.curve = Curve(curve, tension);
   }
--- a/src/marks/area.js
+++ b/src/marks/area.js
@@ -1,37 +1,22 @@
-import {group} from "d3";
-import {create} from "d3";
-import {area as shapeArea} from "d3";
+import {area as shapeArea, create, group} from "d3";
 import {Curve} from "../curve.js";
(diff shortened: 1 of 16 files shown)


src/marks/area.js
```diff
@@ -1,37 +1,22 @@
...
       y1,
       x2,
       y2,
-      z, // optional grouping for multiple series
-      title,
-      fill,
-      fillOpacity,
-      stroke,
-      strokeOpacity,
+      z = findStyle(channels, "fill", "stroke"), // optional grouping for multiple series
```
mbostock, line 16: It might be possible to use maybeZ here instead… and then it might be possible to move the call to styles(…) up to the Mark constructor, which would be really nice.

mbostock: Okay, that worked. 😄

Fil:
Yes that's a big relief, and besides making Plot smaller and more consistent, it will also make custom marks much easier.

I stopped short of:
*  adding the channels x, x1, x2, y, y1, y2 in the default filter (and then maybe also r and text?)
* renaming filter to definedFilter, which would be clearer about the purpose

Fil:
And https://github.com/observablehq/plot/pull/322 became so much simpler!

Fil (approved):

Fil:
Wait, I forgot to apply this in the facet's call to mark.render

mbostock:
Here’s a quirk that I’m thinking about currently.

Some marks default to a fill (with no stroke), while others default to a stroke (with no fill). For example, bar and area default to a fill, while dot and line default to a stroke. (Also some marks don’t support a fill, such as ticks and rules, but for now we can treat that as the same as marks that default a stroke.)

But that’s not all. For marks that have a default fill (such as bar and area), the default fill only applies if the stroke is (constant) none; if you set a stroke, then the default fill switches to none. Similarly for marks that have a default stroke (such as dot), the default stroke only applies if the fill is (constant) none. But line is not consistent with this: if you apply a fill, it retains the default stroke.

So, do we want to standardize on the behavior of dot which is symmetrical, or do we want to standardize on the behavior of line which is (arguably) simpler? And in any case if we make dot and line behave the same way, it’s not strictly backwards-compatible. 🤔

mbostock:
I vote that we standardize on the symmetrical behavior (dot’s), and therefore change the behavior for line. Setting a fill on line is very rare anyway so I think it’s unlikely to cause trouble if the default stroke goes away when you set a fill.

PR #491, opened 2021-08-09
Fil/reduce empty bins to null

Fil:
supersedes #489

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -228,7 +228,7 @@ function bincumset([bin], j, bins) {
 }
 
 function binfilter([{x0, x1}, set]) {
-  return [x0, x1, I => I.filter(set.has, set)]; // TODO optimize
+  return [x0, x1, set.size ? I => I.filter(set.has, set) : () => []]; // TODO optimize
 }
 
 function maybeInset(inset, inset1, inset2) {
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -96,10 +96,8 @@ function binn(
           for (const [k, g] of maybeGroup(I, K)) {
             for (const [x1, x2, fx] of BX) {
               const bb = fx(g);
-              if (bb.length === 0) continue;
               for (const [y1, y2, fy] of BY) {
                 const b = fy(bb);
-                if (b.length === 0) continue;
                 groupFacet.push(i++);
                 groupData.push(reduceData.reduce(b, data));
                 if (K) GK.push(k);
@@ -165,7 +163,7 @@ function maybeBin(options) {
     }
     let bins = bin(range(data)).map(binset);
     if (cumulative) bins = (cumulative < 0 ? bins.reverse() : bins).map(bincumset);
-    return bins.filter(nonempty2).map(binfilter);
+    return bins.map(binfilter);
   };
   bin.label = labelof(value);
   return bin;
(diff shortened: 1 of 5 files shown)


Fil:
(tested with #490 : 👍)

src/transforms/group.js
```diff
@@ -199,13 +199,14 @@ const reduceLast = {
 const reduceCount = {
   label: "Frequency",
   reduce(I) {
-    return I.length;
+    return I.length || NaN;
```
mbostock, line 202: This feels wrong. The count is zero for an empty bin, not NaN. Same with distinct and proportion. I agree that other reducers (such as _first_) should return undefined for an empty bin, but they already do, right?

src/transforms/bin.js
```diff
@@ -230,11 +228,7 @@ function bincumset([bin], j, bins) {
 }
 
 function binfilter([{x0, x1}, set]) {
-  return [x0, x1, I => I.filter(set.has, set)]; // TODO optimize
-}
-
-function nonempty2([, {size}]) {
-  return size > 0;
+  return [x0, x1, set.size ? I => I.filter(set.has, set) : () => []]; // TODO optimize
```
mbostock, line 231: This seems like a nice optimization (if we want to return empty bins). 👍

mbostock (changes requested):
I think we want to return the empty bins, then we should incorporate the binfilter optimization here, and possibly we optimize rendering such that zero-width or zero-height rects and bars are not instantiated in SVG. We should not change the semantics of the reducers that should return zero for empty bins.

src/transforms/group.js
Fil: Returning NaN is analogous to ignoring the empty bins and has the same result, except for line and area where it's semantically better (an empty bin results in a cut of the line rather than an interpolation between full bins). But I'll check if the check on emptiness can be done before the reducer, because that indeed feels wrong.

Note: the only other reducer which returned 0 was sum. I believe there's a tiny difference between a sum of empty and a sum of 1 and -1.

mbostock: I understand that the effect is (roughly) the same on the resulting chart, but it changes the semantics of the bin transform, which I feel is undesirable.

Fil:
In terms of information, we want to be able to tell "there is no data" separately from "the result is zero", because in one case the result can be skipped by a line / ignored by a bar, whereas in the other it should be drawn even if it's a zero.  Would you have the same gut feeling if the reducers returned _null_ instead of _NaN_ for "perhaps invalid bins"?

The code could work with nulls, even though it would mean a special case in the stack transform. (What to do with nulls when stacking? If it's bars that we're stacking, we can ignore nulls, but if if it's areas we probably want the ribbon to get thinner at that point—each mark would have to call the default stack transform with the relevant option to "ignore nulls" or "zero nulls").

Fil:
TODO: this filtering of invalid/empty spots in Plot.stack should be an option: we'll want it if calling from Plot.barY but not from Plot.areaY.

mbostock:
I see. I think what I’m objecting to is that it feels “icky” to encoding the empty bit into the value returned by the reducer, and instead there should be some out-of-band way to communicate which bins are empty to things that care. I’ll think about this some more and try to come up with an alternative.

mbostock:
I’ve mulled on this and I’d like to proceed with #495. I do think it makes sense that the bin transform filters by default (like the group transform, which does not return empty groups). And I feel an explicit filter option is more straightforward than giving the reducers special behavior for the empty case, and then likewise the stack transform needing special behavior for null and NaN. Thank you for engaging in this exploration with me and let’s chat about this tomorrow!

PR #494, opened 2021-08-10
Explicit transform options

### LINKED ISSUE #479: Plot.normalizeX and Plot.mapX have different structure

mbostock:
This PR allows transform options to be kept separate from passthrough options, as well as shorthand options for the normalize and window transforms. Explicit transform options can disambiguate the intended recipient of the options, which could be a specific transform in a sequence of transforms or the output mark. This is especially important for the stack option’s _reverse_ option, which overlaps with the basic reverse transform.

Supersedes #480.
Supersedes #439.
Related #472.
Fixes #479.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1386,7 +1386,7 @@ If a given stack has zero total value, the *expand* offset will not adjust the s
 
 In addition to the **y1** and **y2** output channels, Plot.stackY computers a **y** output channel that represents the midpoint of **y1** and **y2**. Plot.stackX does the same for **x**. This can be used to position a label or a dot in the center of a stacked layer. The **x** and **y** output channels are lazy: they are only computed if needed by a downstream mark or transform.
 
-If two arguments are passed to the stack transform functions below, the stack-specific options (**offset**, **order**, and **reverse**) are pulled exclusively from the first *options* argument, while any channels (*e.g.*, *x*, *y*, and *z*) are pulled from second *options* argument. Options from the second argument that are not consumed by the stack transform will be passed through. Using two arguments is sometimes necessary is disambiguate the option recipient when chaining transforms.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1258,6 +1258,8 @@ The following window reducers are supported:
 * *difference* - the difference between the last and first window value
 * *ratio* - the ratio of the last and first window value
 
+By default, **shift** is *centered* and **reduce** is *mean*.
```
Fil, line 1261: related: #333

README.md
```diff
@@ -1282,37 +1284,37 @@ Plot.mapY("cumsum", {y: d3.randomNormal()})
...
-#### Plot.windowX(*options*)
+#### Plot.windowX(*k*, *options*)
 
 ```js
-Plot.windowX({y: "Date", x: "Anomaly", k: 24})
+Plot.windowX(24, {y: "Date", x: "Anomaly"})
 ```
 
-Like [Plot.mapX](#plotmapxmap-options), but applies the window map method with the given *options*.
+Like [Plot.mapX](#plotmapxmap-options), but applies the window map method with the given window size *k*. If *k* is an object, separate *shift* and *reduce* window options can also be specified.
```
Fil, line 1309: does this wording implies that if k is an object, it is to be written as {k: 24, shift: xxx, reduce: yyy} ? An alternative wording could be:
> more options can be passed by using the longhand format, a as object with properties *k*, *shift* and *reduce*.

Fil (approved):

README.md
mbostock: Yep, correct!

PR #495, opened 2021-08-11
sort and filter for bin and group

mbostock:
This introduces a **filter** output for the bin transform which defaults to *count* such that by default only non-empty bins are returned. By setting the filter to null you can return all bins; by setting it to a function you can apply whatever test you like on the bins before any other output channels are evaluated. Like the other reducers, this is specified on *outputs* rather than *options*.

I’m not sure this is strictly better than #489. It does allow more fine-grained control over which bins to return, and it is faster than applying the filter transform after binning (since it skips evaluating the other channels if the bin will be dropped). However, it is likely (a tiny bit?) slower than #489 since we cannot special-case the test for empty bins. And it is less explicit than a dedicated **empty** option.

Supersedes #491.
Supersedes #489.


mbostock:
This works well in the athletesHeightWeightBinStroke example: rather than setting the strokeOpacity to zero or one, we can filter the elements we don’t want to show.

mbostock:
I see two limitations of this **filter** option *vis à vis* the **sort** option you propose in #334:

1. I only added support to the bin transform, but we should probably also support the filter option on the group transform with the same semantics (although it still won’t be possible to return empty groups yet for the full Cartesian product).
2. The filter option always takes the data as input; you can’t specify a *filter* input channel to pass to the reducer (e.g., if you want the sum of weights to be greater than zero).

I think both of these are possible and would love assistance.

mbostock:
The transform option feels like new semantics (and could arguably be called a *map* option); if we’re going to support the sort option #334 then the filter option here feels more simpatico. And we can make filter and sort more similar if allow an input channel to the filter reducer, and extend the filter option to the group transform (though as I said, for group setting filter: null won’t have any effect because you can’t generate empty groups, but you could still use it to suppress other groups).

mbostock:
The only thing that this is now missing is supporting the *sort* option being specified as a comparator rather than reducer. I think that’s possible: you’d compute the bins or groups for the sort’s input channel (or the data), and then you’d pass those arrays to the comparator during sorting. But it would require some extra finagling to get this to work, and it doesn’t seem urgent since the common case will be things like sorting by count or some other accessor. So, I’ll punt! Forward!

Fil:
An image for the CHANGELOG
penguins-bins

mbostock:
I was thinking we could use the sfCovidDeaths example because that shows the necessity of returning empty bins.

mbostock:
I prefer less contrived if possible but I’ll take a look.

Issue #351, opened 2021-05-01
Add a "full series" option to the stack transforms


Fil:
See discussion at #325 and https://github.com/observablehq/plot/pull/348#issuecomment-830659582

mbostock:
Datadog calls this “default zero” interpolation: https://docs.datadoghq.com/dashboards/functions/interpolation/#default-zero

I wonder to what degree this is specific to time series. I can certainly imagine cases where it’s _not_ specific to time series, but when it is, it seems like the bin transform with filter: null is an option for fixing the missing data.

Fil:
It's worse than this: using the empty bin approach is necessary for a continuous ("binnable") domain, but far from sufficient—as soon as you have *z* or facets, you need a point (real or fake) for each element of the domain times each of the series.

mbostock:
But the bin domain is the same across all groups and facets, so as long as you have at least one data point in a given group,  you’ll get all the bins?

Fil:
In https://observablehq.com/d/f6a7975f2ad4519a there is no empty bin, but if we push up the number of bins to 200, we start to get the same issue of "interpolating above missing bins". Outlined in red in the image below, all the areas should drop to zero since there is no data point in this position.

<img width="624" alt="Capture d’écran 2021-08-12 à 22 39 00" src="https://user-images.githubusercontent.com/7001/129266908-a8500fb9-bf3f-49e1-b9e9-f20c0cd5e889.png">

(I'm not sure that we can find a generic way to do both operations, maybe imputing missing values is something that should be left to the data-wrangling section?)

mbostock:
But that example doesn’t use filter: null on the bin transform, right?

Fil:
possibly superseded by #499 and #500

Fil:
closing since this example is solved with _filter: null_.

Issue #197, opened 2021-03-08
stackR transform?


mbostock:
I could see a stackR transform which takes an *r* (radius) channel in and produces *r1* and *r2* output channels, which you would use in conjunction with an annulus type mark. Maybe the dot mark should allow *r1* and *r2* input channels and have similar behavior to barY’s *y*, *y1*, and *y2* channels. The sqrt transform should ensure that the rings have area proportional to value.

PR #496, opened 2021-08-11
Cars MPG example plot

Fil:
This example plot computes the median of cars' economy (mpg), grouped by number of cylinders

Capture d’écran 2021-08-11 à 12 21 54

The bins are sorted by decreasing r, so that they are all visible (using Plot.sort, not #334 / #349!).

The example would benefit from stackR (#197).

It could also benefit from a strategy to create missing values for the line, so that it's broken when there are no data. However, it won't work with an approach such as "return empty bins" (#495), because returning empty bins will not create the *z* values for each and every category, which would be necessary if we wanted to create broken lines. This shows that a generic foolproof solution to #351 will require much more than #495 (and #489 and #491 are not better in that regard).

It's also crying for a color legend (#23).


test/plots/cars-mpg.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default async function() {
+  const data = await d3.csv("data/cars.csv", d3.autoType);
+
+  return Plot.plot({
+    x: {inset: 20, ticks: 10},
+    y: {grid: true},
+    color: {
+      type: "categorical",
+      legend: true // as soon as it's available
```
mbostock, line 12:
```suggestion
      type: "categorical"
```

I 100% agree this is wanting a color legend, but I’d prefer not to add options until they’re implemented. We don’t want tests to break when legends are implemented.

test/plots/cars-mpg.js
```diff
@@ -0,0 +1,35 @@
...
+    },
+    marks: [
+      Plot.ruleY([0], {filter: false}),
```
mbostock, line 15: Is this filter doing anything?

```suggestion
      Plot.ruleY([0]),
```

test/plots/cars-mpg.js
```diff
@@ -0,0 +1,35 @@
...
+      Plot.line(data,
+        Plot.binX({y: "mean"}, {
+          x: d => +d["year"],
```
mbostock, line 18: Is the coercion and function accessor necessary here?

```suggestion
          x: "year",
```

test/plots/cars-mpg.js
```diff
@@ -0,0 +1,35 @@
...
+          y: "economy (mpg)",
+          stroke: "cylinders",
+          curve: "basis",
+          thresholds: 20
+        })
+      ),
+      Plot.dot(data,
+        Plot.reverse(Plot.sort("length", Plot.binY({ r: "count" }, {
```
mbostock, line 26: Will be nice to pass sort and reverse to bin directly when #334 lands.

```suggestion
        Plot.reverse(Plot.sort("length", Plot.binY({r: "count"}, {
```

test/plots/cars-mpg.js
```diff
@@ -0,0 +1,35 @@
...
+import * as d3 from "d3";
+
+export default async function() {
+  const data = await d3.csv("data/cars.csv", d3.autoType);
+
+  return Plot.plot({
+    x: {inset: 20, ticks: 10},
+    y: {grid: true},
+    color: {
+      type: "categorical",
```
mbostock, line 11: Any reason we can’t use ordinal colors here? The number of cylinders feels ordinal. 🙂

mbostock (changes requested):
Thanks for the test!

test/plots/cars-mpg.js
Fil: yes it's adding 0 in the domain, but no mark

test/plots/cars-mpg.js
Fil: or #349 :)

test/plots/cars-mpg.js
Fil: writing this I suddenly remember the zero: option :)

mbostock: Oh, fascinating. I wasn’t expecting this to work, but it does because Array.from(false) returns []. I think the zero option is more semantic so I switched to that.

mbostock:
If you switch *x* to a point scale here, you’ll notice something interesting which is that the Plot.binX transform for the line is generating more bins than there are years:

Screen Shot 2021-08-11 at 7 57 18 AM

I think what’s happening here is that the bin transform is generating half-year intervals [[70, 70.5], [70.5, 71], …] and then Plot.binX computes the midpoint of each interval [70.25, 70.75, …] but then only half of the bins have any data, so the line is defined at [70.25, 71.25, 72.25, …] whereas the dots are drawn on integers [70, 71, 72, …].

Therefore I think it’s more appropriate to use Plot.groupX for the line here, rather than Plot.binX:

```js
Plot.line(data,
  Plot.groupX({y: "mean"}, {
    x: "year",
    y: "economy (mpg)",
    stroke: "cylinders",
    curve: "basis"
  })
)
```

Screen Shot 2021-08-11 at 8 02 55 AM

If you use Plot.groupX, it doesn’t make much of a difference whether you use a linear or point *x*-scale (since there are no gaps in the *x*-domain here).

mbostock:
Also, since the dots are overlapping here, I feel it’s desirable to minimize occlusion by using a stroke instead of a fill. And then you don’t need the sort and reverse transforms, so the code is simpler to boot. 🙂 

```js
Plot.dot(data, Plot.binY({r: "count"}, {
  x: "year",
  y: "economy (mpg)",
  stroke: "cylinders",
  thresholds: 20
}))
```

Screen Shot 2021-08-11 at 8 09 25 AM

reubano:
Can you please add this info on using `Plot.sort` to sort binned data on its `z` access to the readme? It's mentioned but I had no idea how to actually use it until finding this issue.

Fil:
The documentation mentions two things:

In the *Plot.dot* section:
> Dots are drawn in input order, with the last data drawn on top. If sorting is needed, say to mitigate overplotting by drawing the smallest dots on top, consider a [sort and reverse transform](https://github.com/observablehq/plot#transforms).

In the *Plot.group* section
> By default, all (non-empty) groups are generated in ascending natural order.

(please feel free to suggest a better wording)

Here is an example:
```
Plot.dot(
  { length: 10000 },
  Plot.group(
    { r: "count", sort: "count", reverse: false },
    {
      x: d3.randomBinomial(40, 0.5),
      y: d3.randomBinomial(30, 0.5),
      fill: "red",
      stroke: "white"
    }
  )
).plot()
```

untitled (21)

reubano:
I saw it in one of the commits. The relevant code is here...

```js
Plot.dot(data,
  Plot.reverse(
    Plot.sort(
      "length", 
      Plot.binY(
        { r: "count" }, 
        {
          x: "year",
          y: "economy (mpg)",
          fill: "cylinders",
          thresholds: 20
        }
      )
    )
  )
)
```

Fil:
I've added that example to the notebook https://observablehq.com/@fil/plot-group-and-sort

PR #500, opened 2021-08-13
Plot.groupX({filter: () => true})

Fil:
follow-up on #499: using Plot.groupX({filter: () => true}…) creates empty points where there is no data

todo:
* [ ] the code is a bit ugly
* [ ] for feature parity with binX, the filter should receive the data, not just the index; and maybe the group too?
* [x] use InternSet
* [ ] motivate with a meaningful example
* [ ] use filter: "all" as a shorthand for `{filter: () => true}`?
* [x] optimize: we can probably sort once instead of sorting each group
* [ ] optimize: probably no need to loop through X &times; Y if there is no filter
* [ ] can probably do without sorting (ref. #414)


mbostock:
Yeah, nice, this is what I was imagining with the Cartesian product idea. Not sure how urgently we should tackle this but I like the direction. 🙂

Fil:
obsolete since v0.2!

PR #505, opened 2021-08-18
symmetric diverging

### LINKED ISSUE #504: Symmetric diverging domains centered around the pivot
Please see the example [here](https://observablehq.com/@zanarmstrong/pivot0-expected-equaldarknessfromcenter). 

When I set a diverging color scheme w/ pivot 0, I expect it to look like this. With +4 and -4 to be similarly dark (one a light green and the other a light pink). 

Screen Shot 2021-08-17 at 2 10 09 PM

Instead I saw this: 

Screen Shot 2021-08-17 at 2 11 10 PM

My hunch is that it's calculating the ramp from the pivot to the min and max, leading to a different ramp if the min is closer to the pivot than the max.

mbostock:
Fixes #504. The *symmetric* option ensures that a diverging scale encodes symmetric importance to changes below the pivot and above the pivot.

For example, the choropleth below exaggerates West Virginia’s population decline of −3% relative to Alaska’s gain of +3%. This exaggeration is caused by the domain of [−3%, +17%]: −3% is mapped to the darkest purple while +17% is mapped to the darkest green. (The pivot of 0 is mapped to the midpoint of the scheme, which is light gray here.)

Screen Shot 2021-08-17 at 8 57 05 PM

With the new *symmetric* option, the apparent difference between −3% and 0 is equal to the apparent difference between +3% and 0; equivalently, −17% is mapped to the darkest purple.

Screen Shot 2021-08-17 at 8 57 15 PM

The *symmetric* option is enabled by default. To minimize churn in the test snapshots, I disabled it on a few of the tests. Also, if the color encoding is used primarily for aesthetic purposes, maybe it’s okay to use an asymmetric encoding? I’m not entirely sure whether it’s a good option to retain if *symmetric* = false will encourage distorted encodings.

Asymmetric (before):

Screen Shot 2021-08-17 at 9 01 37 PM

Symmetric (after):

Screen Shot 2021-08-17 at 9 01 51 PM

Unfortunately, this PR is not ready to land because the *symmetric* option only works as-is on linear diverging scales. To support symmetric *diverging-log*,  *diverging-sqrt*,  *diverging-pow*, and *diverging-symlog* scales, some additional math will be needed.


Fil:
👍 Colors are not symmetric per se, but the _legends_ for asymmetric diverging scales are surprising, so I agree that it's a good default.

To be pedantic the scale for a (diverging symmetric) percentage change should be a log1p scale—we would want +50% to be symmetric with -33.3% (or in our case, +17% with -14.5%).

mbostock:
Okay, added support for transformed diverging scales in 0403cb3.

mbostock:
Here’s a little visual explainer of how the truncated interpolator works:

https://observablehq.com/@mbostock/truncated-interpolator

test/plots/grid-choropleth.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    y: {
+      axis: null
+    },
+    color: {
+      type: "diverging",
+      scheme: "piyg"
+    },
+    marks: [
+      Plot.cell(states, {x: "x", y: "y", fill: change}),
```
Fil, line 25:
```suggestion
    color: {
      type: "diverging-log",
      scheme: "piyg"
    },
    marks: [
      Plot.cell(states, {x: "x", y: "y", fill: d => 1 + change(d)}),
```

percentage changes should be log1p-scaled; it's pedantic here, since there is no noticeable difference with the values in this plot. But if you multiply line 37 by 30, the lowest value becomes -99%, undoubtedly the strongest, and it should determine the scale more than the value for DC (which has then become a mere +519%). Besides being a "best practice", this also gives us an opportunity to have a unit test for diverging-log color scales.

Fil (approved):
I've added some API documentation. The example is 👍

PR #508, opened 2021-08-18
in area and line marks, reduce grouped styles and titles

### LINKED ISSUE #503: stroke reduce for lines
in applyGroupedChannelStyles we reduce the stroke (or fill) with the "first" reducer: `([i]) => S[i]`

what syntax could we use to make this reduction explicit and configurable? By analogy with Plot.group we might have:

> Plot.line(data, {stroke: "median"}, {stroke: "value", x: …})

Fil:
closes #503

a typical example would be:

> Plot.line(data, {stroke: {value: "value", reduce: "median"}, title: {value: "value", reduce: "mode"}, x: …})

Note: a solution to #503 with the current Plot is:
```js
Plot.line(data, Plot.map({stroke: d => {
  const r = d3.median(d);
  return Array.from(d, () => r);
}}, {stroke: "value", x: …})
```

shorter version (changing only the first element of the channel, since we know that's what's used later):
```js
Plot.line(data, Plot.map({stroke: d => (d[0] = d3.median(d), d) }, {stroke: "value", x: …})
```

This PR introduces a slightly simpler method to call Plot.map. It may not be worth the extra code, documentation, and might be replaced by a link to a good notebook demonstrating Plot.map for this use case!

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,7 +683,23 @@ In addition to the [standard mark options](#marks), the following optional chann
 
 By default, the data is assumed to represent a single series (a single value that varies over time, *e.g.*). If the **z** channel is specified, data is grouped by *z* to form separate series. Typically *z* is a categorical value such as a series name. If **z** is not specified, it defaults to **stroke** if a channel, or **fill** if a channel.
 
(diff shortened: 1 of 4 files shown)


Fil:
See #724.

PR #511, opened 2021-08-19
a few changes

Fil:
* README document sort x by y
* first ladies example plot (by @tophtucker )
* add limit: [lo, hi]
* throw the correct error if a mistaken sort option is given as {x: "-y", limit: 10}

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -72,7 +72,7 @@ function Channel(data, {scale, type, value}) {
 
 function channelSort(channels, x, y) {
   let reverse, reduce, limit;
-  ({value: y, reverse = /^[-]/.test(y), reduce = true, limit = Infinity} = maybeValue(y));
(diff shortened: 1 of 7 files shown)


src/mark.js
```diff
@@ -85,7 +85,9 @@ function channelSort(channels, x, y) {
   X[1].domain = () => {
     let domain = rollup(range(XV), I => reduce.reduce(I, YV), i => XV[i]);
     domain = sort(domain, reverse ? descendingGroup : ascendingGroup);
-    if (limit < Infinity) domain = domain.slice(0, limit);
+    if (limit !== undefined) {
+      domain = domain.slice(...typeof limit === "object" ? [limit[0], limit[1]] : [0, limit]);
+    }
```
Fil, line 90: here in fact we could be more lazy, since slice(0, undefined) returns the same array.

I the case of limit being an array, I chose to pass [limit[0], limit[1]] rather than limit, but it may be too superstitious

PR #516, opened 2021-08-20
empty facets

Fil:
empty facet logic, broken out of #332 

- in axis.js, do not extend the grid on empty slots
- in facet.js, only render marks if the facet exists


Fil:
image for the CHANGELOG

empty-facet

PR #517, opened 2021-08-20
allow sort to reduce data

mbostock:
If a *sort* option is expressed as a function rather than a channel name, treat it as a data reducer. In the ballotStatusRace example this allows the facets to be sorted by percent accepted.

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -85,16 +85,15 @@ function channelSort(channels, facetChannels, data, options) {
         return domain;
       };
     } else {
-      let YV, reducer;
-      if (typeof y === "string") {
+      let YV;
+      if (y === "data") {
+        YV = data;
+      } else {
         const Y = channels.find(([name]) => name === y);
         if (!Y) throw new Error(`missing channel: ${y}`);
         YV = Y[1].value;
-        reducer = maybeReduce(reduce === true ? "max" : reduce, YV);
-      } else {
-        YV = data;
-        reducer = maybeReduce(y, YV);
       }
+      const reducer = maybeReduce(reduce === true ? "max" : reduce, YV);
       X[1].domain = () => {
         let domain = rollup(range(XV), I => reducer.reduce(I, YV), i => XV[i]);
         domain = sort(domain, reverse ? descendingGroup : ascendingGroup);
--- a/src/mark.js
+++ b/src/mark.js
@@ -49,7 +49,7 @@ export class Mark {
       const {name} = channel;
       return [name == null ? undefined : name + "", Channel(data, channel)];
     });
-    if (this.sort != null) channelSort(channels, facetChannels, this.sort);
+    if (this.sort != null) channelSort(channels, facetChannels, data, this.sort);
     return {index, channels};
   }
   plot({marks = [], ...options} = {}) {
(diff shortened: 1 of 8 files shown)


mbostock:
If you’d prefer, I also considered:

```js
sort: {
  fy: "data",
  reduce: data => data.find(d => d.status === "ACCEPTED").percent
}
```

It’s a little bit more verbose, but maybe it’s more consistent?

Fil:
👍 

I think I prefer the first version {fy: data => …}, it feels more intuitive. In the longhand this could be specified with an undefined value:

```
 sort: {
  fy: { reduce: data => data.find(d => d.status === "ACCEPTED").percent }
}

```

mbostock:
Unfortunately this

```js
sort: {fy: {reduce: data => data.find(d => d.status === "ACCEPTED").percent}}
```

doesn’t work because it is longhand for

```js
sort: {fy: {value: null, reduce: data => data.find(d => d.status === "ACCEPTED").percent}}
``` 

which is equivalent to

```js
sort: {fy: null}
``` 

which means that the *fy* scale’s domain should be driven by the _input_ order of data.

I can support your request with a patch like this:

```diff
diff --git a/src/mark.js b/src/mark.js
index b0694a7..898a68c 100644
--- a/src/mark.js
+++ b/src/mark.js
@@ -71,13 +71,13 @@ function channelSort(channels, facetChannels, data, options) {
   const {reverse: defaultReverse, reduce: defaultReduce = true, limit: defaultLimit} = options;
   for (const x in options) {
     if (!registry.has(x)) continue; // ignore unknown scale keys
-    const {value: y, reverse = defaultReverse, reduce = defaultReduce, limit = defaultLimit} = maybeValue(options[x]);
-    if (reduce == null || reduce === false) continue; // disabled reducer
+    const {value: y, reverse = defaultReverse, reduce = y == null ? null : defaultReduce, limit = defaultLimit} = maybeValue(options[x]);
+    if (y != null && (reduce == null || reduce === false)) continue; // disabled reducer
     const X = channels.find(([, {scale}]) => scale === x) || facetChannels && facetChannels.find(([, {scale}]) => scale === x);
     if (!X) throw new Error(`missing channel for scale: ${x}`);
     const XV = X[1].value;
     const [lo = 0, hi = Infinity] = limit && typeof limit[Symbol.iterator] === "function" ? limit : limit < 0 ? [limit] : [0, limit];
-    if (y == null) {
+    if (y == null && (reduce == null || reduce === false)) {
       X[1].domain = () => {
         let domain = XV;
         if (reverse) domain = domain.slice().reverse();
@@ -93,7 +93,7 @@ function channelSort(channels, facetChannels, data, options) {
         reducer = maybeReduce(reduce === true ? "max" : reduce, YV);
       } else {
         YV = data;
-        reducer = maybeReduce(y, YV);
+        reducer = maybeReduce(y == null ? reduce : y, YV);
       }
       X[1].domain = () => {
         let domain = rollup(range(XV), I => reducer.reduce(I, YV), i => XV[i]);
```

But it’s starting to feel a little convoluted and I worry about having too much overloading. We have a lot already, and it’s what gives Plot much of its conciseness, but it’s possible to go to far. 😄 

Also, note that there’s already ambiguity with a reducer object which implements reduce(I, X), so sometimes you need to say `{reduce: {reduce: (I, X) => …}}` which is a little clunky:

https://github.com/observablehq/plot/blob/6af2740d656fb6e0cb80b7f9630b147b21e0ef23/src/transforms/group.js#L173

Fil (approved):

mbostock:
Also that patch would allow this

```js
sort: {fy: {value: data => data.find(d => d.status === "ACCEPTED").percent}} 
```

which feels a little confusing: here *value* is interpreted as the *reduce* function. 🙂

Fil:
I was just thinking that value: undefined could be treated as "data" but since it's an advanced feature it doesn't matter that much.

PR #524, opened 2021-08-30
Keep style.css in ES module

juba:
The `src/style.css` file is not kept in the distributed ES module. If I understand correctly, keeping it would allow to import it directly and keep it up to date when using Plot from a package manager.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
   },
   "files": [
     "dist/**/*.js",
-    "src/**/*.js"
+    "src/**/*.js",
+    "src/**/*.css"    
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


Fil (approved):
Yes I think you're correct.

mbostock:
We can include this in the distribution, but you still won’t be able to import or require the style.css because it’s not part of the declared exports. I’m not really sure how to support this type of usage.

juba:
I'm using webpack, and with a css-loader it is possible to "import" a CSS file in a bundle. I'm doing it in another project with the Observable runtime inspector CSS, so I think it should work with Plot too.

Issue #318, opened 2021-04-09
API docs and cheat sheet


dsrivast80:
We got strong feedback from friends preview that we need to have the API reference docs as well as a API cheatsheet so folks can quickly create Plot charts.

mbostock:
The API docs are covered by the README (and notebooks); the cheat sheet is covered by #329.

PR #529, opened 2021-09-03
document the default height

Fil:
see #337 and https://observablehq.com/@observablehq/plot-default-height

(cheat sheet #318)


mbostock:
I like the notebook, but I’m tempted to make the README description more hand-wavy. I feel like either we explicit document the logic in pseudocode, or we just leave it vague like: “If a height is not specified, Plot will attempt to chose a reasonable default height based on the *y* and *fy* scale types and, for ordinal scales, the cardinality of the domain. The default height ranges from TK to TK, with 400 for the common case where *y* is quantitative and *fy* is not present.”

Fil:
We could publish the notebook and link to it for details.

Fil:
(Notebook further updated to take #837 into account.)

Fil:
18 months later, it's published :) https://observablehq.com/@observablehq/plot-default-height

Fil:
didn't mean to close the PR!

docs/features/plots.md
```diff
@@ -212,7 +212,7 @@ Plot.plot({
...
+The default **width** is 640. On Observable, the width can be set to the [standard width](https://github.com/observablehq/stdlib/blob/main/README.md#width) to make responsive plots. The default **height** is chosen automatically based on the plot’s associated scales; for example, if *y* is linear and there is no *fy* scale, it might be 396. See [“Plot default height”](https://observablehq.com/@observablehq/plot-default-height) for details. The default margins depend on the maximum margins of the plot’s constituent [marks](./plots.md#marks). While most marks default to zero margins (because they are drawn inside the chart area), Plot’s [axis mark](../marks/axis.md) has non-zero default margins.
```
Fil, line 215:
```suggestion
The default **width** is 640. On Observable, the width can be set to the [standard width](https://github.com/observablehq/stdlib/blob/main/README.md#width) to make responsive plots. The default **height** is chosen automatically based on the plot’s associated scales; for example, if *y* is linear and there is no *fy* scale, it might be 396. See [“Plot default height”](https://observablehq.com/@observablehq/plot-default-height) for details. The default margins depend on the maximum margins of the plot’s constituent [marks](#marks-option). While most marks default to zero margins (because they are drawn inside the chart area), Plot’s [axis mark](../marks/axis.md) has non-zero default margins.
```

Fil:
closing, it seems ok to not document this more in depth

PR #530, opened 2021-09-03
polylinear (aka piecewise linear) scales with an unspecified range

Fil:
**Now merged in #538**

A polylinear scale (or piecewise linear) scale is defined by example as:

~~~js
d3.scaleLinear()
  .domain([0, 100, 200, 300, 400])
  .range(["red", "blue", "green", "lime", "yellow"]) // or pixels for a position scale
~~~

In Plot, specifying the domain _and_ range works as expected, but we must also specify `type: "linear"`, otherwise Plot decides (rightly) that it might be an ordinal scale (a point scale).

However if we specify only the domain, and let the range be determined by the dimensions of the Plot (say, for the *y* axis, by autoScaleRangeY), only the two first elements of the domain are mapped to [ymin, ymax], and the resulting chart is broken.

This pull-request closes that gap.

Capture d’écran 2021-09-03 à 16 55 51

The (made-up) example could be adapted to a real-life scenario of a mission which has long periods of idle time and shorter periods of more intense activity.

~~~js
const times = [
  new Date(2013, 3, 5),
  new Date(2013, 3, 11),
  new Date(2013, 3, 14),
  new Date(2013, 3, 16),
  new Date(2013, 3, 18),
  new Date(2013, 3, 21),
  new Date(2013, 3, 27)
];

const events = [
  { date: new Date(2013, 3, 5, 13, 0), text: "Initiate" },
  { date: new Date(2013, 3, 11, 13, 0), text: "Begin" },
  { date: new Date(2013, 3, 13, 20, 0), text: "Entry" },
  { date: new Date(2013, 3, 15, 0, 0), text: "Test" },
  { date: new Date(2013, 3, 16, 0, 0), text: "Drive" },
  { date: new Date(2013, 3, 17, 8, 0), text: "Drive" },
  { date: new Date(2013, 3, 18, 15, 0), text: "Brake" },
  { date: new Date(2013, 3, 20, 10, 0), text: "Stop" },
  { date: new Date(2013, 3, 23, 14, 0), text: "Shutdown" }
];

export default async function() {
  return Plot.plot({
    grid: true,
    x: {
      domain: times,
      type: "linear",
      ticks: [...d3.timeDays(...d3.extent(times)), times[times.length-1]],
      tickFormat: d3.timeFormat("%d"),
      inset: 20,
      label: "date →"
    },
    color: { scheme: "cool" },
    marks: [
      Plot.barX(d3.pairs(times), {x1: "0", x2: "1", fill: (_,i) => i}),
      Plot.dotX(events, {x: "date", fill: "white"}),
      Plot.textX(events, {x: "date", text: "text", dx: -5, dy: -10, fill: "white", textAnchor: "start"})
    ],
    height: 90
  });
}
~~~


mbostock (approved):
👍

test/plots/polylinear.js
```diff
@@ -0,0 +1,45 @@
...
+const events = [
+  { date: new Date(2013, 3, 5, 13, 0), text: "Initiate" },
+  { date: new Date(2013, 3, 11, 13, 0), text: "Begin" },
+  { date: new Date(2013, 3, 13, 20, 0), text: "Entry" },
+  { date: new Date(2013, 3, 15, 0, 0), text: "Test" },
+  { date: new Date(2013, 3, 16, 0, 0), text: "Drive" },
+  { date: new Date(2013, 3, 17, 8, 0), text: "Drive" },
+  { date: new Date(2013, 3, 18, 15, 0), text: "Brake" },
+  { date: new Date(2013, 3, 20, 10, 0), text: "Stop" },
+  { date: new Date(2013, 3, 23, 14, 0), text: "Shutdown" }
```
mbostock, line 23: These are local times; could we use UTC to avoid being dependent on the local timezone? (And then use d3.utcDays instead of d3.timeDays _etc._ below.)

I’d also prefer if you could change your default prettier settings to avoid the spaces inside the curly braces. 🙂

Fil:
This PR is not ready because it doesn't work for color schemes. The problem is upstream, see https://observablehq.com/d/0659c57ac79f0bbc

Capture d’écran 2021-09-07 à 20 02 35


I have a patch that does this inside Plot, but we should probably fix d3-scale instead (though it might be quite a bit more complicated!).

Fil:
I've added a fix for polylinear color schemes. (It seems more difficult to fix in d3-scale where .domain() can be called several times and the original interpolate function would be lost.)

Fil:
The new fix is correct. Pffeeww!

Fil:
OOps! not quite yet

Fil:
I've verified that it's compatible with #484.

mbostock (changes requested):
This has changed quite a bit since I last reviewed, so retracting my approval until I have a chance to review again.

mbostock:
Actually, it looks like you folded this work into #538, so we should probably close this PR. (Though in the future I think it would still be my preference to keep them separate.)

PR #538, opened 2021-09-10
expose scales

Fil:
exposing scales, extracted from #484 

build at https://observablehq.com/@observablehq/expose-plot-scales-538

With many tests!


src/plot.js
```diff
@@ -90,11 +94,15 @@ export function plot(options = {}) {
...
-  figcaption.appendChild(caption instanceof Node ? caption : document.createTextNode(caption));
+  let figure = svg;
+  if (caption != null) {
+    figure = document.createElement("figure");
+    figure.appendChild(svg);
+    const figcaption = figure.appendChild(document.createElement("figcaption"));
+    figcaption.appendChild(caption instanceof Node ? caption : document.createTextNode(caption));
+  }
+
+  figure.scales = exposeScales(scaleDescriptors);
```
mbostock, line 105:
```suggestion
  figure.scale = exposeScales(scaleDescriptors);
```

Since this returns a single scale now.

src/scales.js
```diff
@@ -205,3 +206,24 @@ function coerceDate(x) {
     : x == null || isNaN(x = +x) ? undefined
     : new Date(x);
 }
+
+// prepare scales for exposure through the plot's scales() function
+export function exposeScales(scaleDescriptors) {
+  return function(key) {
```
mbostock, line 212:
```suggestion
  return key => {
```

src/scales.js
```diff
@@ -205,3 +206,24 @@ function coerceDate(x) {
...
+    if (registry.has(key))
+      return key in scaleDescriptors ? exposeScale(scaleDescriptors[key]) : undefined;
+    throw new Error(`no such scale ${key}`);
```
mbostock, line 215:
```suggestion
    throw new Error(`unknown scale: ${key}`);
```

src/scales.js
```diff
@@ -205,3 +206,24 @@ function coerceDate(x) {
     : x == null || isNaN(x = +x) ? undefined
     : new Date(x);
 }
+
+// prepare scales for exposure through the plot's scales() function
+export function exposeScales(scaleDescriptors) {
+  return function(key) {
+    if (registry.has(key))
```
mbostock, line 213:
```suggestion
    if (registry.has(key += ""))
```

src/scales.js
```diff
@@ -205,3 +206,24 @@ function coerceDate(x) {
...
+      return key in scaleDescriptors ? exposeScale(scaleDescriptors[key]) : undefined;
+    throw new Error(`no such scale ${key}`);
+  };
+}
+
+function exposeScale({scale, ...options}) {
+  for (const remove of ["domain", "range", "interpolate", "clamp", "round", "nice", "padding", "inset", "reverse", "family"]) delete options[remove];
```
mbostock, line 220: This will expose any field _not_ in this list but I think we want to be explicit about which fields are exposed, i.e., I’d prefer to invert the logic here and have an allowlist (default exclude) instead of a blocklist (default include).

mbostock (commented):
Instead of materializing a _family_ field on the internal scale object, I wonder if we can have a function that takes a specific scale type and returns either a corresponding family, or a boolean. For example:

```js
function isTemporalType(type) {
  return type === "time" || type === "utc";
}
```

Or the other approach:

```js
const quantitative = Symbol("quantitative");
const temporal = Symbol("temporal");
const ordinal = Symbol("ordinal");

function family(type, key) {
  switch (type) {
    case "diverging":
    case "diverging-sqrt":
    case "diverging-pow":
    case "diverging-log":
    case "diverging-symlog":
    case "cyclical":
    case "sequential":
    case "linear":
    case "sqrt":
    case "threshold":
    case "quantile":
    case "pow":
    case "log":
    case "symlog":
      return quantitative;
    case "identity":
      return registry.get(key) === position ? quantitative : ordinal;
    case "utc":
    case "time":
      return temporal;
    case "ordinal":
    case "categorical":
    case "point":
    case "band":
      return ordinal;
  }
}
```

This would be useful when coercing channel values based on the scale type, for example. (I haven’t reviewed the suggested code above closely…)

src/scales.js
```diff
@@ -140,8 +141,16 @@ function inferScaleType(key, channels, {type, domain, range}) {
 }
 
 // Positional scales default to a point scale instead of an ordinal scale.
-function asOrdinalType(key) {
-  return registry.get(key) === position ? "point" : "ordinal";
+function asOrdinalType(key, type = "categorical") {
```
mbostock, line 144: It’s a little surprising to see categorical here as the default, as it only applies to color scales.

Fil: True. And in any case my analysis was flawed, since in this part type is always undefined!

src/scales.js
```diff
@@ -236,17 +238,17 @@ export function exposeScales(scaleDescriptors) {
...
-function exposeScale({scale, type, label, percent, diverging, align, paddingInner, paddingOuter, transform, round}) {
+function exposeScale({scale, type, label, range, percent, diverging, align, paddingInner, paddingOuter, transform, round}) {
   if (type === "identity") return {type: "identity"};
   const domain = scale.domain();
   const pivot = diverging && domain.splice(1,1)[0];
   return {
     type,
     domain,
-    range: scale.range(),
+    ...range !== undefined && {range},
```
mbostock, line 248: This will need a defensive copy.

```suggestion
    ...range !== undefined && {range: range.slice()},
```

mbostock:
I’m going to start working on this. I rebased against main and I got a few test failures. I skipped one and regenerated the polylinear test. Is this the expected output?

Screen Shot 2021-09-24 at 5 45 23 PM

Fil:
All tests are fixed now. For reference the correct polylinear test is:
Capture d’écran 2021-09-25 à 08 18 36

(The error was in maybePiecewiseRange which was checking the old type==quantitative.)

mbostock:
Thanks for the help. I’m going to continue rewriting the tests today (schedule permitting).

src/axes.js
```diff
@@ -70,24 +72,34 @@ export function autoAxisLabels(channels, scales, {x, y, fx, fy}, dimensions) {
...
-    axis.labelAnchor = scale.type === "ordinal" ? "center"
-      : scale.reverse ? "left"
+    axis.labelAnchor = isOrdinalScale(scale) ? "center"
+      : isDescendingDomain(scale) ? "left"
       : "right";
   }
   if (axis.label === undefined) {
     axis.label = inferLabel(channels, scale, axis, "x");
   }
+  scale.label = axis.label;
```
mbostock, line 82: I think there’s a little internal confusion here on whether the label and labelAnchor properties live on the scale options, or on the axis options. Related to this, we should probably materialize the labelAnchor property in the exposed scale object, under the philosophy of materializing defaults? I’m going to think about this a bit more.

src/scales/quantitative.js
```diff
@@ -49,86 +50,119 @@ export function ScaleQ(key, scale, channels, {
...
+      scale.interpolate(interpolate);
+    }
+  } else {
+    interpolate = interpolateAuto;
   }
 
+  // TODO describe zero option
+  if (zero) {
+    if (domain[0] > 0) {
+      domain[0] = 0;
```
mbostock, line 99: This was mutating input! We should ideally have tests to protect against this.

mbostock:
Okay, I think this is 99% ready. The last things I want to think through are:

1. The semantics of the isDescendingDomain test versus the _reverse_ options. Thank you for replying to my notebook! Probably either semantics are fine, I just want to feel comfortable with it before committing.
2. The internal confusion around scale options vs. axis options for _label_ and _labelAnchor_, and specifically whether we should also materialize axis-related options in the returned scale object.

The latter one embodies the difference between Plot’s internal and external representations. Externally, scales and axes options are on the same object, but internally, we split them up. Probably we want to favor the external representation when we materialize scale objects and return them from plot.scale. Although I suppose alternatively we could add a plot.axis method in the future, and you’d be expected to splat them together if needed. But at any rate since we are already exposing the _label_ axis option on the returned scale object, I’m guessing that we do in fact want to include all the axis options in the exposed scale and use a unified representation.

PR #541, opened 2021-09-13
pass the groupings to reducers

Fil:
{x, y, z, fill, stroke} for Plot.bin, and {z, fill, stroke} for Plot.group.

ref. https://github.com/observablehq/plot/pull/437#issuecomment-917453986

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -111,15 +111,22 @@ function binn(
             for (const [x1, x2, fx] of BX) {
               const bb = fx(g);
               for (const [y1, y2, fy] of BY) {
-                const extent = {x1, x2, y1, y2};
                 const b = fy(bb);
+                const extent = {
+                  ...K && {[gk]: k},
+                  ...Z && {z: G === Z ? f : Z[b[0]]},
+                  ...F && {fill: G === F ? f : F[b[0]]},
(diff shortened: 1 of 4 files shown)


Fil:
If this looks good, I'll work on the examples (adding a title on a bar chart, for example).

src/transforms/bin.js
```diff
@@ -111,15 +111,22 @@ function binn(
...
               for (const [y1, y2, fy] of BY) {
-                const extent = {x1, x2, y1, y2};
                 const b = fy(bb);
+                const extent = {
+                  ...K && {[gk]: k},
+                  ...Z && {z: G === Z ? f : Z[b[0]]},
+                  ...F && {fill: G === F ? f : F[b[0]]},
+                  ...S && {stroke: G === S ? f : S[b[0]]},
+                  ...x1 != null && {x1, x2},
+                  ...y1 != null && {y1, y2}
```
mbostock, line 121:
```suggestion
                  ...BX1 && {x1, x2},
                  ...BY1 && {y1, y2}
```

But I’m not sure we need to do this conditionally. I think it’s fine if as before these fields are always present but sometimes undefined.

Fil:
obsolete, and can be done differently (e.g. by passing the group key in the title channel, or having a custom reducers).

PR #545, opened 2021-09-20
handle no ticks when rotating

### LINKED ISSUE #544: Error with no data and nonzero tickRotate
Drawing an empty plot with nonzero tick rotation throws an error:

```
Plot.plot({
  marks: [Plot.frame()],
  x: { tickRotate: 1 }
})
```

Example: https://observablehq.com/d/de0d642085846c07

mbostock:
Fixes #544.

Since I was there, I also took a moment to rewrite the code so to make slightly fewer assumptions about the structure of the axis: namely, rather than only looking at the first element’s attributes, look at each element individually. I don’t seeing this having any immediate practical effect, but it does seem slightly cleaner.

There aren’t any tests for the tickRotate option. Any chance you’d be willing to add some to make sure I haven’t broken anything here, @Fil?

(I tested it in [this notebook](https://observablehq.com/d/d3feddc757b564e2) and it seemed to still work…)

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -220,23 +220,24 @@ function createAxis(axis, scale, {ticks, tickSize, tickPadding, tickFormat}) {
     .tickValues(Array.isArray(ticks) ? ticks : null);
 }
 
+const radians = Math.PI / 180;
+
 function maybeTickRotate(g, rotate) {
   if (!(rotate = +rotate)) return;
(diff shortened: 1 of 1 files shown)


mbostock:
I have to do a release since I just bumped D3 to 7.0.3, so I’m going to land this while I’m at it.

Fil:
looks good! thank you. test added in #546

PR #551, opened 2021-09-24
expose window, normalize

### LINKED ISSUE #549: Allow several reduce in Plot.window / Plot.map
In https://github.com/observablehq/plot/issues/548, when we create a bollinger band we would need y2 and y1 to be two different reducers on the same window.

Proposed syntax:
```js
Plot.windowY({
  y1: { value: "close", reduce: data => d3.mean(data) - K * d3.deviation(data) },
  y2: { value: "close", reduce: data => d3.mean(data) + K * d3.deviation(data) },
  k: …
})
```

or, keeping reduce at the top level of the options:

```js
Plot.windowY({
  reduce: {
    y1: data => d3.mean(data) - K * d3.deviation(data),
    y2: data => d3.mean(data) + K * d3.deviation(data)
  },
  y1: "close",
  y2: "close",
  k: …
})
```

mbostock:
Fixes #549. This way we can build on top of the window (and normalize) transform.


Fil (approved):
👍 👍 👍

just a suggestion to zoom in the test plot

test/plots/aapl-bollinger.js
```diff
@@ -0,0 +1,20 @@
...
+    marks: [
+      Plot.areaY(AAPL, Plot.map({y1: bollinger(20, -2), y2: bollinger(20, 2)}, {x: "Date", y: "Close", fillOpacity: 0.2})),
+      Plot.line(AAPL, Plot.map({y: bollinger(20, 0)}, {x: "Date", y: "Close", stroke: "blue"})),
+      Plot.line(AAPL, {x: "Date", y: "Close", strokeWidth: 1})
+    ]
+  });
+}
+
+function bollinger(N, K) {
+  return Plot.window({k: N, reduce: Y => d3.mean(Y) + K * d3.deviation(Y)});
```
Fil, line 19:
```suggestion
  return Plot.window({k: N, reduce: Y => d3.mean(Y) + K * d3.deviation(Y), anchor: "end"});
```

I think BB are a trailing moving-average?

Also I see this note in wikipedia:
> Since Bollinger Bands use the population method of calculating standard deviation, the proper divisor for the sigma calculation is n, not n − 1, which would mean:

```suggestion
  return Plot.window({k: N, reduce: Y => d3.mean(Y) + K * d3.deviation(Y) * Math.sqrt(N/(N-1)), anchor: "end"});
```

(I don't think we need that much detail tho?)

mbostock: The second point is fascinating but I think I’m just going to ignore it for now.

test/plots/aapl-bollinger.js
```diff
@@ -0,0 +1,20 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export default async function() {
+  const AAPL = await d3.csv("data/aapl.csv", d3.autoType);
```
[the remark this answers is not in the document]
mbostock: Thanks for the suggestion, but I think it’s fine as-is for a test. 👍

PR #552, opened 2021-09-24
numeric interval & documentation

Fil:


src/transforms/interval.js
```diff
@@ -6,6 +6,10 @@ import {maybeInsetX, maybeInsetY} from "./inset.js";
 // chose between UTC and local time (or better, an explicit timeZone option).
 function maybeInterval(interval) {
   if (interval == null) return;
+  if (typeof interval === "number" && interval) {
```
mbostock, line 9: What’s the truthy `interval` test for? Testing for nonzero, non-NaN? I think I’d prefer to drop this because it’s not sufficient to ensure good behavior (for example, Infinity or -Infinity will produce NaN below), and it’s just simpler to avoid trying.

```suggestion
  if (typeof interval === "number") {
```

Alternatively, we could ignore NaN specifically:

```suggestion
  if (typeof interval === "number" && !isNaN(interval)) {
```

This will still break for zero and infinities, but those are at least considered defined values, so it’s probably reasonable that it breaks.

src/transforms/interval.js
```diff
@@ -6,6 +6,10 @@ import {maybeInsetX, maybeInsetY} from "./inset.js";
...
+    const n = interval;
+    interval = { floor: d => n * Math.floor(d / n), offset: d => d + n };
```
mbostock, line 11:
```suggestion
    // Note: this offset doesn’t support the optional step argument for simplicity.
    interval = {floor: d => n * Math.floor(d / n), offset: d => d + n};
```

README.md
```diff
@@ -801,7 +801,9 @@ The following channels are optional:
 * **x2** - the ending horizontal position; bound to the *x* scale
 * **y2** - the ending vertical position; bound to the *y* scale
 
-Typically either **x1** and **x2** are specified, or **y1** and **y2**, or both. The rect mark supports the [standard mark options](#marks), including insets and rounded corners. The **stroke** defaults to none. The **fill** defaults to currentColor if the stroke is none, and to none otherwise.
+Typically either **x1** and **x2** are specified, or **y1** and **y2**, or both. **x1** and **x2** can be derived from **x** and an **interval** object (such as d3.utcDay) with a **floor** method that returns *x1* from *x* and an **offset** method that returns *x2* from *x1*. If the interval is specified as a (non-null) number, *x1* and *x2* are taken as the two consecutive multiples of *n* that bracket *x*. The interval may be specified either as as {x, interval} or x: {value, interval}—typically to apply different intervals to x and y.
```
mbostock, line 804:
```suggestion
Typically either **x1** and **x2** are specified, or **y1** and **y2**, or both. **x1** and **x2** can be derived from **x** and an **interval** object (such as d3.utcDay) with a **floor** method that returns *x1* from *x* and an **offset** method that returns *x2* from *x1*. If the interval is specified as a number (and not NaN), *x1* and *x2* are taken as the two consecutive multiples of *n* that bracket *x*. The interval may be specified either as as {x, interval} or x: {value, interval}—typically to apply different intervals to x and y.
```

mbostock (changes requested):
Nice generalization! 👍

src/transforms/interval.js
Fil: I'm OK with not trying :)

mbostock (approved):

PR #554, opened 2021-09-25
scheme subset

Fil:
Often we want to use a scheme such as "greens", but its light end is too light and doesn't contrast against the white background. Tweaking the *domain* means we have to know it in advance. This PR explores tweaking the *range* instead, to change the mapping from the data domain to the interpolator's domain.

Currently, if a range is defined, the scheme is ignored. With this PR, when both are present the scheme is not ignored and defines the color; the range then defines the subset of the scheme that we want to interpolate on.

Capture d’écran 2021-09-25 à 11 39 38

Note: WIP. There are no tests yet; no documentation; and this builds up on #538.

Try it in https://observablehq.com/d/e5cbada7ee2a257c

Related: https://observablehq.com/@d3/piecewise-color-interpolation


mbostock:
Yes, I was thinking the same as I was working on this last night. 👍 I need to think through the exact syntax but I like the idea of supporting this!

mbostock:
Fixed in #560 and followup commit https://github.com/observablehq/plot/commit/c92b02211d39ab90a56db884ee9a0d16fd5a8170.

PR #559, opened 2021-09-26
scale unknown

### LINKED ISSUE #446: Can't set unknown domain in colour scale
I'm using the color scale with explicit domain/range to associate status names with colours:

```
  color: {
    type: "categorical",
    domain: ["success", "failed"],
    range: ["green", "red"]
  },
```
But status names I forgot to list come out invisible.  A way of setting the `unknown()` value would be useful, it looks like it is hardcoded to `undefined` at https://github.com/observablehq/plot/blob/068f8d8125a910ee1a528c8b299a79c27ca1a9a5/src/scales/ordinal.js#L206.

mbostock:
Fixes #446.


README.md
```diff
@@ -171,7 +171,8 @@ A scale’s domain (the extent of its inputs, abstract values) and range (the ex
 
 * *scale*.**domain** - typically [*min*, *max*], or an array of ordinal or categorical values
 * *scale*.**range** - typically [*min*, *max*], or an array of ordinal or categorical values
-* *scale*.**reverse** - reverses the domain, say to flip the chart along *x* or *y*
+* *scale*.**unknown** - the desired output value (defaults to undefined) for input values outside the domain
```
Fil, line 174: to be more precise unknown only applies to:
- ordinal or categorical scales (not point or band); for values that are outside the domain.
- quantitative scales; for undefined, null or NaN.

Fil (approved):

README.md
mbostock: Good point. Added documentation.

PR #560, opened 2021-09-26
expose range for continuous color scales

mbostock:
Supersedes #554.


Fil (approved):
will definitely be useful when we have #554 (or something similar)

PR #563, opened 2021-09-27
symmetric diverging scales by domain extension

### LINKED ISSUE #562: The symmetric transform gets applied twice when reusing diverging scales
Since we don’t expose whether a diverging scale is symmetric, and instead bake the behavior into the interpolator, the symmetric transform (which defaults to true) gets applied twice.

mbostock:
Rather than truncating the interpolator, this extends the domain so that it is symmetric around the pivot. While the effect is the same internally, this has two improvements.

First, it fixes #562 by returning a domain that is already symmetric, and hence applying the (default) symmetric transform again is idempotent and has no effect.

Second, it makes it easier to produce a good color legend for symmetric diverging scales. With the previous approach, you’d see something like this:

Screen Shot 2021-09-26 at 8 08 24 PM

While strictly accurate, it’s a bit visually unsettling and difficult to interpret. The lower half of the legend is given equal visual space to the upper half, but covers a smaller range of values ([-0.78, 0] vs. [0, 1.35]). Under the new approach in this PR, you’d instead get:

Screen Shot 2021-09-26 at 8 10 05 PM

The extended domain includes values that are not present in the chart, but the resulting interpolator no longer needs transformation, and the legend is much easier to interpret.

In retrospect, I now realize why you had symmetric = false when exposing diverging scales. 😅 But while returning symmetric = false would also fix #562, I think it’s likely to be confusing if people inspect the exposed diverging scale, and wonder why is says it’s asymmetric when you asked it to be symmetric. The approach here avoids that confusion, and makes it easier to produce a good color legend.

### DIFF
--- a/src/scales/diverging.js
+++ b/src/scales/diverging.js
@@ -51,10 +51,11 @@ function ScaleD(key, scale, transform, channels, {
 
   // Normalize the interpolator for symmetric difference around the pivot.
   if (symmetric) {
-    const mindelta = Math.abs(transform(min) - transform(pivot));
-    const maxdelta = Math.abs(transform(max) - transform(pivot));
-    if (mindelta < maxdelta) interpolate = truncateLower(interpolate, mindelta / maxdelta);
-    else if (mindelta > maxdelta) interpolate = truncateUpper(interpolate, maxdelta / mindelta);
+    const mid = transform.apply(pivot);
+    const mindelta = mid - transform.apply(min);
+    const maxdelta = transform.apply(max) - mid;
+    if (mindelta < maxdelta) min = transform.invert(mid - maxdelta);
+    else if (mindelta > maxdelta) max = transform.invert(mid + mindelta);
   }
 
   scale.domain([min, pivot, max]).unknown(unknown).interpolator(interpolate);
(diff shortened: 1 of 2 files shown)


Fil:
A third possibility for the legend would be to show it with the data domain, but on a uniform x:

Capture d’écran 2021-09-27 à 11 32 29

But this would mean changing *Legend* rather than Plot?

mbostock:
That would require exporting the symmetric flag, and then we’d need to display the legend as a _sequential_ rather than diverging scale, which feels confusing. 

Alternatively, we would need to expose the numeric range representing the slice of the interpolator (e.g., [0.25, 1]), and then somehow incorporate that into the diverging scale to produce a legend. That would require changes to d3-scale that I don’t want to tackle now.

mbostock:
Also, the way the truncated interpolator worked previously, the pivot is still always exactly in the middle of the interpolator (t = 0.5) so we would need to make other changes… like probably at that point it would make the most sense to map a diverging scale to a linear (sequential) scale, and then you wouldn’t be able to inspect the pivot.

Fil (approved):

Fil:
ok — we can get back to this idea later

mbostock:
I think mapping diverging scales to linear (sequential) scales is probably a nice simplification for the future.

PR #564, opened 2021-09-28
add Plot.shuffle

mbostock:
I was confused why {sort: Math.random} didn’t work, but it’s because Math.random.length is zero, and so it’s interpreted as a comparator rather than an accessor. 😞 But anyway, now you can say Plot.shuffle(_options_) which seems nice.


src/transforms/sort.js
```diff
@@ -2,6 +2,10 @@ import {ascendingDefined} from "../defined.js";
 import {valueof} from "../mark.js";
 import {basic} from "./basic.js";
 
+export function shuffle(options) {
+  return basic(options, sortValue(Math.random));
+}
```
Fil, line 7:
```suggestion
import {basic} from "./basic.js";
import {randomUniform, randomLcg} from "d3";

export function shuffle({seed, ...options}) {
  return basic(options, sortValue(randomUniform.source(randomLcg(seed))()));
}
```

Fil (changes requested):
Do we have a test plot where it's useful?

For the documentation, just after Plot.sort:

~~~
### Plot.shuffle(*options*)

```js
Plot.shuffle({seed: 42, ...options})
```

Randomizes the data order. Pass a seed to make the shuffling reproducible.


~~~

src/transforms/sort.js
Fil: maybe
`sortValue((seed ? randomUniform.source(randomLcg(seed)) : randomUniform)())`

mbostock: Okay, added a seed option. We don’t need randomUniform here; we can use Math.random directly.

Fil (approved):

PR #565, opened 2021-09-30
consume bin options; allow bin options on output

mbostock:
To facilitate the composition of bin transforms:

1. Options to the bin transform (_domain_, _cumulative_, and _thresholds_) may now be specified on the _outputs_ object, similar to the other transforms. This makes it clear which transform is the intended recipient of the option without needing to use the spread operator.

2. Options to the bin transform (_domain_, _cumulative_, and _thresholds_) are now consumed rather than passing through. For consistency, the stack transform now likewise consumes its options (_order_, _offset_, and _reverse_).

3. The *x1* and *x2* input channels are now consumed if *x* is an output; likewise *y1* and *y2* are consumed if *y* is an output. Thus, Plot.binX when outputting *x* will no longer produce an *x1* and *x2* channel.

4. The *count* reducer no longer inherits the input channel’s label. For example, if we’re producing a count of binned dates, we want to label the count as a frequency, not as a date.


Fil (approved):
I've added a PR to do the same with groups—it doesn't need to be merged immediately but made me more confident that this was all good.

PR #567, opened 2021-10-01
asymmetric insets

### LINKED ISSUE #520: Asymmetric insets for scales
Currently the scale.inset option only allows for symmetric insets on both sides. But we could easily allow separate insetLeft and insetRight for the *x* scale, and insetTop and insetBottom for the *y* scale.

https://github.com/observablehq/plot/blob/cc24e85f81745e88bbedcf7f1b29e6345ba1a28a/src/scales.js#L13

mbostock:
Fixes #520.


Fil (approved):
added documentation and tests

mbostock:
I like having tests, but I think it makes them less effective as examples if we use obscure features solely for the purpose of testing. So I’m going to revert these changes to tests. It’d be nice to have either standalone or unit tests (rather than snapshot tests). But I’m also not too worried about this feature breaking.

Fil:
I thought it made these two charts a bit nicer :) But yeah… unit tests will be fine

PR #569, opened 2021-10-01
same with groups

Fil:


test/plots/stargazers-hourly-group.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+    marks: [
+      Plot.barY(stargazers,
+        Plot.groupX({y: "count", title: "first"},
+          Plot.binY(
+            {x: d => d.length < 5 ? `${d.length}` : "5+", thresholds: d3.utcHour, title: "first"},
+            {y: "date", fill: d => d.date.getDay(), title: d => "SMTWTFS"[d.date.getDay()]}
+          )
+        )
+      ),
```
mbostock, line 21: If we write this as…

```suggestion
      Plot.barY(stargazers,
        Plot.groupX({y: "count", title: "first"},
          Plot.binX(
            {x: d => d.length < 5 ? `${d.length}` : "5+", thresholds: d3.utcHour, title: "first"},
            {x: "date", fill: d => d.date.getDay(), title: d => "SMTWTFS"[d.date.getDay()]}
          )
        )
      ),
```

Then this example works as-is in the parent PR. But I think it still makes sense that if the group transform outputs a _y_ channel, that this would mask the input _y1_ and _y2_ channels, even though the group transform didn’t use those channels for anything. Just thinking it through.

PR #570, opened 2021-10-01
rank, quantile

mbostock:
Screen Shot 2021-10-01 at 8 26 30 AM

```js
{
  const randomNormal = d3.randomNormal.source(d3.randomLcg(42))();
  const randoms = Array.from({length: 300}, randomNormal);
  return Plot.plot({
    marks: [
      Plot.dotX(randoms, Plot.mapY("quantile", {y: randoms}))
    ]
  });
}
```


Fil:
Cool :)

We already had
`Plot.dotX(d3.sort(randoms), { y: (d, i) => i / (randoms.length - 1) }).plot()`

but this is obviously nicer since it doesn't need to re-order the data (facets!), and knows what to do with NaNs.

Fil (approved):

PR #572, opened 2021-10-01
coerce to date for temporal thresholds

mbostock:
If a time interval or an array of dates is specified as the _thresholds_ option to the bin transform, the bin transform now implicitly coerces the values to the dates in the same fashion as a temporal scale.


Fil (approved):
Very cool. In README I wonder if the definition of "interval" could be a little more explicit.

PR #573, opened 2021-10-01
allow quantile or threshold position scales

mbostock:
Fixes a crash where we tried to pass an undefined *range* to the scaleQuantile and scaleThreshold constructors, and then fixes an off-by-one in autoScaleRange for threshold scales, which require a range of length domain.length + 1.


Fil (approved):
Oops! and yes.

Issue #356, opened 2021-05-03
Adding a caption breaks the width option


Fil:
At least in Observable:
https://observablehq.com/d/d43f43d1f047af73

git-ashish:
@Fil The referenced notebook doesn't appear to be public. Could you please share that? 
I tried adding captions to some plots along with width and it seemed to work.

Fil:
I have just shared it.

git-ashish:
Looks like its an issue with Observable's CSS which restricts the width of a `figure` element to 640px. 
![image](https://user-images.githubusercontent.com/2697421/117465392-4774ce00-af6f-11eb-8f2e-d4d65c3bc03a.png)

I have tested it locally and it seems to be working as expected. The specified width gets applied to the plot.

git-ashish:
However, wondering if the dimensions should be applied to outermost element i.e. `figure`

Fil:
see also #397

PR #583, opened 2021-10-26
legends - 2

### LINKED ISSUE #23: Color legend.

Fil:
Continuation of https://github.com/observablehq/plot/pull/539 ; build at https://observablehq.com/d/58e2c0596402e08f

In this PR the legends <del>are</del> [EDIT: were initially] implemented as
```js
Plot.legend({color: mycolor, width: 640})
```
where mycolor is a plot or a plot scale obtained with plot.scale("color").

The same goes with {opacity:} or {r:}.

I tend to find this API less usable than the previous attempt, in the sense that:
* we can't just say "legend: true" in the color scale's definition and have a color legend embedded in the chart
* we can't override the function used to draw the legend
* it's good to be able to pass options here, but frustrating that we cannot pass them in the scale definition directly. 
* passing plot.scale("color") seems redundant

UPDATE: they are now implemented as
```js
chart.legend("color", {width: 640})
```

The legends themselves have been adapted from d3's color-legend notebook (with the ramp or swatches); the opacity legend is a simple greyscale ramp; the radius legend is a bunch of based circles.

(It's also very possible that not every type of scale works—there will be a lot of variations to test.)

TODO:
- [x] rebase on main (needs to embed the swatches css in the legend)
- [x] a bug that can be fixed quickly: the scale's label isn't used
- [x] unique css selector for swatches
- [x] accept a label for swatches?
- [x] remove Plot.legend({color: chart}) and replace with chart.legend("color", {})
- [x] test all types of color scales
- [x] remove radius legend, opacity legend
<del>-  remove the scoped css from swatches?</del>

LATER
- [x] add {legend: true} option to the scale description (#588)
- [x] introduce radius legend (#586), opacity legend (#587), x, y, fx, fx legends in separate PRs

Fixes #23.


src/legends/swatches.js
```diff
@@ -0,0 +1,119 @@
...
+    for (const value of color.domain()) {
+      const d = elems.append("div").classed(`${uid}-item`, true);
+      d.append("div")
+        .classed(`${uid}-block`, true)
+        .style("background", color(value));
+      const label = format(value);
+      d.append("div")
+        .classed(`${uid}-label`, true)
+        .text(label)
+        .attr("title", label.replace(/["&]/g, entity));
```
mbostock, line 95: This escaping shouldn’t be necessary, and in fact I think it introduces escaping problems? Escaping happens internally when you call element.setAttribute (within selection.attr).

src/legends/ramp.js
```diff
@@ -21,8 +23,9 @@ export function legendRamp(scale, {
       .attr("width", width)
       .attr("height", height)
       .attr("viewBox", [0, 0, width, height])
-      .style("overflow", "visible")
-      .style("display", "block");
+      .each(addStyle(style));
+    if (svg.style("overflow") === "") svg.style("overflow", "visible");
+    if (svg.style("display") === "") svg.style("display", "block");
```
mbostock, line 28: Can we move the overflow: visible; and display: block; to the inline stylesheet so that we don’t need this conditional check?

Fil:
remaining issues:

- there's still a bit of duck-typing in ramp.js—need to identify continuous scales
- the style option doesn't allow to set font-family, because it's not cascading

src/legends/ramp.js
Fil: hmm… ramps don't have an internal css yet (and no className either); and there might be a similar problem with css priorities.

src/legends.js
```diff
@@ -0,0 +1,30 @@
...
+    marginLeft: color.marginLeft,
+    ticks: color.ticks,
+    tickFormat: color.tickFormat,
+    tickValues: color.tickValues,
+    format: color.format,
+    className: color.className,
+    swatchSize: color.swatchSize,
+    swatchWidth: color.swatchWidth,
+    swatchHeight: color.swatchHeight,
+    columns: color.columns,
```
mbostock, line 25: Why are these options also allowed under *color*? Like, I think we want you to say

```js
Plot.legend({color: …, width: 640})
```

rather than 

```js
Plot.legend({color: {…, width: 640}})
```

Fil: OK, the confusion was coming from my unit tests, where I had a mix of scale options and legend options. I've cleaned this. I've left _ticks_, _tickFormat_, _tickValues_ and _format_ in comments as it would feel consistent to have them as scale options (as they already are for positional scales). But it's not necessary.

Fil:
I've tried to remove as much as I could, but we need a d3 scale to display the horizontal axis.

For instance, it doesn't seem useful to replace
```
    x = color.copy().rangeRound(quantize(interpolate(marginLeft, width - marginRight), n));
```

with 
```
    x = Scale("x", undefined, {
      ...scale,
      interpolate: undefined,
      range: quantize(interpolate(marginLeft, width - marginRight), n),
      round: true
    }).scale;
```

src/figure.js
```diff
@@ -0,0 +1,12 @@
+
+// Wrap the plot in a figure with a caption, if desired.
+export function figureWrap(svg, {width}, caption) {
+  if (caption == null) return svg;
+  const figure = document.createElement("figure");
+  figure.style = `max-width: ${width}px`;
```
mbostock, line 6: This appears to be sneaking in a fix for #356? I’m going to remove this; I’d rather address it separately.

mbostock:
Looks like asymmetric diverging scales aren’t rendered correctly. Haven’t figured out why yet, though.

Screen Shot 2021-11-24 at 2 07 23 PM
```js
Plot.legend({
  color: {
    type: "diverging",
    symmetric: false,
    domain: [1, 4],
    pivot: 3,
    scheme: "PiYG"
  }
})
```

mbostock:
I’m also noticing that options on the color scale aren’t passed through to the legend in the _plot_.legend("color") case, which is because _plot_.scale("color") doesn’t return those options on the exposed scale. I’d guess that we should expose the axis/legend options on the exposed scale, though? Because if you’re sharing a scale across charts, I think you’d expect the legends to be the same, too, at least by default.

mbostock:
It looks like sqrt transforms (and possibly others) aren’t rendered correctly: the sqrt transform should apply to the tick positions, not the interpolator, but it seems to be applying to both. For example, here’s a sqrt color scale:

Screen Shot 2021-11-24 at 2 21 31 PM

```js
Plot.plot({
  color: {
    domain: [0, 10],
    type: "sqrt"
  },
  marks: [
    Plot.cellX(d3.range(10))
  ]
})
```

Note that the value 2 is bright green. The legend for this suggests it should be orange:

Screen Shot 2021-11-24 at 2 24 13 PM

```js
Plot.plot({color: {type: "sqrt", domain: [0, 10]}}).legend("color")
```

src/legends.js
```diff
@@ -0,0 +1,21 @@
+import {exposeScale, Scale} from "./scales.js";
+import {legendColor} from "./legends/color.js";
+
+export function legend({color, ...options}) {
+  if (color != null) {
+    return legendColor(exposeScale(Scale("color", undefined, color)), {
```
mbostock, line 6: This will mutate color.type. 😬

mbostock (approved):
I made [a few changes](https://github.com/observablehq/plot/compare/5e6ef47c329db3a950e9eba39541635392594589~...fil/legends-2) 😉 but this looks good to me now! To summarize:

* The legend implementations now take a scale descriptor (the internal representation), rather than the external representation returned by _plot_.scale(…). In the case where you call Plot.legend directly, the external representation is promoted to the internal one.
* When _plot_.legend(…) is used, the _label_, _ticks_, and _tickFormat_ scale options are propagated to the legend.
* I made the legend styles consistent with Plot (adopting tabular-nums and system-ui), and adopted the same strategy of an embedded style element to allow inline styles to take precedence.
* I rewrote a good part of the legend code; mostly to take advantage of Plot’s stricter scale definitions and options, and to clean up the swatches implementation.
* I fixed the exposed representation of diverging scales to include symmetric: false so that the symmetric transform isn’t erroneously re-applied.
* I removed the “normalize the options” code in Plot. The normalized options weren’t used anywhere, and I want the non-normalized options available as legend options.
* I avoided the mutation of options.type in Scale.
* I cleaned up the internal representation of diverging scales (and niced scales) so that the _domain_ is always available.
* I split the color legend tests into separate snapshots which made it easier for me to debug and review changes to the snapshots. Even just having names is helpful to describe what the test is trying to do (though comments are always welcome).

Fil:
YES

PR #584, opened 2021-10-28
Allow the color: {scheme, range} syntax for ordinal scales

Fil:
to select a part of the domain of an interpolator.

(Only continuous color schemes are authorized in that case, by construction: it's not possible to select part of the category10 scheme.)

Issue raised by @zanarmstrong in https://observablehq.com/d/3a2e33eeffe5b95d


mbostock (approved):
This feels a little overloaded to me—I would have expected the explicit range to take priority over the scheme. But, seems convenient nonetheless!

PR #587, opened 2021-10-29
opacity legend

Fil:
A ramp with increasing opacity.

We need to test various possibilities; maybe a checkered background would be nicer; maybe swatches might be useful for some types of opacity scales.

### DIFF
--- a/src/legends/opacity.js
+++ b/src/legends/opacity.js
@@ -1,11 +1,19 @@
+import {rgb} from "d3";
 import {legendColor} from "./color.js";
 
-export function legendOpacity({type, interpolate, ...scale}, {legend = "ramp", ...options}) {
+const black = rgb(0, 0, 0);
+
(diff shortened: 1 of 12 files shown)


src/legends/opacity.js
```diff
@@ -0,0 +1,19 @@
...
+
+const black = rgb(0, 0, 0);
+
+export function legendOpacity({type, interpolate, ...scale}, {
+  legend = "ramp",
+  color = black,
+  ...options
+}) {
+  if (!interpolate) throw new Error(`${type} opacity scales are not supported`);
+  if (`${legend}`.toLowerCase() !== "ramp") throw new Error(`${legend} opacity legends are not supported`);
```
mbostock, line 12: I added some explicit limitations here; I imagine we’ll add support for other opacity scale types in the future.

mbostock (approved):
Seems like a useful start to me, even if limited!

PR #588, opened 2021-11-02
- legend: true

Fil:
preparing for legend: true after #583


mbostock:
Superseded by 952fc08.

PR #594, opened 2021-11-24
constant numeric x and y for text marks

### LINKED ISSUE #523: For text, allow x and y to be specified as constant numbers
For manual labels, it’d be nice if you could specify *x* and *y* as constant numbers. Currently they can either be a channel or null; in the latter case the labels are placed in the center of the chart.

Related, it might also be nice to have an *anchor* option which allows the text to be positioned somewhere else relative to the surrounding frame; this would eliminate the need to hard-code *dx* and *dy* (e.g., in the [geofacet example](https://observablehq.com/@data-workflows/geofacet)). The *anchor* option could accept values such as *top*, *top-left*, *top-right*, *right*, etc.

Fil:
position the text relative to the anchor

anchor can be any of "top", "bottom", "left", "right" or combinations such as "top-left"

demo: https://observablehq.com/@observablehq/text-facet-alignment-523

closes #523

todo:
- [x] create unit tests for the possible anchors
- [ ] should "right" and "bottom" be counted "inwards" like in css?
- [ ] generalize to all marks?


src/marks/text.js
```diff
@@ -44,16 +47,17 @@ export class Text extends Mark {
     this.fontStyle = string(fontStyle);
     this.fontVariant = string(fontVariant);
     this.fontWeight = string(fontWeight);
+    this.cx = cx;
+    this.cy = cy;
+    this.anchor = anchor;
```
mbostock, line 52: We should parse and validate the anchor here in the constructor, rather than deferring validation until use.

src/marks/text.js
```diff
@@ -44,16 +47,17 @@ export class Text extends Mark {
...
     this.dy = string(dy);
   }
   render(I, {x, y}, channels, dimensions) {
     const {x: X, y: Y, rotate: R, text: T, fontSize: FS} = channels;
-    const {width, height, marginTop, marginRight, marginBottom, marginLeft} = dimensions;
     const {rotate} = this;
     const index = filter(I, X, Y, R).filter(i => nonempty(T[i]));
-    const cx = (marginLeft + width - marginRight) / 2;
-    const cy = (marginTop + height - marginBottom) / 2;
+    const c = textPosition(dimensions, this.anchor, this.cx, this.cy);
```
mbostock, line 60:
```suggestion
    const [cx, cy] = textPosition(dimensions, this.anchor, this.cx, this.cy);
```

Fil:
A question: here we add x and y regardless of the anchor, but we could decide that the sign of the "right" and "bottom" positions should be negated (to go "inset"). I don't know which is the most intuitive.

Fil:
converted to draft with the two questions that came up in review

Fil:
I've now generalized the feature to dots and images. Not sure if it can/should be done for other marks?

Fil:
Demo https://observablehq.com/@fil/constant-numeric-x-y-594

Fil:
Obsolete. This can be achieved with frameAnchor: left and dx: 10 instead of x: 10.

PR #599, opened 2021-11-29
Image mark

### LINKED ISSUE #157: Image mark?

mbostock:
Supersedes #459. Fixes #157.

TODO

- [x] width and height instead of r
- [x] width defaults to height and vice versa
- [x] src instead of href
- [x] ignore fill and stroke attributes
- [x] opacity attribute
- [x] allow src to be a constant
- [x] documentation


RLesser:
Thanks for incorporating this Mike!

mbostock:
> How would we allow for clipping—for example, for the very common task of making rounded avatars?

My thinking is we could reintroduce the _r_ channel for this. Or we could introduce another mark type. Or we could add an _image_ channel for the dot mark.

PR #603, opened 2021-12-08
Add normalize reducers min, max, deviation

mkfreeman:
Added four more reducers that can be set as the `basis` for `Plot.normalize`:
- min
- max
- deviation
- <del>variance

See [this notebook](https://observablehq.com/@observablehq/normalize-reducer-testing) for a demo.


mbostock:
If you use the “deviation” basis, it means that you’re taking a set *X* = {*x1*, *x2*, *x3*, …} and mapping them to *X*′ = {*x1* / *d*, *x2* / *d*, *x3* / *d*, …} where *d* = deviation(*X*). Is that what you expect? That seems surprising to me. I think that I would expect something more like {(*x1* - *m*) / *d*, (*x2* - *m*) / *d*, …} where *m* = median(*X*).

Fil (changes requested):
max is a nice addition (make it fit!), and min is symmetric for negative values. <del>We might also want to normalize by _extent_ (box all the values between min and max). (EDITED: we already have extent)</del>

I can also see the use-case for deviation, in particular with a fixed domain of, say, [-3, 3]. Agree with @mbostock, that it should be centered, but I'd argue for centering around the _mean_ not the median, per the std dev definition.

Re: variance, unless there is a real world use-case that I don't know of, I don't think it's a good idea: the variance is not in the same unit as the original data, but in unit squared, so the normalized value of data in meters would result in values expressed in m^-1, which I think is very confusing.

mbostock:
I agree with Fil’s recommendations.

mkfreeman:
The `min` normalization creates a plot with the domain `[Infinity, Infinity]` if the minimum value is zero -- is this an expected (acceptable) behavior?
Screen Shot 2021-12-08 at 8 43 28 PM

src/transforms/normalize.js
```diff
@@ -61,6 +64,18 @@ const normalizeLast = normalizeBasis((I, S) => {
...
+    }
+  }
+};
+
+const normalizeMax = normalizeBasis((I, S) => max(I, i => S[i]));
 const normalizeMean = normalizeBasis((I, S) => mean(I, i => S[i]));
 const normalizeMedian = normalizeBasis((I, S) => median(I, i => S[i]));
-const normalizeSum = normalizeBasis((I, S) => sum(I, i => S[i]));
+const normalizeMin = normalizeBasis((I, S) => min(I, i => S[i]));
+const normalizeSum = normalizeBasis((I, S) => sum(I, i => S[i]));
```
mbostock, line 81: Please preserve the trailing newline at the end of this file. 🙏

mbostock (commented):
Implementation looks good. 👍 Would you mind also updating the README, too?

src/transforms/normalize.js
mkfreeman: Updated (and preserved!). Let me know if you prefer a different phrasing or ordering (I left them as alphabetical, with `extent` and `deviation` at the end, as they follow a different pattern than those above)

README.md
```diff
@@ -1429,12 +1429,16 @@ If a function is used, it must return an array of the same length as the given i
 
 The Plot.normalizeX and Plot.normalizeY transforms normalize series values relative to the given basis. For example, if the series values are [*y₀*, *y₁*, *y₂*, …] and the *first* basis is used, the mapped series values would be [*y₀* / *y₀*, *y₁* / *y₀*, *y₂* / *y₀*, …] as in an index chart. The **basis** option specifies how to normalize the series values. The following basis methods are supported:
 
+
```
mbostock, line 1432: Now there’s an extra newline. 😂

mkfreeman: Fixed 🤦

mbostock (approved):
👍

Fil (approved):
I think it's normal to crash on min if the minimum is 0 (the max normalization will also give NaNs if the max is zero.)

For deviation, we can avoid a crash since in that case, all values are equal, and it makes sense to normalize the value to 0 (it will happen, for example, on a facet for which there is only one data point: better to show it that not).

I've added a few unit tests, and have decorated your commits with your actual email (instead of <xxx@xxxx-MacBook-Pro.local>).

In your terminal, you can do:
> $ git config --global user.name "John Doe"
> $ git config --global user.email johndoe@example.com

PR #604, opened 2021-12-08
figure max-width: initial

### LINKED ISSUE #356: Adding a caption breaks the width option

mbostock:
I’m not sure this is the “right” fix, but it feels like the safest and least invasive change to start. This way the figure max-width matches the SVG max-width. (It also roughly matches [what we do with Inputs.table](https://github.com/observablehq/inputs/blob/fac7439b463e0d3fdba9f633f6ec2d6259a99d41/src/style.css#L207).)

Fixes #356.


Fil:
why do you prefer 100% in this case rather than max-width: initial like in the Inputs? (both seem to work in Observable)

mbostock:
I thought you might ask that… No particular reason, really. The initial value of max-width is none, so we could just set it to that, too. initial, none, and 100% should be equivalent if there isn’t an explicit width set on the figure element. I suppose using max-width: none or max-width: initial would be less opinionated than max-width: 100% if there is, though, so I can switch to that.

Fil (approved):

PR #606, opened 2021-12-10
font-variant on the axes

### LINKED ISSUE #601: Only apply font-variant: tabular-nums to axis ticks
We’re currently applying it to everything, which doesn’t look great on axis labels and text marks that aren’t strictly numbers. I think it’d probably be a better default to only apply to axis ticks (which are typically numeric or dates, both of which would benefit from tabular numerals) and perhaps selectively text marks where the _text_ channel is formatted numbers (and perhaps dates). It doesn’t look good in swatches, either.

Fil:
remove font-variant from the global styles; introduce the axis fontVariant option, which defaults to tabular-nums

closes #601

### DIFF
--- a/src/axes.js
+++ b/src/axes.js
@@ -16,8 +16,8 @@ export function Axes(
   if (!fxScale) fxAxis = null; else if (fxAxis === true) fxAxis = xAxis === "bottom" ? "top" : "bottom";
   if (!fyScale) fyAxis = null; else if (fyAxis === true) fyAxis = yAxis === "left" ? "right" : "left";
   return {
-    ...xAxis && {x: new AxisX({grid, line, label, ...inferFontVariant(xScale), ...x, axis: xAxis})},
-    ...yAxis && {y: new AxisY({grid, line, label, ...inferFontVariant(yScale), ...y, axis: yAxis})},
+    ...xAxis && {x: new AxisX({grid, line, label, fontVariant: inferFontVariant(xScale), ...x, axis: xAxis})},
+    ...yAxis && {y: new AxisY({grid, line, label, fontVariant: inferFontVariant(yScale), ...y, axis: yAxis})},
     ...fxAxis && {fx: new AxisX({name: "fx", grid: facetGrid, label: facetLabel, ...fx, axis: fxAxis})},
     ...fyAxis && {fy: new AxisY({name: "fy", grid: facetGrid, label: facetLabel, ...fy, axis: fyAxis})}
   };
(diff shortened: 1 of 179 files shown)


Fil:
A comment: for Plot.text, I think it would be too magical to do auto-detection, and it's easy enough to use the fontVariant option. This PR removes font-variant from the static styles, and introduces a new axis option called fontVariant, which defaults to tabular-nums. (Just moving the static style to the axis wouldn't allow to undo the feature… and this approach seemed more consistent with our general patterns.)

src/axis.js
```diff
@@ -23,6 +24,7 @@ export class AxisX {
     this.tickSize = number(tickSize);
     this.tickPadding = number(tickPadding);
     this.tickFormat = tickFormat;
+    this.fontVariant = string(fontVariant);
```
mbostock, line 27:
```suggestion
    this.fontVariant = impliedString(fontVariant, "normal");
```

src/axis.js
```diff
@@ -106,6 +111,7 @@ export class AxisY {
...
```
mbostock, line 114:
```suggestion
    this.fontVariant = impliedString(fontVariant, "normal");
```

mbostock (changes requested):
Nice! But I think we should only default to tabular-nums for quantitative (and probably also temporal?) scales. For categorical/ordinal scales, we should default to normal.

src/axes.js
```diff
@@ -16,8 +16,8 @@ export function Axes(
   if (!fxScale) fxAxis = null; else if (fxAxis === true) fxAxis = xAxis === "bottom" ? "top" : "bottom";
   if (!fyScale) fyAxis = null; else if (fyAxis === true) fyAxis = yAxis === "left" ? "right" : "left";
   return {
-    ...xAxis && {x: new AxisX({grid, line, label, ...x, axis: xAxis})},
-    ...yAxis && {y: new AxisY({grid, line, label, ...y, axis: yAxis})},
+    ...xAxis && {x: new AxisX({grid, line, label, ...inferFontVariant(xScale), ...x, axis: xAxis})},
+    ...yAxis && {y: new AxisY({grid, line, label, ...inferFontVariant(yScale), ...y, axis: yAxis})},
```
mbostock, line 20: Am I missing something, or doesn’t this need to pass xScale.fontVariant into the axis (if present)?

mbostock: Ah yes, I missed the “...x” and “...y”.

mbostock (approved):

Issue #460, opened 2021-07-23
Multi-line tick labels


yurivish:
It would be useful to be able to make axes like this:

<br /><br />

![image](https://user-images.githubusercontent.com/114388/126826951-c60e6b6c-6836-4b40-9868-3d29966bcf4c.png)

One way would be to support multi-line text labels. 

Another option would be to allow SVG nodes to be used as ticks rather than coercing the output of `tickFormat` to a string, which is the current behavior.

Here is a notebook which "almost works": https://observablehq.com/d/7de079c962f419f8

mbostock:
Datawrapper does this well (see multi-line axis ticks):

https://blog.datawrapper.de/new-axis-ticks/

Fil:
The code is not a problem once we agree on a spec. It's not clear to me if datawrapper's format is the way to go, but I can see how we would extend tickFormat to support \n for multiline, and \n? for a non-repeating second line, enabling this:

`x: {tickFormat: "%b\n?%Y"}`

and getting this:
<img width="651" alt="Capture d’écran 2021-12-11 à 00 57 29" src="https://user-images.githubusercontent.com/7001/145655294-463a3c83-5448-4d96-84e9-e9c443934117.png">

thumb up, down, or provide alternative ideas :)

I'll send the PR tomorrow because it's already too late in this time zone.

This should arguably be done in d3-axis instead of Plot. Acting on a \n in tick formats might be considered as a breaking change (at least visually speaking)—but it seems like a highly useful change?

Fil:
related: #327 #506

mbostock:
I think supporting \n is probably the way to go, yeah. I don’t understand how \n? works in your example above. How does it know not to render the year outside of January? I think more likely we’d implement a “multi-scale” format that’s similar to what d3-scale does, but using new lines.

Fil:
The ? will avoid _repetitions_, like DataWrapper does:
![](https://blog.datawrapper.de/wp-content/uploads/2021/03/full-210303_grid4.png)

In the example above it's not clear because the first tick is January, but if it was, say, "**Apr** Jul Oct **Jan** Apr" we'd show 2020 under the first **Apr**, and 2021 under **Jan**.

A multiscale format would work too, maybe by introducing a notation `%0Y` which would show the year only for January. I'll send my PR as a draft so we can tinker a bit.

mbostock:
Ah, I see! Neat. I like that. The repetition rule applies to each line distinctly, then (not to % fields)?

[...]

PR #609, opened 2021-12-11
multiline ticks labels, draft

Fil:
On the x axis, split text labels on \n; use a ? at the start of a line to cancel consecutive repeats

It's a draft that helps tinkering, but it will probably not be merged, because if we're taking this route, it should happen in d3-axis instead (and we'll add the "line" option in there as well).

#460


Fil:
A (better) iteration in d3-axis: https://github.com/d3/d3-axis/pull/83

Issue #114, opened 2021-01-31
Grid color, opacity, and dash-array


mbostock:
There’s not an easy way to change the color of grid lines (without also changing the ticks, tick labels, axis labels, etc.). There should be, and it’d be nice if we could demonstrate a ggplot2-style theme.

ericemc3:
It is aesthetically interesting to differentiate the appearance of certain grid lines:

![image](https://user-images.githubusercontent.com/17407796/118007499-7bc00400-b34c-11eb-818b-0625ea3ca5ea.png)

Note : this is possible with vega-lite: with the properties `gridOpacity` or `gridDash` for example, to which one can associate a function.

PR #612, opened 2021-12-15
custom grid ticks

### LINKED ISSUE #7: Custom grid density or values.
Currently *grid* is strictly a boolean, but true could be equivalent to the corresponding ticks, a number could compute new ticks (*e.g.*, for higher density) and an array could contain explicit tick values.

Fil:
Fixes #7.

Demo: https://observablehq.com/@observablehq/plot-custom-grid-612


Fil:
I might go back to one mark per grid tick, which would also help support grid lines with various styles (as in https://github.com/observablehq/plot/issues/114#issuecomment-839899434)

Fil:
Not sure how to design it so that the styles (opacity: 0.1) could be modified through the plot's root _style_ option. It's a generic question though (see #397 #253).

mbostock (changes requested):
If you specify explicit values for the scale’s grid, we should extend the default domain to cover those values. I’d also prefer to stick to the existing attributes, rather than extending the inline stylesheets, for consistency with d3-axis. I’m cool with adding a “grid” class to differentiate these grid lines from ticks, though.

Fil:
superseded by #985

PR #618, opened 2021-12-17
throw an error if there is not enough space for the x or y scales

### LINKED ISSUE #616: Insets should not allow the range to invert
If the inset is big enough, it will invert the range, which can be very confusing!

Fil:
demo: https://observablehq.com/@observablehq/collapsing-ranges-618

the most difficult part was to come up with a somehow relevant error message

closes #616


Fil:
superseded by #640

PR #620, opened 2021-12-17
tickFormat: null should not create ticks

Fil:
(unifies ramp and swatches)

this is equivalent to setting 0 ticks; an alternative could be to set tickFormat to () => "". I have no preference on the matter.


mbostock:
Hmm, not sure I agree. Maybe you just want to draw the tick marks without the labels? If you want to suppress the ticks I would say that `ticks: null` is more semantic than `tickFormat: null`.

Fil:
PR switched to () => null; the idea is to have a similar behavior between ramp and swatches when using the tickFormat: null option.

PR #621, opened 2021-12-17
labels for swatches

Fil:
Note: the scale’s label is not shown for swatches, since most of the time it is obvious from the context. But it can be nice to write Plot.legend(…, label:"Industry"):

Capture d’écran 2021-12-17 à 23 57 18


mbostock (changes requested):
I think my preference is that we show the scale’s label by default, consistently across legend types, but that you can disable it by setting label: null. That way we don’t need to special-case the behavior for swatches.

Fil:
rebased

src/legends/swatches.js
```diff
@@ -93,6 +93,14 @@ function legendItems(scale, {
         --swatchHeight: ${+swatchHeight}px;
       `);
 
+  const palette = label
+      ? swatches.call(div => div.append("p")
+          .text(label)
+          .style("font-weight", "bold")
+          .style("width", "100%"))
```
mbostock, line 100: The parent element here has flex because the code expects that it contains the swatches. We need to change the styles to match the structure—we shouldn’t be setting the width style.

src/legends/swatches.js
```diff
@@ -93,6 +93,14 @@ function legendItems(scale, {
         --swatchHeight: ${+swatchHeight}px;
       `);
 
+  const palette = label
+      ? swatches.call(div => div.append("p")
```
mbostock, line 97: I think we probably don’t want to use a P element here, and instead should have a DIV with more specific margins as needed. Probably slightly less vertical spacing between the label and the swatches than in this current example.

Screen Shot 2022-03-30 at 8 08 45 AM

mbostock:
I’d like to revert the unrelated formatting changes…

Fil:
I'll do that. The goal of the formatting changes was to produce an output with consistent alignement.

Fil:
rebased—I still need to figure out why some alignments in the tests seem wrong

Fil:
superseded by #1885

PR #625, opened 2021-12-20
Expose the scale on the stand-alone Plot.legend

### LINKED ISSUE #624: can the standalone `Plot.legend` returned object have `.scale` to get the scale Descriptor?
sometimes it's better to create a standalone legend (with `Plot.legend`), not from any pre-existing plot, and use `htl.html ...` with css to organize better layout,

from https://observablehq.com/@observablehq/plot-legends?collection=@observablehq/plot I'm playing around a bit:

image

you see the sportsPlot (returned value from `Plot.plot`) is an `HTMLElement` with `.scale` and `.legend`

> HTMLElement {scale: ƒ(e), legend: ƒ(n, r)}

but if I give it a name to `wx = Plot.legend({...})` to whatever value returned from `Plot.legend`, it has nothing other than an empty `SVGSVGElement {}`

```js
wx = Plot.legend({
  color: {
    // interpolate: (t) => wavelengthToColor(400 + t * 350),
    domain: [400, 750]
  },
  width: 350,
  ticks: 10,
  label: "Wavelength (nm) →"
})
```

then what's an easy way to get the scale object? and where to get d3-scale function from it?

image

I'm expecting it can have a way to return the scale object, something like `sportsScale` in the example,

```
> lgd.scale("color")
Object {
  type: "ordinal",
  domain: Array(4) ["A", "B", "H", "I", "J", ...],
  range: Array(10) ["#1f77b4", "#ff7f0e", ...],
  label: "...",
  apply: ƒ(e),
}
```

current workaround may be to use `fakePlot = Plot.plot(...)` create a fake plot first, and get the scale object from `... = fakePlot.scale(...)`, and then the `fakePlot` is not needed and wasted

Fil:
Closes #624

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -1,4 +1,4 @@
-import {normalizeScale} from "./scales.js";
+import {normalizeScale, exposeScales} from "./scales.js";
 import {legendColor} from "./legends/color.js";
 import {legendOpacity} from "./legends/opacity.js";
 import {isObject} from "./mark.js";
@@ -10,9 +10,13 @@ const legendRegistry = new Map([
 
 export function legend(options = {}) {
   for (const [key, value] of legendRegistry) {
-    const scale = options[key];
-    if (isObject(scale)) { // e.g., ignore {color: "red"}
-      return value(normalizeScale(key, scale), legendOptions(scale, options));
+    const desc = options[key];
+    if (isObject(desc)) { // e.g., ignore {color: "red"}
+      const scale = normalizeScale(key, desc);
+      return Object.assign(
+        value(scale, legendOptions(desc, options)),
+        {scale: exposeScales({[key]: scale})}
+      );
     }
   }
   throw new Error("unknown legend type");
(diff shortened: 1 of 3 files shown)


src/legends.js
```diff
@@ -10,9 +10,13 @@ const legendRegistry = new Map([
...
   for (const [key, value] of legendRegistry) {
-    const scale = options[key];
-    if (isObject(scale)) { // e.g., ignore {color: "red"}
-      return value(normalizeScale(key, scale), legendOptions(scale, options));
+    const desc = options[key];
+    if (isObject(desc)) { // e.g., ignore {color: "red"}
+      const scale = normalizeScale(key, desc);
+      return Object.assign(
+        value(scale, legendOptions(desc, options)),
+        {scale: exposeScales({[key]: scale})}
```
mbostock, line 18: It’s a little more complicated than this now that we have the combined _symbol_ + _color_ legend. That said I’m not sure it’s worth the trouble of implementing this feature.

mbostock (changes requested):

Fil:
Closing for now.

Issue #597, opened 2021-11-26
Interval (or series?) transform for line and area


mbostock:
It’d be nice to support the interval transform for line and area marks such that if there’s missing data, a gap is automatically drawn rather than interpolating. I imagine that it would only be supported on lineX, lineY, areaX, and areaY, where it would apply to y, x, y, and x respectively.

Fil:
It would also be useful for textY: https://github.com/observablehq/plot/discussions/611

Fil:
I love the idea, which seems simpler than the bin+empty bins filter that we currently have. The difficulty is that instead of a _point_ being defined, we want to control when _two consecutive points_ are connected, and create a multiline. Consecutive points are determined in the final call to shapeLine, but their original values are already scaled and can't be compared with the interval. (It's the first time that an issue comes from having applied the scales on the channels.) 🤔

mbostock:
Here’s a “real world” example: https://observablehq.com/@mbostock/npm-daily-downloads?name=@observablehq/mtcars

This package is rarely downloaded, and the npm endpoint doesn’t return an entry when there were zero downloads. So you see this:

![untitled (17)](https://user-images.githubusercontent.com/230541/155907555-5607373d-81fc-40ce-bcdd-6d67e690fb25.png)

But what you should see is this:

![untitled (18)](https://user-images.githubusercontent.com/230541/155907583-1c41f051-dbf9-4135-a695-58a9267d0a34.png)

This can be done by changing this:

```js
Plot.areaY(data, {x: "date", y: "value", fill: "steelblue", curve: "step-before"})
```

To this:

```js
Plot.areaY(data, Plot.binX({y: "sum", filter: null}, {x: "date", interval: d3.utcDay, y: "value", fill: "steelblue", curve: "step-before"}))
```

Which makes me think maybe we just need a convenience shorthand? Something like…

```js
function seriesX(interval, options) {
  return binX({y: "sum", filter: null}, {...options, interval});
}
```

Then you’d say

```js
Plot.areaY(data, Plot.seriesX(d3.utcDay, {x: "date", y: "value", fill: "steelblue", curve: "step-before"}))
```

Or maybe it’d be something supported by the mark, e.g.,

```js
Plot.areaY(data, {x: "date", series: d3.utcDay, y: "value", fill: "steelblue", curve: "step-before"})
```

[...]

PR #627, opened 2021-12-21
extend interval to dot and text marks

Fil:
closes #611 

(not related to #597 which requires a very different kind of interval)

### DIFF
--- a/src/transforms/interval.js
+++ b/src/transforms/interval.js
@@ -1,5 +1,5 @@
 import {range} from "d3";
-import {labelof, maybeValue, valueof} from "../options.js";
+import {isTemporal, labelof, maybeValue, valueof} from "../options.js";
 import {maybeInsetX, maybeInsetY} from "./inset.js";
 
 // TODO Allow the interval to be specified as a string, e.g. “day” or “hour”?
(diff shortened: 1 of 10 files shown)


src/transforms/interval.js
```diff
@@ -9,9 +9,19 @@ function maybeInterval(interval) {
   if (typeof interval === "number") {
     const n = interval;
     // Note: this offset doesn’t support the optional step argument for simplicity.
-    interval = {floor: d => n * Math.floor(d / n), offset: d => d + n};
+    interval = {
+      floor: d => n * Math.floor(d / n),
+      mid: d => n * (Math.floor(d / n) + 0.5),
+      offset: d => d + n
+    };
```
mbostock, line 16: Also, this isn’t sufficient to test for a numeric interval, since you could provide your own implementation of the floor and offset functions, and we’d have to call those to determine whether or not your interval is numeric or temporal. This code assumes every implementation is a temporal interval. (See isTimeInterval, as I mentioned.)

mbostock:
Thanks for incorporating my suggested changes.

I’d love to have a better motivating example of how this should be used. In the case of plotting athletes by birthday month, I think I would probably recommend this instead:

Screen Shot 2022-01-01 at 10 21 33 AM

```js
Plot.plot({
  marginLeft: 60,
  y: {
    reverse: true
  },
  marks: [
    Plot.rectX(athletes, Plot.binY({x: "count"}, {y: d => d.date_of_birth, thresholds: d3.utcMonth})),
    Plot.text(athletes, Plot.binY({x: "count", text: "count"}, {y: d => d.date_of_birth, thresholds: d3.utcMonth, dx: 4, textAnchor: "start"})),
    Plot.ruleX([0])
  ]
})
```

Alternatively, if I wanted an ordinal _y_-scale using the group transform:

Screen Shot 2022-01-01 at 10 21 54 AM

```js
Plot.plot({
  y: {
    tickFormat: Plot.formatMonth()
  },
  marks: [
    Plot.barX(athletes, Plot.groupY({x: "count"}, {y: d => d.date_of_birth.getUTCMonth()})),
    Plot.text(athletes, Plot.groupY({x: "count", text: "count"}, {y: d => d.date_of_birth.getUTCMonth(), dx: 4, textAnchor: "start"})),
    Plot.ruleX([0])
  ]
})
```

I can reduce your example to this, which produces identical output to the first:

```js
Plot.plot({
  marginLeft: 60,
  y: {
    reverse: true
  },
  marks: [
    Plot.rectX(athletes, Plot.groupY({x: "count"}, {y: d => d3.utcMonth.floor(d.date_of_birth), interval: d3.utcMonth})),
    Plot.textX(athletes, Plot.groupY({x: "count", text: "count"}, {y: d => d3.utcMonth.floor(d.date_of_birth), interval: d3.utcMonth, dx: 4, textAnchor: "start"}))
  ]
})
```

But, it still seems more complicated since you need to apply the month interval twice, whereas with the bin transform, you only need to specify it once, and the bin transform already computes the midpoint. Though… I guess that alone is a reason for the interval transform to do the same, for symmetry.

mbostock:
Also, and apologies for not realizing this sooner, we should find a way to reuse the existing `mid` helper (used by the bin transform) for the interval transform so that the two behave consistently:

https://github.com/observablehq/plot/blob/72b2adbc28ed12a2fb2de800af8b57bd14a63154/src/mark.js#L277-L292

mbostock:
This came up again here: https://talk.observablehq.com/t/text-marks-for-the-sum-of-last-n-days-when-using-plot-windowy/6627/2

PR #632, opened 2021-12-23
fix buylrd

Fil:
before/after:

Capture d’écran 2021-12-23 à 18 32 12

Capture d’écran 2021-12-23 à 18 32 23


mbostock (approved):
Oops! Good catch.

Might be worth adding a few more tests if you have a moment.

PR #633, opened 2021-12-23
test schemes

Fil:
tests generated at https://observablehq.com/@observablehq/test-plot-schemes

follow-up on https://github.com/observablehq/plot/pull/632#pullrequestreview-839548033


mbostock:
I don’t think we would notice if the hexadecimal values wouldn’t match. I was thinking of a visual test? E.g., if these are snapshot tests, then we can look at the SVG visually and see if it matches the expectations. Though, it’d be awkward to have many variations of each color scheme.

Fil:
True. This is a non-regression test *after* I did a visual control, by looking at the (forthcoming) schemes cheatsheet—which is how I discovered the bug.

mbostock:
My point was that these tests would not reveal a bug, only a change, so I’d rather have visual tests—I’d rather be able to see the colors. I can put something together.

mbostock:
Closing in favor of #637 (and #638).

PR #636, opened 2021-12-30
add first and last reducers for window

mbostock:
Ref. https://github.com/observablehq/plot/discussions/635#discussioncomment-1888720


Fil (approved):
I reached for it and it wasn't there—had put it on my todo-list :)

PR #637, opened 2021-12-31
test all color schemes

mbostock:
Alternative to #633.

Screen Shot 2021-12-30 at 4 25 56 PM

Screen Shot 2021-12-30 at 4 25 59 PM

### DIFF
--- a/src/scales/schemes.js
+++ b/src/scales/schemes.js
@@ -143,7 +143,7 @@ const ordinalSchemes = new Map([
 
 function scheme9(scheme, interpolate) {
   return ({length: n}) => {
-    n = n > 3 ? Math.floor(n) : 3;
+    n = Math.max(3, Math.floor(n));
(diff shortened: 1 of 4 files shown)


mbostock:
I think this identified a few more bugs…

Burd and buylrd are inverted with n=2 ordinal domains:

Screen Shot 2021-12-30 at 4 28 11 PM

These quantitative schemes don’t handle n=1 ordinal domains well:

Screen Shot 2021-12-30 at 4 28 56 PM

We should probably truncate cubehelix since it’s unlikely that you’ll want to go from 100% black to 100% white, especially with n=2:

Screen Shot 2021-12-30 at 4 29 34 PM

Warm and cool appear to have a divide-by-zero bug with n=1:

Screen Shot 2021-12-30 at 4 29 58 PM

Fil (approved):

PR #639, opened 2021-12-31
standalone scales with Plot.scale

### LINKED ISSUE #624: can the standalone `Plot.legend` returned object have `.scale` to get the scale Descriptor?

mbostock:
Alternative to #625. This lets you create the standalone scale directly, without requiring either a legend or a chart (but you can pass the scale to either as desired). Fixes #624.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,12 @@ const color = plot.scale("color"); // retrieve the color scale object
 console.log(color.range); // inspect the color scale’s range, ["red", "blue"]
 ```
 
+You can also create a standalone scale with Plot.**scale**(*options*):
+
+```js
+const color = Plot.scale({color: {type: "linear"}});
+```
+
 To reuse a scale across plots, pass the scale object into another plot specification:
 
 ```js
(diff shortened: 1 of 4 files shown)


Fil (approved):
Excellent. I don't think it invalidates #625 though (but it makes it less urgent).

PR #641, opened 2021-12-31
“mid” interval transform

mbostock:
I prefer to keep the two transforms distinct: we’ll compute the midpoint of the interval for dot, text, and image, and the extent of the interval for other marks. This also allows the “mid” interval transform to avoid memoization.

In addition to your fix to pass *data* to the interval’s transform function in case x2 in computed before x1, I’ve also fixed the memoization to be bypassed if the *data* changes. This isn’t currently possible in Plot, but I could imagine us wanting to support data changing in the future, so fixing while I’m here.

### DIFF
--- a/src/transforms/interval.js
+++ b/src/transforms/interval.js
@@ -9,9 +9,19 @@ function maybeInterval(interval) {
   if (typeof interval === "number") {
     const n = interval;
     // Note: this offset doesn’t support the optional step argument for simplicity.
-    interval = {floor: d => n * Math.floor(d / n), offset: d => d + n};
+    interval = {
+      floor: d => n * Math.floor(d / n),
+      mid: d => n * (Math.floor(d / n) + 0.5),
+      offset: d => d + n
+    };
   }
   if (typeof interval.floor !== "function" || typeof interval.offset !== "function") throw new Error("invalid interval");
+  if (typeof interval.mid !== "function") {
+    interval = {
+      ...interval,
+      mid: x => (x = interval.floor(x), new Date((+x + +interval.offset(x)) / 2))
+    };
+  }
   return interval;
 }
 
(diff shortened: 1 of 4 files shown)


src/transforms/interval.js
```diff
@@ -27,14 +27,34 @@ function maybeIntervalK(k, maybeInsetK, options) {
...
     ...options,
-    [k]: {transform: (data) => tv1(data).map(v => mid(interval.offset(v), v)), label},
-    [`${k}1`]: v1 === undefined ? {transform: tv1, label} : v1,
-    [`${k}2`]: v2 === undefined ? {transform: (data) => tv1(data).map(v => interval.offset(v)), label} : v2
+    [k]: {
+      label: labelof(v),
+      transform: data => valueof(data, value).map(v => {
+        const a = interval.floor(v);
+        const b = interval.offset(a);
+        return a instanceof Date ? new Date((+a + +b) / 2) : (a + b) / 2;
```
mbostock, line 55: We could use isTimeInterval here (and thus perform the check once, outside the loop).

Fil: Instead I've added an interval.mid function—is this ok (#627)?

src/transforms/interval.js
```diff
@@ -9,9 +9,19 @@ function maybeInterval(interval) {
...
     // Note: this offset doesn’t support the optional step argument for simplicity.
-    interval = {floor: d => n * Math.floor(d / n), offset: d => d + n};
+    interval = {
+      floor: d => n * Math.floor(d / n),
+      mid: d => n * (Math.floor(d / n) + 0.5),
+      offset: d => d + n
+    };
   }
   if (typeof interval.floor !== "function" || typeof interval.offset !== "function") throw new Error("invalid interval");
+  if (typeof interval.mid !== "function") {
```
mbostock, line 19: This means we’ll have to wrap every interval with this object, even though we won’t end up using it in most places. I think it’d be better to only create the mid function if it’s needed.

PR #642, opened 2022-01-01
fix mid interval computation

mbostock:


src/transforms/interval.js
```diff
@@ -59,7 +49,13 @@ function maybeIntervalMidK(k, maybeInsetK, options) {
     ...options,
     [k]: {
       label: labelof(v),
-      transform: data => valueof(data, value).map(interval.mid)
+      transform: data => {
+        const V1 = valueof(data, value).map(v => interval.floor(v));
+        const V2 = V1.map(v => interval.offset(v));
+        return V1.map(isTemporal(V1)
+          ? (v1, i) => new Date((+v1 + +V2[i]) / 2)
+          : (v1, i) => (+v1 + +V2[i]) / 2);
```
mbostock, line 57: This probably needs to handle `v1` or `V2[i]` being null here. I’d guess in that case the transform should produce undefined for dates (to match coerceDate) and NaN for numbers (to match coerceNumber).

src/transforms/interval.js
```diff
@@ -59,7 +49,13 @@ function maybeIntervalMidK(k, maybeInsetK, options) {
     ...options,
     [k]: {
       label: labelof(v),
-      transform: data => valueof(data, value).map(interval.mid)
+      transform: data => {
+        const V1 = valueof(data, value).map(v => interval.floor(v));
```
mbostock: It might not be safe to use valueof(data, value).map here if the channel is provided as a typed array and the interval returns something else, such as a date. I’ve added a commit that switches to Array.from.

src/transforms/interval.js
mbostock: I’ve added a commit that handles missing data here. But, it occurs to me that interval.floor and interval.offset won’t handle missing data, either, so… 🤷

Issue #577, opened 2021-10-14
Plot.text should default to tabular figures / numbers


visnup:
If you plot some numbers as text with a `text-anchor: end`, it'd be nice if the numbers were [tabular and aligned horizontally](https://www.fonts.com/content/learning/fontology/level-3/numbers/proportional-vs-tabular-figures) by default.

For example:

```javascript
Plot.plot({
  marks: [
    Plot.ruleX([0], { stroke: 'lightgray' }),
    Plot.text(
      [
        ["a", 3.4],
        ["b", 2.1],
        ["c", 12.8]
      ],
      { y: "0", text: "1", textAnchor: "end", dx: 7.5 }
    )
  ],
  width: 100
})
```

Produces:
![untitled](https://user-images.githubusercontent.com/10444/137234852-85782242-d45f-4abe-a2a2-83db06cd1fb2.png)

Where the decimal is not quite aligned due to the thinner `1` in `2.1`.

I'm pretty sure there are some arcane CSS styles that may fix this.

tophtucker:
imo even regardless of textAnchor!

PR #643, opened 2022-01-01
implicit format for numbers and dates

### LINKED ISSUE #476: Format dates and numbers in title and text
Not sure if applies is only to titles or more generally for all dates when coercing them to strings, but we'd want to avoid this:

Capture d’écran 2021-08-03 à 13 31 32

when we have e.g.:
`Plot.dot(data, { x: "date", y: "count", title: "date" })`

mbostock:
Fixes #476. Related #577.


src/memoize.js
```diff
@@ -0,0 +1,20 @@
...
+    let cached = cacheKey.length === keys.length;
+    if (cached) {
+      for (let i = 0; i < cacheKey.length; ++i) {
+        if (cacheKey[i] !== keys[i]) {
+          cached = false;
+          break;
+        }
+      }
+    }
+    if (!cached) {
```
Fil:
```suggestion
    if (cacheKey.length !== keys.length || cacheKey.some((k, i) => k !== keys[i])) {
```

is there any drawback to using array.some or array.every?

mbostock: Simplified as you suggested. 👍

Fil (approved):

PR #645, opened 2022-01-01
href channel

### LINKED ISSUE #283: Hyperlinks?

mbostock:
Adds a generic href channel that wraps the specified mark elements with an A element with the appropriate xlink:href attribute. A target option is also available (e.g., “_blank”). I didn’t implement it for grouped marks (lines and areas), but maybe I should. Fixes #283.


src/style.js
```diff
@@ -141,6 +147,17 @@ export function applyDirectStyles(selection, mark) {
   applyAttr(selection, "opacity", mark.opacity);
 }
 
+function applyHref(selection, href, target) {
+  selection.each(function(i) {
+    if (href[i] != null) {
+      const a = document.createElementNS(namespaces.svg, "a");
+      a.setAttributeNS(namespaces.xlink, "href", href[i]);
+      if (target != null) a.setAttribute("target", target);
+      this.parentNode.insertBefore(a, this).appendChild(this);
```
Fil, line 156: makes we wish we could do selection.wrap("svg:a") (related: https://github.com/d3/d3-selection/pull/296)

Fil:
wonder if we shouldn't have target: "_blank" by default?

mbostock:
> wonder if we shouldn't have target: "_blank" by default?

I don’t think so, no. I do wonder whether it’s possible to set the equivalent of a `<base target>` for the SVG though, so we don’t have to repeat the target option everywhere? It’d be nice to set it on the base plot somehow… but it’s not a huge deal to set it on each link, too.

mbostock:
Any other changes, @Fil? Or LGTY?

Fil (approved):
it's good, we'll just have to add tests

src/style.js
Fil: selection.wrap: https://github.com/d3/d3-selection/pull/297

PR #646, opened 2022-01-02
line+dots

### LINKED ISSUE #628: Line with dot
It’d be nice to have a Line+Dot as a built-in primitive, rather than needing a separate Line and Dot mark. Mainly because if you have a multi-series line, you want the dots to be z-interleaved with the lines, rather than drawing all the dots on top of all the lines. Also it’s nice because you don’t have to repeat yourself (and transfer the line’s stroke to the dot’s fill).

Fil:
It's so much complication I don't think we should do it.

demo https://observablehq.com/@observablehq/plot-line-and-dots-628

closes #628

### DIFF
--- a/src/marks/line.js
+++ b/src/marks/line.js
@@ -1,8 +1,8 @@
 import {create, group, line as shapeLine} from "d3";
 import {Curve} from "../curve.js";
-import {defined} from "../defined.js";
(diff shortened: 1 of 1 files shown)


Fil:
superseded by #731

PR #648, opened 2022-01-04
dodge (beeswarms)

### LINKED ISSUE #164: Beeswarm plot?
Some sort of dodge transform?

mbostock:
Inspired by @yurivish’s [Building a Better Beeswarm](https://observablehq.com/@yurivish/building-a-better-beeswarm) and @jtrim-ons’s [non-force-directed beeswarms](https://observablehq.com/@jtrim-ons/non-force-directed-beeswarm-of-mobility-version-3-no-union-o), but using [interval-tree-1d](https://github.com/mikolalysenko/interval-tree-1d) for simpler (and hopefully faster?) intersection testing.

Screen Shot 2022-01-03 at 7 49 46 PM

```js
Plot.plot({
  width: 764,
  height: 400,
  color: {
    scheme: "purd",
    range: [0.15, 0.95]
  },
  r: {
    type: "linear",
    range: [0, 20]
  },
  marks: [
    Plot.dot(dataset, Plot.dodgeY({x: "x", r: "r", fill: "r"}))
  ]
})
```

Screen Shot 2022-01-03 at 7 49 01 PM

```js
Plot.plot({
  width: 1152,
  height: 400,
  x: {
    domain: [-1, 0],
    grid: true,
    clamp: true
  },
  r: {
    range: [0, 24]
  },
  marks: [
    Plot.dot(mobility, Plot.dodgeY("middle", {
      filter: d => d.in_lockdown === "TRUE",
      sort: "pop",
      reverse: true,
      x: "pct_change",
      r: "pop",
      fill: "currentColor"
    }))
  ]
})
```

Demo: https://observablehq.com/d/2055ab012188f381

Still to figure out:

* How to specify the *Rsep* (separation radius option)? Perhaps `dodge: {anchor: "center", separation: 2}`?
* Would it be better to specify `y: "dodge"` or some such (i.e., tied to the channel), instead of `dodge: true`?
* Avoid materializing the R array if `r` is a constant?
* Generalize this to other marks? Does Plot need a concept of “layouts” (in pixel space), similar to transforms (data space)?

Fixes #164.


Fil (commented):
I love the feature, but I'd prefer if we could have a minimal code footprint in src/dot.js, since that's one of the first files that someone would read to understand how to write a mark. Also if we introduce layouts, we'll probably have more of them with time :)

I can see two options:

1. Make it a plugin, rather than a core feature (as in https://observablehq.com/@fil/experimental-plot-beeswarm, or better—and that is a good opportunity to define what a plugin is)
2. Move the specific code to a different file in src (layouts/dodge.js ?), but still call it from within dot.js

mbostock:
I think this is useful enough that I want it included by default, rather than folks having to go look for a plugin and learn how to load it.

I don’t think we should optimize the readability of the implementation. Or at least, it’s not a priority relative to the exposed functionality of Plot which will be used by many more people. Of course I will endeavor to build a beautiful implementation but I do not expect Plot’s implementation to be “easy” to understand, or simple (see the bin transform, for example).

I do think there will be other layouts, and that we may want this functionality shared across marks, and thus it’s useful to think about both how the code should be structured within Plot, and whether we can generalize how it is exposed, too. The idea of binding a layout to a channel is interesting. Though, I think it may be limiting, as I expect there will be other cases where a layout may produce more than one channel. So, perhaps there’s a “layout” option, similar to a transform? Like `layout: Plot.dodge(…)`?

jtrim-ons:
This looks good to me, and thanks for the mention.

i hope this isn't just shamelessly promoting my own idea, but I wonder if it might be worth considering a "compact beeswarm" option if the radii are all equal. Some examples of compact beeswarm vs dodge are in my initial post [here](https://github.com/aroneklund/beeswarm/issues/9), and an explanation of the method is [here](https://observablehq.com/@jtrim-ons/compact-beeswarm-plot?collection=@jtrim-ons/compact-beeswarm-plots). For many datasets, I prefer it to the dodge function, just because it doesn't have a windblown look.

I don't think any of my compact beeswarm implementations have particularly good code style, but I could potentially contribute to an implementation if it would be useful. I also completely understand if it's not something you want to add to Plot at the moment.

src/marks/dot.js
```diff
@@ -50,6 +75,71 @@ export class Dot extends Mark {
...
+      const dx = X[i] - X[j];
+      const dr = R[i] + Rsep + R[j];
+      const dy = Math.sqrt(dr * dr - dx * dx);
+      intervals.push([yj - dy, yj + dy]);
+    });
+
+    // Find the best y-value where this circle can fit.
+    for (let y of intervals.flat().sort(compare)) {
+      if (intervals.every(([lo, hi]) => y <= lo || y >= hi)) {
+        Y[i] = y;
```
mbostock, line 124: Currently we’re setting Y[i] = 0 if there’s no conflict, but I wonder at least in the case of asymmetric swarms whether we should instead be setting it to R[i] + Rsep so that the bottom of the swarm is aligned (rather than the centers of the circles).

mbostock:
Thanks @jtrim-ons! I think this initial algorithm is “good enough” to start, but I’m definitely willing to swap it out with something that performs better in the future. I’d prefer if the algorithm supported varying radius, though Plot could chose a special algorithm for fixed radius dots if we wanted.

The main thing to figure out with this PR is how much we want to generalize the concept of layouts, or if this will be a bespoke feature for dots. (Or if it can start as bespoke and evolve into a more general approach in the future.)

Fil:
Also related, Dorling cartograms (for Plot.Carto :-) )

mbostock:
I’ve now implemented this as a layout independent of marks, as a counterproposal to #691. A layout takes the same arguments as a mark’s render function and returns transformed values rather, providing a hook for transforming channels before rendering. Layouts are invoked similarly to option transforms, and thus can access (and even transform!) mark options, or take separate layout-specific options as desired.

Fil (commented):
It's really fun and powerful to see how we can add new layouts (I've created a circlePacking layout in less than 20 lines of code).

The dodge layout looks good to me.

As a follow-up or as part of this PR we want to extend it to work with a channel *r*, even if the underlying mark has no use for it, as this would enable labelling bubble charts with a variable radius. (As discussed in #708).

I'd also prefer if the default *r* for a symbol dot was consistent across the dot mark and the dodge layout (4.5 instead of 3), but not at the cost of overcomplicating things.

I'll try and work on these two issues, which are not blocking.

Besides that, it needs documentation. A starting point is available in #708.

Fil:
During review we mentioned a possible (structural) change which would be to pass all the facets at once instead of calling the function once for each facet—thus allowing a layout to worked "globally" or "locally" on faceted data.

mbostock:
Another layout idea: a strategy for avoiding occlusion with text labels, like [@fil/occlusion](https://observablehq.com/@fil/occlusion). Though I’m not sure how the layout would determine the bounding box for the text labels. Would it use a heuristic like we do for line wrapping? Would it use an offscreen canvas and measureText?

src/plot.js
```diff
@@ -95,8 +95,9 @@ export function plot(options = {}) {
 
   for (const mark of marks) {
     const channels = markChannels.get(mark) ?? [];
-    const values = applyScales(channels, scales);
+    let values = applyScales(channels, scales);
     const index = filter(markIndex.get(mark), channels, values);
+    if (mark.layout != null) values = mark.layout(index, scales, values, dimensions);
```
Fil, line 100:
```suggestion
    if (mark.layout != null) values = mark.layout(index.slice(), scales, values, dimensions);
```

should we make a defensive copy?

Alternatively, if we assume that it is legal for the layout to modify the index, write it as:

```suggestion
    if (mark.layout != null) ({index, values} = mark.layout(index, scales, values, dimensions));
```

(and same in the facet call).

I think my preference would go to explicitly allowing a layout to change the index, for example a hexbin layout could group points into hexagons, and a unit chart/isotype layout could create several symbols out of (say) a bar.

Fil:
Suggestions:
-  #711
-  #712
-  #713 (needs a small update if we approve 712)

Besides these suggestions:
- the question whether a layout is explicitly allowed or forbidden to modify the index (https://github.com/observablehq/plot/pull/648#discussion_r793643082); this is blocking if we expose the API (in the sense that it would mean a change to the API); but not blocking if we don't yet expose it.
- the need for Plot.text(dodgeY({r: channel})) — TBD, but probably not blocking

mbostock:
> I think my preference would go to explicitly allowing a layout to change the index, for example a hexbin layout could group points into hexagons, and a unit chart/isotype layout could create several symbols out of (say) a bar.

I think we should try implementing these and see where that takes us.

For hexbin, I’m guessing that’s more of a mark than a layout (or transform)? Because if you’re producing hexagonal bins, you almost certainly want to render them as hexagons, meaning that the representation is dictated and it’s not simply transforming channel values. I’ve been thinking about density contours similarly, which perhaps we could also implement.

For isotype and waffle charts (is that what you meant by unit chart?), it feels like it might be more of a transform than a layout. Or maybe that’s a mark, too? Or a pattern? In any case it should be easier to answer these questions through prototyping. 😁

mbostock:
Superseded by #775.

Issue #41, opened 2020-12-11
Proportional symbol mark?


Fil:
Speaking in general, we have several possibilities for proportional symbols:
- svg:circle is concise (and possibly the most cpu-efficient); scale with r/size.
- svg:path + d3.symbol is useful for certain types of scatterplots, but difficult to extend; scale with transform:scale
- svg:text + emoji works if we're careful to center the mark properly; scale with transform:scale

Note that svg:use + defs covers the three cases, and removes the difficulty of having to define complex symbols when we're not happy with the set in d3.symbols.
See https://observablehq.com/@fil/svg-use

So if we want a variant of Plot.dot that allows various symbols, I would try to code it with `<use>`. The range of the symbol scale is a list of objects that would have an id and be added to the defs, or a list of ids—but then how do we pass the objects themselves? The mark could default to `<use href=#square>`. And it would accept an orientation channel!

(discarding svg:polygon which has no advantage that I know of over svg:path.)

Fil:
Linked to https://github.com/observablehq/plot/issues/170

Fil:
See also the preattentive symbols by Heman Robinson https://observablehq.com/@heman/preattentive-symbols-176

mbostock:
Fixed in #652 #649.

Issue #39, opened 2020-12-09
Arrow mark (like link, but with an arrowhead) and swoop mark (like link, but with a curved line).


Fil:
partially implemented in https://github.com/observablehq/plot/pull/38 and 
https://observablehq.com/d/c364dd180391294a

EE2dev:
I thought of arrowhead more to be a constant option for Plot.link() and Plot.line(), with the choice of several arrowhead markers

tx0c:
> partially implemented in #38 and https://observablehq.com/d/c364dd180391294a

https://github.com/observablehq/plot/pull/38#issuecomment-830551122
> superseded by plugin https://observablehq.com/@data-workflows/plot-arrow

both links are no longer there, is #39 #38 still in progress? stuck for a whole year

and ideas in #41 for any symbol mark put on the link would be awesome

Fil:
You should now be able to access the notebook with the suggested plugin at https://observablehq.com/@fil/plot-arrow

I have not worked on arrowhead symbols. Suggestions and contributions are welcome.

PR #649, opened 2022-01-04
vector + symbol

### LINKED ISSUE #407: Vector field plots : x, y, size, color and orientation channels
Wondering wether Plot could, eventually, take advantage of  all the set of x, y, size, color and orientation visual properties such as in:
https://observablehq.com/@d3/vector-field

mbostock:
Adds a new Plot.vector mark, similar to Plot.dot, for drawing little arrows. In addition to *x* and *y*, it exposes *length* (magnitude) and *rotate* (direction, in degrees) channels. The *anchor* option (start, middle, or end) controls where the arrow is anchored to *x* and *y*; it defaults to start.

Screen Shot 2022-01-04 at 10 09 59 AM

```js
Plot.plot({
  inset: 12,
  height: 600,
  marks: [
    Plot.vector(
      (T => d3.cross(T, T))(d3.ticks(0, 2 * Math.PI, 20)),
      {
        anchor: "middle",
        length: ([x, y]) => (x + y) * 2 + 2,
        rotate: ([x, y]) => (Math.sin(x) - Math.sin(y)) * 60
      }
    )
  ]
})
```

TODO
- [x] Add documentation for Plot.vector.
- [x] Add documentation for Plot.dot’s new _symbol_ option.
- [x] Bind Plot.vector’s _length_ channel to a _length_ scale.

Fixes #407.


mbostock:
(I really wanted to use radians here, but I wanted to be consistent with Plot.text’s *rotate* option.)

mbostock:
Maybe anchor should default to middle instead?

mbostock:
I tried implementing *vx* and *vy* channels as alternatives to *length* and *rotate*, but it’s awkward because *vx* and *vy* would likely need to be in pixel space, rather than data space as *x* and *y* are. So I’m going to punt on that idea.

Fil:
Yes I imagine vx, vy to be working in data space; maybe it needs a "time resolution" factor _t_, so if x and y are in kilometers, and vx, vy in km/h, the actual length of a vector represents _t_ hours of driving?

Fil:
A few thoughts as I'm playing with this branch:

Should we filter on positive lengths? Or do we keep it generalized to negative lengths being "reversed"?

Do we want a way to personalize the arrows' shapes? (In https://github.com/observablehq/plot/issues/39 there is some demand for arrowheads symbols, and in https://observablehq.com/@fil/plot-arrow I've experimented with a variable arrowhead angle and length.)

If we wanted to generalize, Plot.vector could be an instance of Plot.symbol (https://github.com/observablehq/plot/issues/41), with an arrow symbol and a linear size. In that case, we wouldn't compute the rotated path's *d*, but only a scaled path, and we would leave the rotation to a svg transform.

(Also, it needs a bit of documentation.)

mbostock:
> Should we filter on positive lengths? Or do we keep it generalized to negative lengths being "reversed"?

I originally filtered on positive lengths, but I think it’s preferable to support inverted vectors for negative lengths. You can always filter if you don’t want this behavior.

> Do we want a way to personalize the arrows' shapes? … If we wanted to generalize, Plot.vector could be an instance of Plot.symbol

I think we want this _eventually_, but I don’t think it’s a blocker for now. I also think there is a difference between Plot.symbol having an *r* channel with default sqrt scale (or *area* channel with linear scale) versus Plot.vector’s *length* channel. Having magnitude and direction feels most natural for a vector field, but for a symbol I would expect an areal encoding (and rarely need *rotate*).

> In that case, we wouldn't compute the rotated path's d, but only a scaled path, and we would leave the rotation to a svg transform.

I originally implemented Plot.vector with a transform attribute such that the *d* attribute always pointed up with the given *length*. But I suspected that the performance would be better to bake everything into the path string, or at least, it felt more concise to do it in one place. But, I’m happy to change it. I don’t see it as being necessary, though, since it’s just an implementation detail of the mark.

I could certainly take a crack at Plot.symbol as a PR to this PR so that we can feel confident that we’re designing holistically.

> (Also, it needs a bit of documentation.)

Yep!

mbostock:
This is ready for review, @Fil. 🙏

README.md
```diff
@@ -1093,6 +1110,41 @@ The following optional channels are supported:
...
+* **length** - the length in pixels; bound to the *length* scale; defaults to 12
+* **rotate** - the rotation angle in degrees clockwise; defaults to 0
+
+The following options are also supported:
+
+* **anchor** - one of *start*, *middle*, or *end*; defaults to *middle*
+
+If the **anchor** is *start*, the arrow will start at the given *xy* position and point in the direction given by the rotation angle. If the **anchor** is *end*, the arrow will maintain the same orientation, but be positioned such that it ends in the given *xy* position. If the **anchor** is *middle*, the arrow will be likewise be positioned such that its midpoint intersects the given *xy* position.
+
+If the **x** channel is not specified, vectors will be horizontally centered in the plot (or facet). Likewise if the **y** channel is not specified, vectors will vertically centered in the plot (or facet). Typically either *x*, *y*, or both are specified.
```
Fil, line 1132:
```suggestion
If the **x** channel is not specified, vectors will be horizontally centered in the plot (or facet). Likewise if the **y** channel is not specified, vectors will be vertically centered in the plot (or facet). Typically either *x*, *y*, or both are specified.
```

README.md
```diff
@@ -1093,6 +1110,41 @@ The following optional channels are supported:
...
+
+The **rotate** and **length** options can be specified as either channels or constants. When specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel. The length defaults to 12 pixels, and the rotate defaults to 0 degrees (pointing up↑). Vectors with a negative length will be drawn inverted. Positive angles proceed clockwise from noon.
```
Fil, line 1134:
```suggestion
The **rotate** and **length** options can be specified as either channels or constants. When specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel. The length defaults to 12 pixels, and the rotate defaults to 0 degrees (pointing up↑). Vectors with a negative length will be drawn inverted. Vectors of length 0 will show as a single pixel; use NaN to remove. Positive angles proceed clockwise from noon.
```

(not sure about this wording)

Fil (approved):
two small remarks in the README

README.md
mbostock: Yeah, I don’t think saying a single pixel is correct… since it depends on the strokeWidth? Or possibly you see nothing? In any case, probably better to elaborate in the Marks section above, near here:

> Plot.dot will not generate circles with null, undefined or negative radius, or null or undefined coordinates.

PR #652, opened 2022-01-05
symbol

### LINKED ISSUE #41: Proportional symbol mark?

mbostock:
TODO

- [x] add a test for categorical symbol encoding
- [x] optimize rendering for the common case of circles
- [x] optimize rendering for the common case of no rotation
- [x] allow a symbol scale’s domain to be specified as symbol names
- [x] a symbol legend? not very useful without a legend…
- [x] use better symbols https://github.com/d3/d3-shape/pull/189
- [ ] add documentation

Fixes #41.


Fil:
I like where this is going. We can do the legend (a bit) later; I need to revisit the radius legend also…

mbostock:
Implementing a symbol legend was surprisingly easy! Hooray. It would be nice if we could combine the legends for the symbol and color scales into one, in the case where they’re used for the same encoding, but I’m not sure yet how we’d determine that and at any rate it feels like a future problem.

Screen Shot 2022-01-05 at 12 42 21 PM

mbostock:
Screen Shot 2022-01-05 at 1 34 12 PM

Fil:
wow!

mbostock:
The adoption of new (preattentive) symbols is probably a blocker for this PR, unless we can come up with good and non-conflicting names for those symbols. I wouldn’t want people to have to say `symbol: "diamond2"` to get the preferred diamond.

Fil:
Re: symbols names, just like you can have text: "abc" and various fonts, we could imagine that symbol: "diamond" looks up the diamond symbol in the current symbolFamily / symbolSet. (In other words I don't believe there can be a flat namespace for symbols, see the unicode/emoji mess.)

This approach would also allow "styling", by switching to a different symbol set. It would be cool to systematically have the symbols aliased as symbol1 symbol2 symbol3, on top of their descriptive names.

mbostock:
I would prefer symbol names that are context-free so that it’s unambiguous and you can mix and match symbols from different sets. Also, I’d generally like Plot to use D3’s names simply to avoid introducing confusion, and we’ll need to pick unique names for D3 (in https://github.com/d3/d3-shape/pull/189, currently: asterisk, diamond2, plus, square2, triangle2, and x).

It’ll also be possible to say `symbol: {domain: d3.symbolsFill}` or `symbol: {domain: d3.symbolsStroke}` to choose a different domain (the set of fillable symbols and strokeable symbols, respectively). And it’d be great if Plot could chose the right one based on whether the symbols will be filled or stroked—probably by allowing marks to specify a default domain for a particular scale, similar to how they can specify a default scale type.

I can start prototyping this here but it’d probably be easier to land and release the new d3-shape first, once you review.

PR #654, opened 2022-01-06
select valid

### LINKED ISSUE #515: Generalize the select transform
It could be made more generic, allowing you to specify a custom selector and optional associated input channel.

Fil:
demo: https://observablehq.com/@observablehq/plot-select-generalize-515

simple examples would be:
* Plot.selectFirst({valid: "text"}) 
* Plot.selectLast({valid: "text"}) 
* Plot.selectValid({valid: "text"}) 

questions:
- [ ] the wording is cringy, it's probably not "valid" that we want
- [ ] test for empty rather than defined?
- [ ] documentation, tests
- [ ] maybe this is not enough: for example, it doesn't allow to select the median value; should we add a generic function?

closes #515

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ export {group, groupX, groupY, groupZ} from "./transforms/group.js";
 export {normalize, normalizeX, normalizeY} from "./transforms/normalize.js";
 export {map, mapX, mapY} from "./transforms/map.js";
 export {window, windowX, windowY} from "./transforms/window.js";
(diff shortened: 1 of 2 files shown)


mbostock:
Isn’t this equivalent to the existing _filter_ option? E.g.,

```js
Plot.text(
  data,
  Plot.selectFirst({
    text: ([a, b]) => `${a}-${b}`,
    filter: ([a, b]) => (a % 2 && b % 2) || NaN
  })
).plot()
```

Fil:
true! 😅

then I guess we should just remove the comments
`// TODO If the value (for some required channel) is undefined, scan forward?`

PR #656, opened 2022-01-07
generalize Plot.select

### LINKED ISSUE #515: Generalize the select transform

Fil:
demo: https://observablehq.com/@observablehq/plot-select-generalize-656

closes #515

*select a sample:*
Capture d’écran 2022-01-11 à 11 31 12

*select min and max of a color:*
Capture d’écran 2022-01-11 à 11 31 30

### DIFF
--- a/README.md
+++ b/README.md
@@ -1633,6 +1633,31 @@ Selects the rightmost point of each series.
 
 Selects the highest point of each series.
 
+### Plot.selectAny(*select*, [*channel*, ]*options*)
+
+Selects the points of each series selected by the *select* selector. The selector can be specified as a built-in ("min" and "max"), or as a function, which receives as inputs the index of the series and the optional *channel*, if specified.
+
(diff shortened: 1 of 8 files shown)


mbostock (changes requested):
How about this?

```js
Plot.select({x: "min"}, options)
```

And we could throw an error if you pass an object with more than one (or less than one) key as the first argument.

mbostock:
On the topic of having a selector interface {select(I, V)} instead of just a function(I, V), so that we could potentially have a more convenient shorthand form for functions, I couldn’t think of a better shorthand for functions in this case, so I’m going to punt on that even though it’s a little different than what we did for reduce and map. Unlike, say, the filter transform, the select transform operates on a group of data at a time, so we need a way to identify a subset of that data, and a subset of the index feels like the most natural form for that.

PR #664, opened 2022-01-12
select edits

mbostock:


README.md
```diff
@@ -1581,35 +1613,6 @@ Selects the rightmost point of each series.
...
-    }, {
-      x: "year",
-      y: "revenue",
-      fill: "format"
-    })
-```
-
-To pick three points at random in each series:
-```js
-  Plot.select(I => d3.shuffle(I).slice(0, 3), {z: "year", ...})
```
mbostock, line 1605: Note that d3.shuffle(I) mutates in place!

README.md
```diff
@@ -1581,35 +1613,6 @@ Selects the rightmost point of each series.
...
-  Plot.select({
-    y: (I, V) => {
-      const median = d3.median(I, i => V[i]);
-      const i = d3.least(I, (i) => Math.abs(V[i] - median));
-      return [i];
-    }, {
-      x: "year",
-      y: "revenue",
-      fill: "format"
-    })
```
mbostock, line 1600: I fixed the missing brace here, and removed the extra indentation (which was repeated below).

src/transforms/select.js
```diff
@@ -3,62 +3,83 @@ import {maybeZ, valueof} from "../mark.js";
...
+  return selectChannel(null, selectorFirst, options);
 }
 
 export function selectLast(options) {
-  return selectAny(last, undefined, options);
+  return selectChannel(null, selectorLast, options);
 }
 
 export function selectMinX(options) {
-  return select({x: "min"}, options);
```
mbostock, line 30: Note that this one went through a different code path, but I think it worked out the same. (I changed it so that all the convenience methods are now consistent.)

src/transforms/select.js
```diff
@@ -3,62 +3,83 @@ import {maybeZ, valueof} from "../mark.js";
...
+  return selectChannel("x", selectorMin, options);
 }
 
 export function selectMinY(options) {
-  return selectAny({y: "min"}, options);
```
mbostock, line 34: Note that this code path (and those below) would not have generated a good error message if the _y_ channel is missing. (I changed it so that all code paths will throw the same “missing channel” error.)

mbostock:
I also want to think a little more if we should have a more specialized interface like we do for reducers (a {select} object instead of just a function), and then perhaps you could have a more convenient form for functions. Like, as a function of data that returns an array of zero-based indexes.

README.md
Fil: Ah, yes. Maybe we should be defensive in `for (const I of Z ? group(facet, i => Z[i]).values() : [facet])` instead, by using `facet.slice()`?

Fil: or, we could be defensive just in the case that we pass the facet to a custom function (since our internal functions are not mutating I).

Fil:
Series subsampling will be a neat application of Plot.select:
https://observablehq.com/@fil/time-series-topological-subsampling

README.md
mbostock: I don’t think we want to be defensive here because it will introduce overhead. We can document that reducers should not mutate their inputs, if you like (or risk undefined behavior). And we should be careful to not give any examples that mutate their inputs. 🙂

mbostock:
I am going to merge this into your branch so that I can rebase against main.

PR #665, opened 2022-01-12
radius legends

Fil:
Capture d’écran 2022-01-12 à 12 43 46

It's just one possibility among several

(I've based this on mbostock/arrow since it involves quite a few changes in the legends section)


Fil:
I've fixed the conflict but I now have a circular dependency which I don't know how to fix. Should I rewrite this legend as "pure d3" rather than using Plot?

Fil:
As an alternative here's an independent implementation
https://observablehq.com/@recifs/a-radius-legend-for-plot-665

PR #667, opened 2022-01-12
text halo option

### LINKED ISSUE #666: Plot.text halo option(s)
It's a CSS design question + an API design question.

    stroke-width: 4;
    fill: none;
    stroke: white; 

The stroke: white is controversial, since we don't know that currentColor is black (or any dark color).

API design question: do we need three options (halo: true, haloStroke = "white", haloWidth = 4)?

(This can't be done by applying two text marks, because the halos are also there to fight overlapping texts.)

Ref. #27 and https://observablehq.com/@mbostock/inequality-in-american-cities

Fil:
tbd:
- [ ] The stroke: white is controversial, since we don't know that currentColor is black (or any dark color).
- [ ] API design question: do we need three options (halo: true, haloStroke = "white", haloWidth = 4)?

closes #666


mbostock:
Another possibility is that we could add to #701 such that if a text mark has both a *stroke* and a *fill*, the *paintOrder* defaults to “stroke” (so that the stroke is drawn underneath the fill, rather than vice versa). That’s a little bit opinionated, but it means we don’t need a dedicated *halo* option, and we require the user to specify what the color will be. We already default *strokeLinejoin* to round, so you’d probably also want to specify *strokeWidth*… otherwise you only get effectively a 0.5px halo. But either way it’s a lot easier than having to duplicate the entire mark!

Fil:
superseded by #703

PR #668, opened 2022-01-14
Add info to README on what's available for scale.tickFormat

flother:
It took me a little while to track down exactly what format specifiers I can use with `scale.tickFormat`. I thought it would be useful if links to the [d3-format](https://github.com/d3/d3-format#locale_format) and [d3-time-format](https://github.com/d3/d3-time-format#locale_format) format specifier docs were included in the README, to help fellow travellers.


Fil (approved):
thanks!

PR #671, opened 2022-01-16
universal channel filter

mbostock:
Takes the filter(index, ...channels) function and incorporates it into plot.js (and facet.js) so that filtering is more declarative and happens by default. You can opt-out of filtering by declaring filter: null (as for lines and areas) or to specify a custom filter definition.

This is groundwork for layouts (e.g., dodge #648)


Fil (approved):
I dig this structural commit, that removes more lines than it adds.

PR #672, opened 2022-01-16
add svg namespace

Fil:
if only because it helps making plots of plots https://observablehq.com/@fil/plot-of-plots

- [ ] Question: should we also add xlink?
- [ ] Remove redundant add from test/plot.js


mbostock:
I don’t think this is necessary for the Plot-of-plots techniques; it’s only necessary when using data URL serialization, and in that case, you could inject the xmlns attribute yourself (as you do in the notebook). I think we’ll want some sort of “glyph” mark for Plot-of-plots that allows you to generate arbitrary SVG content at a given xy location, or alternatively, perhaps we can improve the function shorthand or custom mark implementation to make this sort of thing easier.

Fil:
We also do it in the snapshot tests. But agree it's not necessary, just a a bit of help.

PR #673, opened 2022-01-16
automatic color identity scale

mbostock:
Alternative to #660. For the *color* scale, if at least one value is defined and all defined values (from the domain if specified, and otherwise from channels) are valid color strings, default to an identity scale rather than some other type.


Fil:
There is always going to be a situation in which these default colors are surprising. In this case, if the categories are ["orange", "apple", "banana"] only the orange will show; but if they are  ["apple", "banana", "orange"], everything will show (and the "orange" category will be blue or something).

This situation is what I had in mind when I suggested that there could be a different type of default ordinal color scale which would be (roughly) a pass-thru for valid css colors, and turbo10 for the others.

Also identity doesn't allow to have a legend, so we can't use the legend to inspect what the invalid fields are.

mbostock:
> There is always going to be a situation in which these default colors are surprising.

Totally, but the idea is that you can address those by setting the scale’s type explicitly. And this test is cheap and consistent with our other tests (e.g., to disambiguate ordinal from quantitative or temporal data).

> Also identity doesn't allow to have a legend, so we can't use the legend to inspect what the invalid fields are.

I think that’s also expected and desirable. And if you want a legend, you just set the scale type to ordinal or whatever.

mbostock:
One other thing is that we could explicitly disallow named CSS colors (and thereby require rgb(…) or #rrggbb etc.). I suspect named colors will be responsible for nearly all the ambiguous cases. Of course then folks still might get confused why () => "red" doesn’t give you red.

src/options.js
```diff
@@ -210,6 +207,38 @@ export function isNumeric(values) {
...
+  return true;
+}
+
+// Mostly relies on d3-color, with a few extra color keywords. Currently this
+// strictly requires that the value be a string; we might want to apply string
+// coercion here, though note that d3-color instances would need to support
+// valueOf to work correctly with InternMap.
+export function isColor(value) {
+  if (!(typeof value === "string")) return false;
+  value = value.toLowerCase();
```
Fil, line 235: that's new. css colors are case-insensitive, so should we mention it as a bugfix?

Fil (approved):
👍 this feels quite comfortable.

src/options.js
mbostock: Yes, this does fix a bug. I’d be surprised if anyone encountered it, but we could mention it.

Issue #384, opened 2021-05-12
A global locale option


ericemc3:
Wherever English formatting appears by default, a simple possibility of taking into account any other language could/should be proposed.

E.g.: formatting of numbers or dates on the axes.

Suggestion: a high level `locale` property in a Plot specification, applying to the whole plot. By default, locale is set to "en". 

This would work in a similar way to what has very usefully appeared in the Table component.

Note: vega-lite allows you to globally specify a locale for numbers and for dates:
```
vl.vega.timeFormatLocale(locale);
vl.vega.formatLocale(locale);
```

PS: Plot is a great library, very intellectually stimulating, and the 21 beautifully written articles in the https://observablehq.com/collection/@observablehq/plot collection should be read by all data scientists!

mbostock:
I propose this is a top-level Plot.plot option, i.e.,

```js
Plot.plot({
  locale: "de",
  marks: …
})
```

The plot should then inherit this locale rather than the default "en-US". This is possibly a little tricky with things like [Plot.formatWeekday](https://github.com/observablehq/plot#plotformatweekdaylocale-format) since that function currently needs to be supplied the locale upon creation, whereas here we would need to somehow pass the plot’s locale into the format function. One way to do that might be to pass the plot’s locale to the format function instead:

```js
function formatMonth(month = "short") {
  return (i, locale = "en-US") => {
    if (i != null && !isNaN(i = new Date(Date.UTC(2000, +i)))) {
      return i.toLocaleString(locale, {timeZone: "UTC", month});
    }
  };
}
```

(I think this would also be slower since we couldn’t reuse the same Intl.DateFormatInstance, but we could address that separately using memoization.)

In addition, we could do something like Plot.defaults:

```js
plot = Plot.defaults({locale: "de"})
```
```js
plot({marks: …})
```

mbostock:
And more specifically: let’s _not_ use global mutation, _e.g._ Plot.setLocale(*locale*), since this makes it hard to support multiple locales on the page and introduces ordering problems where the behavior of Plot.plot will change before or after a call to Plot.setLocale.

Fil:
This would also impact the automatic word "frequency", #395, and maybe other language elements.

(Regarding autodetection of "date"/"year" to omit the axis label, it is based on the field name so I would tend to not consider it in this context.)

rusosnith:
This would be very important for us spanish users! Date parse as mentioned for axis it's a must!

Fil:
[...]

PR #674, opened 2022-01-16
propagate title and href by default

### LINKED ISSUE #617: The bin and group transform could propagate the title channel, if any, by default
E.g.,

```js
Plot.rectY(data, Plot.binX({y: "sum"}, {x: "date", y: "count", fill: "name", title: "name"}))
```

would be equivalent to

```js
Plot.rectY(data, Plot.binX({y: "sum", title: "first"}, {x: "date", y: "count", fill: "name", title: "name"}))
```

And possibly it could even be smart by showing the distinct values rather than just the first? Or it could be like `Fred (42)` and show both the distinct values and their respective counts by default? I feel like that would be very useful.

mbostock:
Fixes #617.

### DIFF
--- a/src/transforms/group.js
+++ b/src/transforms/group.js
@@ -44,10 +44,6 @@ function groupn(
   inputs = {} // input channels and options
 ) {
 
-  // Propagate standard mark channels by default.
-  if (inputs.title != null && outputs.title === undefined) outputs.title = reduceTitle;
-  if (inputs.href != null && outputs.href === undefined) outputs.href = reduceFirst;
-
   // Compute the outputs.
   outputs = maybeOutputs(outputs, inputs);
   reduceData = maybeReduce(reduceData, identity);
@@ -139,7 +135,11 @@ export function hasOutput(outputs, ...names) {
 }
 
 export function maybeOutputs(outputs, inputs) {
-  return Object.entries(outputs).map(([name, reduce]) => {
+  const entries = Object.entries(outputs);
+  // Propagate standard mark channels by default.
+  if (inputs.title != null && outputs.title === undefined) entries.push(["title", reduceTitle]);
+  if (inputs.href != null && outputs.href === undefined) entries.push(["href", reduceFirst]);
+  return entries.map(([name, reduce]) => {
     return reduce == null
       ? {name, initialize() {}, scope() {}, reduce() {}}
       : maybeOutput(name, reduce, inputs);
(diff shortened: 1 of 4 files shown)


src/transforms/group.js
```diff
@@ -266,6 +270,16 @@ const reduceFirst = {
...
 };
 
+const reduceTitle = {
+  reduce(I, X) {
+    return sort(rollup(I, V => V.length, i => X[i]), second)
+      .reverse()
+      .slice(0, 5)
+      .map(([key, value]) => `${key} (${value.toLocaleString("en-US")})`)
+      .join("\n");
+  }
```
Fil, line 280:
```suggestion
    const labels = sort(rollup(I, V => V.length, i => X[i]), second)
      .reverse();
    return labels
      .slice(0, 5)
      .map(([key, value], i) => i === 4 && labels.length > 5 ? "…" : `${key} (${value.toLocaleString("en-US")})`)
      .join("\n");
  }
```

to replace the 5th element by continuation dots if there are more?

mbostock: Probably worth counting the number of “other” then.

mbostock: If there are more than five distinct values, these are now combined into an “Other” category.

src/transforms/group.js
```diff
@@ -266,6 +270,15 @@ const reduceFirst = {
   }
 };
 
+const reduceTitle = {
+  reduce(I, X) {
+    const groups = sort(rollup(I, V => V.length, i => X[i]), second);
+    const top = groups.slice(-5).reverse();
+    if (top.length < groups.length) top.push(["Other", sum(groups.slice(0, -5), second)]);
```
Fil, line 277:
```suggestion
    const N = 5;
    const groups = sort(rollup(I, V => V.length, i => X[i]), second);
    const top = groups.slice(-5).reverse();
    if (top.length < groups.length) top.push([groups.length === N + 1 ? groups[N][0] : "Other", sum(groups.slice(0, -N), second)]);
```

Fil: (if there are exactly 6 groups we can use the 6th's name.)

Fil:
The locale (https://github.com/observablehq/plot/issues/384) should also apply to "Other".

src/transforms/group.js
mbostock: Changed to show the number of other categories, as well as the number of constituent elements.

Fil (approved):
👍

PR #675, opened 2022-01-16
automatic color identity range

mbostock:
Alternative to #673 and #660. For ordinal or categorical color scales, if the domain contains only valid color values (or undefined or null), and neither a scheme nor a range is set, then the range defaults to the domain.


mbostock:
I think I prefer #676 to this.

PR #679, opened 2022-01-18
better fontSize disambiguation

### LINKED ISSUE #378: Allow fontSize to be specified as a CSS length
It’d be nice if `fontSize: "12pt"` worked. But to do this we’ll need a smarter version of maybeNumber that can detect CSS length strings — we don’t want to interpret `"12pt"` as a field name.

mbostock:
Fixes #378.

### DIFF
--- a/src/marks/text.js
+++ b/src/marks/text.js
@@ -25,13 +25,13 @@ export class Text extends Mark {
       rotate
     } = options;
     const [vrotate, crotate] = maybeNumberChannel(rotate, 0);
-    const [vfontSize, cfontSize] = maybeNumberChannel(fontSize);
+    const [vfontSize, cfontSize] = maybeFontSizeChannel(fontSize);
     super(
       data,
       [
         {name: "x", value: x, scale: "x", optional: true},
         {name: "y", value: y, scale: "y", optional: true},
-        {name: "fontSize", value: numberChannel(vfontSize), optional: true},
+        {name: "fontSize", value: vfontSize, optional: true},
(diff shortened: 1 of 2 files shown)


src/marks/text.js
```diff
@@ -105,3 +105,39 @@ function applyDirectTextStyles(selection, mark) {
...
+// - number in pixels
+// - string keyword: see above
+// - string <length>: e.g., "12px"
+// - string <percentage>: e.g., "80%"
+// Anything else is assumed to be a channel definition.
+function maybeFontSizeChannel(fontSize) {
+  if (fontSize == null || typeof fontSize === "number") return [undefined, fontSize];
+  if (typeof fontSize !== "string") return [fontSize, undefined];
+  fontSize = fontSize.trim().toLowerCase();
+  return fontSizes.has(fontSize) || /^[+-]?\d*\.?\d+([eE][+-]?\d+)?(\w*|%)$/.test(fontSize)
```
Fil, line 140:
```suggestion
  return fontSizes.has(fontSize) || /^[+-]?\d*\.?\d+(e[+-]?\d+)?(\w*|%)$/.test(fontSize)
```

e is already lowercased (note: I wouldn't have thought to check the scientific notation form for numbers!)

Fil (approved):
with a minor nit

src/marks/text.js
mbostock: I found this helpful:

https://www.w3.org/TR/css-syntax-3/#number-token-diagram
https://drafts.csswg.org/css-values/#numbers

PR #682, opened 2022-01-19
use x, y and dy when there is no rotate

Fil:
test multiline (with 1 empty line)
document


mbostock (commented):
Doesn’t this break the layout of multiline text, since the tspans assume that x and y are relative to the transform, which is no longer present? That’s why I adopted the transform in the first place.

And perhaps more importantly, is there a reason to do this? I was trying to decide the answer to this question. I don’t know if there’s a compelling reason to do this (such as performance) and hoped you’d help me think that through.

Also I am reminded that the changelog will need to mention the change in behavior for dx and dy. If I’m not mistaken, in addition to the loss of units, they now apply before the rotation rather than after. This makes me wonder if we want something like the old dx and dy in addition to the new one?

Fil:
In my test notebook https://observablehq.com/@fil/test-text-multiline-677  it seems to work, but maybe there's a case I didn't see? (I don't have a strong opinion on transform vs x/y, I must say.)

PR #686, opened 2022-01-19
frameAnchor; fix multiline text

### LINKED ISSUE #683: A frameAnchor option for positioning otherwise position-less marks relative to the frame
When marks don’t have a position, we currently default them to the center of the frame. For example:

https://github.com/observablehq/plot/blob/136fc2f54a4d31043868e584073d28d3680f7169/src/marks/dot.js#L57-L58

https://github.com/observablehq/plot/blob/136fc2f54a4d31043868e584073d28d3680f7169/src/marks/text.js#L56-L57

The *frameAnchor* option would allow more control over this behavior, for example when [labeling facet frames](https://observablehq.com/@data-workflows/geofacet). It would support nine possible values:

* top-left
* top
* top-right
* right
* bottom-right
* bottom
* bottom-left
* left
* middle (default)

mbostock:
Fixes #683.


src/style.js
```diff
@@ -225,3 +225,10 @@ export function applyInlineStyles(selection, style) {
     }
   }
 }
+
+export function applyFrameAnchor({frameAnchor}, {width, height, marginTop, marginRight, marginBottom, marginLeft}) {
+  return [
+    /-?left$/.test(frameAnchor) ? marginLeft : /-?right$/.test(frameAnchor) ? width - marginRight : (marginLeft + width - marginRight) / 2,
+    /^top-?/.test(frameAnchor) ? marginTop : /^bottom-?/.test(frameAnchor) ? height - marginBottom : (marginTop + height - marginBottom) / 2
```
Fil, line 232: I don't think we need the `-?` here since the keywords have already been validated.

mbostock: We do, because you can have e.g. “top” or “top-left” and we want to match on both.

mbostock: Oh, I see what you mean, because “top” will match both, and nothing else. That’s true, but I was a little worried that we’d introduce another name in the future that would coincidentally match the same prefix. I think it’s clearer to have the hyphen as an explicit delimiter even though it’s not strictly necessary.

Fil: I meant that matching "top" or "left" seems to be enough. Just a nit.

mbostock: Okay, I thought about `/^top($|-)/` instead, but we’d have to change this code if we introduce new frameAnchor values anyway, so it feels fine to take your suggestion.

Fil:
I'm confused by something—writing it quickly since we're having dinner.

I'm trying to add a multiline explainer on top of the aapl-bollinger example.

Ideally I should be able to say
`Plot.text([`Multiline\nexplainer`], {frameAnchor: "top-right", text: d => d})`

But this fails in two ways:
1. If my text is a single line, I have to pass explicit x: null, y: null options, otherwise things just don't work
2. Multiline just doesn't work, the x/y is not used.

For the point 1, I suppose we could have frameAnchor setting the default x and y. For point 2 it's probably due to something I didn't see in #682.

Capture d’écran 2022-01-19 à 20 20 05

bbl

mbostock:
For the first, that’s expected because Plot.text defaults x and y to first and second, respectively, using maybeTuple.

https://github.com/observablehq/plot/blob/b5a1724035cc48821754b1092b9b59e1cfe15913/src/marks/text.js#L105

https://github.com/observablehq/plot/blob/b5a1724035cc48821754b1092b9b59e1cfe15913/src/options.js#L96-L99

If you want to change that, we’d need to do something like type inference on the data, and default x and y to null if it’s an array of strings. Maybe that’s useful for text annotations? 🤷 

For the second, that’s the brokenness I was expecting here https://github.com/observablehq/plot/pull/682#pullrequestreview-856962806. With #682, you get something like this:

```svg
<text x="639.5" y="0.5">
  <tspan x="0" y="-0.18em">Multiline</tspan>
  <tspan x="0" y="0.8200000000000001em">explainer</tspan>
</text>
```

But the *x* and *y* on the tspans aren’t additive with those on the text: the *x* and *y* on the text are now ignored. So we have to use a *transform* attribute when we generate multiline text. I’ll revert this part.

mbostock:
Another thing we could do here is change the defaults for *textAnchor* and *lineAnchor* based on the specified *frameAnchor*… but I think we probably don’t want to do this as it would be more difficult to remember the behavior. This seems to work pretty well:

untitled-245

```js
Plot.plot({
  marks: [
    Plot.frame(),
    Plot.text([`Multiline\nexplainer`], {
      x: null,
      y: null,
      textAnchor: "end",
      lineAnchor: "top",
      frameAnchor: "top-right",
      text: d => d
    })
  ]
})
```

mbostock:
Okay, changed it so that if you specify a *frameAnchor* option, *x* and *y* now default to null instead of maybeTuple. That makes sense to me, since you’d never use the two of them together.

Fil (approved):

PR #691, opened 2022-01-20
first tentative for a layouts option

Fil:
two layouts are available:
- "darker" changes the fill channel
- "collide" collides the x/y/r channels

The layouts can be given as a function, which receives the *scaled* values, and returns a function of (index, dimensions), which is called for each facet. You can pass several layouts which will be called one after the other.

(Very rough: no error checking, mutation of the values arrays… no options…)

testing: https://observablehq.com/@fil/plot-layouts-initial-tests

issues/questions:
- [ ] currently doesn't allow dynamic layouts (where we would immediately return the <g> then proceed to compute and update the positions in an async fashion

### DIFF
--- a/src/layouts.js
+++ b/src/layouts.js
@@ -24,7 +24,7 @@ function layoutCollide({x: X, y: Y, r: R}) {
       .force("x", forceX(({x}) => x))
       .force("y", forceY(({y}) => y))
       .force("collide", forceCollide(({r}) => 1 + r));
-    simulation.nodes(nodes).tick(30);
+    simulation.nodes(nodes).tick(30).stop();
     for (const node of nodes) X[node.i] = node.x, Y[node.i] = node.y;
   };
 }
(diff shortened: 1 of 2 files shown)


mbostock (commented):
Quick thoughts:

1. I think we could get by with just a single mark.layout, rather than an array of layouts, and have layouts be responsible for composing (as we do with transforms).

2. I don’t think we’ll want named layouts because I expect that layouts will almost always want accompanying options.

3. I’d guess that the user would apply layouts the same way they apply transforms today, _e.g._ `Plot.dot(data, Plot.collide({…}))`, allowing layout options and assigning the _layout_ option. While it would also be possible to set the _layout_ option directly (again, as you can do with a transform), I think this is rare; by wrapping instead we can compose layouts and potentially a layout can change behavior based on shared options.

mbostock:
> currently doesn't allow dynamic layouts

Yeah, I think we should only consider synchronous layouts to start.

Fil:
superseded by #648

PR #692, opened 2022-01-20
promote boolean to constant

mbostock:
We do this for number and date, so we should for boolean, too. Previously #346.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -7,13 +7,14 @@ const TypedArray = Object.getPrototypeOf(Uint8Array);
 const objectToString = Object.prototype.toString;
 
 // This allows transforms to behave equivalently to channels.
-export function valueof(data, value, type) {
-  const array = type === undefined ? Array : type;
-  return typeof value === "string" ? array.from(data, field(value))
-    : typeof value === "function" ? array.from(data, value)
-    : typeof value === "number" || value instanceof Date ? array.from(data, constant(value))
-    : value && typeof value.transform === "function" ? arrayify(value.transform(data), type)
-    : arrayify(value, type); // preserve undefined type
+export function valueof(data, value, arrayType) {
+  const array = arrayType === undefined ? Array : arrayType;
+  const type = typeof value;
+  return type === "string" ? array.from(data, field(value))
+    : type === "function" ? array.from(data, value)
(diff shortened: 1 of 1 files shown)


Fil:
Agree that it's better than to filter them out.

But it's a little bit surprising at least in the following case:
> Plot.dot(data, {x…, fill: true})

Reading this code, you might imagine that fill: true means you want the dots to be filled, and fill: false that you prefer the default stroked shapes. Conversely for bars, you might thin that fill: false would reverse the default.

Instead what this does is give them a category which makes them filled blue — in both cases: true and false appear filled blue, at least if there's nothing else that drives more colors.

mbostock:
> Plot.dot(data, {x…, fill: true})

It’s equivalent to this, though, right?

```js
Plot.dot(data, {x…, fill: () => true})
```

And when that’s defined by data…

```js
Plot.dot(data, {x…, fill: d => d.valid})
```

You almost certainly do want the categorical color encoding. So, I think this makes sense for consistency.

An alternative would be, like we do for the *identity* color scale, if all the associated defined values are booleans, we could map that to “currentColor” and “none” or some such. But then the false values would typically not be visible at all, unless you also had a stroke. Do you prefer that?

Fil:
It's logical in all the cases to have fill: true create filled dots. The fact that they appear to be blue is secondary (and could be fixed by defaulting the ordinal scale to ["currentColor"] when the domain is [true]). The surprising bit is that fill: false makes them blue too.

What if, when the domain is comprised only of booleans+undefined, we would map true to "currentColor" and false to maybe a middle grey #888—and undefined to invalid? Then I think it would do the expected thing (at least "expected in my mind", I reckon that there are several ways to think about this).

mbostock:
I did roughly that in #693, but mapping false to “none”. Wouldn’t saying fill: false and getting a (non-none) fill be confusing?

mbostock:
Another possibility is that we could special-case the default scheme if the associated domain is entirely booleans. So, it’d still be an ordinal scale, but you’d set dark/light gray instead of tableau10. I’ll give that a shot…

Fil:
This last option (special case a boolean-only ordinal scale to two greys) has my vote.

Fil (approved):
great in combination with #694

PR #693, opened 2022-01-20
shorthand boolean colors

mbostock:
Not sure if this is a good idea, but this implements https://github.com/observablehq/plot/pull/692#issuecomment-1017795011. I think my preference is that we consider booleans to be abstract data rather than literal colors, independent of whether they are specified as constants or channels.


mbostock:
Closing in favor of #694.

PR #694, opened 2022-01-20
boolean greys

mbostock:
When the defined values associated with an ordinal or categorical color scale are entirely boolean (true or false), construct a default range from the *greys* color scheme, assigning the light gray to false and the dark gray to true (rather than *turbo* or *tableau10*). In addition, make the *n* = 2 ordinal schemes higher contrast by favoring the extrema.

I considered having categorical color scales continue to use *tableau10*, restricting this behavior to ordinal color scales. This would require changing the logic in asOrdinalType to similarly check whether the associated defined values are boolean. I wasn’t sure if it was desirable to do this check twice. And I went back and forth over whether this should be an exhaustive check or a fast check (looking only at the first defined value). But I think it’s better to do an exhaustive check here since the the *greys* color scheme won’t work very well if there are many values.

Screen Shot 2022-01-20 at 11 27 54 AM

### DIFF
--- a/src/scales/ordinal.js
+++ b/src/scales/ordinal.js
@@ -4,7 +4,7 @@ import {ascendingDefined} from "../defined.js";
 import {maybeSymbol} from "../options.js";
 import {none} from "../style.js";
 import {registry, color, symbol} from "./index.js";
-import {ordinalScheme, quantitativeScheme} from "./schemes.js";
+import {maybeBooleanRange, ordinalScheme, quantitativeScheme} from "./schemes.js";
 
 export function ScaleO(scale, channels, {
   type,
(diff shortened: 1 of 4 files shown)


Fil:
Yes. Can we extend this to having systematically **two** greys when the domain is boolean, always associating false with the lighter grey and true with the darker one? Right now if everything is true or everything is false, we get the same output. (This is what I had in mind in https://github.com/observablehq/plot/pull/692#issuecomment-1017819124)

mbostock:
Ah, whoops. I thought I did that, but then it got lost in the iterations… Let me think some more.

mbostock:
Okay, take another look @Fil? 🙏

Fil (approved):
In combination with #692 it's 👍

And a great solution for the legend when only one boolean is present

PR #695, opened 2022-01-20
Add plot cheatsheets

mkfreeman:
Adds a link to the cheatsheets landing notebook and uploads the PDF. Unsure if we want to feature them this prominently, though I didn't see another obvious place for them in the `README.md` file. 

Note, this is a **draft**, as the PDF is not yet finalized and the notebook is not yet published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
 * [Changelog](./CHANGELOG.md)
 * [Contributing](./CONTRIBUTING.md)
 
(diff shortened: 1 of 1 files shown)


mbostock:
Is this the primary place where they will be hosted? I think since they’re large and may change over time, it might be desirable to host them somewhere else where they can change independently from Plot’s source.

mkfreeman:
I'm happy to have them live somewhere else -- not sure if there's a way to just host them in a notebook and download them from there?

mkfreeman:
The branch for this PR contains a file that is larger than desired and will be deleted.

Issue #383, opened 2021-05-12
Automatic margins based on tick label length


ericemc3:
A Plot could be internally composed in a harmonious and adaptive way, taking into account the actual bounding box of tick labels and axis titles for example, rather than having you to manually specify margins. 
 
Here I need to enter a guestimate `marginLeft:110` to accomodate labels width:

![image](https://user-images.githubusercontent.com/17407796/117985421-f7b05100-b338-11eb-987a-1603d5795de0.png)

A manual specification of margins is not robust: if unit or data change, or the font family/size used for the axes, overlaps may occur, or unsightly gaps may appear.

Note: ggplot2 and vega-lite automatically handle margins/padding inferred from axis (and legends) bounding boxes.

Plus, an option comparable to `labelOverlap` in vega-lite would allow the density of tick labels on a numerical or temporal axis to be managed harmoniously, avoiding any overlap.

PS: Plot is a great library, very intellectually stimulating, and the 21 beautifully written articles in the https://observablehq.com/collection/@observablehq/plot collection should be read by all data scientists!

Fil:
Adapting to labels' lengths would also make the chart behave unpredictably. There should be a better method than guesstimating though.

ericemc3:
Well i don't see ggplot2 or vega-lite behave unpredictably to that regard, and i have already handle this with D3 in the past (SVG or canvas provides methods for measuring BB and texts length after painting). Could you please elaborate on what you fear might become unpredictable?

mbostock:
We chose not to do this both to keep things simple and because we want the margins to be consistent across plots for vertical alignment. However, I would like to truncate long labels with ellipsis rather than letting them get cropped by the edge of the chart #394, and I would also like to explore smarter defaults for large numbers (e.g., automatically converting to thousands, millions, or billions) #395.

That said, I could see opting-in to this somehow if the other approaches are insufficient.

ericemc3:
I agree that in any case a maximum label width should prevail and ellipsis are a convenient and elegant truncation option.

enjalot:
What about a compromise where the default marginLeft for ordinal scales is a bit wider than 60px, as that's often much too short for categorical labels.

Every time I make a Plot with ordinal Y axis I change the margin, because reading the labels is higher priority than the consistency in vertical alignment, especially while I'm in the process of exploring the data and trying to generate many Plots rapidly.

Fil:
[...]

PR #699, opened 2022-01-22
text wrap (lineWidth)

### LINKED ISSUE #688: Automatic text wrapping
Now that we support multiline text, what if we had a *lineWidth* option that wrapped text automatically to produce multiple lines? We could a simple lookup table of character widths (ref. [Gregor Aisch](https://observablehq.com/@gka/cheap-fit-text-to-circle)) or we could assume monospaced text.

mbostock:
Implements text wrapping using a *lineWidth* option on the text mark. Fixes #688. Tasks:

- [x] Don’t wrap by default (when *lineWidth* is Infinity).
- [x] Allow *lineWidth* to be specified in ems.
- [x] ~~Allow *lineWidth* to be specified in pixels? (But how?)~~ No.
- [x] Preserve newlines in input.
- [x] ~~Optionally ignore newlines in input?~~ No.
- [x] ~~Optionally coalesce multiple spaces in input?~~ No.
- [x] Support monospace text layout (a *monospace* option, say)?
- [x] ~~Support tabular numerals?~~ No.
- [x] ~~Measure text metrics dynamically with Canvas, if fast enough?~~ No.
- [x] ~~Allow custom implementations for text metrics?~~ No.
- [x] ~~Implement Unicode line breaking algorithm.~~ No, too hard.
- [x] ~~Implement Unicode grapheme cluster breaking algorithm.~~ No, too hard.
- [x] ~~Implement Knuth–Plass line breaking algorithm.~~ No, too hard.
- [x] Documentation
- [x] Tests

Even though this is very rudimentary, I think it will be super useful and can be improved in the future based on usage.

I think the biggest blocker is just whether it’s okay to have *lineWidth* in ems. That’s the easiest thing for us to implement based on the baked-in rudimentary text metrics, but I think people are likely to get confused because they are more likely to be thinking in pixels. The problem with pixels is that the *fontSize* isn’t necessarily in pixels, either #378 and so we’d almost certainly need to invoke the browser’s full text metrics in order to compute it. And that also means an element will need to be attached to the DOM, and it’s slow, etc. So, I think maybe *lineWidth* in ems is the best option under the “worse is better” philosophy. We could also expose a hook for computing text metrics yourself, but that feels super low-level, and also not necessary since you could just specify text with newlines if you want to do the wrapping yourself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1109,6 +1109,8 @@ The following text-specific constant options are also supported:
 * **textAnchor** - the [text anchor](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor) for horizontal position; start, end, or middle
 * **lineAnchor** - the line anchor for vertical position; top, bottom, or middle
 * **lineHeight** - the line height in ems; defaults to 1
+* **lineWidth** - the line width in ems, for wrapping; defaults to Infinity
+* **monospace** - if true, changes the default fontFamily and metrics to monospace
 * **fontFamily** - the font name; defaults to [system-ui](https://drafts.csswg.org/css-fonts-4/#valdef-font-family-system-ui)
 * **fontSize** - the font size in pixels; defaults to 10
 * **fontStyle** - the [font style](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style); defaults to normal
@@ -1116,6 +1118,8 @@ The following text-specific constant options are also supported:
 * **fontWeight** - the [font weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight); defaults to normal
 * **frameAnchor** - the frame anchor; top-left, top, top-right, right, bottom-right, bottom, bottom-left, left, or middle (default)
 * **rotate** - the rotation angle in degrees clockwise; defaults to 0
+
+If a **lineWidth** is specified, input text values will be wrapped as needed to fit while preserving existing newlines. The line wrapping implementation is rudimentary; for non-ASCII, non-U.S. English text, or for when a different font is used, you may get better results by hard-wrapping the text yourself (by supplying newlines in the input). If the **monospace** option is truthy, the default **fontFamily** changes to “ui-monospace, monospace”, and the **lineWidth** option is interpreted as characters (ch) rather than ems.
 
 The **fontSize** and **rotate** options can be specified as either channels or constants. When fontSize or rotate is specified as a number, it is interpreted as a constant; otherwise it is interpreted as a channel.
 
--- a/src/marks/text.js
+++ b/src/marks/text.js
@@ -20,7 +20,8 @@ export class Text extends Mark {
       lineAnchor = /^top/i.test(frameAnchor) ? "top" : /^bottom/i.test(frameAnchor) ? "bottom" : "middle",
       lineHeight = 1,
       lineWidth = Infinity,
-      fontFamily,
+      monospace,
+      fontFamily = monospace ? "ui-monospace, monospace" : undefined,
       fontSize,
       fontStyle,
       fontVariant,
@@ -46,6 +47,7 @@ export class Text extends Mark {
     this.lineAnchor = keyword(lineAnchor, "lineAnchor", ["top", "middle", "bottom"]);
     this.lineHeight = +lineHeight;
     this.lineWidth = +lineWidth;
+    this.monospace = !!monospace;
     this.fontFamily = string(fontFamily);
     this.fontSize = cfontSize;
     this.fontStyle = string(fontStyle);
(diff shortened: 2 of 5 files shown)


Fil:
- lineWidth in ems make sense—maybe we could name the option lineChars to make that clearer? The monospace option is also very welcome.

- Optionally coalesce multiple spaces in input: yes this is useful if we're using the date formatter that produces a space for the day's numbers: i've sometimes had to use `d3.utcFormat("%b %e")(date).replace(/  +/, " ")` and it's ugly.

- Optionally ignore newlines in input? Doesn't seem necessary, it's easy enough to replace(\n, " ").

- Would there be a way to support non-breaking spaces, which can be quite important when the text contains numbers (e.g. "1,000~units" should be preserved, if possible)?

mbostock:
> maybe we could name the option lineChars to make that clearer

I think this requires a little prognostication. Here are some possible futures:

1. We support lineWidth in px, em, and perhaps other CSS units; px is preferred.
2. We support lineWidth in px, em, and perhaps other CSS units; em is preferred.
3. We don’t support px; em is implied.

In future 1, we’d want `lineWidth: "40em"` for ems and `lineWidth: 200` or `lineWidth: "200px"` for pixels. 

In future 2, we’d want `lineWidth: 40` or `lineWidth: "40em"` for ems and `lineWidth: "200px"` for pixels. 

In future 3, we’d want `lineWidth: 40` for ems… or possibly some name other than “lineWidth”, but “lineWidth” is symmetric with “lineHeight” in the sense that both are multiplied by the font size. And I would avoid “chars” because it implies counting characters, which is different from measuring ems (unless you’re using a monospaced font).

So if we think there’s a reasonably high probability that we want `lineWidth: 200` to mean 200 pixels in the future (future 1), then it’s probably worth doing `lineWidth: "40em"` for now and throwing an error on other units. On the other hand if we’re not sure that we’ll implement lineWidth in pixels in the nearish future (future 3), *or* if we think that ems will be the preferred unit for specifying the length of lines (future 2), then we should leave the current behavior as-is.

I think em is probably preferred to px, given that Plot is mostly about positioning in abstract coordinates anyway (scaled data space). I vote we do nothing.

> The monospace option is also very welcome

Yes. I think for due diligence it’s probably also worth benchmarking what the performance would be like if we measured the text with canvas rather than using a hard-coded character width map. That would also make it possible to support lineWidth in px, though as I implied above I don’t think that’s a priority.

> Optionally coalesce multiple spaces in input: yes this is useful if we're using the date formatter that produces a space for the day's numbers: i've sometimes had to use d3.utcFormat("%b %e")(date).replace(/ +/, " ") and it's ugly.

You can use `d3.utcFormat("%b %-e")`. 🙂

> Optionally ignore newlines in input? Doesn't seem necessary, it's easy enough to replace(\n, " ").

This seems conflicting with your previous input… I think I would only do these together or not at all, similar to [CSS white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space), i.e., `whiteSpace: "normal"` to coalesce spaces and ignore newlines, `whiteSpace: "pre-line"` to coalesce spaces and preserve newlines, and `whiteSpace: "pre-wrap"` to preserve spaces and newlines (the current behavior). This would also allow `whiteSpace: "pre"` which would ignore the lineWidth option. But in any case, I also don’t think it’s hard to replace spaces in the input, and so don’t consider this pressing and feel we shouldn’t implement it now.

> Would there be a way to support non-breaking spaces, which can be quite important when the text contains numbers (e.g. "1,000~units" should be preserved, if possible)?

Yes, this is already supported. You can type option-space for a non-breaking space, or use `\xa0`.

Fil (approved):
TIL about %-e

Thanks for the detailed explanations—it's really neat!

mbostock:
I did some exploration using context.measureText, which would allow several things:

- We could specify lineWidth in pixels (and possibly any other CSS unit).
- We could support any fontSize, fontFamily, and fontVariant accurately.

However, it was also about 10× slower than the current approach, and added some complexity (such as needing to create an invisible canvas element and attaching it to the DOM; you can’t set font-variant-numeric via the *context*.font attribute). So, I’m inclined to just keep it simple to start and revisit this in the future according to demand.

Fil:
yes. I was thinking that in node we don't have canvas as a dependency (only as devDependency)

mbostock:
I fixed the definition of [em](https://en.wikipedia.org/wiki/Em_(typography)). In CSS, 1em is not the width of a capital M, but instead an alias for the current fontSize. I had computed the default width map using 100px ui-system, but the layout is different at 10px ui-system (presumably because of hinting to align with pixel boundaries)? So, I recomputed the default width map at 10px ui-system, and then multiplied by 10 to avoid to produce integers. This will make the default metrics less accurate at larger font sizes, but I assume that most of the time text will be rendered with the default font size, and so we should prioritize that.

I’ve also implemented the monospace option. I’ll add a test or two and then merge.

yurivish:
Would the `defaultWidth` function implemented here also potentially work as a heuristic for providing default margin sizes based on tick label lengths (https://github.com/observablehq/plot/issues/383)?

Issue #501, opened 2021-08-14
render API


Fil:
https://github.com/observablehq/plot/pull/498 has introduced function marks: `() => svg node`

That render function is called like any other and receives as input (data, scales, channels, dimensions), with empty data and channels, all the instantiated scales, and complete dimensions.

It would be useful I think to design and document this API (for all marks). The changes I'd suggest would be:

1. pass the facet's "content" (i.e. "Chinstrap")
2. order the API as (dimensions, scales, facet, data, channels)
3. change the dimensions in such a way that we can ignore margins (ie width would be the current width-marginLeft-marginRight, and the <g> would already be translated by marginLeft, marginTop)

https://observablehq.com/d/cfdce446a293784f

mbostock:
We can’t change this API without breaking backwards compatibility, so I do not wish to redesign it. I agree we should document _mark_.render though.

Fil:
Would you be against passing *facet* as a fifth argument?

mbostock:
I realized we are inconsistent with what we pass mark.render. Without faceting, we pass {index, scales, values, dimensions, axes}; with faceting we pass {index, scales, values, dimensions}. So, if we want to pass the current facet it would need to be a sixth argument, or we’d somehow need to squirrel it into one of the existing arguments. But aside from how I’m onboard with somehow exposing the current facet values to marks during rendering.

Fil:
One of the recurring frustrations with the current API is that _render_ doesn't know who the parent SVG and the parent FIGURE will be.

This makes it awkward to build interactivity, since we have to wait (asynchronously) for the `g` we return to be appended to the plot before we can bind listeners, or know where to dispatch our custom events to update the viewof.

It's still probably too early to cast this in concrete, but it should be included in the discussion about the API. Having the 6th argument be {facet} would allow to extend it in the future to {facet, parentSVG, figure} or something else (a callback when the figure is added to the DOM, maybe).

Fil:
I did a bit of documentation in https://observablehq.com/d/cfdce446a293784f ; but def. not a blocker for 0.2

Fil:
superseded by #1263

PR #700, opened 2022-01-24
color <funciri>

### LINKED ISSUE #362: Support SVG patterns
`fill: "url(#hatched)"` should be detected by the color matcher, and treated as a constant rather than a field

reference: https://www.w3.org/TR/SVG11/painting.html#SpecifyingPaint

mbostock:
Fixes #362. Also fixes an unfiled bug where “none” or “currentColor” with surrounding whitespace would not properly be recognized as a constant color.


Fil:
This is so cool!

A side issue with the function marks in the <del>examples</del> snapshot tests is that the technique isn't perfect when using faceting: in that case the `<defs>` end up being repeated in each facet.

We could do something like
~~~js
 (function() { let i = 0; return () => i++ ? null : svg`…` })()
~~~

but it's quite a lot to expect of the user. Related: #501

mbostock:
> A side issue with the function marks in the examples snapshot tests is that the technique isn't perfect when using faceting

True, but that doesn’t seem like it should be a blocker for this PR.

I’m not sure what approach we want to take here. Some possibilities…

1. I don’t think that marks should be stateful (which would make them harder to compose and reuse in novel contexts). 
2. We could have a “defs” property that is parallel to marks where you can put hooks for rendering into a defs element, but that seems very special purpose for this feature, and I expect we want something higher-level for creating patterns and gradients (eventually). 
3. We could pass an additional argument to mark.render that indicates faceting. Technically this is already true because the _axes_ argument is undefined when faceting, but we still don’t know which facet is being rendered. I do feel like mark.render already takes too many arguments though.
4. A render hook for scales? That’s really what we want here—the patterns are to support the _color_ scale. If we wanted to support textures as a built-in scale type, we’d need to provide a way for scales to “render” their supporting elements.

It’s probably worth moving this discussion to another (possibly existing) issue. I don’t see that there’s anything blocking us from supporting funciri for color.

Fil (approved):

PR #702, opened 2022-01-25
labelArrow

### LINKED ISSUE #510: Axis label helper?
If I make a plot with custom axis labels I want them to match the style of the built-in labels, and often wind up searching for unicode arrows to copy-and-paste.

Might it make sense for Plot to provide helper functions to make it a little easier to stay consistent with the default convention?

```javascript
Plot.formatAxisX = text => text + ' →'
Plot.formatAxisY = text => '↑ ' + text
````

mbostock:
Fixes #510. TODO:

- [ ] I’m getting a little confused as to whether *label* and *labelArrow* are options on scales or axes. For the most part, we only need labels (and arrows) on axes. But, labels are also used for legends. So, I probably shouldn’t only implement *labelArrow* on axes, as we will probably also want them on legends. But (color) legends are always oriented left-right, and we don’t support labels on all legends (related #621).

This isn’t a high priority so I’m just going to let this sit. I thought it would be easier. 🙁

### DIFF
--- a/src/axes.js
+++ b/src/axes.js
@@ -71,7 +71,7 @@ export function autoScaleLabels(channels, scales, {x, y, fx, fy}, dimensions, op
   }
   for (const [key, type] of registry) {
     if (type !== position && scales[key]) { // not already handled above
-      autoScaleLabel(key, scales[key], channels.get(key), options[key]);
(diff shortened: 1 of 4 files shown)


mbostock:
This might want to be incorporated into #1197.

mbostock:
Superseded by #1566.

PR #703, opened 2022-01-25
make it easier to create a text halo 

### LINKED ISSUE #666: Plot.text halo option(s)

Fil:
by defaulting paintOrder to stroke and strokeWidth to 3 if the stroke option is specified.

supersedes #667
closes #666

Note: I balanced between 3 and 4 for the default, and took the conservative approach. It creates a 1.5px "halo" which is just enough, maybe sometimes a bit too small, but at least never looks gross. Not setting strokeWidth results (as you mentioned) in a 0.5px stroke, which is not a very good default.


src/marks/text.js
```diff
@@ -39,7 +39,10 @@ export class Text extends Mark {
         {name: "rotate", value: numberChannel(vrotate), optional: true},
         {name: "text", value: text, filter: nonempty}
       ],
-      options,
+      {
+        ...options.stroke && {paintOrder: "stroke", strokeWidth: 3},
```
mbostock, line 43: The strokeWidth only has an effect if there is a stroke, so we can simply move this to up the `defaults` declaration. Likewise, we could move the `paintOrder` up to `defaults` too and add some logic to style.js so only apply the default `paintOrder` if there’s both a fill and a stroke.

Alternatively, I was thinking that the strokeWidth would be 3 only if _both_ a fill and stroke are specified, and that it would be 1.5 (or 1?) if only the stroke were specified.

PR #706, opened 2022-01-25
Update boolean scheme9 scales

mkfreeman:
The lighter color of the boolean color scales is difficult to see against a white background. This branch updates the default scale. See [this notebook](https://observablehq.com/@observablehq/boolean-colors-pr) to compare results.

Screen Shot 2022-01-25 at 5 29 37 PM

### DIFF
--- a/src/scales/schemes.js
+++ b/src/scales/schemes.js
@@ -144,7 +144,7 @@ const ordinalSchemes = new Map([
 function scheme9(scheme, interpolate) {
   return ({length: n}) => {
     if (n === 1) return [scheme[3][1]]; // favor midpoint
-    if (n === 2) return [scheme[3][0], scheme[3][2]]; // favor extrema
+    if (n === 2) return [scheme[5][1], scheme[5][3]]; // favor extrema, avoid lightness
     n = Math.max(3, Math.floor(n));
     return n > 9 ? quantize(interpolate, n) : scheme[n];
   };
--- a/src/scales/schemes.js
+++ b/src/scales/schemes.js
@@ -144,7 +144,7 @@ const ordinalSchemes = new Map([
 function scheme9(scheme, interpolate) {
   return ({length: n}) => {
     if (n === 1) return [scheme[3][1]]; // favor midpoint
-    if (n === 2) return [scheme[5][1], scheme[5][3]]; // favor extrema, avoid lightness
+    if (n === 2) return [scheme[3][1], scheme[3][2]]; // favor darker
     n = Math.max(3, Math.floor(n));
     return n > 9 ? quantize(interpolate, n) : scheme[n];
   };
--- a/test/output/colorSchemesOrdinal.html
+++ b/test/output/colorSchemesOrdinal.html
@@ -3635,7 +3635,7 @@
         align-items: center;
         margin-right: 1em;
       }
-    </style><span class="plot-107-swatch" style="--color: #deebf7;">blues</span><span class="plot-107-swatch" style="--color: #3182bd;">1</span>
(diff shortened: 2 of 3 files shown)


zanarmstrong:
+1 to this feature!

(how light the hue was came up in data viz dev pairing this morning, and @mkfreeman ran with it!)

mkfreeman:
I see that I need to update the test, looking into that now.

src/scales/schemes.js
```diff
@@ -144,7 +144,7 @@ const ordinalSchemes = new Map([
 function scheme9(scheme, interpolate) {
   return ({length: n}) => {
     if (n === 1) return [scheme[3][1]]; // favor midpoint
-    if (n === 2) return [scheme[3][0], scheme[3][2]]; // favor extrema
+    if (n === 2) return [scheme[5][1], scheme[5][3]]; // favor extrema, avoid lightness
```
mbostock, line 147:
```suggestion
    if (n === 2) return [scheme[3][1], scheme[3][2]]; // favor darker
```

What about this instead?

mkfreeman: I like it! Updated [here](https://observablehq.com/@observablehq/boolean-colors-pr). 

This will now fail the test that `colorSchemesOrdinal must match snapshot`. I'm not sure how to go about updating a test that matches HTML output files (do I change those directly)?

mbostock: You can regenerate the snapshots by saying `rm -rf test/output` and then `yarn test`. Or you can copy over the `-changed` file to replace the snapshot.

mbostock: (This is covered in [CONTRIBUTING.md](https://github.com/observablehq/plot/blob/main/CONTRIBUTING.md).)

mkfreeman: Ah, great - thank you. Will do that now.

mbostock:
This PR looks good. I’ve also now enabled branch restrictions for this repo to make it explicit that approval is expected before merging. 🙏

mkfreeman:
Thanks! I wasn’t sure if the review above was explicit approval, but was recently told that the Author of the PR should merge (so when I saw the button enabled, I thought it was on me).

mbostock:
No, that was a comment. An approval looks like this:

Screen Shot 2022-01-25 at 3 49 56 PM

You should merge, but only after it’s approved. (I’ve now made that restriction explicit.)

zanarmstrong:
Just curious about the motivation for the switch from: 
` if (n === 2) return [scheme[5][1], scheme[5][3]];`
to
`if (n === 2) return [scheme[3][1], scheme[3][2]]`
?

The most important goal (to me) was to avoid the lightest end of the spectrum, which is now solved for*.

But, sometimes the dark end is pretty dark too. So, when we were testing things, I thought that [5][1] / [5][3] was a nice balance between getting a lot of intensity difference between the two boolean colors while also avoiding the darkest darks and lightest lights. 

All in all, I'm super happy with this change and am 100% fine leaving it as `if (n === 2) return [scheme[3][1], scheme[3][2]]`. 

That said, I am curious to understand why you prefer `if (n === 2) return [scheme[5][1], scheme[5][3]];` :). Thank you!

_* except the Magma / Viridis / Plasma family which go from dark-to-light by default instead of light-to-dark like "greys", etc._

mkfreeman:
This was @mbostock's suggestion, but to me it `[3][1] - [3][2]` feels less arbitrary: ` [5][1] - [5][3]` felt like a guess (I still wasn't sure that the colors were dark _enough_).

Fil:
The general hand-wavy argument is that if a scheme is good for 3 colors, then a pair extracted from that scheme should be good for 2 colors?

zanarmstrong:
sounds good - thanks!

PR #708, opened 2022-01-26
extend to text and vector; document

Fil:
- [x] Pending issue: the default is 3 for dodge, but can be 4.5 for dot if there is a symbol channel.


mbostock (commented):
It seems wrong for marks to define channels they don’t use, just so they can be made available to an unrelated layout.

In the case of a constant *r*, this change isn’t necessary as the dodge layout already reads the *r* option even if the downstream mark doesn’t. If we want do do a channel (variable) *r*, then we’ll need a way for layouts to be able to register extra channels. But, do you think that’s required?

README.md
```diff
@@ -1824,6 +1824,35 @@ Plot.stackX2({y: "year", x: "revenue", z: "format", fill: "group"})
 
 Equivalent to [Plot.stackX](#plotstackxstack-options), except that the **x2** channel is returned as the **x** channel. This can be used, for example, to draw a line at the right edge of each stacked area.
 
+## Layouts
+
+A layout processes the transformed and scaled values of a mark before rendering. A layout might, for example, modify the marks’ positions to avoid occlusion. A layout operates in the representation space (such as pixels, *i.e.* after scales have been applied) rather than data space.
+
+### Dodge
+
+The dodge layout can be applied to the Dot, Text and Vector marks.
```
mbostock, line 1833: It can be applied to any mark.

Fil:
I want to support the same layout on text marks, so we can do this:
Capture d’écran 2022-01-26 à 18 00 19

Maybe marking r optional in the mark itself is the wrong approach, and what we need instead is to allow the layout to ask for the (scaled) channels it wants to use? It would seem more appropriate in terms of separation of concerns.

We also don't want users to have to pass the *r* option differently—and maybe *r* is going to be the result of a bin transform—, so it shouldn't be a dodgeOption but part of the normal options.

mbostock:
> I want to support the same layout on text marks, so we can do this

As I said, your test works as-is in the mbostock/dot-dodge branch (because the dodge layout will check for the *r* option even if it’s ignored by the mark). And similarly in the case of a bin transform that produces an *r* channel, the dodge layout will consume it.

Fil:
> if you want an r channel (i.e., variable radius) but the downstream mark doesn’t support it

Yes, this is what I was aiming for with these changes (the test plot I added is simpler, and it already works). I like the idea of adding a layout.channels.

The default of 3 vs 4.5 for symbols is not a huge issue, even though it would be nicer if we respected the same defaults.

I realize I forgot to mention images. Beyond that, I haven't been able to make it work with any other mark yet, like ticks, rules, lines… not that it needs to be working, but I think the documentation needs to capture where a layout can be applied.

Closing for now.

mbostock:
> The default of 3 vs 4.5 for symbols is not a huge issue, even though it would be nicer if we respected the same defaults.

Yes. One possibility is that, rather than consuming the passed-in *options*, the layout could consult `this.r` when it is invoked. This way it would be able to check if the mark provides a default radius. But in the case that a mark doesn’t implement the *r* option, the dodge layout would presumably still need to fallback to the passed in *r* option.

Fil:
An example here https://observablehq.com/@recifs/dodge-text

PR #711, opened 2022-01-27
Dodge: use the mark's this.r if present

Fil:
otherwise default to reading the options, then 3.

follows [7a845f0]


src/layouts/dodge.js
```diff
@@ -39,8 +39,9 @@ export function dodgeY(dodgeOptions = {}, options = {}) {
 }
 
 function dodge(y, x, anchor, padding, options) {
-  const [, r] = maybeNumberChannel(options.r, 3);
-  return layout(options, (I, scales, {[x]: X, r: R}, dimensions) => {
+  const [, r0] = maybeNumberChannel(options.r, 3);
+  return layout(options, function(I, scales, {[x]: X, r: R}, dimensions) {
+    const r = this.r == null ? r0 : this.r;
```
mbostock, line 44: In cases of option “inheritance” or defaults, we typically check against undefined rather than null. Also, we should try to avoid computing values that aren’t needed. The precedence order here is:

* The `r` channel (`R`)
* The `r` mark property (`this.r`)
* The `r` option (`options.r`)

We may not even need maybeNumberChannel here, since if `options.r` were interpreted as a channel, then `R` would be defined, so we can use it.

```suggestion
  return layout(options, function(I, scales, {[x]: X, r: R}, dimensions) {
    const r = R ? undefined : this.r !== undefined ? +this.r : options.r !== undefined ? +options.r : 3;
```

(I kinda want to use the null coalescing operator `??` here, but I think the correct interpretation of null here is zero rather than undefined, since null represents a defined value. I kinda wish they made it consistent with default arguments.)

src/layouts/dodge.js
```diff
@@ -39,8 +39,8 @@ export function dodgeY(dodgeOptions = {}, options = {}) {
 }
 
 function dodge(y, x, anchor, padding, options) {
-  const [, r] = maybeNumberChannel(options.r, 3);
-  return layout(options, (I, scales, {[x]: X, r: R}, dimensions) => {
+  return layout(options, function(I, scales, {[x]: X, r: R}, dimensions) {
+    const r = R ? undefined : this.r !== undefined ? +this.r : options.r !== undefined ? +options.r : 3;
```
Fil, line 43:
```suggestion
    const r = R ? undefined : this.r !== undefined ? this.r : options.r !== undefined ? +options.r : 3;
```

I think this.r is necessarily a number?

mbostock: It is in the case of all the marks we’ve implemented, but theoretically it might not be for some other custom mark someone else implements. Similarly, it’s theoretically possible that the `r` channel is not numbers—and we don’t protect against that. So, sure, we can remove the `+` here if you like.

test/plots/penguin-dodge-symbol.js
```diff
@@ -0,0 +1,19 @@
...
+    marks: [
+      Plot.dot(penguins, Plot.dodgeY({
+        x: "body_mass_g",
+        symbol: "species",
+        fill: "species",
+        stroke: "white",
+        strokeWidth: .5
+      }))
+    ],
+    symbol: {legend: true}
```
mbostock, line 17: For Plot options, I prefer the following order:

* Layout (e.g., height)
* Scale
* Facets
* Marks

I think (hope) nearly all the tests conform to this pattern, and would like to be consistent.

test/plots/penguin-dodge-symbol.js
```diff
@@ -0,0 +1,19 @@
...
+  const penguins = await d3.csv("data/penguins.csv", d3.autoType);
+  return Plot.plot({
+    height: 220,
+    marks: [
+      Plot.dot(penguins, Plot.dodgeY({
+        x: "body_mass_g",
+        symbol: "species",
+        fill: "species",
+        stroke: "white",
+        strokeWidth: .5
```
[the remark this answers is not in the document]
mbostock: Or about instead

```js
Plot.dot(penguins, Plot.dodgeY({x: "body_mass_g", symbol: "species", stroke: "species"}))
```
Screen Shot 2022-01-27 at 11 05 48 AM

Then we don’t need the little stroke for separation.

mbostock:
Superseded by #775.

PR #712, opened 2022-01-27
call the layout once with all the frame indices

Fil:
implements my comment https://github.com/observablehq/plot/pull/648#issuecomment-1022492736

The layout is called once, with an index that is an array of arrays, one for each facet.

this addresses two concerns:
1. makes it possible to create "smarter" layouts that can work on either the elements of each facet or the whole set of elements
2. performance (memory allocation is done once for all the facets)

the cost is a slightly more complex function to write, because it needs to iterate on the arrays in index; but this is consistent with what we do in transforms.

### DIFF
--- a/src/layouts/dodge.js
+++ b/src/layouts/dodge.js
@@ -40,40 +40,43 @@ export function dodgeY(dodgeOptions = {}, options = {}) {
 
 function dodge(y, x, anchor, padding, options) {
   const [, r] = maybeNumberChannel(options.r, 3);
-  return layout(options, (I, scales, {[x]: X, r: R}, dimensions) => {
+  return layout(options, (index, scales, {[x]: X, r: R}, dimensions) => {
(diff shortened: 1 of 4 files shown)


mbostock:
As you point out, there is a cost to doing this: an increase in the complexity of every layout. Therefore we should be confident that the added expressiveness (or performance) is worth the added complexity. In the case of transforms, we knew that transforms had to be facet-aware _e.g._ so that the same set of bins would be used across facets. It is not clear to me that the same is true of layouts: none of the current or proposed layouts appear to require knowledge of facets. Also it is not clear that the performance cost of the current approach is appreciable.

Even if we want facet-aware layouts in the future, it seems likely that this will be the less common case. So there is likely a way we could design that is backwards-compatible where some layouts could opt-in to the added complexity of facets when needed, letting most layouts remain simple.

Issue #540, opened 2021-09-12
Contrasting text color?


mbostock:
Often it’s desirable for a text label to have a strong contrast against its background. If the background color is constant (e.g., white in the empty space of the chart), then it’s fine to hard-code the text color. But if the background color is variable, as when labeling cells, then it’s more difficult. For example, the extreme values (e.g., 4.5 and 9.2) are hard to read here:

<img width="648" alt="Screen Shot 2021-09-12 at 10 09 06 AM" src="https://user-images.githubusercontent.com/230541/132996521-16c918fd-8d38-417e-be2b-124d3761d535.png">

It’s difficult to do this well in Plot currently.

One approach is to define the fill color as a channel. However, since the fill channel is bound to the color scale, you cannot express the color literally; you must specify an abstract value that is passed through the color scale. Hence, this is generally not an option. This problem was discussed in #56, and identity scales were added in #305, however, this isn’t really a good fix for this problem because we don’t want to disable the color scale entirely; we just want to disable it for this specific mark channel.

Another approach is to repeat the text mark definition twice, once for light text and once for dark text, and use the _filter_ option to control which data gets which color. This produces the desired result, but it’s tedious.

<img width="645" alt="Screen Shot 2021-09-12 at 10 17 36 AM" src="https://user-images.githubusercontent.com/230541/132996765-41cd86f2-dace-4287-9c89-30bfacd64040.png">

```js
Plot.text(simpsons, {
  x: "season",
  y: "number_in_season",
  text: d => d.imdb_rating?.toFixed(1),
  title: "title",
  filter: d => d.imdb_rating >= 5.5 && d.imdb_rating <= 8.5,
  fill: "black"
}),
Plot.text(simpsons, {
  x: "season",
  y: "number_in_season",
  text: d => d.imdb_rating?.toFixed(1),
  title: "title",
  filter: d => d.imdb_rating < 5.5 || d.imdb_rating > 8.5,
  fill: "white"
})
```

This can be alleviated with a helper mark:

```js
function ctext(data, {invert, ...options} = {}) {
  const filter = Plot.valueof(data, invert);
  return [
    Plot.text(data, {...options, filter, fill: "white"}),
    Plot.text(data, {...options, filter: filter.map(d => !d)})
  ];
}
```

[...]

PR #713, opened 2022-01-27
document layouts

Fil:
note: the custom layout example will change a little if we do https://github.com/observablehq/plot/pull/712

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,9 @@ When drawing a single mark, you can call *mark*.**plot**(*options*) as shorthand
 ```js
 Plot.barY(alphabet, {x: "letter", y: "frequency"}).plot()
 ```
-### Layout options
+### Geometry options
 
-These options determine the overall layout of the plot; all are specified as numbers in pixels:
+These options determine the overall geometry of the plot; all are specified as numbers in pixels:
 
 * **marginTop** - the top margin
 * **marginRight** - the right margin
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1824,6 +1824,54 @@ Plot.stackX2({y: "year", x: "revenue", z: "format", fill: "group"})
...
+When it *options* have a *layout* property, the layout function is called after the data has been faceted and scaled; it receives as inputs the index, scales, values (scaled channels) and dimensions, and the mark as this. It must return the (possibly modified) values.
+
+For example, the following custom layout makes every fill color darker:
+
+```js
+Plot.dot(data, {
+  layout: (index, scales, values, dimension) => {
+    if (values.fill) {
+      for (const i of index) {
+        values.fill[i] = d3.rgb(values.fill[i]).darker();
```
mbostock, line 1867: We shouldn’t use/recommend mutation.

README.md
```diff
@@ -1824,6 +1824,54 @@ Plot.stackX2({y: "year", x: "revenue", z: "format", fill: "group"})
...
+
+#### Custom layouts
+
+When it *options* have a *layout* property, the layout function is called after the data has been faceted and scaled; it receives as inputs the index, scales, values (scaled channels) and dimensions, and the mark as this. It must return the (possibly modified) values.
+
+For example, the following custom layout makes every fill color darker:
+
+```js
+Plot.dot(data, {
+  layout: (index, scales, values, dimension) => {
```
mbostock, line 1864:
```suggestion
  layout(index, scales, values, dimension) {
```

README.md
```diff
@@ -1824,6 +1824,54 @@ Plot.stackX2({y: "year", x: "revenue", z: "format", fill: "group"})
...
+    if (values.fill) {
```
mbostock, line 1865: This example does nothing out of the box, because values.fill will always be undefined. Perhaps a more realistic example would be better? For example, maybe you could have a helper function…

```js
function darker(key = "fill", amount = 1) {
  return (index, scales, {fill, ...values}, dimension) => {
    fill = fill.slice();
    for (const i of index) fill[i] = d3.rgb(fill[i]).darker(amount);
    return {...values, fill};
  };
}
```

And then

```js
Plot.dot(penguins, {fill: "body_mass_g", layout: darker()})
```

I guess even this feels very contrived, since it’d be better to do this on the scale definition. Maybe we should remove the custom layouts section until we can think of a better motivating example?

Fil: layout: contrasting on Plot.text could be short (and fix #540). A good contrasting color not as simple as checking a color's lightness, though, so an official implementation would likely diverge from such an example.

Removing this part for now.

mbostock: Choosing a contrasting color sounds interesting and useful. Perhaps the _fill_ input channel specifies the background color (e.g., `fill: "imdb_rating"` in the Simpsons example), and then the layout outputs a new _fill_ channel where each output color is either light or dark depending on the lightness of the input? Something like:

```js
Plot.text(simpsons, Plot.contrast({
  x: "season",
  y: "number_in_season",
  text: d => d.imdb_rating?.toFixed(1),
  title: "title",
  fill: "imdb_rating"
}))
```

Though, some sort of magic like this might be better…

```js
Plot.text(simpsons, {
  x: "season",
  y: "number_in_season",
  text: d => d.imdb_rating?.toFixed(1),
  title: "title",
  fill: Plot.contrast("imdb_rating")
})
```

Which, I guess, would be like a per-channel layout.

Fil: I'll open a new issue and prototype :)

mbostock:
Superseded by #775.

PR #717, opened 2022-01-28
brush facets; target root element

Fil:


mbostock:
Thanks. I’m going to try a slightly different approach, but this is the functionality that I want.

mbostock:
Implemented in 46c496e179a5cc38cb9c2cbb1ba9352d70bc05ab in the parent branch. Thanks for the assistance!

PR #719, opened 2022-01-28
prototype channel alias

mbostock:
This is a quick sketch at allowing channels to be shared across marks as mentioned in https://github.com/observablehq/plot/pull/708#issuecomment-1022415533. We should think through the syntax and the semantics a bit more.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -14,6 +14,7 @@ export function valueof(data, value, arrayType) {
     : type === "function" ? array.from(data, value)
     : type === "number" || value instanceof Date || type === "boolean" ? array.from(data, constant(value))
     : value && typeof value.transform === "function" ? arrayify(value.transform(data), arrayType)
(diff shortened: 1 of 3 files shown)


Fil:
I made this helper to allow inlining the functionality:

~~~js
combine = (data, ...rest) => {
  const known = new Map();
  return Array.from(rest, ({ mark, ...options }) => {
    for (let k in options) {
      if (known.has(k)) options[k] = { alias: known.get(k) };
    }
    const m = mark(data, options);
    for (let k in options) {
      if (!known.has(k)) known.set(k, m);
    }
    return m;
  });
}
~~~

and the mark becomes:

```js
marks: [
 combine(
  data,
  { mark: Plot.dot, x: "bill_length", y: "body_mass", symbol: "species" },
  { mark: Plot.text, x: "bill_length", y: "body_mass", text: "sex" }
 )
]
```

mbostock:
Superseded by #798.

PR #721, opened 2022-01-29
pointer

mbostock:
Related #4. Like the new brush mark, except whatever is near the pointer is considered selected.

https://user-images.githubusercontent.com/230541/151642007-86ba0519-7292-49fd-a375-1b31f2ffc29f.mov

TODO

- [ ] Should the pointer mark display anything when there’s a selection?
- [ ] Should you be able to control what the pointer mark displays, or is it just 4px-radius circles?
- [ ] Validate the *mode* option
- [ ] Optimize the display so that only the visible circles are instantiated
- [ ] Do something nice for band scales?
- [ ] A mode where things are only selected while the pointer is down?
- [ ] … and where shift-dragging will add to the current selection?
- [ ] … and maybe option-dragging removes from the current selection?


Fil:
From the screen capture my feeling is that this is a different system than the one implemented for brush. With brush, the interaction mark shows the selected region, and is not responsible for highlighting the selected data points. Here the interaction mark shows the individual selected dots.

We should aim for a consistent approach (brush, lasso, pointer) and just show the selection region as an overlay of the corresponding shape.

The task of highlighting the selected marks is the same in all cases, and would be handled by the dot mark instead, maybe with the suggested `selected: { stroke: "red" }` option?

Rather than mode, could we use pointer, pointerX and pointerY? Conversely, would we prefer to remove brushX brushY and have only one brush with a mode option? (I think I like the mode approach better, if it works consistently.)

mbostock:
> We should aim for a consistent approach (brush, lasso, pointer) and just show the selection region as an overlay of the corresponding shape.

Consistency is generally a good thing, but it’s worth thinking this through.

I find it helpful to think of two forms of selection: the _geometric_ selection and the _logical_ selection.

The geometric selection describes the area—some region of the plane—that is selected. For a brush that is an axis-aligned rectangular region defined by _x1_, _x2_, _y1_, and _y2_. For a lasso it is some complex polygon or set of polygons. For a pointer it might be a circle around the mouse, or an infinitely tall or wide region for one-dimensional pointing. Or, it might be a buffered region of a given radius around the path traveled by the pointer (in the case of the more persistent “spray paint” pointing that I’d like to implement). In some cases, say if we select the *n* points that are closest to the pointer, there might not be a simple definition of the geometric selection!

The logical selection describes which data are selected. In Plot this is represented exclusively by a subset of the mark’s index (*node*[Plot.selection]). This is the value that the plot exposes to the rest of the notebook, so you can e.g. show a table or a coordinated visualization. The logical selection, in a sense, is what matters to the user. It’s what the user seeks to control through interaction. The geometric selection is the means to that end: _how_ to indicate that selection (conveniently). So, showing the logical selection, perhaps _in addition to_ the geometric selection, is usually a good thing because it provides more direct feedback as to the effect of the user’s input.

In the case of a brush, the geometric selection is persistent: after brushing a rectangular region, you can drag that rectangular region around. We could, possibly, do this behavior with the lasso as well, where you draw an arbitrary polygon, and then you can pick up and move that polygon around to change your selection. But I don’t think we want or need that. Instead, my expectation with a lasso is that the geometric selection (the polygon) is ephemeral: it dissolves as soon as I lift the pointer, and I’m left only with a logical selection as a set of selected points. Similarly when I’m using the pointer, if we want to support a persistent selection where I can click and drag to select some points near the mouse, and then shift-click and drag to select some other points, I’d rather just see the logical selection. I don’t think I want a visual representation of a buffered region around the path of the pointer while it was down.

The challenge with showing the logical selection is that, unlike the geometric selection, we may not know an appropriate visual representation. If interaction marks only understand that you’re selecting infinitesimal points in *x* and *y*, then we could draw circles around those points, say, but that might not be a great representation say when you’re selecting circles of varying radii, or bars, images, rects, etc. But, maybe that’s fine? Maybe it just needs to convey a hint that you’ve selected something, even if it’s not perfect. The goal of Plot is just to get you something “good enough”, quickly.

As we’ve discussed, I like the idea of an internal selection channel maintained by Plot and some mechanism for marks to express style overrides when selected. Maybe it’s even possible for Plot to have defaults so that all marks have a default selected state. But there remain a variety of technical and design challenges to get this to work, and we’ve already merged the brush mark and I would like to add pointer, lasso, and perhaps others. Perhaps when we add selection styles in the future, we change the interaction marks so that they don’t have a visual representation of the logical selection? Or at least a way to turn it off.

> Rather than mode, could we use pointer, pointerX and pointerY? Conversely, would we prefer to remove brushX brushY and have only one brush with a mode option? (I think I like the mode approach better, if it works consistently.)

It’s useful to control separately (1) whether the points to be selected are defined in *x* or *y* or both and (2) whether the selection region around the pointer is defined in *x*  or *y* or both. As I showed in the video, the points are defined in *x* and *y*, but I only want to select in *x*. I think this will be a common pattern when there is an independent and dependent variable (e.g., time and temperature, respectively). Also, the separate constructors (pointer, pointerX, and pointerY) don’t just set the mode, they also providing convenient defaults for *x* and *y*: maybeTuple if both *x* and *y* are specified, and otherwise identity if only one of *x* and *y* is specified. I think we want to maintain this pattern for consistency with other marks in Plot. I think the defaults are setup correctly that you won’t have to think about it most of the time, I hope.

If we want the brush to show the logical selection, we’ll need to take the *mode* option implemented here and implement that for brush, too.

mbostock:
Here’s my other idea for a visual representation of the logical selection. The first video shows mode *xy* and the second mode *x*.


https://user-images.githubusercontent.com/230541/151676770-28a3b5de-60a3-4b66-b072-c241f5725aa1.mov


https://user-images.githubusercontent.com/230541/151676771-e8837560-25aa-493d-aea6-72100fcd4cb0.mov


It looks kinda neat, but I don’t think it will work as well when we make the pointer selection persistent using click-and-drag (and letting you point to multiple areas with shift-click-and-drag, or deselect with option-click-and-drag). And also this style wouldn’t work with the brush, where the position of the pointer isn’t relevant. And, perhaps most importantly, while looking cool it does put most of the emphasis on the pointer location which is already visible with the pointer anyway. But leaving it here for the record. 🙂

mbostock:
> Rather than mode, could we use pointer, pointerX and pointerY?

Maybe a different interpretation of your comment is that pointer means _mode_ = *xy*, pointerX means _mode_ = *x*, and pointerY means _mode_ = *y*. So, you’d never need to specify the *mode* option (or perhaps, you’d only specify the *mode* option with pointer, but we’d expect you to use pointerX or pointerY instead). However, unlike Plot’s other similar constructors such as dotX and dotY, you’d be allowed to specify a *y* channel for pointerX, or an *x* channel for pointerY. Maybe that’s confusing? I think maybe that‘s confusing but it’s not a strong preference so let me know what you prefer.

Fil:
I'm confused by your remark since dotY does accept both x and y (it's just defaulting y to identity, and x to undefined) ;-)

Anyway, yes, I'd be inclined to use pointer as 2d (mode=xy), pointerY as mode=y and pointerX as mode=x, and in that case there is no need to expose "mode" as an option.

In terms of visual feedback, I don't like the circle much since it moves in sporadic jumps, whereas the link version is a bit less jarring as it's never static (it's still very jumpy though); but this can probably be smoothed out with a (quick) transition, a less aggressive shape and color, I don't know. link + circle would work too, the link easing the visual movement, but stopping before it actually occludes the target. 

I think i'd use it most often to select the (n = 1) closest element to the pointer rather than a whole range, but it's more gut feeling than any experience with this type of selection mechanism. My hunch is that with a large n the flock of lines will tend to occlude a majority of the selected dots, and the ones that are not occluded will be those farthest from the pointer, which seems a bit opposite to the goal.

mbostock:
> I'm confused by your remark since dotY does accept both x and y

Oh, whoops. I got confused because it wasn’t always that way; prior to 5b81573b374129f7d1811d0644ba6e2625841472 dotX forced _y_ to be null, and dotY forced _x_ to be null.

So, does this mean brushX should allow _y_, and brushY should allow _x_? 🤔 

https://github.com/observablehq/plot/blob/551f53a1b1a63cd9c099e125f4c8929ce13f52f6/src/marks/brush.js#L81-L87

Fil:
I think the difference between brushX(… {y}) and brush(… {y}) would be if the brush highlights the logical selection; in the first case brushX(… {y}) the brush is horizontal, but the highlighted values are dots positioned in x and y. In the second case it is a 2-d brush.

mbostock:
Okay, I think I got it in the latest commit. Now brushX and brushY allow you to specify _y_ and _x_, respectively. And pointerX and pointerY similarly only provide defaults for _mode_ and _x_ and _y_, respectively, while allowing you to specify _y_ and _x_ respectively.

mbostock:
I’ve implemented persistent selection (click and drag, and shift-click and drag). Let me know what you think, @Fil!

mbostock:
Some open tasks:

- [ ] Allow the radius of pointer circles to be specified as *r* (a constant? a channel?).
- [ ] Rename the current *r* option to *distance*, or something else, maybe *maxR*?
- [ ] Allow *fill* on pointer. (It’s currently overridden by fill="none".)
- [ ] Allow the *fill* and *stroke* be specified as channels, not just constants?
- [ ] Making clearing the persistent pointer selection less twitchy. (Any pointermove will interfere.)
- [ ] Allow *dx* and *dy* on pointer.
- [ ] For one-dimensional pointers, draw blue rules rather than dots?
- [ ] Support data transforms (*e.g.*, bin and group).

The last task also applies to the brush—it’ll require some core changes, not just changes to pointer. The problem is that *mark*.data isn’t the same as the data returned by the transform. We have to decide whether we’re selecting from the original (untransformed) data or, perhaps more likely, selecting from the transformed data (_e.g._, selecting bins).

Fil:
This branch currently breaks brushX and brushY on a 2-d scatterplot.

can be fixed by 020a1ea5bec4776698193a57eac44b6faa4d132d

mbostock:
> This branch currently breaks brushX and brushY on a 2-d scatterplot.

Based on our earlier discussion, it is behaving as expected. (Or, it’s not clear what you mean by “breaks”.) In your example you are specifying *x* and *y*, so brushX behaves the same as brush. The only difference between brush, brushX, and brushY is the defaults for *x* and *y*.

If you want one-dimensional brushing:

```js
Plot.brushX(data, {x: "culmen_depth_mm"})
```

Or:

```js
Plot.brush(data, {x: "culmen_depth_mm"})
```

If you want two-dimensional brushing:

```js
Plot.brush(data, {x: "culmen_depth_mm", y: "culmen_length_mm"})
```

Fil:
The case I found "broken" is when you want the data to be a 2-d scatterplot, and the brush to be on X only. This (in my mind) is what you want when you say brushX(data, {x, y}). It's also what's currently implemented in main.

Since we're not highlighting the logical selection, it's true that we can use brush(data, {x}) instead. Maybe I just need to adapt my mental model… but for now I don't see what is the purpose of brushX and brushY anymore, if the mode is driven by the x and y options. 🤔

mbostock:
> but for now I don't see what is the purpose of brushX and brushY anymore, if the mode is driven by the x and y options

It’s the same as dotX and dotY. The only purpose is to specify different defaults.

https://github.com/observablehq/plot/blob/94d69a54eea6a085657e66c04b8412da51bacff8/src/marks/dot.js#L95-L101

The brush doesn’t support a *mode* option currently. We could add one for symmetry with the other marks, but it’s a little redundant because brush doesn’t need the *x* and *y* channels in the other modes. Meaning if we did, you’d be passing in (and computing) the *y* channel with *mode* = *x*, but you wouldn’t be using it for anything.

Fil:
I've reverted my commit and updated my mental model!

mbostock:
I’ve made a few small improvements (to handle band scales, to improve the display of one-dimensional pointing using rules instead of circles, and to change the default *n* to 1). The big outstanding tasks are:

- [ ] Support data transforms (*e.g.*, bin and group)
- [ ] Support *x1*, *x2*, *y1*, and *y2* (*e.g.*, for selecting bins or rects)

I’m currently thinking that when you use brush or pointer with a transform, that you’re still selecting the source (input) data, i.e., the data prior to aggregation; you won’t be selecting e.g. bins.

mbostock:
This branch now reverts #748, restoring the brush #71 and fixes #5.

src/marks/brush.js
```diff
@@ -0,0 +1,87 @@
...
+              }
+              if (Y) {
+                let [y0, y1] = X ? [extent[0][1], extent[1][1]] : extent;
+                if (y.bandwidth) y0 -= y.bandwidth();
+                S = S.filter(i => y0 <= Y[i] && Y[i] <= y1);
+              }
+            }
+            if (!selectionEquals(this[selection], S)) {
+              this[selection] = S;
+              this.dispatchEvent(new Event("input", {bubbles: true}));
```
enjoylife, line 63: Not sure how much of a "public" api your wanting to expose in these events... but access to the selected data indexes or the range of the data is needed. e.g.
```js
this.dispatchEvent(new CustomEvent("input", {bubbles: true, detail: {selected: S, range}}));
```
where range is something like `if(X && x.invert) range = extent.map(x.invert);`

Then in the listener, you can use the indexes for filtering right away or show the textual range to to the user. 

Without access to the indexes used, the only thing callers have is the` plot.value`, which only has what was selected. It does not provide what criteria was used for the selection, nor what was left out.

kevinschaich:
Hey guys – any update on this? I saw [this demo on observable](https://observablehq.com/@fil/plot-brush-71) not sure if it ever got released) but would be keen to see this feature in a recent version on NPM 🚀

PR #725, opened 2022-01-29
axis ariaDescription

mbostock:
This…

* Consolidates code for setting the aria-label and aria-description of an axis
* Allows the user to specify a custom aria-description, while retaining the default aria-label
* By default, uses the (visible) scale label as the aria-description rather than the aria-label

I know it’s tempting to put the scale label in the aria-label of the axis because they’re both “labels”, but the current default axis label (e.g., “x-axis”) is a better label than the scale label which in Plot indicates only the abstract dimension of data, i.e. the input, not the output visual channel such as *x* or *y*. And if we try to combine the scale label into the axis label, then we have to separate them somehow (e.g., with a comma “x-axis, Date →” or with parens “x-axis (Date)”, though the latter won’t work well since we already use parens for percentages etc.). And then the user can’t set one without affecting the other.

I believe we are supposed to use aria-describedby here, but I would prefer to avoid a random/incrementing id, so I’m hoping that using aria-description will be equivalent. We could switch to aria-describedby if necessary.

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -2,7 +2,7 @@ import {axisTop, axisBottom, axisRight, axisLeft, create, format, utcFormat} fro
 import {boolean, take, number, string, keyword, maybeKeyword, constant, isTemporal} from "./options.js";
 import {formatIsoDate} from "./format.js";
 import {radians} from "./math.js";
-import {impliedString} from "./style.js";
+import {applyAttr, impliedString} from "./style.js";
 
 export class AxisX {
   constructor({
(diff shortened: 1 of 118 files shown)


Fil:
It's getting better :)

It looks like aria-description is fairly recent and not well supported at the moment (https://html5accessibility.com/stuff/2021/02/09/aria-description-by-public-demand-and-to-thunderous-applause/ ; https://caniuse.com/?search=aria-description). I haven't been able to get assistive tools to say the aria-description. In the meantime, the aria-label works well ((and everywhere https://caniuse.com/?search=aria-label).

mbostock:
Well, I’m all for being forward-looking and embracing aria-description. Being forward-looking has worked out well for me in the past. 😁

Fil:
I've opened a test account on https://assistivlabs.com/ and tried with NVDA and Narrator, without much success either. But I think that this is going in the right direction, and that independently of what assistive tools already offer or not, we can build on these elements (for example, a script that would add a keyboard navigation/tooltip behavior).

PR #730, opened 2022-01-31
lasso

Fil:


mbostock:
Fun! Let’s continue this discussion: https://github.com/observablehq/plot/pull/721#issuecomment-1024955303

src/marks/lasso.js
```diff
@@ -0,0 +1,165 @@
...
+    this.activeElement = null;
+  }
+
+  // The lasso polygons follow the even-odd rule in css, matching the way
+  // they are computed by polygonContains.
+  render(index, scales, {x: X, y: Y}, dimensions) {
+    const margin = 5;
+    const {ariaLabel, ariaDescription, ariaHidden, fill, fillOpacity, stroke, strokeWidth} = this;
+    const {marginLeft, width, marginRight, marginTop, height, marginBottom} = dimensions;
+    if (!this.lasso) this.lasso = [];
```
mbostock, line 35: Am I missing something? I don’t see this.lasso being read anywhere. We only push onto it during render. So I’m guessing it could be removed.

More generally, I expect marks to be able to be used across plots. Like, you can create a mark instance, and then pass it to multiple plot instances. So we probably should avoid storing any interaction state on the mark instance itself. Possibly we’ll need Plot to provide a state store that is initialized when Plot.plot is called, so that a mark has a place to store interaction state.

Fil: that was a bit of cruft, now removed; thanks!

Fil:
There is an issue sometimes where the pointer is not released appropriately.

PR #732, opened 2022-02-01
grouped reduce in area and line marks

### LINKED ISSUE #724: Reduce for line/area values?
In lines and areas, the stroke, fill, fillOpacity, strokeOpacity, strokeWidth, title, and ariaLabel channels are reduced to the first element of the series in https://github.com/observablehq/plot/blob/5ce6cc0595b29839174b804c98f85ffbc4335770/src/style.js#L163

It would be interesting to be allowed to customize this reducer, in particular for title and ariaLabel, where we might want to display something like "10 data points with a mean of 3.3".

But this requires a structural change, because at present the scales are applied before the call to mark.render, and the grouping is done in render.

A solution could be to make any mark that needs grouping request that grouping (by *z* and by facet), by adding a "group: true" option for the mark, a grouper: true option for the grouping channel (*z*), and a *grouped: true* option on the channels which must be reduced (currently all defined in style.js). Then we could imagine that a line mark would be defined as

> Plot.line(data, {x, y, z, ariaLabel: { value: "name", reduce: "mode" }})

or

> Plot.line(data, {x, y, z, ariaLabel: { value: "name", reduce: series => `some stats about the series` }})

Fil:
closes #724 (and #503)

todo:
- [ ] avoid the multiplication of the reduced value(s) to the whole channel
- [ ] reuse the group reducers if possible?

supersedes #508

### DIFF
--- a/README.md
+++ b/README.md
@@ -945,7 +945,25 @@ In addition to the [standard mark options](#marks), the following optional chann
 
 By default, the data is assumed to represent a single series (a single value that varies over time, *e.g.*). If the **z** channel is specified, data is grouped by *z* to form separate series. Typically *z* is a categorical value such as a series name. If **z** is not specified, it defaults to **stroke** if a channel, or **fill** if a channel.
 
-The **fill** defaults to none. The **stroke** defaults to currentColor if the fill is none, and to none otherwise. If both the stroke and fill are defined as channels, or if the *z* channel is also specified, it is possible for the stroke or fill to vary within a series; varying color within a series is not supported, however, so only the first channel value for each series is considered. This limitation also applies to the **fillOpacity**, **strokeOpacity**, **strokeWidth**, and **title** channels. The **strokeWidth** defaults to 1.5 and the **strokeMiterlimit** defaults to 1.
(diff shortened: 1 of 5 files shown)


mbostock:
This is still materializing a mapped channel even though we only need a single value. And as you noted it doesn’t use the existing reducers. I’ll try to take a crack at this soon, but I want something that does both of these things before merging.

Fil:
yes a bit more work is needed. I wanted to rebase and clean up the previous PR so we could make progress on this.

To sum up the current status: this is suboptimal because it populates a channel with *n* copies of the reduced value. But I've been struggling with the fact that we can't reduce to a single value either: we need a channel, because we have to carry an element for each line or area. The difficulty is that the reduction must to happen in a grouped transform (z + facet, like map does), in a manner that is compatible with Plot.window, before we apply scales, and before we group.

mbostock:
Should we close this, since we landed #761?

Fil:
Yes. I still think there is value in the Plot.reduce helper described in https://github.com/observablehq/plot/issues/724#issuecomment-1024937329.

Issue #733, opened 2022-02-03
When explicit bin thresholds are specified, but no domain is specified, the last bin is dropped


mbostock:
E.g., https://observablehq.com/d/748823354c2435e1

![untitled-246](https://user-images.githubusercontent.com/230541/152270664-fc7efa67-257f-4d47-9d0c-7eb05cc80ce1.png)

```js
Plot.plot({
  x: {
    domain: [0, 24]
  },
  marks: [
    Plot.rectY(
      data,
      Plot.binX(
        { y: "count" },
        {
          x: (d) => d.timestamp.getHours(),
          thresholds: d3.range(1, 24)
        }
      )
    )
  ]
})
```

mbostock:
It’s not a bug. You have to specify the domain explicitly or there’s no way to know how big the last bin should be.

PR #735, opened 2022-02-03
bin interval

### LINKED ISSUE #734: Intervals for binning
When you have data with a certain regularity, e.g. integers, it’d be nice if you could indicate that to the bin transform. This would allow it to “nice” the domain to contain that interval, and compute the appropriate thresholds. For example, say you have hourly data, you might want to specify the *interval* option instead of *thresholds*:

```js
Plot.plot({
  marks: [
    Plot.rectY(data, Plot.binX({y: "count"}, {
      x: d => d.timestamp.getHours(),
      domain: [0, 24],
      interval: 1
    })),
    Plot.ruleY([0])
  ]
})
```

Previously #733.

mbostock:
This…

* Allows numeric intervals (such as integers) to be specified as the *interval* option, while avoiding ambiguity with the *thresholds* option where a number represents the desired number of thresholds.
* When *thresholds* is specified as an numeric interval (typically indirectly by using the *interval* option, but also if *thresholds* is specified explicitly as an interval object), extends the domain to cover the first and last interval automatically. This avoids zero-width bins as seen in #733, and we already do this for temporal data.

Fixes #734. For example, with hourly data:

Screen Shot 2022-02-02 at 7 40 43 PM

```js
Plot.plot({
  marks: [
    Plot.rectY(timestamps, Plot.binX({y: "count"}, {x: d => d.timestamp.getUTCHours(), interval: 1}))
  ]
})
```


Fil (approved):
👌

PR #740, opened 2022-02-08
hexbin

Fil:
demo:  https://observablehq.com/@observablehq/plot-hexbin-740

- [x] ideally this will evolve to be a transform (or layout-transform), which would allow to pass the reducers cleanly, and to scale cleanly. Currently the legend will not display the correct value.
- [x] test all reducers, forbid those that don't work.
- [x] check what happens if x or y are missing
- [x] add a mesh (needs the clip: true option of v0.4.1)
- [ ] documentation


Fil:
Status: everything seems to work, I'm particularly happy with the color reducers (for example a hex can have a fill: min, stroke: max, title: "proportion-reduce").

However, I don't like the quantity of messy code I had to put in the layout logic to derive the new indices (in part because of the "one call per facet" logic), and deal with the new and updated/rescaled data, which sometimes wants to create new scales. There are several cases to consider (for example, when applying a fill: min reducer, we need to make sure that fill is not scaled before it's reduced), and I think there must be a simpler approach that I'm failing to see. (Also, by way of consequence, it's hard to merge this branch with main.)

I'll pause here for a moment.

mbostock:
Superseded by #775.

PR #743, opened 2022-02-09
Plot.scale(x) exposes bandwidth and step when applicable

### LINKED ISSUE #741: Get bandwidth for band scale
I might be missing something, but it seems that there's no simple way to get the bandwidth used by a band scale. I'm trying to align additional elements to the center of each band, and currently it involves more math than when a band scale is used in D3.

It could be added as a field within here:
Screen Shot 2022-02-08 at 9 29 14 PM

Fil:
closes #741

ping @rlesser

### DIFF
--- a/src/scales.js
+++ b/src/scales.js
@@ -413,8 +413,7 @@ function exposeScale({
     // band, point
     ...scale.align && {align: scale.align(), round: scale.round()},
     ...scale.padding && (scale.paddingInner ? {paddingInner: scale.paddingInner(), paddingOuter: scale.paddingOuter()} : {padding: scale.padding()}),
-    ...type === "point" && {step: scale.step()},
-    ...type === "band" && {bandwidth: scale.bandwidth(), step: scale.step()},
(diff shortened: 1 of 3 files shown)


src/scales.js
```diff
@@ -413,6 +413,8 @@ function exposeScale({
     // band, point
     ...scale.align && {align: scale.align(), round: scale.round()},
     ...scale.padding && (scale.paddingInner ? {paddingInner: scale.paddingInner(), paddingOuter: scale.paddingOuter()} : {padding: scale.padding()}),
+    ...type === "point" && {step: scale.step()},
+    ...type === "band" && {bandwidth: scale.bandwidth(), step: scale.step()},
```
mbostock, line 417:
```suggestion
    ...scale.bandwidth && {bandwidth: scale.bandwidth(), step: scale.step()},
```

I think it’s fine (and possibly helpful) to return a bandwidth of zero for point scales and to reduce the variation in response shapes.

mbostock (approved):

PR #747, opened 2022-02-09
a chart with no colors defaults to the identity color scale

### LINKED ISSUE #744: default color scale should be identity
When no color channel is activated, the color scale defaults to linear-turbo, which is not a problem in the chart (since it's not used), but then shows as a turbo color *legend* coming out of nowhere.
image

(reported by @yurivish)

Fil:
- [ ] Should this be undefined instead? See discussion in #744.

fixes #744

cc: @yurivish


Fil:
superseded by #756 and #757

PR #751, opened 2022-02-10
warnings!

### LINKED ISSUE #536: Warning / logging system?
In some cases we'd want Plot to warn the user about some kind of misconfiguration or unfit data that causes issues.

- data points mapped to nothing (#493)
- negative data on a positive log scale (#534)
- too many ordinal values (#74)
- cycling the ordinal color scale because we have more categories than colors
- infinite values (#528)
- on-the-fly data coercion (#532)
- (more to add)

This meta-issue tries to capture all the cases when we'd want such a logging/warning system, and discussions on how they could be rendered.

A design issue is to know when it would be appropriate to display this — with a distinction between the case in which the user is actively coding (data exploration, development), vs. when the user passively consumes a chart, and the developer might want the chart to be more robust to errors.

mbostock:
Fixes #536 by introducing a warning mechanism. When a warning is triggered, a ⚠️ icon is displayed in the top-right corner of the plot.

Screen Shot 2022-02-09 at 9 02 24 PM

Mousing over the warning indicator gives the tooltip _e.g._ “1 warning. Please check the console.” The console shows the warnings:

Screen Shot 2022-02-09 at 9 48 08 PM

I have implemented three warnings so far:

* When dates are associated with an implicit band scale (typically because of a bar mark)
* When strings that appear to be dates are associated with an implicit band scale
* When strings that appear to be dates are associated with an implicit ordinal or point scale

I imagine we will think of lots of additional warnings in the future, but I think these three cover the most common mistake I’ve seen in Plot. We may want to have another tracking issue to cover ideas for other warnings (e.g., insets resulting in zero-width rects or bars, mixed-sign values with log scales, undefined or invalid data, etc.).

Ref. https://talk.observablehq.com/t/bar-chart-with-dates/6194/3

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -1,4 +1,4 @@
-import {create, cross, difference, groups, InternMap} from "d3";
+import {create, cross, difference, groups, InternMap, select} from "d3";
 import {Axes, autoAxisTicks, autoScaleLabels} from "./axes.js";
 import {Channel, channelSort} from "./channel.js";
 import {defined} from "./defined.js";
@@ -8,6 +8,7 @@ import {arrayify, isOptions, keyword, range, first, second, where} from "./optio
 import {Scales, ScaleFunctions, autoScaleRange, applyScales, exposeScales} from "./scales.js";
 import {applyInlineStyles, maybeClassName, maybeClip, styles} from "./style.js";
 import {basic} from "./transforms/basic.js";
+import {consumeWarnings} from "./warnings.js";
 
 export function plot(options = {}) {
   const {facet, style, caption, ariaLabel, ariaDescription} = options;
@@ -119,6 +120,19 @@ export function plot(options = {}) {
 
   figure.scale = exposeScales(scaleDescriptors);
   figure.legend = exposeLegends(scaleDescriptors, options);
+
+  const w = consumeWarnings();
+  if (w > 0) {
+    select(svg).append("text")
+        .attr("x", width)
+        .attr("y", 20)
+        .attr("dy", "-1em")
+        .attr("text-anchor", "end")
+        .text("⚠️")
+      .append("title")
+        .text(`${w.toLocaleString("en-US")} warning${w === 1 ? "" : "s"}. Please check the console.`);
+  }
+
   return figure;
 }
 
--- a/src/options.js
+++ b/src/options.js
@@ -222,6 +222,15 @@ export function isTemporalString(values) {
   }
 }
 
+// Are these strings that might represent numbers? This is stricter than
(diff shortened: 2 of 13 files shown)


src/plot.js
```diff
@@ -119,6 +120,19 @@ export function plot(options = {}) {
...
   figure.legend = exposeLegends(scaleDescriptors, options);
+
+  const w = consumeWarnings();
+  if (w > 0) {
+    select(svg).append("text")
+        .attr("x", width)
+        .attr("y", 20)
+        .attr("dy", "-1em")
+        .attr("text-anchor", "end")
+        .text("⚠️")
```
Fil, line 131:
```suggestion
        .attr("cursor", "help")
        .text("⚠️")
```

semantically this would feel more appropriate than the text cursor, but it occludes the ⚠️ symbol

Fil (approved):
Excellent!

I've edited the issue number in your comment to #536. I've also uplifted an old console.warn.

Suggestions for follow-ups:
* I wonder if the messages should start with `Plot warning: ` or  `Observable Plot warning: ` for easier filtering.
* I'd like to have a top-level option to disable any warning (or use the classic tier system of debug, info, warn, error?)
* how does https://github.com/observablehq/plot/pull/593 fit into this? We want the user to know that some indices have been ignored, but maybe they now it and want to ignore the warning
* at some point we'll want to expose warn and give access to this system to third party code

mbostock:
> I've edited the issue number in your comment to https://github.com/observablehq/plot/issues/536.

I was thinking we’d close that with this PR since we’ve implemented the warning system, and then have a new tracking issue for specific warnings. (Or separate issues or sets of related warnings.) But it’s not important. I can live without the joy of closing another issue. 😉

mbostock:
I wonder if I can similarly detect when strings are numbers? That would catch errors like using CSV data without typed: true.

Screen Shot 2022-02-10 at 3 24 40 PM

mbostock:
> I wonder if the messages should start with Plot warning: or Observable Plot warning: for easier filtering.

For what it’s worth, browser consoles will also filter by file name, so as long as Plot is loaded as e.g. “plot.js”, filtering by “plot” will work even if the messages themselves do not contain the word.

mbostock:
FWIW I considered adding a warning if you have numeric data with an implicitly ordinal scale (we already have a warning for temporal data with an implicitly ordinal scale). However it seemed like it might introduce some false positives. This will be something to keep an eye on when we release.

I also thought about having the warning include a link to a webpage that gives more detail on how to fix the warning. I think React does this pretty well?

Issue #181, opened 2021-03-03
Clipping a Text mark?


mbostock:
Should the Text mark support a clipping rect? A clipping circle? A way to reference another mark and use it as a clipping mask?

![untitled-25](https://user-images.githubusercontent.com/230541/109749712-0982b480-7b90-11eb-91db-5496d71c8c37.png)

Fil:
Maybe introduce a **clip**: "rect" option that consumes x1, x2, y1, y2?

Fil:
browser support for clip-path: polygon is not great…
Current best effort notebook: https://observablehq.com/@observablehq/plot-treemap-clip

PR #752, opened 2022-02-10
clip: true clips the mark to the frame

### LINKED ISSUE #165: Optionally clip marks to the canvas area?
And related, you might want overflow: visible on the SVG element?

Fil:
closes #165

this is just a first step, which does not address #181 where an individual mark (a text) needs to be clipped by a data-driven clip-path.


src/style.js
```diff
@@ -172,6 +174,29 @@ export function applyGroupedChannelStyles(selection, {target}, {ariaLabel: AL, t
...
+// https://github.com/observablehq/plot/issues/181
+export function maybeClip(clip) {
+  if (clip === true) return "frame";
+  if (clip == null || clip === false) return false;
+  throw new Error(`clip method not implemented: ${clip}`);
+}
+
+export function applyClip(selection, mark, {width, height, marginLeft, marginRight, marginTop, marginBottom}) {
+  if (mark.clip === "frame") {
+    const id = `plot-clippath-${++nextClipPathId}`;
```
mbostock, line 188: We’ll need to reindex these in the test framework so that the tests are independent. You should be able to mirror what we did for markers which have similar behavior.

mbostock (changes requested):
Good work! Thanks for doing this.

To reduce duplication, I think we can either (1) fold the call to applyClip into applyIndirectStyles (this would require passing dimensions to applyIndirectStyles) or (2) move the clipping logic up to plot.js so that it is part of the base Mark functionality (and the behavior cannot be overridden or ignored by Mark subclasses).

src/style.js
```diff
@@ -187,6 +198,17 @@ export function applyIndirectStyles(selection, mark) {
...
   applyAttr(selection, "paint-order", mark.paintOrder);
+  if (mark.clip === "frame") {
+    const id = `plot-clippath-${++nextClipPathId}`;
+    const w = width - marginRight - marginLeft;
+    const h = height - marginTop - marginBottom;
+    selection.append("clipPath")
+      .attr("id", id)
+      .append("rect")
+      .attr("width", w)
+      .attr("height", h);
```
mbostock, line 209:
```suggestion
    selection.append("clipPath").attr("id", id).append("rect").attr("width", w).attr("height", h);
```

src/style.js
```diff
@@ -187,6 +198,17 @@ export function applyIndirectStyles(selection, mark) {
   applyAttr(selection, "stroke-dasharray", mark.strokeDasharray);
   applyAttr(selection, "shape-rendering", mark.shapeRendering);
   applyAttr(selection, "paint-order", mark.paintOrder);
+  if (mark.clip === "frame") {
+    const id = `plot-clippath-${++nextClipPathId}`;
```
mbostock, line 202:
```suggestion
    const id = `plot-clip-${++nextClipPathId}`;
```

mbostock (approved):
A few minor suggestions.

I would also like to reformat the traffic-horizon.js example (no spaces within braces, and reordering the Plot.plot options to match the convention).

Also, it’d probably make sense to sort the data once outside the Plot rather than using the Plot.sort operator on each area mark? But either is fine.

test/plots/traffic-horizon.js
```diff
@@ -0,0 +1,31 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export default async function() {
+  const data = await d3.csv("data/traffic.csv", d3.autoType);
+  const bands = 7;
+  const step = +(d3.max(data, d => d.value) / bands).toPrecision(2);
```
mbostock, line 7: You could use d3.tickStep here (or d3.tickIncrement if you want to be really fancy).

test/plots/traffic-horizon.js
```diff
@@ -0,0 +1,31 @@
...
+  return Plot.plot({
+    facet: { data, y: "name" },
+    fy: { axis: null, domain: new Set(data.map((d) => d.name)) },
+    x: { axis: "top" },
+    y: { axis: null, domain: [0, step] },
+    color: { scheme: "blues", range: [0.05, 1], type: "ordinal", legend: true },
```
mbostock, line 13: Feels like a bit of a bummer than the range is needed here, like these color schemes don’t work out of the box? Is there something about this application that makes the ColorBrewer schemes to light, compared to say a choropleth?

test/plots/traffic-horizon.js
```diff
@@ -0,0 +1,31 @@
...
+          data,
+          Plot.sort("date", {
+            x: "date",
+            y: d => d.value - i * step,
+            fill: i * step,
+            clip: true
+          })
+        )
+      ),
+      Plot.text(data, Plot.selectFirst({ text: "name", frameAnchor: "left" }))
```
mbostock, line 26: I wonder if there’s some sort of axis style option that would make this slightly easier. This is fine so. Just musing.

Fil: It's because the default axis's is below the mark. And no z-index :-/

test/plots/traffic-horizon.js
Fil: Just a tiny aesthetic preference I had in the notebook—we can remove it.

Fil: I remember now why I played with this: the thing is that the first band represents values between 0 and 1 * step, and the paler shade of blue should represent 0; here what I'd really want is the color representing step/2. I'll make this more explicit.

test/plots/traffic-horizon.js
Fil: Not sure how? With 7 bands and a maximum value = 9616, the step is rounded to 1400, but d3.ticks(0, max, bands + 1) returns 9 bands (i.e. 10 ticks = [0, 1000, 2000, …, 9000]).

Besides the technical point, I like the original example which clearly shows (with the select for the number of bands) how horizon charts are built.

mbostock: In my PR #753 you can say:

```js
const bands = 7; // just a hint; not guaranteed
const max = d3.max(data, d => d.value);
const step = d3.tickStep(0, max, bands);
const ticks = d3.ticks(0, max, bands);
```

This results in 10 bands spaced every 1,000:

Screen Shot 2022-02-10 at 10 39 21 AM

mbostock: Or asking for five bands and labeling the lower bound rather than upper bound:

Screen Shot 2022-02-10 at 10 41 55 AM

PR #753, opened 2022-02-10
tweak horizon example

mbostock:
It feels much more natural to me that the _color_ scale would be defined in ~~band index~~ offset space (0 = the first offset, 1,400 = the second offset, and so on). The previous construction was dependent on the labels being in lexicographic order.

I also don’t grok why we can’t use the ordinal sequential color schemes as they are out of the box? Why do we need to recrop them using `range: [1 / (bands + 1), 1 - 1 / (bands + 1)]`? That feels like a complex thing to remember when you want to make a horizon chart.

### DIFF
--- a/test/plots/traffic-horizon.js
+++ b/test/plots/traffic-horizon.js
@@ -5,6 +5,7 @@ export default async function() {
   const data = d3.sort(await d3.csv("data/traffic.csv", d3.autoType), d => d.date);
   const bands = 7;
   const step = +(d3.max(data, d => d.value) / bands).toPrecision(2);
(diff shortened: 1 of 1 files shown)


Fil:
The idea behing `range: [1 / (bands + 1), 1 - 1 / (bands + 1)]` is that I wanted the color to reflect the value in the middle of the band.

Not super important when you make one chart like here. However if you have several charts with different number of bands, I felt it would be more consistent to have the first band of 3 have the same value as the second band of 9.

Also, you might want to mix a horizon mark with other marks—e.g., a few points that you'd want to highlight for some reason.

But again, these two use cases are not present in that example :)

PR #757, opened 2022-02-13
fix crash on null color legend

Fil:
(refs: #746 & #756)

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -46,8 +46,8 @@ function legendColor(color, {
   legend = true,
   ...options
 }) {
+  if (color?.domain === undefined) return;
   if (legend === true) legend = color.type === "ordinal" ? "swatches" : "ramp";
-  if (color.domain === undefined) return;
   switch (`${legend}`.toLowerCase()) {
     case "swatches": return legendSwatches(color, options);
     case "ramp": return legendRamp(color, options);


mbostock (changes requested):
I don’t think this is the right place to fix this problem (because the issue of a scale being null is not unique to color legends). I think we need to fix it inside `legend`~~or `exposeLegends`~~?

For example, after #756, the `if (isObject(scale))` is no longer sufficient to check whether a scale exists.

mbostock:
Should this return undefined? Or should it throw an error?

```js
Plot.legend({color: {}})
```

I guess it should throw an error (either `unknown legend type` or `missing scale`).

Fil:
I'm fine with both of these choices.

mbostock:
Okay. I think my recommendation is that it behaves the same as `Plot.legend({})` in the sense that there’s no scale defined, which is an error.

mbostock:
This is pretty tricky, because there are three places where we’re using isObject as a proxy for “is a given scale defined?”

At the top-level when iterating over the passed in *options*:

https://github.com/observablehq/plot/blob/08118dd73988250f2b931a43b479ce8f98e07a04/src/legends.js#L16

For a symbol scale to check whether there’s also a color scale:

https://github.com/observablehq/plot/blob/08118dd73988250f2b931a43b479ce8f98e07a04/src/legends.js#L20

In legendsSymbol to check whether there’s also a color scale:

https://github.com/observablehq/plot/blob/08118dd73988250f2b931a43b479ce8f98e07a04/src/legends.js#L26

It makes me wonder if #756 was a mistake since now the test for “is a scale defined” is more involved than testing whether the option exists.

mbostock:
Superseded by #759.

PR #758, opened 2022-02-14
respect margins when clipping

Fil:
(we'd have caught the bug in the test plot if we had added a Plot.frame)


src/style.js
```diff
@@ -205,7 +205,8 @@ export function applyIndirectStyles(selection, mark, {width, height, marginLeft,
     const id = `plot-clip-${++nextClipId}`;
     const w = width - marginRight - marginLeft;
     const h = height - marginTop - marginBottom;
-    selection.append("clipPath").attr("id", id).append("rect").attr("width", w).attr("height", h);
+    selection.append("clipPath").attr("id", id).append("rect").attr("width", w).attr("height", h)
+      .attr("transform", marginLeft || marginTop ? `translate(${marginLeft},${marginTop})` : null);
     applyAttr(selection, "clip-path", `url(#${id})`);
```
mbostock, line 210: There’s no reason to use applyAttr here (since the value is always defined).

src/style.js
```diff
@@ -205,7 +205,8 @@ export function applyIndirectStyles(selection, mark, {width, height, marginLeft,
     const id = `plot-clip-${++nextClipId}`;
     const w = width - marginRight - marginLeft;
     const h = height - marginTop - marginBottom;
-    selection.append("clipPath").attr("id", id).append("rect").attr("width", w).attr("height", h);
+    selection.append("clipPath").attr("id", id).append("rect").attr("width", w).attr("height", h)
+      .attr("transform", marginLeft || marginTop ? `translate(${marginLeft},${marginTop})` : null);
```
mbostock, line 209: I’d prefer to use x and y attributes.

mbostock (changes requested):
I’m going to make some tweaks to address my comments.

PR #759, opened 2022-02-15
fix for null scales

mbostock:
#756 changed the test used to determine whether or not a scale exists. This updates Plot.legend to match. Now `Plot.legend({color: {}})` throws an error because it does not consider the color scale to be defined (there is neither a type nor a domain). Supersedes #757.


mbostock:
`if (isScaleOptions(o) && o.legend)` will break a number of tests, such as decathalon. 🤔

Fil (approved):

PR #761, opened 2022-02-16
group aeshetics

### LINKED ISSUE #724: Reduce for line/area values?

### LINKED ISSUE #760: Filtering on grouped marks (area and line) behaves differently depending on the channel
The area and line marks have special filtering behavior for their positional channels (x and y for line; x1, y1, x2, and y2 for area). If some of the *x* values are NaN, for example, you’ll get gaps:

Screen Shot 2022-02-15 at 2 48 33 PM

Here *x* is defined as:

```js
d => d.unemployment < 14 ? d.date : NaN
```

However, if I do something similar with the *stroke* channel, the line will interpolate across missing values:

Screen Shot 2022-02-15 at 2 46 54 PM

Here *stroke* is defined as:

```js
d => d.unemployment < 14 ? d.division : NaN
```

The latter seems like a bug: it would be preferable to show gaps if any of the values are missing, rather than interpolating over the missing values.

This is related to the desire to specify custom reducers for grouped values, rather than being limited to just showing the first value for each group. #724

mbostock:
This PR allows areas and lines to have varying colors and other aesthetics, such as titles. For example, here is a line with varying stroke and stroke width:

Screen Shot 2022-02-16 at 2 06 41 PM

And here is an area with varying fill:

Screen Shot 2022-02-16 at 2 35 26 PM

This works by first grouping the index by Z, as before, and then iterating over the values. If the aesthetics for the next value do not match the previous value, a new group is created. The current index is repeated in the previous and next groups. In effect this means that the aesthetics for the current value apply to the current value _and the following value_. (In the future we could perhaps have an option to control this behavior.)

A side-effect of this change is that when some data is undefined, multiple path elements are generated. This is because undefined data and varying aesthetics are handled within the same loop, rather than relying on d3-shape’s _line_.defined and _area_.defined. It would probably be slightly better to use a single path element where possible, but this would complicate the grouping. (We’d need some other way to signal to D3 how to interrupt the path.)

This PR also removes the use of the _filter_ hint on channels, since now the area and line marks opt-out of filtering completely on all channels. I’m not super happy with this aspect of the design and would like to think about whether I can make it better. This also necessitated the axis marks implementing _mark_.filter, which I did by way of introducing a Decoration base class.

Fixes #760. Fixes #724.


src/style.js
```diff
@@ -176,6 +176,58 @@ export function applyGroupedChannelStyles(selection, {target}, {ariaLabel: AL, t
   applyTitleGroup(selection, T);
 }
 
+function groupAesthetics({ariaLabel: AL, title: T, fill: F, fillOpacity: FO, stroke: S, strokeOpacity: SO, strokeWidth: SW, opacity: O, href: H}) {
+  return [AL, T, F, FO, S, SO, SW, O, H].filter(c => c !== undefined);
+}
+
+export function* groupIndex(I, position, channels) {
+  const {z: Z} = channels; // group channel
+  const A = groupAesthetics(channels); // aesthetic channels
```
mbostock, line 185: We should probably special-case when there are no aesthetic channels.

mbostock:
I could maybe make this smarter regarding _z_ if I defer the implicit _z_ determination until render. (As an added bonus, this means that it won’t compute _z_ separately from the channel that it is _fill_ or _stroke_ channel it is derived from.) In render, if the _z_ option was undefined (but not null), then check the _fill_ or _stroke_ channel and also test whether the _color_ scale is ordinal: the _fill_ or _stroke_ will only be promoted to _z_ if the _color_ scale is ordinal. That might be a little magical, though…

mbostock:
> doesn't the `scale.scale.type = scale.type; // allow marks to check scale type` change indicate that we should pass the scale descriptors rather than the scale functions

Maybe. Previously the channel values were not passed-in as scaled, so the marks had to apply the scales and I wanted more concise syntax. Now that the channels are passed to marks already scaled, it’s probably equally convenient (and more informational) to pass the whole scale descriptor. However, we shouldn’t pass the internal scale descriptor as-is because that is an internal API—we would need to make it match the same public API that we return in _plot_.scale.

> I don't think it closes #724 and #155

It does in the sense that we won’t need to support these features. The idea is that now that area and line no longer do reducing, you don’t need to specify a reducer. I suppose we could still support these features, but it seems less important if the area and line marks no longer arbitrarily apply the first reducer.

> It takes a moment to remember that if you specify a variable fill (especially a continuous value) you need to specify z explicitly or the chart might display nothing (or group on quantized colors); maybe this could be added as a warning?

That would be nice, but I’m not sure what heuristic we would use. This has similar challenges to making the implicit _z_ smarter based on whether the _fill_ or _stroke_ is ordinal. We shouldn’t warn for all cases of implicit _z_; it’s only a problem if the _fill_ or _stroke_ is continuous _and_ there’s an undeclared channel that is also ordinal or categorical. You could have a warning for a line that has non-monotonic *x*-values to detect when it loops back, but in the case of connected scatterplots this behavior is expected.

> the lines are disjoint doesn't work well with a large stroke-width

This is the expected behavior and you can fix it with a round strokeLinecap. It’s worth considering whether round should be the default (and perhaps whether round should also be the default strokeLinejoin).

> Once segmented by fill color, an area cannot use a stroke

True, but using stroke with area is pretty rare; more commonly a separate line is used for stroke. I think this is fine.

> More generally, the loss of some curves is a bit uncomfortable; I wonder if we could sample the (full) curve and create intermediary control points?

I don’t think this is worth the effort/complexity. Some curves already have problems with series that have a small number of data points. (Some curves also have other problems, like failing to preserve the monotonicity of data.) I prefer to keep this behavior as simple as possible both for us in terms of maintenance and for the user in terms of understanding what’s happening.

mbostock:
> As a more general comment on this PR, I love the clean and radical genericity of the approach!

Thank you! And thank you for the detailed review.

> It takes a moment to remember that if you specify a variable fill (especially a continuous value) you need to specify z explicitly or the chart might display nothing (or group on quantized colors); maybe this could be added as a warning?

I thought of a possible warning heuristic: we could look at the cardinality of the implicit _z_ channel, and if it’s ≥50% of the size of the dataset, that would suggest that it’s likely to be wrong and the user needs to set an explicit _z_ channel (possibly to null).

Fil:
> whether round should be the default (and perhaps whether round should also be the default strokeLinejoin).

Yes, I think so. I don't see a drawback.

Fil (approved):
I think the only remaining question for me is the grouping or path segments by z, https://github.com/observablehq/plot/pull/764, but it can be done later.

nspessot:
@mbostock do you have a sample of Plot.line with multi color depending on the "y" axis value, like if temperature is 80 on day 1, color red, t=60 day=2 color blue, etc.

Fil:
@nspessot  I think [this example](https://observablehq.com/@observablehq/plot-line#cell-570) might be what you're looking for?

PR #770, opened 2022-02-19
sort aliases

### LINKED ISSUE #769: Easier sorting of bins by weight?
In a case like this ([data](https://observablehq.com/@af96e6f2b5626d88/us_energy_production)), where I want the first categorical color to be assigned to the most prominent category, and so on by descending prominence, it’d be nice to have an easier shorthand for referencing the height (y2 - y1) of the rects:

Screen Shot 2022-02-19 at 12 55 38 PM

```js
Plot.plot({
  width,
  x: { tickFormat: "" },
  color: { legend: true },
  marks: [
    Plot.rectY(overview, {
      x: "Year",
      y: "Value",
      fill: "Type",
      interval: 1,
      sort: {
        color: "data",
        reduce: (D) => -d3.mean(D, (d) => d.Value)
      }
    }),
    Plot.ruleY([0])
  ]
})
```

The challenge is that I can’t say `sort: {color: "y"}` because in this context _y_ refers to the midpoint of _y1_ and _y2_. And we don’t support expressions such as `sort: {color: "y2 - y1"}`. Perhaps we could support some magic shorthand for “dy” and “dx” though?

mbostock:
Fixes #769. Introduces a concept of channel aliases and derived channels when sorting. For example, if you request to sort on “y”, but there is no _y_ channel, it will fallback to _y2_. The new “dx” and “dy” refer to implicitly derived channels representing |_x2_ - _x1_| and |_y2_ - _y1_| respectively. For these channels, _reverse_ defaults to true (since they typically represent a length encoding rather than a position encoding).

Screen Shot 2022-02-19 at 4 07 52 PM

```js
Plot.plot({
  x: {
    tickFormat: "d",
    label: null
  },
  y: {
    label: "↑ Annual production (quads)"
  },
  color: {
    tickFormat: t => types.get(t),
    legend: true
  },
  marks: [
    Plot.rectY(energy, {x: "Year", interval: 1, y: "Value", fill: "Description", sort: {color: "dy"}}),
    Plot.ruleY([0])
  ]
})
```


Fil (approved):
Yes. Although I think it would be easier to memorize and explain if it were called width and height rather than dx and dy?

Issue #772, opened 2022-02-21
Plot 0.4.0 -> 0.4.1 introduces breaking change preventing creation of scale objects with undefined `type`. 


MaxwellBo:
Reproduced in a public notebook: https://observablehq.com/@mjbo/plot041-scale-creation-breaking-change.

_In Plot 0.4.0_

```js
Plot40 = require("@observablehq/plot@0.4.0")
Plot40.scale({ color: { scheme: "viridis" } })
>>> { type: "linear", domain: Array(2), range: Array(2), interpolate: ƒ(e), clamp: false, apply: ƒ(e), invert: ƒ(e)}
```

_In Plot 0.4.1_

```js
Plot41 = require("@observablehq/plot@0.4.1")

Plot41.scale({ color: { scheme: "viridis" } })
>>> TypeError: Cannot destructure property 'scale' of 'undefined' as it is undefined.

// explicitly defining the scale `type` suppresses the error
Plot41.scale({ color: { scheme: "viridis", type: "linear" } })
>>> { type: "linear", domain: Array(2), range: Array(2), interpolate: ƒ(e), clamp: false, apply: ƒ(e), invert: ƒ(e)}
```

mbostock:
This was an intentional change; see #756. This is described in the release notes:

> If scale options are declared, but the scale has no defined type, domain, or data, a scale is no longer constructed rather than a default linear scale.

Our intent is that standalone scale definitions should match the behavior of scale definitions in plots. And, if you specify _e.g._ `color: {legend: true}` to a plot definition but the _color_ scale isn’t actually used by any marks, we don’t want to render a spurious color legend. #744

I’ve improved the error message in bc685081.

The fix is, as you discovered, to set the scale type explicitly.

As to whether or not this should be considered a breaking change, I do not think it rises to the level of rescinding the Plot 0.4.1 release. The behavior of Plot.scale given an underspecified scale definition was never fully specified. Also Plot is currently in 0.x in regards to semantic versioning.

That said, I do think Plot.scale(…) should match Plot.plot(…).scale(…), e.g.,

```js
Plot.plot({color: {scheme: "viridis"}}).scale("color")
```

This variation returns undefined rather than throwing an error, so it would probably be appropriate for Plot.scale (and Plot.legend) to return undefined if no scale definition is provided for symmetry.

mbostock:
Closing as intended behavior (now with the corrected error message). Thank you for the report.

PR #773, opened 2022-02-22
undefined instead of error

mbostock:
This makes Plot.scale and Plot.legend more consistent with plot.scale and plot.legend: they all now return undefined if no scale definition is provided. I’m not sure this is really an improvement, though. It seems like you’d want to have an error if your standalone scale or legend definition failed. But putting this up for discussion…

Related #772.

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -27,7 +27,6 @@ export function legend(options = {}) {
       );
     }
   }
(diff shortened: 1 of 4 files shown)


mbostock:
I think the error is helpful here, so we should close this PR.

PR #774, opened 2022-02-22
relax cardinality test

mbostock:
I was thinking “fast divide by two”, not divide by four.


Fil (approved):
oops, I should have spotted that!

Issue #13, opened 2020-10-30
Box plot.


mbostock:
Related https://github.com/d3/d3-array/issues/180.

PR #777, opened 2022-02-23
boxX, boxY

mbostock:
This seems like the sort of convenience thing that we should offer as a built-in, even if it requires baking in a few assumptions and limitations. The implementation can always be copy-pasted and modified if the provided options are insufficient, but I suspect this is a nice starting point and it beats having to go look for a notebook or plugin.

Related #13.


Fil:
It would be nice to be have: `Plot.boxX([0, 3, 4.4, 4.5, 4.6, 5, 7]).plot()`

I think we should use the other definition (in which the whiskers stop at real data points); see https://observablehq.com/@akngs/vega-lite-renders-box-plots-differently and https://commons.wikimedia.org/wiki/File:Michelsonmorley-boxplot.svg

I'm sending a PR #778  for these two items; feel free to cherry-pick :)

Re: the documentation, should this be a "mark" or a "combined mark"? If we introduce several of these marks, it might make sense to group them below the "basic" marks? I have a similar question for Plot.frame and Plot.hexgrid.

(I like it without the tick at the end of the whiskers, but I'm pretty sure some people will ask for this aesthetic—maybe we'll add it when the question comes—it could be a simple option "tickSize" or something.)

Fil (approved):

PR #782, opened 2022-02-24
one-dimensional vectors

mbostock:
I went to use these (admittedly for an obscure purpose of manual axis ticks) and forget they didn’t exist. But I think they should exist for symmetry with Plot.dot. And they’re trivial to add, so…


Fil (approved):

Fil:
I've added docs and a test in https://github.com/observablehq/plot/pull/783

PR #792, opened 2022-02-28
dense interval for line and area

### LINKED ISSUE #351: Add a "full series" option to the stack transforms

### LINKED ISSUE #597: Interval (or series?) transform for line and area

mbostock:
Fixes #597. Fixes #351.

With the default *reduce* = *sum*:

Screen Shot 2022-02-28 at 1 41 13 PM

With *reduce* = *median*:

Screen Shot 2022-02-28 at 1 41 28 PM


mbostock:
In terms of the example in https://github.com/observablehq/plot/issues/351#issuecomment-897956355, I don’t think I would use the new *interval* option introduced by this PR, and would instead add *filter* = null to generate dense bins in the first bin transform.

Screen Shot 2022-02-28 at 1 53 14 PM

```js
Plot.plot({
  y: {
    axis: "right",
  },
  facet: {
    data: penguins,
    y: "species",
    marginLeft: 80,
  },
  marks: [
    Plot.areaY(
      penguins,
      Plot.stackY(
        Plot.binX(
          { y: "count", filter: null },
          { x: "body_mass_g", thresholds: 20, fill: "sex", fillOpacity: 0.2 }
        )
      )
    ),
    Plot.lineY(
      penguins,
      Plot.stackY2(
        Plot.binX(
          { y: "count", filter: null },
          { x: "body_mass_g", thresholds: 20, stroke: "sex" }
        )
      )
    ),
    Plot.dot(
      penguins,
      Plot.stackY2(
        Plot.binX(
          { y: "count" },
          { x: "body_mass_g", thresholds: 20, stroke: "sex" }
        )
      )
    ),
    Plot.ruleY([0]),
  ],
})
```

Fil:
Could we change the default reducer to "first"? "sum" only works with a quantitative *y*, whereas "first" (or "mode", but it's slower and not *that* better) would also work with an ordinal *y*.

mbostock:
> Could we change the default reducer to "first"? 

Okay, done. This means you’ll see gaps by default, which is probably a good thing… and I suppose adding *reduce* = *sum* isn’t too hard to remember?

Fil (approved):

PR #794, opened 2022-03-02
generic formatter for text channels

### LINKED ISSUE #793: Plot.text formatting crashes on mixed-type channel
If the text channel is detected to be numeric or temporal, the formatting crashes if it encounters a non-conform value.

For examples:
~~~js
// TypeError: Cannot read properties of undefined (reading 'split')
Plot.textX([1, "a"], { x: [0, 1], text: (d) => d }).plot()

// TypeError: Cannot read properties of undefined (reading 'length')
Plot.textX([1, "a"], { x: [0, 1], text: (d) => d, lineWidth: 10 }).plot()

// TypeError: t.toISOString is not a function
Plot.textX([new Date(), 1], { x: [0, 1], text: (d) => d }).plot()

// TypeError: t.toISOString is not a function
Plot.textX([new Date(), "a"], { x: [0, 1], text: (d) => d }).plot()

// TypeError: t.toISOString is not a function
Plot.dotX([new Date(), "a"], { x: [0, 1], title: (d) => d }).plot()
~~~

Because the filtering of T is done before the formatting, a svg text element (or title) is created for the offending value.

There are two ways to solve this issue: 1. to filter more drastically, by applying formatting in the filtering for T;  or 2. to be more relaxed about formatting, i.e. instead of assuming that the channel has a uniform type, apply conditional formatting for each value.

I prefer the second approach: since T is not scaled we should not be strict about typing.

Fil:
fixes #793

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -1,4 +1,5 @@
 import {format as isoFormat} from "isoformat";
+import {string} from "./options.js";
 import {memoize1} from "./memoize.js";
 
 const numberFormat = memoize1(locale => new Intl.NumberFormat(locale));
@@ -23,3 +24,6 @@ export function formatWeekday(locale = "en-US", weekday = "short") {
 export function formatIsoDate(date) {
   return isoFormat(date, "Invalid Date");
 }
+
+const fNum = formatNumber();
+export const formatGeneric = v => v instanceof Date ? formatIsoDate(v) : typeof v === "number" ? fNum(v) : string(v);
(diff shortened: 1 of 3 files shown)


mbostock (approved):
It’s a bit of a bummer that we now have to check every value, but it’s definitely better than crashing and I’d guess the overhead is likely negligible.

I renamed it to formatDefault and added a TODO since in the future, we’ll need to avoid this sort of pattern when Plot supports a top-level locale option #384. But we already have that problem so it’s not a blocker. 👍

PR #799, opened 2022-03-09
inline facets and axes

mbostock:
A few internal improvements derived from the mbostock/layout branch:

* Don’t wrap all marks in a Facet mark when faceting; handle faceting directly.
* Don’t try to treat AxisX and AxisY as marks; render them directly.
* Override _mark_.filter instead of passing defaults.filter to the constructor.
* Rename a few local variables.


src/plot.js
```diff
@@ -96,13 +125,69 @@ export function plot(options = {}) {
...
+          .append("g")
+            .attr("transform", facetTranslate(fx, fy))
+            .each(function(key) {
+              const j = indexByFacet.get(key);
+              for (const [mark, {channels, values, index, faceted}] of stateByMark) {
+                const renderIndex = mark.filter(faceted ? index[j] : index, channels, values);
+                const node = mark.render(renderIndex, scales, values, subdimensions);
+                if (node != null) this.appendChild(node);
+              }
+            }));
```
Fil, line 184: This pattern makes the code a bit hard to read with the two successive anonymous functions. Suggestion:
```suggestion
    if (fy && axes.y) {
      const axis1 = axes.y, axis2 = nolabel(axis1);
      const j = axis1.labelAnchor === "bottom" ? fyDomain.length - 1 : axis1.labelAnchor === "center" ? fyDomain.length >> 1 : 0;
      select(svg).selectAll()
        .data(fyDomain)
        .enter()
        .append((ky, i) => (i === j ? axis1 : axis2).render(
          fx && where(fxDomain, kx => indexByFacet.has([kx, ky])),
          scales,
          {...dimensions, ...fyMargins, offsetTop: fy(ky)}
        ));
    }
    if (fx && axes.x) {
      const axis1 = axes.x, axis2 = nolabel(axis1);
      const j = axis1.labelAnchor === "right" ? fxDomain.length - 1 : axis1.labelAnchor === "center" ? fxDomain.length >> 1 : 0;
      const {marginLeft, marginRight} = dimensions;
      select(svg).selectAll()
        .data(fxDomain)
        .enter()
        .append((kx, i) => (i === j ? axis1 : axis2).render(
          fy && where(fyDomain, ky => indexByFacet.has([kx, ky])),
          scales,
          {...dimensions, ...fxMargins, labelMarginLeft: marginLeft, labelMarginRight: marginRight, offsetLeft: fx(kx)}
        ));
    }
    select(svg).selectAll()
      .data(facetKeys(scales).filter(indexByFacet.has, indexByFacet))
      .enter()
      .append("g")
        .attr("transform", facetTranslate(fx, fy))
        .each(function(key) {
          const j = indexByFacet.get(key);
          for (const [mark, {channels, values, index, faceted}] of stateByMark) {
            const renderIndex = mark.filter(faceted ? index[j] : index, channels, values);
            const node = mark.render(renderIndex, scales, values, subdimensions);
            if (node != null) this.appendChild(node);
          }
        });
```

src/plot.js
```diff
@@ -96,13 +125,69 @@ export function plot(options = {}) {
...
+  } else {
+    for (const [mark, {channels, values, index}] of stateByMark) {
```
Fil, line 186: For code readability also, it would be nice to have the simpler case (non-faceted) come first, so that the reader understands what this part is doing before diving in (or skipping) the faceted part.

Fil (approved):
This is beautiful!

src/plot.js
mbostock: That’s appealing, but it also has the consequence of further separating these two similar loops, and I think it’s nice if they’re close to each other so it’s more obvious the logic must be keep consistent.

https://github.com/observablehq/plot/blob/6a2c559a22c7164c387d0d9cae6de5824e5b2857/src/plot.js#L179-L183

https://github.com/observablehq/plot/blob/6a2c559a22c7164c387d0d9cae6de5824e5b2857/src/plot.js#L186-L190

That’s because in the faceting case there is a bunch of additional work to render the facet axes.

src/plot.js
mbostock: True, the _selection_.call is unnecessary. We can use a local variable instead if we don’t want to use chaining.

src/plot.js
Fil: yes—it's fine like that

PR #801, opened 2022-03-10
mark initializers

mbostock:
This introduces a concept of a mark “initializer” which is allowed to redefine the mark’s (possibly faceted) index and channels, and by extension, the plot’s scales. The mark’s _initialize_ option is a function that is passed the _facets_ and _channels_ returned previously by _mark_.initialize, and returns an object which may specify new _facets_ and _channels_. This PR includes an example of implementing hexbins as an initializer. This is a refinement of the previous design of “layouts” #648 #711 #713 #740 #775; I wanted to move away from the term “layout” since it implies position whilst initializers are more generic. But I am open to going back to “layout” or other alternatives.

Building on #799, this also makes a few improvements to internal APIs:

* _mark_.initialize now returns a {facets, channels} object rather than {index, channels} (like transforms).
* _channels_ are now represented as an object e.g. {x: {value}} rather than an array ["x", {value}].

While it is possible to call _mark_.initialize externally and hence this could be considered a backwards-incompatible change, this method is undocumented so I think it is acceptable to make this change. Also, it is now more convenient for debugging.

TODO:

- [x] ~~Throw an error if initializers define new positional (*x* and *y*) scaled channels?~~ Not going to worry about it.
- [x] When initializers return new scaled channels, merge with existing scaled channels when reconstructing scales.
- [x] Add hexagon symbol.
- [x] Decide on the definition of radius for hexagons.
- [x] Add hexgrid decoration mark.
- [x] Add circle and hexagon dot shorthand.
- [x] Implement the hexbin transform as an initializer (including support for faceting).
- [x] ~~Implement the _filter_ option for the hexbin transform (to show empty hexbins)?~~ Not now.
- [x] Implement implicit group by _z_ in the hexbin transform? #810
- [x] Implement sorting of hexbins? #810
- [x] Is _radius_ the right name of the option that controls hexbin size? _binWidth_? #810
- [x] Inline and optimize the implementation of d3-hexbin #810
- [x] Implement the dodge transform as an initializer. #809
- [x] Add documentation. #819
- [x] Add tests. #810
- [x] Allow composition/chaining of initializers? #818


Fil:
#795 for the radius of hexagons

Fil:
#804 tries to implement hexbin as an options transform (adding an initializer)

src/transforms/hexbin.js
```diff
@@ -0,0 +1,57 @@
...
+    ...!hasOutput(outputs, "r") && {r: radius},
+    ...hasOutput(outputs, "fill") && {stroke: "none"},
+    ...basic(options, (data, facets) => {
+      for (const o of outputs) o.initialize(data);
+      return {data, facets};
+    }),
+    initialize(facets, {x: X, y: Y}, {x, y}) {
+      if (X === undefined) throw new Error("missing channel: x");
+      if (Y === undefined) throw new Error("missing channel: y");
+      ({value: X} = X);
```
yurivish, line 29: Is this equivalent to saying `X = X.value`?

mbostock: Yes.

yurivish: Perhaps writing it that way would make the code easier to read (unless there's a reason to keep it this way).

mbostock: Maybe? But it’s not like we can avoid destructuring entirely—we use it throughout the codebase. And I don’t think destructuring in general is an antipattern or something that is hard to read.

Fil:
I have a question re: "inline and optimize binning"; is the idea to salvage the code in the previous PR, or to optimize d3-hexbin?https://github.com/observablehq/plot/blob/352d08670b8dfda49f60f5595d079a7b9808399c/src/layouts/hexbin.js#L14

Still on the todo-list
* "filter: null" (not actually a filter, it means we'll have to actively enumerate the centers)
* *z*

Ready:
- #808
- #809

Fil:
Browsed a base of synonyms, here are a few words that could replace _layout_:

design, layout, encoding/**encode**, variation/**variant**, **derivation**, **profile**, blueprint, organization, **plan**, outline, formation, pattern, construction, structure, **disposition**, makeup, skeleton, anatomy, architecture, stencil, markings, mold, cast, stamp, arrangement/**arrange**.

In bold a few of the ones that could "work".

Fil:
For the hexbin radius, another name could be the *step* (ie distance between neighboring hex centers); however in that case it should be sqrt(3) times the current radius. But it might be easier to explain!

Fil:
For composition of initializers, see #818

Fil:
I've added refs to several pull-requests in the original description.

src/plot.js
```diff
@@ -91,9 +90,32 @@ export function plot(options = {}) {
   autoScaleLabels(channelsByScale, scaleDescriptors, axes, dimensions, options);
   autoAxisTicks(scaleDescriptors, axes);
 
+  // Reinitialize; for deriving channels dependent on other channels.
+  const newByScale = new Set();
+  for (const [mark, state] of stateByMark) {
+    if (mark.reinitialize != null) {
+      const {facets, channels} = mark.reinitialize(state.data, state.facets, state.channels, scales);
```
mbostock, line 97: If we pass state.data to mark.reinitialize here, we also need to allow the initializer to return new data…

src/transforms/basic.js
```diff
@@ -32,6 +34,20 @@ function composeTransform(t1, t2) {
...
 
+export function composeInitialize({initialize: i1, ...options} = {}, i2) {
+  return i1 == null
+  ? {...options, initialize: i2}
+  : {
+    ...options,
+    initialize(data, facets, channels, scales, dimensions) {
+      let newChannels;
+      ({data, facets, channels: newChannels} = i1.call(this, data, facets, channels, scales, dimensions));
+      Object.assign(channels, newChannels);
```
mbostock, line 45: This is mutating the input argument _channels_. We should avoid mutating arguments.

yurivish:
I'm very excited about initializers – they open up a really large design space for marks.

With Fil's help I've been experimenting with porting my [density plot library](https://observablehq.com/@twitter/density-scatterplot?collection=@twitter/density) to Plot (current iteration [here](https://observablehq.com/d/f1c510c8517a53e9)), which requires the use of initializers since, like Fil's [heatmap](https://observablehq.com/@observablehq/plot-heatmap-823) mark, the color scale depends upon the bin values.

image

The scatterplots work great (including faceting, which is really useful!), but I've noticed there is a fairly steep performance penalty (~10x slowdown for 1M points) to the Plot mark implementation versus the original. 

I haven't investigated very deeply yet but it seems that a substantial part of the performance gap has to do with the fact that Plot materializes the array of bin colors and that these are represented as strings. I did some basic caching to avoid unneeded `d3.rgb` calls, but more could be done if the mark was allowed to be fully in contol of invoking the color scale.

Outside of Plot, it would be possible to only require three intermediate representations to build a density scatterplot: 1. The original data array, 2. The dense bin counts array, and 3. The dense canvas image data array (a `Uint8ClampedArray` of length 4 times the number of bins). 

I'm not familiar with Plot internals so don't have any specific suggestions, but hope that the density plot might serve as a useful example of what's possible with initializers and where perf concerns might start to creep in.

Note on measurement: My performance analysis was based on scrubbing the slider back and forth in the [density mark notebook](https://observablehq.com/d/f1c510c8517a53e9) and comparing timings to the original [density scatterplot notebook](https://observablehq.com/@twitter/density-scatterplot?collection=@twitter/density). The original design could also be optimized further, but I tried to apply some basic performance-oriented design techniques when working on it.

mbostock:
> I haven't investigated very deeply yet but it seems that a substantial part of the performance gap has to do with the fact that Plot materializes the array of bin colors and that these are represented as strings. I did some basic caching to avoid unneeded d3.rgb calls, but more could be done if the mark was allowed to be fully in contol of invoking the color scale.

Would it be possible to add some caching to d3-scale (or on top of d3-scale) to make this faster? Or maybe we could make quantize color scales extra fast?

yurivish:
Good question, but I don't have an answer right now since I did some _very_ cursory investigation only. In fact, I was curious and dialed down the numberof bins, which reduces the size of the derived data to a 10×10 array of bins.

Here's the profile for plotting 2 million points in 100 buckets, which takes 1.35 seconds:

image

The bottommost orange bars are GCs:

image

I haven't been able to get clicking on source line numbers to work in the Chrome dev tools so I'm not sure what code path this is using (neither importing Plot from the heatmap notebook nor attaching the file and requiring it works), but here are some definitions I think are involved:

```js
function coerceArray(array, coerce, type = Array) {
  if (array !== undefined) return type.from(array, coerce);
}
```

presumably called from the big switch statement in Scale for the linear scale, though I'm not sure what the allocations are:

```
options = coerceType(channels, options, coerceNumber, Float64Array);
```

```js
function coerceNumber(x) {
  return x == null ? NaN : +x;
}
```

yurivish:
[This change](https://github.com/yurivish/plot/commit/0f26a1e6a551e9e8250a46618a57e9d6c392b1d6) significantly improves the runtime of the `range` call.

The other performance bottleneck right now is [coerceType](https://github.com/observablehq/plot/blob/fbee842086439dc3c56d5a80d28cf0d73f9d71ce/src/scales.js#L336-L350) which is called  from `Scale` on (I believe) [this line](https://github.com/observablehq/plot/blob/fbee842086439dc3c56d5a80d28cf0d73f9d71ce/src/scales.js#L172).

Fil:
I've built this branch locally (merged with main) and the speed gains from #861 and #862 are very noticeable:
https://observablehq.com/d/53181fe069c3f6ca (from ~700ms to ~300ms).

src/transforms/hexbin.js
```diff
@@ -0,0 +1,113 @@
...
+import {identity, maybeColorChannel, valueof} from "../options.js";
+import {hasOutput, maybeGroup, maybeOutputs, maybeSubgroup} from "./group.js";
+import {initialize} from "./initialize.js";
+
+// We don’t want the hexagons to align with the edges of the plot frame, as that
+// would cause extreme x-values (the upper bound of the default x-scale domain)
+// to be rounded up into a floating bin to the right of the plot. Therefore,
+// rather than centering the origin hexagon around ⟨0,0⟩ in screen coordinates,
+// we offset slightly to ⟨0.5,0⟩. The hexgrid mark uses the same origin.
+export const ox = 0.5, oy = 0;
```
mbostock, line 11: These are no longer even used by this file. 😦 It looks like it was removed in 886615b40d8d490d1612914843cfb27f0255731f (as part of #810) but it’s hard to trace since this code has been rewritten so many times. In any case, looks like we’ll need to change the implementation of `hbin` below to actually respect these values, or at the very least, and a big warning comment to the code below that it needs to be consistent with the values of `ox` and `oy` above (if it even is?).

mbostock:
I have extensively reviewed (and amended) this PR and plan on merging tomorrow. If anyone wishes to make any final words before I merge please speak now! :pray:

mbostock:
We should probably draw the axes on top? Although the grid lines should still be underneath…

untitled (23)

Fil:
> draw the axes on top?

This should be a generic issue (in my view the axes should be displayed on top of areas and below lines… which is neither top nor bottom). Also you should be able to use `clip: true` in this example?

mbostock:
I’ve reverted the default sort by descending _r_ for the hexbin transform because it feels like it should be handled more appropriately at the dot mark level, or possibly at the generic mark level. And in a sense it is already by the basic sort transform, but that isn’t supported after an initializer, and maybe we want to have default sorting for dots, too.

PR #804, opened 2022-03-11
hexbin as a transform

Fil:
am i on the right track?

(I can't see how to access Z )

### DIFF
--- a/src/scales/ordinal.js
+++ b/src/scales/ordinal.js
@@ -116,7 +116,7 @@ function inferDomain(channels) {
 // If all channels provide a consistent hint, propagate it to the scale.
 function inferSymbolHint(channels) {
   const hint = {};
-  for (const {hint: channelHint} of channels) {
+  for (const {hint: channelHint = {}} of channels) {
     for (const key of ["fill", "stroke"]) {
       const value = channelHint[key];
       if (!(key in hint)) hint[key] = value;
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ export {valueof, channel} from "./options.js";
 export {filter, reverse, sort, shuffle, basic as transform} from "./transforms/basic.js";
 export {bin, binX, binY} from "./transforms/bin.js";
 export {group, groupX, groupY, groupZ} from "./transforms/group.js";
+export {hexbin} from "./transforms/hexbin.js";
 export {normalize, normalizeX, normalizeY} from "./transforms/normalize.js";
 export {map, mapX, mapY} from "./transforms/map.js";
 export {window, windowX, windowY} from "./transforms/window.js";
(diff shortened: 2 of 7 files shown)


mbostock:
I don’t think this approach will work… it’ll lead to the same duplicate computation of channels problem we had in the mbostock/layout branch. I need to think about this more.

mbostock:
To elaborate, I think we’ll need something more like this:

```js
{
  const z = "species";
  const [Z, setZ] = Plot.channel(z);
  return Plot.plot({
    grid: true,
    marks: [
      Plot.dot(penguins, {
        x: "culmen_depth_mm",
        y: "culmen_length_mm",
        z: Z, // optionally make the z channel available to dot (no effect here)
        symbol: "hexagon",
        transform(data, facets) {
          setZ(Plot.valueof(data, z)); // compute the z channel from the data
          return {data, facets};
        },
        initialize([index], {x: {value: X}, y: {value: Y}}, {x, y}) {
          console.log(Z.transform()); // access the previously-computed z values
          const bins = Hexbin().x(i => x(X[i])).y(i => y(Y[i])).radius(20)(index);
          return {
            facets: [d3.range(bins.length)],
            channels: {
              x: {value: bins.map(bin => bin.x)},
              y: {value: bins.map(bin => bin.y)},
              r: {value: bins.map(bin => bin.length), radius: 20, scale: "r"}
            }
          };
        }
      })
    ]
  });
}
```

To break it down:

1. The options transform would (optionally) declare the _z_ channel as a “lazy” channel via Plot.channel.
2. The mark transform would compute the _z_ channel. (This part is composable with other transforms.)
3. The mark initializer could then access the previously-computed _z_ channel values.

Because the channels are computed in a mark transform, they’ll have access to the (possibly transformed) data, so we won’t need to pass that data separately to the initializer. And because the options transform can declare the channel as an option, it won’t be computed twice.

src/scales/ordinal.js
```diff
@@ -116,7 +116,7 @@ function inferDomain(channels) {
 // If all channels provide a consistent hint, propagate it to the scale.
 function inferSymbolHint(channels) {
   const hint = {};
-  for (const {hint: channelHint} of channels) {
+  for (const {hint: channelHint = {}} of channels) {
```
mbostock, line 119: It looks like this code requires that each channel specify a hint. If the hint is optional, then I think we’ll need to change this loop to instead ignore channels that don’t have a hint. Currently it looks like this code will now consider an undefined hint a hint, which might not be what we want.

mbostock: Also, if this is fixing a bug, it’d be helpful to have a comment or an issue to reference. 🙏

Fil: yes, I should have done this before!  issue #806

Fil: ty!

Fil:
superseded by #810 #819

PR #809, opened 2022-03-15
dodge

Fil:
only API change is to pass dimensions as the fifth argument


src/plot.js
```diff
@@ -90,11 +90,16 @@ export function plot(options = {}) {
...
+  const fyMargins = fy && {marginTop: 0, marginBottom: 0, height: fy.bandwidth()};
+  const fxMargins = fx && {marginRight: 0, marginLeft: 0, width: fx.bandwidth()};
+  const subdimensions = {...dimensions, ...fxMargins, ...fyMargins};
+
   // Reinitialize; for deriving channels dependent on other channels.
   const newByScale = new Set();
   for (const [mark, state] of stateByMark) {
     if (mark.reinitialize != null) {
-      const {facets, channels} = mark.reinitialize(state.facets, state.channels, scales);
+      const {facets, channels} = mark.reinitialize(state.facets, state.channels, scales, subdimensions);
```
mbostock, line 102: Here we’re effectively using `fx || fy` to test for faceting, but else where we use `facets !== undefined`. The two aren’t currently guaranteed to be equivalent, so we should either make that guarantee, or we should be consistent with which test we use.

Fil: I think it's equivalent? This is not using options.fx and options.fy, but scales.fx and scales.fy which are set only if there are corresponding channels in channelsByScale, and defined at the same time as facets in L.49.

mbostock:

PR #810, opened 2022-03-16
z hexbins

Fil:
(includes the sort by r branch #808)

inlines an optimized version of hexbin

Capture d’écran 2022-03-16 à 13 51 15


src/marks/hexgrid.js
```diff
@@ -8,23 +8,23 @@ const defaults = {
   ariaLabel: "hexgrid",
   fill: "none",
   stroke: "currentColor",
-  strokeOpacity: 0.1
+  strokeWidth: 0.1
```
mbostock, line 11: I designed this to match the default axis grid.

mbostock:

src/transforms/hexbin.js
```diff
@@ -10,52 +11,99 @@ import {hasOutput, maybeOutputs} from "./group.js";
...
+      if (setGS) setGS(valueof(data, vstroke));
+      if (setGZ) setGZ(valueof(data, z));
+      for (const o of outputs) o.initialize(data);
       if (X === undefined) throw new Error("missing channel: x");
       if (Y === undefined) throw new Error("missing channel: y");
-      X = X.value;
-      Y = Y.value;
-      const binsof = Hexbin().x(i => x(X[i]) - ox).y(i => y(Y[i]) - oy).radius(radius * sqrt4_3);
+      X = X.value.map(x);
+      Y = Y.value.map(y);
```
mbostock, line 46: It isn’t safe to use X.value.map here because _array_.map returns the same type of _array_ which may be a typed array such as Uint8Array. Or to put it another way, X.value is in data space, whereas the result of applying the _x_ scale is in screen space and can be any number (such as Float64Array).

PR #814, opened 2022-03-17
the Plot.stack offset option can be a function

Fil:
use case: diverging semantic bars following a Likert scale
see https://talk.observablehq.com/t/diverging-stacked-bar-chart-in-plot/6028

Capture d’écran 2022-03-17 à 16 35 10


README.md
```diff
@@ -1866,6 +1866,7 @@ After all values have been stacked from zero, an optional **offset** can be appl
 - *expand* (or *normalize*) - rescale each stack to fill [0, 1]
 - *center* (or *silhouette*) - align the centers of all stacks
 - *wiggle* - translate stacks to minimize apparent movement
+- a function that receives as arguments a nested index of stacks, the X1 and X2 arrays (resp. Y1 and Y2 for stackY), and Z, and can modify X1 and X2, typically by subtracting the same offset from each of the X1 and X2 values that pertain to a stack.
```
mbostock, line 1869: I’d like this bulleted list to fit on one line. Can you use a short description here, and move the longer description to the paragraph below that talks about offsets? E.g., “If *offset* is a function, blah blah blah.” We use this style elsewhere in the README if you want to find precedent.

mbostock (commented):
I like this.

I also wonder if we should have a _diverging_ offset, where the _z_ channel is defined as -2, -1, 0, 1, 2, etc. values, and then you can set the order to _z_, too, and not need any custom code to do this.

Fil:
I wonder if allowing the user to mutate values in-place is the correct API here. Maybe when it's a function we want them to return [X1, X2]?

mbostock:
> I wonder if allowing the user to mutate values in-place is the correct API here.

I had the same thought. For precedent, you could take a look at the map transform: when the map is implemented a function, it returns an array of values:

https://github.com/observablehq/plot/blob/66142e2c096439cffea0ed390de13c6cf1545388/src/transforms/map.js#L60

However if the map is implemented a map object with a map method, we use mutation by assigning to the passed `T` array:

https://github.com/observablehq/plot/blob/66142e2c096439cffea0ed390de13c6cf1545388/src/transforms/map.js#L67-L72

Also d3-shape’s _stack_.offset uses mutation:

https://github.com/d3/d3-shape/blob/main/README.md#stack_offset

I think mutation is fine here, mostly because it feels like a pretty low-level API already given the arguments.

mbostock (approved):
I refactored the example to make it easier to reuse this technique. (I expect in the future we could make these even easier with a new option for the stack transform, but this is fine for now.)

I also switched the dataset to CSV and added attribution.

PR #815, opened 2022-03-18
dodge compact

Fil:
the "compact" option applies a force simulation after dodge (thus benefitting from a good initial position); 

Capture d’écran 2022-03-18 à 11 04 03

(Not considering this for integration yet, but just to show how we could expand on dodge in the future)


mbostock:
I don’t think we should offer this option as it makes the position encoding inaccurate.

Fil:
Another drawback is the performance/convergence issue when there are thousands of dots. But I like the aesthetic, and in some cases when the *x* value is the product of a fuzzy measurement, it doesn't matter that the position is not exact. So, maybe I'll provide this as a plugin :)

PR #818, opened 2022-03-22
exploring initializers composition

Fil:
1. Plot.selectFirst should be able to select the first hexbin
3. Plot.hexbin should be able to operate on scaled X and Y channels

Note: I've also tested merging this branch with `fil/reinitialize-dodge` and composing Plot.hexbin (Plot.dodgeX(…)).


mbostock (commented):
Do we need to implement initializer composition within the basic transform helper? My impression is that could be implemented in a separate helper like we did before. The `basic` transform helper could still check that the incoming _initialize_ option is nullish and error if any transform is specified.

mbostock:
I don’t see the problem. The hexbin transform will declare both a transform (data transform) and an initializer (channel transform); the former will error if there’s already an initializer declared. If someone tries to declare another transform (data transform) after the hexbin transform, that will also error because the hexbin transform declared an initializer.

Fil:
A concrete use case would be Plot.hexbin + Plot.dodgeY (which may or may not make sense).

Capture d’écran 2022-03-23 à 17 50 38

Maybe the problem here is that Plot.hexbin uses a data transform to initialize its reducers—it could probably do this in the initializer part of the code instead.

I'm also starting to get lost between all the branches I have opened :)

mbostock:
Okay, done in 42ac4f08093c5819e9fa2dcecf3e09e12785990e. Seem good?

Fil:
yes!

src/transforms/hexbin.js
```diff
@@ -22,10 +23,11 @@ function hexbinn(outputs, {radius = 10, ...options}) {
     ...basic(options, (data, facets) => {
       for (const o of outputs) o.initialize(data);
       return {data, facets};
-    }),
-    initialize(facets, {x: X, y: Y}, {x, y}) {
+    }, (facets, {x: X, y: Y}, scales) => {
       if (X === undefined) throw new Error("missing channel: x");
       if (Y === undefined) throw new Error("missing channel: y");
+      const x = X.scale !== undefined ? scales[X.scale] : identity.transform;
+      const y = Y.scale !== undefined ? scales[Y.scale] : identity.transform;
```
mbostock, line 30: This is fixing the behavior of the hexbin transform when identity scales are used?

mbostock: Actually, no, since we still have scale objects when position scales use the identity scale type…

So this is instead respecting the incoming channel’s scale definition, rather than assuming that the _x_ channel is bound to the _x_ scale and the _y_ channel is bound to the _y_ scale. That’s interesting and I hadn’t considered that possibility; all the built-in marks bind _x_ and _y_ to the corresponding scales, but it is theoretically possible for a custom mark not too.

Fil: yes, that's for when you compose layouts, and the previous layout has returned channels that are already scaled. Typically hexbin + dodgeY leads to this situation.

mbostock: Aha. Thanks for the context.

mbostock:

PR #819, opened 2022-03-22
document layouts (as "scale-aware transforms")

Fil:
covers dodge (#809), hexbin (#810)

PR #823, opened 2022-03-24
reinitialize

Fil:
merges on #801:
- #809 
- #810
- #819
- #818  (in a simpler version that doesn't mix data transforms and channel transforms)

This is mainly in order to test the composition of channel transforms (wanted to have both hexbin and dodge in the same branch). My goal is also to have a base on which to build new channel transforms.


Fil:
Wondering if we should not merge new channels with the old channels when doing composition?

Fil:
the composition is not fully working…

Fil:
Another example of something we can only do with reinitializer:

Capture d’écran 2022-03-30 à 13 58 40

Capture d’écran 2022-03-30 à 13 59 47


We need to be able to read in the x and y scales, the compute all the numbers on the cross-domain x⨂y, then create the color scale (and associated legend).

PR #829, opened 2022-03-25
quantize scale

### LINKED ISSUE #828: Quantize scales
We should support _quantize_ scales like d3-scale does, so that you can more easily convert a continuous color scale into a discrete one.

mbostock:
Fixes #828.

Screen Shot 2022-03-24 at 8 21 14 PM

TODO

- [x] handle descending domains?
- [x] what if you want an exact number of range values?
- [x] what if the range is specified an array of values?


Fil:
I've added a bit of documentation.

> handle descending domains?

see #830

> what if you want an exact number of range values?

Since we're using d3.ticks, I don't think we can guarantee that the values will be "round". In that case, I think we should leave it to the user to analyze the domain, and return the appropriate threshold scale.

It might help if we accepted a _domain_ function as a scale option (independently of its type); the input would be the flattened set of channel values (or maybe just a nested array of values to make it more efficient, replacing flatMap by map):
~~~diff
--- a/src/scales.js
+++ b/src/scales.js
@@ -133,6 +133,9 @@ export function normalizeScale(key, scale, hint) {
 }

 function Scale(key, channels = [], options = {}) {
+  if (typeof options.domain === "function") options.domain = options.domain(channels.flatMap(({value}) => value));
   const type = inferScaleType(key, channels, options);

   // Warn for common misuses of implicit ordinal scales. We disable this test if
~~~

then in simpson-views.js we could specify:
~~~js
    color: {
      type: "threshold",
      domain: (data) => [2, 7, 12, 20],
      legend: true
    },
~~~

> what if the range is specified an array of values?

Currently this works if the range's length is equal to the computed number of thresholds + 1 — otherwise we get blank values to the right; we could add this to remove the thresholds that have no matching color:
~~~js
if (range) thresholds = thresholds.slice(0, range.length - 1);
~~~
but it would create a false impression of what the domain is (the last bin would be much larger than the previous ones), and it's probably better if the chart fails visibly in that case. And if we add the domain function of the previous item, it would still allow users to set the thresholds efficiently.

mbostock:
I’ve now made it so you can pass an explicit range to a *quantile* or *quantize* scale, and it will infer a domain of the appropriate corresponding cardinality.

Fil (approved):

src/scales/quantitative.js
```diff
@@ -123,23 +124,50 @@ export function ScaleLog(key, channels, {base = 10, domain = inferLogDomain(chan
   return ScaleQ(key, scaleLog().base(base), channels, {...options, domain});
 }
 
+export function ScaleSymlog(key, channels, {constant = 1, ...options}) {
+  return ScaleQ(key, scaleSymlog().constant(constant), channels, options);
+}
+
 export function ScaleQuantile(key, channels, {
-  quantiles = 5,
+  quantiles,
```
mbostock, line 132: Makes me wonder if we should support *n* for quantile scales, too (renaming the *quantiles* option)?

PR #830, opened 2022-03-25
handle descending domain for quantize scales

Fil:
Capture d’écran 2022-03-25 à 12 50 34
Capture d’écran 2022-03-25 à 12 50 40
Capture d’écran 2022-03-25 à 12 50 46
Capture d’écran 2022-03-25 à 12 50 43

### DIFF
--- a/src/legends/ramp.js
+++ b/src/legends/ramp.js
@@ -93,8 +93,7 @@ export function legendRamp(color, {
 
   // Threshold
   else if (type === "threshold") {
-    const thresholds = domain.slice();
-    if (color.descending) thresholds.reverse();
+    const thresholds = domain;
 
     const thresholdFormat
         = tickFormat === undefined ? d => d
--- a/src/legends/ramp.js
+++ b/src/legends/ramp.js
@@ -93,7 +93,8 @@ export function legendRamp(color, {
 
   // Threshold
   else if (type === "threshold") {
-    const thresholds = domain;
+    const thresholds = domain.slice();
+    if (color.descending) thresholds.reverse();
 
     const thresholdFormat
         = tickFormat === undefined ? d => d
--- a/src/scales/quantitative.js
+++ b/src/scales/quantitative.js
@@ -1,5 +1,5 @@
 import {
-  ascending,
+  descending,
   extent,
   interpolateHcl,
   interpolateHsl,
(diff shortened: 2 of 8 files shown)


src/legends/ramp.js
```diff
@@ -93,7 +93,8 @@ export function legendRamp(color, {
 
   // Threshold
   else if (type === "threshold") {
-    const thresholds = domain;
+    const thresholds = domain.slice();
+    if (color.descending) thresholds.reverse();
```
mbostock, line 97: It doesn’t appear that color.descending is ever set?

src/scales/quantitative.js
```diff
@@ -151,8 +151,10 @@ export function ScaleQuantize(key, channels, {
   range,
   reverse
 }) {
-  domain = extent(domain); // TODO preserve descending domains?
+  const descending = order(domain) < 0; // preserve descending domain
```
mbostock, line 154: I’m not sure we can assume that domain is an array here. I believe it’s allowed to be an iterable, so we’d need to arrayify it first. Also, would it be simpler to flip `reverse` here if the `domain` is decending, rather than reversing the thresholds later? Then we won’t need to support descending thresholds for reversed scales generally.

```js
  if (order(domain) < 0) reverse = !reverse; // preserve descending domain
```

src/scales/quantitative.js
```diff
@@ -168,9 +170,16 @@ export function ScaleThreshold(key, channels, {
   range = interpolate !== undefined ? quantize(interpolate, domain.length + 1) : registry.get(key) === color ? ordinalRange(scheme, domain.length + 1) : undefined,
   reverse
 }) {
+  const descending = domain[0] > domain[1];
```
mbostock, line 173: We’d need to use `order(domain)` here again. But as I hinted at in the other comment, I don’t think we need to add support for descending domains for the threshold scales; we only need to do it for quantize scales. Threshold scales require you to specify explicit thresholds rather than an extent, so it seems reasonable to require that those thresholds are in ascending order.

Also, it looks like ScaleThreshold doesn’t correctly handle when the _domain_ is specified as an iterable rather than an array (because it references _domain_.length), and it probably should.

src/scales/quantitative.js
```diff
@@ -168,9 +170,16 @@ export function ScaleThreshold(key, channels, {
...
+  if (descending) domain.reverse();
```
mbostock, line 174: This would mutate the caller’s domain! Don’t mutate inputs! 😬

mbostock (changes requested):
Thanks for the help! 🙏

src/scales/quantitative.js
mbostock: To answer my own question here, flipping `reverse` would produce a different result: it flips the range of the scale rather than the domain. So, you’d get this:

Screen Shot 2022-03-25 at 12 50 54 PM

Instead of this:

Screen Shot 2022-03-25 at 12 50 35 PM

The behavior of the scale is the same, but the appearance of the legend is different. If we want to be consistent with descending domains on continuous scales, then we want to flip the domain rather than the range.

PR #831, opened 2022-03-25
trivial intervals for rect

mbostock:
E.g. promotes *x* to *x1* and *x2* if the latter two are not specified.

~~I’m not sure why the aaplMonthly snapshot changed.~~ 🤔 Update: Fil figured it out.


Fil:
The aaplMonthly snapshot changed because we're now calling maybeInset on the Y dimension, whereas before we were returning earlier at https://github.com/observablehq/plot/blob/bddb4d7fbda5225ce28b43d690ce95dbfc21371c/src/transforms/interval.js#L34

In the cli tests the insets we get are [1, 0], and in the browser, on a retina screen, [0.5, 0.5].

mbostock:
> The aaplMonthly snapshot changed because we're now calling maybeInset on the Y dimension

Ah, thanks for figuring that out! I think we can avoid that when the interval is null.

mbostock:
How’s this, @Fil?

Fil (approved):

PR #833, opened 2022-03-29
fix arrow

### LINKED ISSUE #832: Arrow inset bug
untitled (21)

should be
untitled (26)

Fil:
Capture d’écran 2022-03-29 à 09 47 14

fixes #832


src/marks/arrow.js
```diff
@@ -76,8 +76,9 @@ export class Arrow extends Mark {
               // The start ⟨x1,y1⟩ and end ⟨x2,y2⟩ points may be inset, and the
               // ending line angle may be altered for inset swoopy arrows.
               let x1 = X1[i], y1 = Y1[i], x2 = X2[i], y2 = Y2[i];
-              let lineAngle = Math.atan2(y2 - y1, x2 - x1);
               const lineLength = Math.hypot(x2 - x1, y2 - y1);
+              if (lineLength <= insetStart + insetEnd) return null;
```
mbostock, line 80: This change seems unrelated to the fix. Also, I’m not sure we should do it? We don’t do it for rules, for example.

mbostock (changes requested):
Thanks for fixing. I’m sure I tested this before, so I’m surprised there was a bug here. But I made the mistake of not writing tests, so we’ll never know!

src/marks/arrow.js
Fil: It's very ugly when connecting two nodes that are not far enough… Without insets, if lineLength === 0 you get NaNs in the path definition.

test/plots/markov-chain.js
```diff
@@ -0,0 +1,32 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export default async function() {
+  const M = [[0.3, 0.2, 0.5], [0.1, 0.7, 0.2], [0.1, 0.1, 0.8]];
```
mbostock, line 5: Is `M` short for “matrix” here?

mbostock (approved):
👍

test/plots/markov-chain.js
Fil: yup!

Fil:
I've put more extensive tests here https://observablehq.com/@observablehq/test-more-arrows-833

Issue #836, opened 2022-04-05
Fixed data aspect ratio?


Fil:
Sometimes you want x and y to be comparable, maybe because they both represent similar quantities in the same physical dimension. Plot currently has no way [1] to ask for a fixed "data aspect ratio", since x and y will both fill the available space.

Note that the expected aspect ratio is not necessarily 1:1, for example if your data are longitude and latitude, the correct setting might be to set a dataAspectRatio = [1, cos(median latitude)].

For this, MATLAB offers [daspect](https://fr.mathworks.com/help/matlab/creating_plots/manipulating-axes-aspect-ratio.html); ggplot2 has [coord_fixed](https://ggplot2.tidyverse.org/reference/coord_fixed.html); MASS has [geometrically equal scales](https://rdrr.io/cran/MASS/man/eqscplot.html):
* daspect is expressed as an array of ratios for x, y (and z) : [1 1] or [1 1 1].
* coord_fixed is expressed as a single number y/x

[1] of course you can specify ranges explicitly, but then the axes won't adapt to the chart's dimensions.

mbostock:
Duplicate of #445.

PR #837, opened 2022-04-05
aspectRatio option

### LINKED ISSUE #445: x/y aspect ratio
Sometimes I'd want an exact aspect ratio between the two spatial dimensions, for example when they're in the same physical domain (age and years, azimuth and altitude, meters and meters…).

Since the dimensions take into account the margins, the facets etc, this can be a bit difficult to set up.

I'm not sure how we can approach this task, but there must be something better than computing (y(0)-y(1))/(x(0)-x(1)) and tweaking the parameters until it's (almost) equal to 1.

Fil:
Questions:
- [x] name? (~~daspect,~~ dataAspectRatio, ~~height = square~~)?
- [x] application to other scale types? (time, utc at least, <del>maybe band scales too?</del>)
- [x] <del>position? (middle/left/right/top/bottom, top-left etc)?</del>
- [x] <del>syntax (number, array [x, y], number and position (top-left…)?</del>

closes https://github.com/observablehq/plot/issues/836
closes #445


mbostock:
I was thinking something different for this feature, which is that if you specify this option (which I’d probably want to call _aspectRatio_), it would compute the necessary _height_ based on the given (or default) _width_ and margins. Meaning, we’d compute this in `autoHeight` rather than in `autoScaleRange`.

Another thought: if we don’t want to introduce option is that we could have a special named value for the _height_ option, such as _height_ = _square_ for a 1:1 aspect ratio.

Fil:
Agree, this will result in much nicer charts (my approach was creating a lot of whitespace).

Fil:
the last remaining issue is the name of the option.
- _aspectRatio_ doesn't mention that we're speaking about the data/scales/channels, but not the chart's aspect ratio
- _daspect_ sounds awfully technical
- _dataAspectRatio_ is quite a mouthful
- height: "square" doesn't connect (in my brain) with this concept

Another element that we might want to address at a later stage is if we want to size *r* with the same logic as *x* and *y* (for example, if *x* and *y* represent metric positions, and *r* an euclidian distance).

yurivish:
If there is ever a future option able to set the “data width” and “data height” then maybe the mouthful is worth it. To me dataAspectRatio does a good job of capturing the meaning of the option.

Fil:
rebased on main

Fil:
rebased on main
cc: @mootari

tophtucker:
I love this. I was scared to look at the diff but the actual change without tests/docs is 25 lines! 😅 ❤️ Not an expert but looks good and non-invasive. 

It's restricted to "utc", "time", and "linear" because those are the only linear scales, and this is only possible for linear scales, right? Might add that to the readme. I like being able to mix "utc"/"time" and "linear" (spacetime diagrams! natural units with aspect ratio of _c_!). The milliseconds spanned by a temporal x axis usually dwarfs the number of units on the y axis, but it's wonderful if you have relative time in ms on the y axis!

I agree with Yuri that dataAspectRatio suggests the meaning well. I also think aspectRatio would be fine. It's trivial to control the outer frame's pixel-space aspect ratio (width = height), so maybe it's fine to give the shorter name to the nontrivial option? I guess it’s like the aspect ratio of the “grid”, of the plane, of the fabric of reality lol. What does “aspect” mean anyway? It’s the ratio of 1 in x to 1 in y… unitRatio lol. Idk. Whatever! But very valuable feature!

Fil:
Thank you for the enthusiastic support. The feature could be expanded in the future but we're currently constrained to linear scales because of the wording:
> a variation of one unit in the *x* dimension is represented by the same number of pixels
There may be a case for a non-linear dataAspectRatio where x and y share a common domain.

You can totally daspect time (utc) vs duration (milliseconds), see https://observablehq.com/@recifs/dataaspectratio-837

Wording for the README might be: _both x and y must be linear (including time scales, for which the unit is the millisecond)_?

mbostock:
This would be useful for a lot of raster mark applications, too.

mbostock:
(I am not sure the treatment is 100% correct with respect to padding in ordinal scales, but it’s probably “close enough for now” and we can improve it later.)

Fil:
Re: **padding** in ordinal scales, I don't think we'd want the autoheight to change when the user specifies a padding (or outer/inner padding), which means the correct approach is to ignore these and base the computation on their default values, as you do here.

Same story with the **round** option, whose actual value should be ignored, and we should consider only the default. And that's where the computation is not currently perfect, because the default is _true_, and rounding does not necessarily result in the same bandwidth values for x and y. This is similar to https://github.com/observablehq/plot/issues/1239

PR #840, opened 2022-04-05
"sort fy by x" bug

Fil:
This branch is trying to fix the following issue:

When we have "sort fy by x" and a transform, the series are not indexed properly. The channel to sort (XV, in this case *fy*) is indexed by the original data, and the channel by which to sort (YV, in this case *x*) is indexed by the transformed (grouped) data.

It can be solved, in the case of this unit test, by grouping data by series:
> const XV = x === "fy" || x === "fx" ? [...new Set(X[1].value)] : X[1].value;

However this is not a proper fix, since it should apply only when data is indeed grouped this way. We have to understand how the data is reindexed, or alternatively we have to make the facet information follow along with the grouping.

Note: the broken test is the _expected_ chart, but we can't achieve it yet without breaking a few others:

Capture d’écran 2022-04-05 à 23 24 20

cc: @enjalot

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -25,6 +25,7 @@ export function channelSort(channels, facetChannels, data, options) {
     const X = channels.find(([, {scale}]) => scale === x) || facetChannels && facetChannels.find(([, {scale}]) => scale === x);
     if (!X) throw new Error(`missing channel for scale: ${x}`);
     const XV = X[1].value;
+    // const XV = x === "fy" || x === "fx" ? [...new Set(X[1].value)] : X[1].value;
     const [lo = 0, hi = Infinity] = limit && typeof limit[Symbol.iterator] === "function" ? limit : limit < 0 ? [limit] : [0, limit];
     if (y == null) {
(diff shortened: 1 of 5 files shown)


Fil:
Forgot to mention that a solution is to avoid doing the transform, and use e.g.:
```
    Plot.dotX(agediff, {
      x: "age_difference",
      z: "title",
      stroke: null, // don't display any mark
      sort: {fy: "x", reduce: "median"}
    }),
```

(so maybe the bug can be solved by warning that there is a mismatch?)

mbostock:
I’m having difficulty following this issue. 🤯

mbostock:
I think this is only a bug with the facet channels, right? Normal channels should be transformed consistently; it’s just the facet channels that don’t respect the transforms. I made the same mistake at first in #1304 and fixed it in 41aad8056ebb737f0fade5e028e025ce5b4a4b48. There’s probably some similar special logic we need here to handle the facet channels when there is a transform (reconstructing them from index.fx and index.fy).

Issue #136, opened 2021-02-24
Graphs, hierarchies?


mbostock:

river-matt:
> Related, for graphs:
> 
>     * Plot.compose (#171)  to draw both dots and links
> 
>     * An example of a graph that is already “positioned”: https://observablehq.com/d/c54acc4426268a58
> 
>     * Using forceSimulation in a plugin https://observablehq.com/d/29e9f5c9191789f6

I'm unable to access, or that link is broken. I'd love to check out your approach to 'plugins' for Plot.

Aside: this library is as close as I've seen to providing the same grammar and ergonomics as a product like Tableau. Cool stuff!

Fil:
I've updated and published the two notebooks. Let me know your thoughts!

Fil:
A prototype for circle packing: https://observablehq.com/@fil/plot-circle-packing-136

PR #841, opened 2022-04-06
tree

mbostock:
Introduces two new options transforms: **Plot.treeNode**, for laying out tree nodes, and **Plot.treeLink**, for laying out tree links. The new convenience mark **Plot.tree** combines a link, dot, and text with these options transforms to generate a tree diagram.

The Plot.treeNode options transform supports special evaluators for tree nodes. If a channel is defined as a string starting with “node:” then it is interpreted as one of the built-in node evaluators:

* node:path - the normalized, slash-delimited path to this node
* node:name - the last part of this node’s path
* node:depth - this node’s distance from its furthest ascendant
* node:height - this node’s distance from its furthest descendant
* node:internal - true if this is an internal node; false if this is a leaf node

Alternatively you can specify a channel as an {node} object where _node_ is a function. This function is then passed the d3-hierarchy node instance.

The Plot.treeLink options transform similarly supports special evaluators for tree links. If the “node:” prefix or {node} syntax is used, it refers to child node of the link; if the “parent:” prefix syntax is used, it refers to the parent node of the link. If the {link} syntax is used, the function is passed two arguments (_child_, _parent_) and can compute the desired corresponding value.

TODO

- [x] ~~Use root.count or root.sum and expose a node:value somehow?~~ Not now.
- [x] Support different sort orders (as a _treeSort_ option?)
- [x] Support different separations (as a _treeSeparation_ option?)
- [x] Support different orientations of the tree (as a _treeAnchor_ option?)
- [x] Support faceting
- [x] Support imputed nodes
- [x] Documentation
- [x] Fix escaped backslashes when evaluating *node:name*
- [x] Have the **path** option default to identity
- [x] Ignore data with undefined path values

Related #136.


mbostock:
Okay, I figured out how to get faceting and imputed nodes to work: now the tree transforms generate one element per node or per link, respectively, and for imputed nodes the associated data is null (but you can still access node-related metadata using a node accessor). Here’s a little example of the flare dataset with small nodes (<5000 bytes) on the left and large nodes (≥5000 bytes) on the right. Fun!

Screen Shot 2022-04-10 at 11 31 56 AM

Fil (approved):
Excellent! I've added a few suggestions (which can all be addressed later) in #843.

PR #842, opened 2022-04-07
column, not channel

mbostock:
This helper is for constructing columns, not channels, even though typically those columns feed into channels. So I think the original Plot.column name is better, and we can still change it since we haven’t released it yet.


README.md
```diff
@@ -2024,11 +2024,11 @@ Plot provides a few helpers for implementing transforms.
...
+This helper for constructing derived channels returns a [*channel*, *setColumn*] array. The *channel* object implements *channel*.transform, returning whatever value was most recently passed to *setColumn*. If *setColumn* is not called, then *channel*.transform returns undefined. If a *source* is specified, then *channel*.label exposes the given *source*’s label, if any: if *source* is a string as when representing a named field of data, then *channel*.label is *source*; otherwise *channel*.label propagates *source*.label. This allows derived channels to propagate a human-readable axis or legend label.
```
mbostock, line 2029: I couldn’t decide whether to replace other instances of “channel” here with “column”, too. 🤔

Fil (approved):

PR #843, opened 2022-04-11
A few suggestions for the tree layout

Fil:
* accept arrays as inputs (following the work in https://github.com/d3/d3-hierarchy/pull/185#issuecomment-961563223)
* ignore null paths
* default path = identity
* useful defaults for the Plot.tree().plot() shorthand
* and a test


src/transforms/tree.js
```diff
@@ -158,10 +158,14 @@ function nodeData(field) {
...
 function normalizer(delimiter = "/") {
   return `${delimiter}` === "/"
-    ? P => P // paths are already slash-separated
+    ? P => P.map(pathJoin) // paths are already slash-separated
     : P => P.map(replaceAll(delimiter, "/")); // TODO string.replaceAll when supported
 }
 
+function pathJoin(p) {
+  return Array.isArray(p) ? p.map(str => `${str}`.replace(/\//g, '\\/')).join("/") : p;
+}
```
mbostock, line 167: I don’t think we should accept arrays. I think limiting the path to strings is simpler.

src/transforms/tree.js
```diff
@@ -217,7 +221,7 @@ function nodePath(node) {
 }
 
 function nodeName(node) {
-  return node.id.split("/").pop();
+  return node.id.replace(/\\\//g, "\0").split("/").pop().replace(/\0/g, "/");
```
mbostock, line 224: This makes me realize that we should handle escaped backslashes like we do in d3-hierarchy:

https://github.com/d3/d3-hierarchy/blob/7bea49efdc1093b28a8d60d2f8717d8a17803564/src/stratify.js#L125-L133

We shouldn’t use \0 and hope that it doesn’t exist in the path. 😄

src/transforms/tree.js
```diff
@@ -4,7 +4,7 @@ import {channel, isObject, one, valueof} from "../options.js";
 import {basic} from "./basic.js";
 
 export function treeNode({
-  path, // the delimited path
+  path = d => d, // the delimited path
```
mbostock, line 7: We should use `identity` from ../options.js here.

mbostock (commented):
Thanks for the suggestions! I think I’ll manually incorporate a couple of these ideas into the branch. Here are my thoughts.

mbostock:
> accept arrays as inputs (following the work in https://github.com/d3/d3-hierarchy/pull/185#issuecomment-961563223)

Not added, but I added string coercion and you can use a comma as a _delimiter_ to approximate this. (But that kind of defeats the reason to use an array, which is that you don’t have to escape separators, so I’d still recommend escaping separators and passing strings.)

> ignore null paths

Added.

> default path = identity

Added, using Plot.identity. 

> useful defaults for the Plot.tree().plot() shorthand

Not added; I think these need to be channel hints rather than _mark.plot shorthand, but I don‘t feel like it’s worth the trouble to implement these yet.

> and a test

Added.

PR #845, opened 2022-04-15
default diverging based on scheme

mbostock:
I was thinking of cases like this, where a diverging scheme is specified, but the scale type is not specified and it defaults to _linear_.  It would be nice, when appropriate, to default to a _diverging_ scale instead so that it is symmetric around zero.

https://observablehq.com/@violetr/30daychartchallenge-day-13-correlation

I did this by checking for whether the _scheme_ is a diverging scheme, and whether a _pivot_ is specified or whether the _domain_ includes zero. The nice thing is that most of our tests that specify a diverging scale now merely need to specify a diverging scheme, though there were a couple tests that use a diverging scheme without a meaningful pivot (e.g., for temperature or IMDb rating where you want a diverging scheme simply for “low” and “high”).

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ The normal scale types—*linear*, *sqrt*, *pow*, *log*, *symlog*, and *ordinal*
 
 For a *threshold* scale, the *domain* represents *n* (typically numeric) thresholds which will produce a *range* of *n* + 1 output colors; the *i*th color of the *range* applies to values that are smaller than the *i*th element of the domain and larger or equal to the *i* - 1th element of the domain. For a *quantile* scale, the *domain* represents all input values to the scale, and the *n* option specifies how many quantiles to compute from the *domain*; *n* quantiles will produce *n* - 1 thresholds, and an output range of *n* colors. For a *quantize* scale, the domain will be transformed into approximately *n* quantized values, where *n* is an option that defaults to 5.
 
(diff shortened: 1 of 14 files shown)


mbostock:
One thing I don’t like about this is that it will flip between a _linear_ and a _diverging_ scale based on whether the domain includes zero, so there could be an abrupt change based on the data. But if we don’t look at the data, anytime you want to use a diverging scheme with a linear scale, you’d need to specify the type as _linear_ explicitly. Maybe that’s okay, or even good, because it forces you to think about it. Let me know if you prefer that.

Fil:
Sorry for the belated answer. Yes, I'd prefer the second option—where the user has to explicitly indicate a _type_ to use a diverging scheme on a non-diverging scale. Otherwise, the changes with the data extent will be too unexpected.

It will be a breaking change (which is fine), but I wonder if we should add a warning when none of the conditions are met. I think the list of conditions is: there is no (explicit) scale type, the pivot is undefined, the data is numeric (not temporal or ordinal), and the domain doesn't contain 0?

mbostock:
Can you take another look, @Fil? I’ve implemented your preferred option. I chose not to implement the warning because the warning is also dependent on the domain of the data and I think that could be noisy. (I’d like to know that as long as my data has the same shape as my test date, Plot will not generate a warning.)

If you think this is good, I can include this in the 0.5 release for initializers. Otherwise we probably should wait a while for 0.6 since this is nominally a breaking change.

Fil (approved):
added documentation

PR #852, opened 2022-04-19
date scale interval & warning

Fil:
Specifying a scale interval shows the intent of having ordinal numerical or ordinal dates: suppress warning.

Side note: if a numeric interval was specified, string numerics have already been coerced to numbers by the scale transform; so this in fact only has consequences for ordinal dates, such as in the downloads-ordinal test plot.


src/scales.js
```diff
@@ -136,20 +136,21 @@ function Scale(key, channels = [], options = {}) {
...
-  // you intended for the scale to be ordinal; we also disable it for facet
-  // scales since these are always band scales.
+  // you specify a scale interval or if you set the domain or range explicitly,
+  // since setting the domain or range (typically with a cardinality of more than
+  // two) is another indication that you intended for the scale to be ordinal; we
+  // also disable it for facet scales since these are always band scales.
   if (options.type === undefined
       && options.domain === undefined
       && options.range === undefined
+      && options.interval === undefined
```
mbostock, line 146:
```suggestion
      && options.interval == null
```

src/scales.js
```diff
@@ -136,20 +136,21 @@ function Scale(key, channels = [], options = {}) {
...
-    else if (values.some(isNumericString)) warn(`Warning: some data associated with the ${key} scale are strings that appear to be numbers. If these strings represent numbers, you should parse or coerce them to numbers. Numbers are typically associated with a "linear" scale rather than a "${formatScaleType(type)}" scale. If you want to treat this data as ordinal, you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}".`);
+    if (values.some(isTemporal)) warn(`Warning: some data associated with the ${key} scale are dates. Dates are typically associated with a "utc" or "time" scale rather than a "${formatScaleType(type)}" scale. If you are using a bar mark, you probably want a rect mark with the interval option instead; if you are using a group transform, you probably want a bin transform instead. If you want to treat this data as ordinal, you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}" or by specifying the interval option.`);
```
mbostock, line 151:
```suggestion
    if (values.some(isTemporal)) warn(`Warning: some data associated with the ${key} scale are dates. Dates are typically associated with a "utc" or "time" scale rather than a "${formatScaleType(type)}" scale. If you are using a bar mark, you probably want a rect mark with the interval option instead; if you are using a group transform, you probably want a bin transform instead. If you want to treat this data as ordinal, you can specify the interval of the ${key} scale (e.g., d3.utcDay), or you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}".`);
```

I think it’d be good to show an example of what the interval option is set to, and to always have the “suppress this warning” as the last option since it is intended as a last resort to more semantic alternatives.

src/scales.js
```diff
@@ -136,20 +136,21 @@ function Scale(key, channels = [], options = {}) {
...
+    else if (values.some(isTemporalString)) warn(`Warning: some data associated with the ${key} scale are strings that appear to be dates (e.g., YYYY-MM-DD). If these strings represent dates, you should parse them to Date objects. Dates are typically associated with a "utc" or "time" scale rather than a "${formatScaleType(type)}" scale. If you are using a bar mark, you probably want a rect mark with the interval option instead; if you are using a group transform, you probably want a bin transform instead. If you want to treat this data as ordinal, you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}" or by specifying the interval option.`);
```
mbostock, line 152:
```suggestion
    else if (values.some(isTemporalString)) warn(`Warning: some data associated with the ${key} scale are strings that appear to be dates (e.g., YYYY-MM-DD). If these strings represent dates, you should parse them to Date objects. Dates are typically associated with a "utc" or "time" scale rather than a "${formatScaleType(type)}" scale. If you are using a bar mark, you probably want a rect mark with the interval option instead; if you are using a group transform, you probably want a bin transform instead. If you want to treat this data as ordinal, you can specify the interval of the ${key} scale (e.g., d3.utcDay), or you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}".`);
```

Does this work? I’m not sure that the temporal strings will get converted to dates before the interval transform is applied?

src/scales.js
```diff
@@ -136,20 +136,21 @@ function Scale(key, channels = [], options = {}) {
...
+    else if (values.some(isNumericString)) warn(`Warning: some data associated with the ${key} scale are strings that appear to be numbers. If these strings represent numbers, you should parse or coerce them to numbers. Numbers are typically associated with a "linear" scale rather than a "${formatScaleType(type)}" scale. If you want to treat this data as ordinal, you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}" or by specifying the interval option.`);
```
mbostock, line 153:
```suggestion
    else if (values.some(isNumericString)) warn(`Warning: some data associated with the ${key} scale are strings that appear to be numbers. If these strings represent numbers, you should parse or coerce them to numbers. Numbers are typically associated with a "linear" scale rather than a "${formatScaleType(type)}" scale. If you want to treat this data as ordinal, you can specify the interval of the ${key} scale (e.g., 1 for integers), or you can suppress this warning by setting the type of the ${key} scale to "${formatScaleType(type)}".`);
```

Using integers as the example here.

src/scales.js
Fil: You're correct, utcDay.floor doesn't apply isoParse. Should it? I'd say so, but it's going to be complicated.

A solution (added in the ibmTrading plot test) is to apply _both_ an interval and a custom transform:
```js
      interval: d3.utcDay,
      transform: d => d3.utcDay.floor(d3.isoParse(d)),
```

Fil:
this should be good to go

Fil:
* https://github.com/observablehq/plot/pull/852/commits/759c46f5d1a5ca821e86667afe37c0a3ee6f3a3d fixes a bug in the original feature branch (the scale was not reusable since the *transform* was not exposed)
* https://github.com/observablehq/plot/pull/852/commits/fffe7363645e3b0023a60555453eeb675a2e91c5 helps with #355

src/axes.js
```diff
@@ -33,12 +34,22 @@ export function autoAxisTicks({x, y, fx, fy}, {x: xAxis, y: yAxis, fx: fxAxis, f
...
   if (axis.ticks === undefined) {
     const [min, max] = extent(scale.scale.range());
     axis.ticks = (max - min) / k;
   }
 }
 
+function tickInterval(scale, axis) {
+  const interval = maybeInterval(scale.interval);
+  if (interval != null) {
+    const [min, max] = extent(scale.scale.domain());
```
mbostock, line 47: This should only be computed if axis.ticks is undefined.

src/axes.js
```diff
@@ -33,12 +34,22 @@ export function autoAxisTicks({x, y, fx, fy}, {x: xAxis, y: yAxis, fx: fxAxis, f
...
+    if (axis.ticks === undefined) axis.ticks = interval.range(interval.floor(min), interval.offset(interval.floor(max)));
+    if (scale.type !== "point" && scale.type !== "band" && axis.tickFormat === undefined && typeof scale.interval === "number") axis.tickFormat = ",";
```
mbostock, line 49: This should be using the isBandScale helper probably. Or maybe isOrdinalScale?

More to the point, what is this doing? We shouldn’t assume that a numeric interval is an integer; you can say interval: 0.5 (though maybe fractional intervals are not really recommended—it would probably be better for us to use interval: -2 like we do for d3.tickIncrement). In any case, rather than this default, maybe we should simply default tickFormat to [formatDefault](https://github.com/observablehq/plot/blob/93acd8220d4571506ace9d763847cfe6fad08669/src/format.js#L36) so that it works with number and date ticks?

src/scales/ordinal.js
```diff
@@ -28,7 +28,7 @@ export function ScaleO(scale, channels, {
     if (typeof range === "function") range = range(domain);
     scale.range(range);
   }
-  return {type, domain, range, scale, hint};
+  return {type, domain, range, scale, hint, interval};
```
mbostock, line 31: I think we want to normalize the interval representation here:

```suggestion
  return {type, domain, range, scale, hint, interval: maybeInterval(interval)};
```

I know you weren’t doing that before so you could check if `typeof interval === "number"`, but I think we don’t need that anymore with my formatDefault suggestion, and it’s better to normalize input options as early as possible so we don’t have to worry about alternative representations.

src/scales/quantitative.js
```diff
@@ -105,7 +106,7 @@ export function ScaleQ(key, scale, channels, {
   if (nice) scale.nice(nice === true ? undefined : nice), domain = scale.domain();
   if (range !== undefined) scale.range(range);
   if (clamp) scale.clamp(clamp);
-  return {type, domain, range, scale, interpolate};
+  return {type, domain, range, scale, interpolate, interval};
```
mbostock, line 109: And maybeInterval here, too, I guess?

src/scales.js
```diff
@@ -136,20 +136,21 @@ function Scale(key, channels = [], options = {}) {
...
-  // you intended for the scale to be ordinal; we also disable it for facet
-  // scales since these are always band scales.
+  // you specify a scale interval or if you set the domain or range explicitly,
+  // since setting the domain or range (typically with a cardinality of more than
+  // two) is another indication that you intended for the scale to be ordinal; we
+  // also disable it for facet scales since these are always band scales.
   if (options.type === undefined
       && options.domain === undefined
       && options.range === undefined
+      && options.interval == null
```
mbostock, line 146:
```suggestion
      && options.interval === undefined
```

Assuming we do maybeInterval on scale construction (per other comments) we don’t have to worry about null vs. undefined here, too.

src/scales/ordinal.js
Fil: I also wanted to export it with the scale ( 
https://github.com/observablehq/plot/pull/852/files#diff-1dac81a3ab4fe192a8817bbe7a4580c5703cc70f8ccfba55f33be4aeaf556dc7 ). However, just like interpolators, it can be exported in its normalized (and opaque) form, as long as it's reusable (which is also unit-tested).

mbostock (approved):
I made some changes here:

1. I changed the default tickFormat for all ordinal scales to be Plot’s default formatter. This way, if you specify numbers or dates as your ordinal ticks (or domain), you’ll get nice localized formatting by default. I figure this shouldn’t depend on whether the scale has an interval or not, and also because I don’t think Plot’s default formatter should override the default tick format on quantitative scales. (Maybe that could change in the future if we have a default tick format that is specific to the interval you’ve specified, but that’s for future.)
2. I changed the default fontVariant for ordinal scales with an interval to use _tabular-nums_ so that you get better tick labels (just like you would with a quantitative scale).
3. I moved to interval coercion (maybeInterval call) slightly earlier in scale initialization.
4. I augmented one of the tests to ensure that the scale transform defaults to _interval_.floor.
5. I added a comment to the test saying that using a transform to parse dates is not recommended. (You should do that earlier when you load data.)
6. I adopted more shorthand in the tests where possible.

PR #858, opened 2022-04-29
bail out if the inferred ordinal domain has more than 10k values

Fil:
(see https://github.com/observablehq/plot/pull/849#issuecomment-1105874552)

### DIFF
--- a/src/scales/ordinal.js
+++ b/src/scales/ordinal.js
@@ -4,7 +4,7 @@ import {ascendingDefined} from "../defined.js";
 import {isNoneish, map} from "../options.js";
 import {maybeInterval} from "../transforms/interval.js";
 import {maybeSymbol} from "../symbols.js";
-import {registry, color, symbol} from "./index.js";
+import {registry, color, position, symbol} from "./index.js";
 import {maybeBooleanRange, ordinalScheme, quantitativeScheme} from "./schemes.js";
 
 // This denotes an implicitly ordinal color scale: the scale type was not set,
@@ -13,8 +13,9 @@ import {maybeBooleanRange, ordinalScheme, quantitativeScheme} from "./schemes.js
 // of this by setting the type explicitly.
 export const ordinalImplicit = Symbol("ordinal");
 
-export function ScaleO(scale, channels, {
+function ScaleO(scale, channels, {
   type,
+  key,
   interval,
   domain,
   range,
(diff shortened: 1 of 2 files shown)


mbostock (commented):
I think we only want this restriction for positional ordinal scales (point or band), not for all ordinal scales.

Fil:
I've added this check. I realize I don't know why we're exporting ScaleO (which is not used anywhere else and not part of the API).

src/scales/ordinal.js
```diff
@@ -102,14 +102,14 @@ function maybeRound(scale, channels, options) {
...
-function inferDomain(channels) {
+function inferDomain(channels, type) {
   const values = new InternSet();
   for (const {value, domain} of channels) {
     if (domain !== undefined) return domain();
     if (value === undefined) continue;
     for (const v of value) values.add(v);
   }
-  if (values.size > 10e3) throw new Error("This ordinal domain would have more than 10,000 values. If this is intentional, set the domain explicitly.");
+  if (values.size > 10e3 && ["point", "band"].includes(type)) throw new Error("This ordinal domain would have more than 10,000 values. If this is intentional, set the domain explicitly.");
```
mbostock, line 112: You can still have a strictly ordinal scale (type = `"ordinal"`) on a position dimension, though it’s usage is rare since you need to specify the range values explicitly. And if you did, I think we probably would want an error for that here, too, so I prefer this instead:


 `registry.get(key) === position` here.
```suggestion
  if (values.size > 10e3 && registry.get(key) === position) throw new Error("This ordinal domain would have more than 10,000 values. If this is intentional, set the domain explicitly.");
```

src/scales/ordinal.js
```diff
@@ -60,7 +61,7 @@ export function ScaleOrdinal(key, channels, {
     }
   }
   if (unknown === scaleImplicit) throw new Error("implicit unknown is not supported");
-  return ScaleO(scaleOrdinal().unknown(unknown), channels, {...options, type, domain, range, hint});
+  return ScaleO(scaleOrdinal().unknown(unknown), channels, {...options, key, type, domain, range, hint});
```
mbostock, line 64: I would prefer not to mix the `key` with options.

```suggestion
  return ScaleO(key, scaleOrdinal().unknown(unknown), channels, {...options, type, domain, range, hint});
```

mbostock (approved):

PR #861, opened 2022-05-07
faster array manipulations

Fil:
This follows on the issue described by @yurivish in https://github.com/observablehq/plot/pull/801#issuecomment-1119103573

I had also noted that Array.from is slow: https://observablehq.com/@fil/array-from-is-slower-than-expected

My understanding (after @mythmon showed me how to use a browser ;-) ) is that Array.from treats the data as an iterator, using the ->next method until completion, which means 1) it's necessarily slower than a tight loop and 2) it doesn't know how much memory to allocate (I speculate).

So when data is an array (it might not always be the case, iterators still have to be supported? though it seems that in all cases the data we pass to this function has already been arrayified), it might be worth using the faster path to populate it with indices.

Same story with type coercion (#863).

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -127,7 +127,13 @@ export function maybeZ({z, fill, stroke} = {}) {
 
 // Returns a Uint32Array with elements [0, 1, 2, … data.length - 1].
 export function range(data) {
-  return Uint32Array.from(data, indexOf);
+  const n = data.length;
+  if (n !== undefined) {
+    const r = new Uint32Array(n);
+    for (let i = 0; i < n; ++i) r[i] = i;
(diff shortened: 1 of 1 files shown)


mbostock:
If all calls to range(data) take an array, we should just use the array path and not implement the conditional. Is that the case?

Fil:
I’ll double check but I think it is the case.
(I want to do a bit more work on this but needed a commit to start a pr!)

Fil:
Double checked: arrayify is called both before and after every call to a data transform: all cases of range do receive arrays.

mbostock (approved):

PR #863, opened 2022-05-07
optimize channel type coercion

### LINKED ISSUE #862: performance nit: unnecessary copy in coerceArray?
I'm animating a graph using Plot. This isn't what it's intended for, but I'm still getting 60 frames per second, so not bad.

It uses more CPU than I'd like. I did JavaScript profile in Chrome looking for performance issues and noticed that 230 milliseconds or 8.5% of the time is in [coerceArray](https://github.com/observablehq/plot/blob/81f17eac8013e9269ae962ea726e5ea8d1eeba1f/src/scales.js#L360).

It's being called [from Scale](https://github.com/observablehq/plot/blob/81f17eac8013e9269ae962ea726e5ea8d1eeba1f/src/scales.js#L174) to convert to a Float64Array, and this case it already is one, because that's what I passed in. Maybe an unnecessary copy could be avoided?

mbostock:
Fixes #862.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -80,6 +80,16 @@ export function arrayify(data, type) {
     : (data instanceof type ? data : type.from(data)));
 }
 
+// An optimization of type.from(values, f): if the given values are already an
+// instanceof the desired array type, the faster values.map method is used.
+export function map(values, f, type = Array) {
+  return values instanceof type ? values.map(f) : type.from(values, f);
+}
+
+export function isTypedArray(values) {
+  return values instanceof TypedArray;
+}
+
 // Disambiguates an options object (e.g., {y: "x2"}) from a primitive value.
 export function isObject(option) {
   return option?.toString === objectToString;
--- a/src/channel.js
+++ b/src/channel.js
@@ -1,5 +1,5 @@
 import {ascending, descending, rollup, sort} from "d3";
-import {first, labelof, maybeValue, range, valueof} from "./options.js";
+import {first, labelof, map, maybeValue, range, valueof} from "./options.js";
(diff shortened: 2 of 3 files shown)


src/options.js
```diff
@@ -80,6 +80,16 @@ export function arrayify(data, type) {
     : (data instanceof type ? data : type.from(data)));
 }
 
+// An optimization of type.from(values, f): if the given values are already an
+// instanceof the desired array type, the faster values.map method is used.
+export function map(values, f, type = Array) {
+  return values instanceof type ? values.map(f) : type.from(values, f);
```
yurivish, line 86: If `values` is indexable and has a length, could the same for-loop approach as in [this PR](https://github.com/observablehq/plot/pull/861/files#diff-b3c78ec4a30606d2f642a7425509de6e2942d4ff24f8304d3f40f2aadacad887R130) speed up the "else" branch?

Then we could avoid the slower iterator-based `type.from` unless we find ourselves in the general iterator case.

mbostock: > If values is indexable and has a length

There’s not really a way to test whether a value is “indexable”, and checking for a length property is not sufficient. For example, if `values` is a string, it has a length property and is indexable, but you wouldn’t want to use that here. We could do an `instanceof Array` test and use the faster form, but that would only help in the case of numeric coercion, and in that case, it would still be faster for the caller to provide a typed array and avoid the coercion entirely.

yurivish: That’s a really good and clarifying example. Thanks for explaining!

Fil (approved):

PR #870, opened 2022-05-13
line halo

### LINKED ISSUE #860: Halo for lines?
I wonder if we want something like this…

Screen Shot 2022-05-04 at 2 55 55 PM

```diff
diff --git a/src/marks/line.js b/src/marks/line.js
index f12d798..fba2591 100644
--- a/src/marks/line.js
+++ b/src/marks/line.js
@@ -53,7 +53,11 @@ export class Line extends Mark {
               .curve(this.curve)
               .defined(i => i >= 0)
               .x(i => X[i])
-              .y(i => Y[i])))
+              .y(i => Y[i]))
+          .clone(true)
+            .each(function() { this.parentNode.insertBefore(this, this.previousSibling); })
+            .attr("stroke-width", 4)
+            .attr("stroke", "white"))
       .node();
   }
 }
```

Fil:
An unexpected difficulty was to make it compatible with group aesthetics. I had to pass a flag (in the shape of a counter) to secondary groups. For now I just "walk back to the first element", and don't use the counter per se, because we might want to switch to a real flag that would be absent from the first group—which might be more satisfying?

closes #860

a follow-up could be to add the halo option to other "non-fillable" marks: Plot.link, Plot.arrow, Plot.tick, Plot.rule…


Fil:
A problem currently is that the halo option can be an object containing anything. I don't know how to check that the values are numbers or strings as we usually do for normal options.

Fil:
scratch that, a much better solution will be to use a svg filter

Fil:
new version with the SVG filter strategy

Fil:
manually rebased

src/marks/halo.js
```diff
@@ -0,0 +1,33 @@
...
+  const id = `plot-linehalo-${nextHaloId++}`;
+  g.selectChildren().style("filter", `url(#${id})`);
+  g.append("filter").attr("id", id).html(`
+    <feMorphology in="SourceAlpha" result="DILATED" operator="dilate" radius="${radius}"></feMorphology>
+    <feFlood flood-color="${color}" result="BG"></feFlood>
+    <feComposite in="BG" in2="DILATED" operator="in" result="OUTLINE"></feComposite>
+    <feMerge>
+      <feMergeNode in="OUTLINE" />
+      <feMergeNode in="SourceGraphic" />
+    </feMerge>`);
```
mbostock, line 15: You only need the result+in for *dilated*; otherwise in defaults to the previous filter primitive.

```suggestion
    <feMorphology in="SourceAlpha" result="dilated" operator="dilate" radius="${radius}"></feMorphology>
    <feFlood flood-color="${color}"></feFlood>
    <feComposite in2="dilated" operator="in"></feComposite>
    <feMerge>
      <feMergeNode></feMergeNode>
      <feMergeNode in="SourceGraphic"></feMergeNode>
    </feMerge>`);
```

src/marks/line.js
```diff
@@ -88,8 +90,25 @@ export class Line extends Mark {
                   .x((i) => X[i])
                   .y((i) => Y[i])
           )
-      )
-      .node();
+      );
+
+    if (this.halo) {
+      // With variable aesthetics, we need to regroup segments by line
```
mbostock, line 96: This is hard to follow. I think it would be clearer to initialize the DOM the desired way from the beginning, rather than creating it once (above using groupIndex) and then moving things around afterwards to apply the halo.

mbostock:
This seems like a nice feature. The implementation needs some tidying.

Fil:
A similar situation arises with the line opacity when the line is broken into segments for varying aesthetics: https://talk.observablehq.com/t/opacity-artifacts-in-line-segments/8142/2

PR #871, opened 2022-05-15
Expose more Plot internals

kentr:
Exposes more Plot internals, for observablehq/plot#170.

In addition to the comment by @mbostock to expose all of `options.js`,
this exposes other internals that I used when implementing
a custom mark.


Fil:
We probably want to be more selective, since everything we expose must be documented and (as much as possible) have a "definitive" API. Can you share a concrete example of a custom mark?

kentr:
> We probably want to be more selective, since everything we expose must be documented and (as much as possible) have a "definitive" API. Can you share a concrete example of a custom mark?

I think I found a better way to accomplish what I needed with the custom mark. Sounds like this PR is missing the mark for #170, so I'll close it.

PR #873, opened 2022-05-17
document Plot.valueof

Fil:


mbostock:
Corrected the documentation to reflect that *value* can also be an array, null, or undefined. Also added documentation around whether valueof makes a copy, and what it will return. Also fix the documented behavior for the transform method case.

PR #881, opened 2022-05-24
Custom vite resolver

duaneatat:
Resolves https://github.com/observablehq/plot/pull/880#issuecomment-1136350582

This makes it so we can use "js" extensions for TypeScript imports by adding a custom resolver for Vite (which was the only piece that wasn't happy with the js extensions for files that don't exist. This custom loader won't be necessary once all of the files are TypeScript, since Vite implemented the typescript fallback for [_when importing from a typescript file_.](https://github.com/vitejs/vite/pull/5510).

With this change, I'm able to successfully run `yarn test`, `yarn dev`, as well as import Plot as an ESM into another project.


mythmon (approved):
Ah, excellent. I was starting to look into having TS rewrite the imports, but that proved fruitless. This seems cleaner anyways.

duaneatat:
> I was starting to look into having TS rewrite the imports, but that proved fruitless.

Yeah, the TypeScript folks have taken a position against doing this: https://github.com/microsoft/TypeScript/issues/16577 😅

PR #882, opened 2022-05-26
common.ts, 

Fil:
**TODO: cherry pick!** (This includes many changes from ts-take-2 which were merged differently.)

sharing TypedArray and reusing Intl.DateTimeFormatOptions["month"]

### DIFF
--- a/src/curve.ts
+++ b/src/curve.ts
@@ -76,14 +76,9 @@ export function Curve(
   const c = curves.get(`${curve}`.toLowerCase() as CurveName);
   if (!c) throw new Error(`unknown curve: ${curve}`);
 
-  if (tension !== undefined) {
-    if ("beta" in c) {
-      return c.beta(tension);
-    } else if ("tension" in c) {
-      return c.tension(tension);
-    } else if ("alpha" in c) {
-      return c.alpha(tension);
-    }
-  }
-  return c;
+  return tension === undefined ? c
+    : "beta" in c ? c.beta(tension)
+    : "tension" in c ? c.tension(tension)
+    : "alpha" in c ? c.alpha(tension)
+    : c;
 }


Fil:
Duane is telling me that I should also include es5 instead of esnext in tsconfig

Fil:
why not **es6**? and should we have **esnext** in _target_ and _module_?

duaneatat:
> Duane is telling me that I should also include es5 instead of esnext in tsconfig

This is because the values you were saying were valid for formatMonth (`"numeric" | "2-digit" | "long" | "short" | "narrow" | undefined`) are the ones from es5's `Intl.DateTimeFormatOptions["month"]`. If we want to use es6 or esnext then we should use different values.

Fil:
Ah, I just want us to be current, so probably esnext (or es6?) is the right choice here, and we'll use whatever type definitions exist for the "month" option of  Intl.DateTimeFormatOptions.

mythmon:
The valid options for the keys are es3, es5, es6, es2015 through es2022, and esnext. (es6 and es2015 are aliases).

> The special ESNext value refers to the highest version your version of TypeScript supports. This setting should be used with caution, since it doesn’t mean the same thing between different TypeScript versions and can make upgrades less predictable.

I'd say we should probably use one of the year versions, perhaps es2022 if we just want maximally up to date code.

duaneatat:
I think the confusion is that I'm not seeing the "month" key on `Intl.DateTimeFormatOptions` in versions later than es5

duaneatat:
image
But I'm probably just missing something?

duaneatat (approved):

duaneatat:
> image
> 
> But I'm probably just missing something?

Indeed I was just misunderstanding how namespaces work -- they are merged with previous namespace declarations, so each version builds on the last and isn't supposed to be a complete representation.

Fil:
I'll need to cherry-pick from this, too many changes are not useful anymore.

PR #884, opened 2022-05-27
Fix Plot.hexbin default reducer, and simplify

Fil:
Oops. I realized everything was wrong when I tested _z: null_ (to have 1 center per position independently of the fill and stroke channels). This should be better 😓


mbostock:
I’m still reviewing the implementation of the hexbin transform in the other branch, so I’m just going to blindly merge this now so I don’t have to review it twice. My intuition is there are things I want to change here.

src/transforms/hexbin.js
```diff
@@ -18,45 +17,45 @@ export function hexbin(outputs = {fill: "count"}, options = {}) {
...
       Y = Y.value.map(y);
-      const F = setGF && GF.transform();
-      const S = setGS && GS.transform();
-      const Z = setGZ ? GZ.transform() : (F || S);
+      const Z = valueof(data, z);
+      const F = valueof(data, vfill);
+      const S = valueof(data, vstroke);
+      const G = maybeSubgroup(outputs, Z, F, S);
+      if (Z && !outputs.find(r => r.name === "z")) {
+        outputs.push(...maybeOutputs({z: "first"}, {z: Z}));
```
mbostock, line 42: This is mutating `outputs` which is declared up above (in the `hexbinn` scope). This is bad because it means if the initializer is invoked more than once, it will behave differently. Generally speaking we don’t invoke an initializer more than once (since Plot.plot is often invoked once and then everything is thrown away), but still, we should try to write the transforms to be stateless. Invoking them shouldn’t mutate the internal state of the transform.

mbostock: Also, I’m very confused as to why the logic here is so different than what the bin and group transforms do… Does it have something to do with the fact that initializers run after channels have been constructed?

mbostock: I came up with an approach I like in 2dc40c09aeba3238572b153b8c78c375c2b1fb74.

PR #885, opened 2022-05-27
realign the hexgrid and the hex centers

Fil:
### before
before

### after
after


mbostock:
I merged this but I think this is actually the wrong fix and the bug is in the src/transforms/hexbin.js? https://github.com/observablehq/plot/pull/801#discussion_r883958278 Anyway, I’m investigating, but please hold off on any further contributions to the reinitialize branch unless absolutely necessary as I am trying to review it and merge it.

mbostock:
Confirmed that this is the wrong fix and dropped from the parent branch. I’ll work on fixing the `hbin` implementation that was broken when it was inlined now.

PR #888, opened 2022-05-30
avoid Array.from if possible

mbostock:
Continuation of #863: if we can use *array*.map or *array*.slice instead of Array.from, do so because it’s faster. Supposedly. I didn’t actually check so this could be premature optimization. But this was also relatively easy, so… 🤷.

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -31,7 +31,7 @@ export function valueObject(channels, scales) {
   for (const channelName in channels) {
     const {scale: scaleName, value} = channels[channelName];
     const scale = scales[scaleName];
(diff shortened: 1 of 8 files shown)


Fil:
I tested [a little bit](https://observablehq.com/@fil/array-from-vs-slice). In Chrome this seems indifferent, but in Safari and FF it's 2x faster—and even 10x faster for typed arrays in Safari… (Also, it makes the code a bit nicer to read.)

One question though: slice is never called with the *type* parameter?

mbostock:
> One question though: slice is never called with the _type_ parameter?

Sure, but I wanted symmetry with map.

PR #891, opened 2022-05-30
fix hexbin + dot sort

mbostock:
Alternative to #889 by making the index sort optional if the given channel is not defined.

Though, there is something to the initializer declaring that the channel will exist, just that its values are not yet known when the mark is constructed. Still it feels somewhat misleading to declare it as an empty array.


mbostock:
Taking https://github.com/observablehq/plot/pull/889#issuecomment-1141347903 as tacit approval. 🙂

PR #892, opened 2022-05-31
facet wrap

mbostock:
Rebased #332 against main. Also avoided changing the exposed value of the *fx* and *fy* scales, though it still feels awkward to me. Like, when you turn on facet wrapping, you still want to specify *fy* options potentially, even though there isn’t really an *fy* scale…


Fil:
Thank you! Yes it's hard to decide if this should be represented externally by 1 scale (or projection: facet name to 2-d position), 2 scales (facet name to x, facet name to y), or three scales (facet name to rank, rank to x, rank to y). The main question is reusability (using the scales from one chart in another chart). I guess if we can only salvage the domain it should be fine?

mbostock:
> 2 scales (facet name to x, facet name to y)

Actually, that’s a thought. Maybe two scales *fx* and *fy* makes the most sense here. But I’d have two requirements:

1. Both *fx* and *fy* should have the same domain (like *fx* does now in the working example).
2. Neither *fx* nor *fy* is invertible (because they map multiple values to the same position).

This sort of scale would not be expressible as a band scale but would be expressible as an ordinal scale.

I still don’t know how this would be “round-trippable” as *fx* and *fy* scales though. This definitely isn’t ready to merge—there are some conceptual things to work out first about the meaning of the facet scales and where you can set options.

Fil:
Note to self: another option would be to substitute the logic in `facetTranslate(kx)`. There would be only one _kx_, but _fx_ (or _fxy_) would affect both coordinates of the translate.

PR #893, opened 2022-05-31
disable default sort when explicitly ordering

mbostock:
The Plot.reverse, Plot.sort, and Plot.shuffle transforms weren’t setting the **sort** option, and hence the dot mark’s default sort by descending *r* was still having an effect. Now these transforms set the **sort** option to null, explicitly disabling any default sort implemented by marks.

### DIFF
--- a/src/transforms/basic.js
+++ b/src/transforms/basic.js
@@ -46,7 +46,7 @@ function filterTransform(value) {
 }
 
 export function reverse(options) {
-  return basic(options, reverseTransform);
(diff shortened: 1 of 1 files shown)


Fil:
Made me realize we had 0 tests for Plot.sort, Plot.reverse and Plot.shuffle.

before/after:

Capture d’écran 2022-05-31 à 09 37 22

mbostock:
Thanks for the tests!

PR #894, opened 2022-05-31
maintain faceting after optional sort

Fil:
(follows #891)


mbostock:
The bug is in composeInitializer, not here. I’ll put up an alternative PR.

PR #897, opened 2022-05-31
construct r channel and scale as needed

mbostock:
If the dodge transform is being used with a mark that doesn’t define an _r_ channel, then none will be constructed. So, the dodge transform needs to do it itself?

Alternatively, maybe all marks should expose an _r_ channel even if it’s unused? This PR feels like a bit of a hack and duplicates logic.


mbostock:
Superseded by #899.

PR #901, opened 2022-06-01
faster dodge

Fil:
algorithm unchanged, but using a fixed typed array makes it about 2x faster on a mildly dense chart
https://observablehq.com/@fil/dodge-performance-901

On my computer the problematic `Plot.dot(olympians, Plot901.dodgeX({y: "sport"})).plot()` is now solved (in Chrome) in 12s instead of a whopping 205s with the base branch. Safari needs 17s and Firefox 15s. Still not usable for this particular case, but it expands the range of problems for which it's going to be tolerable.

### DIFF
--- a/src/transforms/dodge.js
+++ b/src/transforms/dodge.js
@@ -73,11 +73,14 @@ function dodge(y, x, anchor, padding, options) {
         });
 
         // Find the best y-value where this circle can fit.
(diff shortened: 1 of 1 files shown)


mbostock (approved):
Nice work! 👏

PR #902, opened 2022-06-01
fix dodge identity

Fil:


src/transforms/dodge.js
```diff
@@ -45,7 +45,7 @@ function dodge(y, x, anchor, padding, options) {
     if (!X) throw new Error(`missing channel: ${x}`);
     X = coerceNumbers(valueof(X.value, X.scale !== undefined ? scales[X.scale] : identity));
```
mbostock, line 46: We should do the same with X.scale here. I’ll add to your branch.

mbostock (approved):

src/transforms/dodge.js
Fil: yes there was no issue with x in my tests (but it's nicer this way)

PR #904, opened 2022-06-01
fix edge case with dodge

mbostock:


src/transforms/dodge.js
```diff
@@ -66,7 +66,7 @@ function dodge(y, x, anchor, padding, options) {
       const intervals = new Float64Array(2 * I.length + 2);
       for (const i of I) {
         const ri = radius(i);
-        const y0 = ky ? ri + padding : 0; // offset baseline for varying radius
+        const y0 = ky && R ? ri + padding : 0; // offset baseline for varying radius
```
mbostock, line 69: I’m not sure we want to behave differently for constant _r_ vs. varying _r_ where all values are the same. I think I’d rather just leave it simple and consistent even if that means some floating point error? Although, maybe we can fix the floating point error somehow.

Fil: Yes. The alternative is to have a small tolerance for a circle that doesn't fit exactly. Since we're in pixel space it's guaranteed that 1e-6 is "small enough" (and if this doesn't work for someone it's possible to add 0.01 or something to the padding). This approach gives a more compact result in all the test plots.

mbostock: Nice! That looks great.

PR #906, opened 2022-06-01
update changelog

Fil:


mbostock:
Thanks for the suggestion, Fil, but I want to use some different images and make further changes to the wording.

PR #908, opened 2022-06-02
sort, reverse and shuffle can be applied as initializers

### LINKED ISSUE #907: Apply Plot.reverse (and sort) after Plot.dodge
This wWould be nice to change the z-order and hide the shirts rather than the faces of the following chart:

Capture d’écran 2022-06-02 à 11 43 41

Fil:
if sort is applied on a string such as "y" and a (possibly derived) y channel exists, it is preferred to sorting on data[i]["y"]; which allows to sort("y", dodgeY()).

closes #907

demo https://observablehq.com/@observablehq/plot-sort-dodge-907


mbostock:
A challenge here is that if you say `{sort: "r"}` and expect it to sort by channel it only works if you’ve already specified an initializer (_i.e._, if the **initializer** option is not null). Otherwise it gets interpreted as a basic transform and isn’t passed the channels argument. So, it means you can use Plot.sort _et al._ after an initializer, but if you try to use Plot.sort before an initializer and reference a channel, it won’t work, and there’s no way to tell Plot.sort that you want to use it as an initializer rather than a transform.

For example if we want to use this new feature within the dodge transform, rather than using a basic transform and repeating the definition of the _r_ channel, we’d have to say:

```diff
diff --git a/src/transforms/dodge.js b/src/transforms/dodge.js
index 7e7fe53..dc84575 100644
--- a/src/transforms/dodge.js
+++ b/src/transforms/dodge.js
@@ -3,6 +3,7 @@ import IntervalTree from "interval-tree-1d";
 import {finite, positive} from "../defined.js";
 import {identity, number, valueof} from "../options.js";
 import {coerceNumbers} from "../scales.js";
+import {sortTransform, reverseTransform} from "./basic.js";
 import {initializer} from "./initializer.js";
 
 const anchorXLeft = ({marginLeft}) => [1, marginLeft];
@@ -50,7 +51,11 @@ function dodge(y, x, anchor, padding, options) {
   if (r != null && typeof r !== "number") {
     const {channels, sort, reverse} = options;
     options = {...options, channels: [...channels ?? [], {name: "r", value: r, scale: "r"}]};
-    if (sort === undefined && reverse === undefined) options.sort = r, options.reverse = true;
+    if (sort === undefined && reverse === undefined) {
+      options = initializer(options, sortTransform("r"));
+      options = initializer(options, reverseTransform);
+      options.sort = null;
+    }
   }
   return initializer(options, function(data, facets, {[x]: X, r: R}, scales, dimensions) {
     if (!X) throw new Error(`missing channel: ${x}`);
```

Maybe if we set the **initializer** option to true, it could work? As a signal that you want the basic transforms to operate as initializers rather than transforms.

mbostock:
Actually it turned out to be simpler than that… we just need to support the same **filter**, **sort**, and **reverse** options when composing initializers, and then the dodge transform can set the **sort** option to _r_ and it will just work!

mbostock (approved):
Pretty happy with how this turned out. You can now pass an object like {channel: "r", order: "descending"} to the Plot.sort transform or **sort** option and it automatically uses an initializer instead of a data transform.

Issue #380, opened 2021-05-12
New stroke-dasharray channel?


ericemc3:
Similar to the `vl.strokedash()` channel in vega-lite, this channel could be used to display, for example, estimated values as dotted lines, i.e. depending on an indicator column in the dataset:
 
![image](https://user-images.githubusercontent.com/17407796/117982269-21b44400-b336-11eb-94d8-b9b470b653b7.png) ![image](https://user-images.githubusercontent.com/17407796/117982304-2b3dac00-b336-11eb-9de5-d6b36882e526.png)
 
Note: a `stroke-width` channel would also prove useful...

PS: Plot is a great library, very intellectually stimulating, and the 21 beautifully written articles in the https://observablehq.com/collection/@observablehq/plot collection should be read by all data scientists!

mbostock:
There is already a strokeDasharray option. See this example:

https://github.com/observablehq/plot/blob/main/test/plots/covid-ihme-projected-deaths.js

It could be made a channel I suppose, but in practice I think it is usually better to have two marks.

Fil:
Here's a common use-case where the user would want the constant to be a channel (cc @espinielli):

`textAnchor: d => d.properties.city === "Moscow" ? "end" : "start",`

yurivish:
I had the same case with textAnchor [here](https://observablehq.com/@yurivish/more-words) and used two marks.

harrislapiroff:
I ran into this issue earlier this week in [this notebook](https://observablehq.com/d/58bde3d66a6f3878) with two lines that were the same type of data, but one is a "current" system and one is a "proposed" system. I did accomplish this using two marks for each set of data,

```js
    Plot.line(scenarios, {
      filter: d => d.system === 'proposed',
      stroke: PROFIT_COLOR,
      y: 'bidaProfit',
      x: 'attendees',
      fx: 'musicians',
      strokeWidth: 3,
    }),
    Plot.line(scenarios, {
      filter: d => d.system === 'current',
      stroke: PROFIT_COLOR,
      y: 'bidaProfit',
      x: 'attendees',
      fx: 'musicians',
      strokeWidth: 3,
      strokeDasharray: '2 10',
    }),
```

but I think it would have been more elegant if I had been able to use one:

```js
    Plot.line(scenarios, {
      stroke: PROFIT_COLOR,
      y: 'bidaProfit',
      x: 'attendees',
      fx: 'musicians',
      strokeWidth: 3,
[...]

PR #909, opened 2022-06-02
variable styles

Fil:
This PR wants to make all of the common styles (or most of them) "variable" (#175, #380)

- [x] **strokeDasharray** (difficult case: we can't consider that an array is a columnar representation of a channel) 🌶
- [x] **strokeDashoffset** (number)
- [ ] **mixBlendMode**
- [ ] etc.

For **dx** and **dy** the situation is a bit different, see https://github.com/observablehq/plot/blob/de66b18292a74384e534524fcd89b26863f1a2e4/src/marks/text.js#L69


Fil:
I'm not sure how to make dx and dy variable when X or Y are not channels; I think I'll punt on this for now, because I don't want to overcomplicate the PR—my first prototype was making changes in each and every mark, depending on X, Y, and possibly cx, cy when X and Y were not defined… booo. The current approach which shifts X seems much better, but doesn't work when X is not a channel.

Fil:
rebased

mbostock:
Fil, thanks for the work here. I would like to put this on hold to focus on animation and interaction. I am concerned that this generalization may make animation more difficult to implement. Let’s discuss this week.

PR #912, opened 2022-06-02
fix font-variant for axis labels

mbostock:
Like #911, but for axes.


Fil (approved):
the diff is visible (for the better) on test/output/covidIhmeProjectedDeaths.svg which mentions "COVID-19"

PR #924, opened 2022-06-08
link insets

### LINKED ISSUE #920: inset for Plot.link
> it would be nice if the link mark supported inset the same way arrows do

suggested by @yurivish

Fil:
For parity with Plot.arrow. I don't think it's possible to do a good job with insets and curves other than linear (step-* and bump-x/y), so I've restricted insets to the linear curve.

cc: @yurivish 

closes #920


mbostock (changes requested):
I don’t think we should limit this to the linear curve. We should just inset the control points regardless of curve. Or we should not do this and tell people to use the arrow mark instead.

Fil:
Here's a build of the current code with the error removed:
https://observablehq.com/@fil/link-inset-924

Various curves on the link with inset points

Is it acceptable? Is there any way we could make it better? (I don't think evaluating the tangent with getPointAtLength is an option—at least, it's not available with jsdom).

We can decide to punt on this until someone figures how to solve this.

mbostock:
> Is it acceptable? Is there any way we could make it better?

I think it’s probably fine, and better that throwing an error. If we wanted to be fancier, you could do a switch statement on the curves and special case a few of them. e.g., bump-x and step should inset horizontally, while bump-y should inset vertically.

If you prefer, we could also ignore the inset option for unknown/unsupported (custom) curves. But I don’t think we should error.

PR #927, opened 2022-06-09
treat missing scale bandwidth as zero

### LINKED ISSUE #918: Dodge error message for non-dodgeable marks could be improved

image

mbostock:
Fixes #918, probably.


Fil:
It fixes the crash, but no ticks are shown. Should we have a default of 30 (instead of 0)?

mbostock:
Closing in favor of #929.

PR #929, opened 2022-06-09
with the dodge transform, the y scale can be undefined even if Y exists

### LINKED ISSUE #918: Dodge error message for non-dodgeable marks could be improved

Fil:
closes #918, alternative to #927


src/scales.js
```diff
@@ -327,6 +327,7 @@ export function scaleOrder({range, domain = range}) {
 // example, a rect will span the full extent of the chart along a collapsed
 // dimension (whereas a dot will simply be drawn in the center).
 export function isCollapsed(scale) {
+  if (!scale || !scale.domain) return true;
```
mbostock, line 330: If scale is truthy, would scale.domain ever be falsey? Can we do this instead:

```suggestion
  if (scale === undefined) return true;
```

src/marks/tick.js
```diff
@@ -61,13 +62,13 @@ export class TickX extends AbstractTick {
...
-  _y1(scales, {y: Y}, {marginTop}) {
+  _y1({y}, {y: Y}, {marginTop}) {
     const {insetTop} = this;
-    return Y ? i => Y[i] + insetTop : marginTop + insetTop;
+    return Y && !isCollapsed(y) ? i => Y[i] + insetTop : marginTop + insetTop;
   }
   _y2({y}, {y: Y}, {height, marginBottom}) {
     const {insetBottom} = this;
-    return Y ? i => Y[i] + y.bandwidth() - insetBottom : height - marginBottom - insetBottom;
+    return Y  && !isCollapsed(y) ? i => Y[i] + y.bandwidth() - insetBottom : height - marginBottom - insetBottom;
```
mbostock, line 71: Okay, next question: do we need to use isCollapsed here? Or should we just check if the scale is defined?

```suggestion
    return Y && y ? i => Y[i] + y.bandwidth() - insetBottom : height - marginBottom - insetBottom;
```

Fil: It can be defined by another mark
```
  Plot.tickX([1, 1.1, 2, 3], Plot.dodgeY()),
  Plot.dotY([1, 2, 3, 4])
```

... in which case it's not enough to check collapsed, we should also check if it has a bandwidth method (next commit).

Fil:
In 47c62ff it is a bit unfortunate that I used isBandScale, which is a function name that also exists in scales.js (but doesn't look at an instanced d3 scale). I should change this for a better name—just have to think of one.

mbostock:
I’ve lost track of what we’re trying to solve here. I understand that the scale can be undefined, but why are we checking if the scale is collapsed?

Fil:
Yes, sorry… I did that by symmetry with the rule mark, since I wanted to have the same behavior. The crash initially happened when the scale didn't have a bandwidth—returning 0 in that case avoided a crash, but the tick then didn't appear at all (it had a zero length).

Now I think the right call is to test `Y && y?.bandwidth` both for _y1 and _y2. In other words, if the y scale isn't defined, or is defined as a continuous scale (by another mark), our tick will span the whole height.

mbostock:
> Now I think the right call is to test Y && y?.bandwidth both for _y1 and _y2.

Yes, that sounds right to me. So no checking for a collapsed domain?

mbostock (approved):
Hopefully this is what you meant by https://github.com/observablehq/plot/pull/929#issuecomment-1152547660. Let me know if otherwise. 🙂

mbostock (changes requested):
I thought about this some more, and I don’t like how initializers can break previously safe assumptions. Like, the mark declared the _y_ channel as being associated with the _y_ scale, but now the initializer can break that assumption and the mark has to code against it? And likewise the mark can declare that the _y_ scale is a _band_ scale, but the initializer can break that assumption and have some other arbitrary scale type, too? That feels like a slippery slope where suddenly “all bets are off” during render and marks have to protect against unknown conditions.

I’m still going to revert #930 but I think we don’t want to land this change, and instead we should think about just having a better error message like was originally suggested in #918, rather than trying to make this work.

mbostock:
How about this: after the initializers are run, we **create any scales** that the marks assume exist based on the channels that the mark declared in its constructor (even though those channels may be optional and may not have been declared with an initial value, and even though the initializer created channels not associated with a scale). Probably these scales should be identity scales, unless one of the marks declared a specific scale type, in which case we should respect that. In the case of ordinal/band/point scales, we can use an empty domain.

I suspect that we could also eliminate [the inferChannelScale function](https://github.com/observablehq/plot/blob/0a38a5873f31e23e5b5afca82cdff77ed03d00d3/src/plot.js#L337-L354), and instead use the information we already have from the channels declared by the mark constructor.

Fil:
This will be superseded by #936

Issue #928, opened 2022-06-09
Follow-ups on Delaunay/Voronoi


Fil:
- [ ] In #917, the Voronoi mark uses the full frame. As a consequence, it doesn't work well with `clip: true` (since it is exactly on the clipping path). the way to fix it might be to add a _padding_ (or _inset_) option.
- [ ] Add a _curve_ option on the voronoi cells
- [ ] Add a _curve_ option on the hull
- [ ] Add a _padding_ option on the hull (see https://observablehq.com/@fil/hull-padding)
- [ ] Add an _alpha_ option on the hull (for alpha shapes)
- [ ] Create a voronoiLabel mark that does occlusion https://observablehq.com/@d3/voronoi-labels
- [ ] ~~Sub-graphs: #941~~
- [x] The Delaunay/Voronoi/Hull marks require x and y, and a such don't accept the result of the dodge transform. Not sure how to approach this generic issue, but a simple work-around is to specify a fake and empty _y_ channel: `Plot.voronoiMesh(penguins, Plot.dodgeY("middle", {x: "body_mass_g", y: []}))` (solved by #931)
- [x] Delaunay and voronoi _can_ work in 1-d (#931)
- [ ] would be nice to have an option to clip the resulting polygons with an arbitrary geometry (#1635 or #1109)

mbostock:
This umbrella issue is over a year old. We should close it and re-open more specific issues as needed.

PR #930, opened 2022-06-09
dodge declares output channel

mbostock:
Fixes one part of #928.

I was hoping there’d be a way to do this for the hexbin transform too, but I couldn’t wrap my head around it. The problem is that the initial options (passed to the Mark constructor) need to declare the channels _prior_ to the initializer running, not the channels _after_ the initializer runs; if we blindly declare the output channels in _options_, then we’ll overwrite the input channels to the hexbin transform. The dodge transform doesn’t have this problem because it doesn’t have a circular channel dependency (it consumes x and outputs y, or consumes y and outputs x; whereas the hexbin transform consumes x and y and outputs x and y, at a minimum).

Anyway, this seems like an improvement. 🤷


Fil (approved):
Yes, we need this in general.

#931 also addresses this issue in (narrower) case of Delaunay and Voronoi diagrams, but it's not an alternative (we need both).

mbostock:
I am reverting this. #931 feels like a better fix, and this was causing issues in #929.

PR #931, opened 2022-06-09
delaunay X and Y are optional

Fil:
A complement to https://github.com/observablehq/plot/pull/930

the X and Y channels should be optional in all the voronoi marks—since d3-delaunay can handle collinear points (and it's not just for the sake of it: the resulting voronoi can be useful, as illustrated in the test case).


mbostock (commented):
I think we want X and Y to default to the center of the frame, not 0. It doesn’t matter for the Voronoi, but it would matter for the degenerate case of a one-dimensional Delaunay (which you could argue, who cares anyway, but I still think we should do it). I’ll fix…

mbostock (approved):

Fil:
thank you!

PR #934, opened 2022-06-11
fix zero for negative domains

### LINKED ISSUE #933: scale.zero option not useful for plotting negative numbers
If I turn on the "zero" option for data that consists entirely of negative numbers, Plot doesn't choose a useful domain. In this case, the data contains values from -2 to -0.5, so a good domain including zero would be [-2, 0]. It seems to be picking something like [0, -0.5] instead, which is backwards and means that most points aren't plotted.

```
Plot.plot({
  width: width,
  height: 200,
  y: {zero: true},
  marks: [
    Plot.lineY([-1, -2, -0.5])],
});
```

Demo [here](https://observablehq.com/@skybrian/plotting-negative-numbers).

mbostock:
Fixes #933, probably.

### DIFF
--- a/src/scales/quantitative.js
+++ b/src/scales/quantitative.js
@@ -98,8 +98,8 @@ export function ScaleQ(key, scale, channels, {
     const [min, max] = extent(domain);
     if ((min > 0) || (max < 0)) {
       domain = slice(domain);
-      if (order(domain) < 0) domain[domain.length - 1] = 0;
-      else domain[0] = 0;
+      if (order(domain) !== Math.sign(min)) domain[domain.length - 1] = 0; // [2, 1] or [-2, -1]
+      else domain[0] = 0; // [1, 2] or [-1, -2]
     }
   }
 


Fil (approved):
Yes it does fix the original bug report. I've added a test case.

test/plots/travelers-covid-drop.js
```diff
@@ -0,0 +1,19 @@
...
+      grid: true,
+      zero: true,
+      label: "↓ Drop in passenger throughput (2020 vs. 2019)",
+      tickFormat: "%"
+    },
+    marks: [
+      Plot.line(travelers, {x: "date", y: d => d.current / d.previous - 1, strokeWidth: 0.25, curve: "step"}),
+      Plot.line(travelers, Plot.windowY({x: "date", y: d => d.current / d.previous - 1, k: 7, stroke: "steelblue"}))
+    ],
+    width: 960
```
mbostock, line 17: Can you move this to the first line of the options? 😬

PR #936, opened 2022-06-13
infer the scale used from the mark's scales definitions

### LINKED ISSUE #918: Dodge error message for non-dodgeable marks could be improved

Fil:
when a new channel is defined by an initializer, infer the scale used from the mark's scales definitions

(see https://github.com/observablehq/plot/pull/929#issuecomment-1152799152)

As it is, however, it doesn't fix #918.

supersedes #929


mbostock:
Closing in favor of #939.

PR #937, opened 2022-06-13
fix date in test

Fil:


mbostock (approved):
I would prefer to use the Date constructor as long as we know the input is valid ISO 8601.

PR #938, opened 2022-06-13
facet requires the data to be specified

Fil:
Note: is it explicit enough, given that a common error is people who see the pattern `{facet: {data, x: "category"} }` and replace `data` with their own variable name, resulting in `{facet: {data2, x: "category"} }`?

another wording might be `the values to facet must be specified as the data property`, or something to that effect.

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -51,6 +51,7 @@ export function plot(options = {}) {
     const {x, y} = facet;
     if (x != null || y != null) {
       const facetData = arrayify(facet.data);
+      if (facetData == null) throw new Error("facet requires the data to be specified");
       facetChannels = {};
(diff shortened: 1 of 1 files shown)


src/plot.js
```diff
@@ -51,6 +51,7 @@ export function plot(options = {}) {
     const {x, y} = facet;
     if (x != null || y != null) {
       const facetData = arrayify(facet.data);
+      if (facetData == null) throw new Error("facet specification requires a data property");
```
mbostock, line 54:
```suggestion
      if (facetData == null) throw new Error("missing facet data");
```

eagereyes (approved):

PR #940, opened 2022-06-14
visited links color

### LINKED ISSUE #935: a href issues
- [ ] a link added to a Plot, without a target=_blank, will crash the observable iframe in Chrome (it works in Safari and Firefox)
- [ ] the constant fill color of a text link anchor is overridden by the default link:visited color.

reported by @yurivish @ https://observablehq.com/d/02cab844233728fb

Fil:
fixes #935

cc @yurivish

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -288,6 +288,7 @@ function applyHref(selection, href, target) {
     const h = href(i);
     if (h != null) {
       const a = document.createElementNS(namespaces.svg, "a");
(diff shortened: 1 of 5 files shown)


mbostock:
We can use an attribute for this rather than a style, and I’d prefer that (because it has lower precedence than a stylesheet, and because it’s easier to see in the DOM, and it minimizes the turnover in the tests).

mbostock (approved):

PR #941, opened 2022-06-14
Gabriel, Urquhart, and minimal spanning tree

Fil:
three subgraphs of Delaunay, for #928

Currently implemented as named meshes. I don't think the API is the most elegant, and I'd rather change it to an option of the delaunayLinks and delaunayMesh marks (?). Anyway, it's fun and it seems useful, maybe?.

The associated code is also at https://observablehq.com/@fil/delaunay-sub-graphs


Fil:
Capture d’écran 2022-06-14 à 17 33 03
Capture d’écran 2022-06-14 à 17 32 54
Capture d’écran 2022-06-14 à 17 32 47

mbostock:
This feels niche to me; it’s cool but I’m not sure how I’d use it? I would be more supportive of this as a general pattern rather than these specific graphs. I.e., what would be a good way to do this without needing it to be built-in to Plot?

Fil:
I don't know, I felt delaunay was niche already. But if we want to measure nicheness, I'd argue that the mst might be more useful (to highlight a group of points) than the delaunay mesh.

The convex hull is obviously the most common (and simplest to understand) subgraph.

The Gabriel and Urquhart subgraphs, if *very nice to look at*, are probably not generically useful. But in a cartographic metaphor, where  the data's densest/clumpy areas are villages and towns, the urquhart graph looks like the roads that would connect them.

In terms of API, I'll have to think a bit more. Currently by construction it doesn't allow types of graphs that aren't subsets of the delaunay.

mbostock:
The only reason I added Delaunay was to support Voronoi and convex hull, which together have several known use cases. Hiding Delaunay when it’s needed to support those cases felt overly strict. But adding more subsets of the Delaunay triangulation seems more niche. Though I’d be more supportive, as I said, if it felt like amore generic extensible mechanism rather than adding a handful of specific subgraphs.

Fil:
cleaning this for now; might revisit some day in the future

PR #943, opened 2022-06-16
density mark

mbostock:
Screen Shot 2022-06-15 at 11 35 39 PM

```js
Plot.plot({
  inset: 20,
  marks: [
    Plot.density(faithful, {x: "waiting", y: "eruptions", stroke: "steelblue", strokeWidth: 0.25}),
    Plot.density(faithful, {x: "waiting", y: "eruptions", thresholds: 4, stroke: "steelblue"}),
    Plot.dot(faithful, {x: "waiting", y: "eruptions", fill: "currentColor", r: 1.5})
  ]
})
```

TODO

- [x] group by _z_ (or _fill_ or _stroke_)

Open questions:

* Should this be called [contour](https://ggplot2.tidyverse.org/reference/geom_contour.html) instead of [density](https://ggplot2.tidyverse.org/reference/geom_density.html) (to match ggplot2)?
* What if you want non-density contours, like for a [topographic map](https://observablehq.com/@d3/volcano-contours) or a [continuous function](https://observablehq.com/@d3/contours)?
* What if you want one-dimensional density (KDE)? Is that a transform?
* Should there be a legend or continuous color encoding for density contours?


Fil:
A few more features in #948.



*Should this be called [contour](https://ggplot2.tidyverse.org/reference/geom_contour.html) instead of [density](https://ggplot2.tidyverse.org/reference/geom_density.html) (to match ggplot2)?*

I agree that _density_ or _density contours_ is a better name than "contours" because we might want to create other types of contours.

*one-dimensional density (KDE)? Is that a transform?*

yes, very similar to 1-d binning

*a legend or continuous color encoding for density contours?*

I don't think it's possible to make a readable "density level" legend (points per pixel square?)

Fil:
superseded by #948

PR #948, opened 2022-06-22
density contours stroke & fill

Fil:
As a complement to https://github.com/observablehq/plot/pull/943:
* group by z, stroke, fill
* apply other styles as well (constant like mixBlendMode, channel-based like title…)
* use fill: "density" or stroke: "density" to color by "density level" (note that the numbers ~~are not interpretable by the reader, so the legend is not recommended—or maybe we need to normalize~~ are normalized by the maximum density across all facets and series)

*todo*:
- [x] add a *weight* channel
  - [x] (and an example) 
- [x] document
- [x] make sense of the density value (as a percentage of maximum density?)
- [x] test initializer composition
- [x] 1d density contours
- [x] wait for a new release of d3-contour? https://github.com/d3/d3-contour/pull/57


src/marks/density.js
```diff
@@ -12,43 +12,117 @@ const defaults = {
...
 }
 
-export function density(data, {x, y, ...options} = {}) {
+export function density(data, {x, y, stroke, fill, bandwidth = 20, thresholds = 20, ...options} = {}) {
   ([x, y] = maybeTuple(x, y));
-  return new Density(data, {...options, x, y});
+  let f, s;
+  if (fill === "density") { fill = undefined; f = true; }
+  if (stroke === "density") { stroke = undefined; s = true; }
+  return new Density(data, {...options, x, y, fill, stroke, initializer: initializer(+bandwidth, +thresholds, f, s)});
```
mbostock, line 57: We should move this initializer to inside the Density class constructor, not the convenience constructor here. (You’d never use the Density class without it.)

Fil:
I've tested the initializer composition with the bin and the hexbin transforms… I don't think we want to add them to the unit tests because they don't have value as dataviz, but it's fun

Capture d’écran 2022-06-23 à 17 17 39
Capture d’écran 2022-06-23 à 17 17 50

src/marks/density.js
```diff
@@ -52,6 +52,7 @@ export class Density extends Mark {
 
 export function density(data, {x, y, ...options} = {}) {
   ([x, y] = maybeTuple(x, y));
+  if (x === undefined || y === undefined) throw new Error("x and y must be defined");
```
mbostock, line 55: We could treat these as cx/cy like we did for Voronoi/Delaunay.

Fil: done!

Fil:
rebased on main

Fil:
I want to have 1-d KDE, but it's not in the scope of this PR, which in my view is complete.

src/marks/density.js
```diff
@@ -0,0 +1,124 @@
...
+    let max = 0, maxn = 0;
+    const density = contourDensity()
+      .x(X ? i => X[i] : constant(cx))
+      .y(Y ? i => Y[i] : constant(cy))
+      .weight(W ? i => W[i] : 1)
+      .size([width, height])
+      .bandwidth(bandwidth)
+      .thresholds(thresholds);
+
+    // First pass: seek the maximum density across all facets and series; memoize for performance.
```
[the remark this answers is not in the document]
Fil: Yes, cache the contours and reuse those that ended up with the same threshold selection.

mbostock:
Okay, I was able to rebase my PR to main, then this PR to my PR, and we’re back to me being able to review all the changes here. 😅 Thanks for being patient with me.

src/marks/density.js
```diff
@@ -11,44 +12,113 @@ const defaults = {
...
 }
+
+function densityInitializer(options, bandwidth, thresholds, f, s) {
+  return initializer(options, function(data, facets, channels, scales, dimensions) {
+    const X = channels.x?.scale ? valueof(channels.x.value, scales[channels.x.scale]) : channels.x?.value;
+    const Y = channels.y?.scale ? valueof(channels.y.value, scales[channels.y.scale]) : channels.y?.value;
+    const W = channels.weight?.value;
+    const Z = channels.z?.value;
+    const {z} = this;
+    const [cx, cy] = applyFrameAnchor(options, dimensions);
```
mbostock, line 65: `applyFrameAnchor` takes `this` as the first argument, not `options`. (I’ll fix this.)

src/marks/density.js
```diff
@@ -11,44 +12,113 @@ const defaults = {
...
+          );
+        }
+      }
+    }
+
+    channels = {contours: {value: contours}, ...Object.fromEntries(newChannels)};
+    // normalize colors to a thresholds scale
+    const m = max * (maxn + 1) / maxn;
+    if (f) channels.fill.value = channels.fill.value.map(v => v / m);
+    if (s) channels.stroke.value = channels.stroke.value.map(v => v / m);
```
mbostock, line 120: I’m thinking maybe we don’t do this, and just leave the units as “points per square pixel” or “value per square pixel” or perhaps “points per 1,000 square pixels” as a more human-readable value. Even though the value is arbitrary and resolution-dependent, it’s still probably more useful than 0 = 0, 1 = maximum observed density because it allows you to have a consistent scale across charts.

src/marks/density.js
```diff
@@ -11,44 +12,113 @@ const defaults = {
...
+          }
+          memo.push({facetIndex, index, c, top: d.value});
+        }
+      }
+    }
+
+    // Second pass: generate contours with the thresholds derived above
+    density.thresholds(thresholds);
+    for (const {facetIndex, index, c: memoc, top} of memo) {
+      const c = top < max ? density(index) : memoc;
```
mbostock, line 103: I’m not sure how much this caching is helping; I think most commonly, only one series (or facet-series) will be cacheable as having the max threshold. All the others will need to be recomputed. I guess that’s still helpful in the common case of only a single series and single facet, but I suspect we could make this significantly faster if d3.contourDensity had a way to compute and store the blurred grid, and then compute individual contours for a particular threshold afterward. That way, we wouldn’t have to recompute the underlying grid when computing new contours, and we could cache individual contours rather than sets.

mbostock (approved):
I made some edits in abfd86e0730774d4e8db7a9d3c1c10adaff4d46d. To summarize:

* For options that are only needed by the initializer, I moved the defaults and coercion into the initializer.
* I fixed a bug where if the fill was specified as “density”, the default stroke is now “none” (not “currentColor”).
* I removed this.path in favor of creating a path in render.
* I added a comment explaining why mark.filter is overridden. I wonder if we also want to change how the channels are grouped: rather than picking the value associated with the first index for each facet-series, we should pick the first _defined_ value? But this is probably fine as-is.
* I adopted applyChannelStyles instead of applyGroupedChannelStyles to avoid constructing wrapper index arrays for each contour.
* I added _z_ to the set of channels that are not automatically grouped (since _z_ is not needed to render the computed contours), and adopted a Set (since the explicit enumeration of channel names in the code was feeling a bit unwieldy with four named channels).
* I adopted the coerceNumbers and identity technique used by the hexbin transform for the _x_, _y_, and _weight_ channels.
* I fixed the first argument to applyFrameAnchor.
* I tweaked how the newChannels are constructed: it’s an object rather than an array, and I handle the _fill_ and _stroke_ = “density” cases separately.
* I removed the use of constant(cx) and constant(cy), since d3-contour does that for us.
* I removed then normalization of the color scale values (in the color = density case) and instead scaled by a fix factor of 100. Mostly because this is simpler and faster and I’m not really sure what we should do here.
* I removed the caching logic in favor of skipping the threshold-finding pass when there’s only one facet-series. If we want to make this faster, we should instead consider making changes to the d3-contour API to avoid recomputing the blurred grid.
* I forced the color scale to include zero in the color = density case, rather than the default which is that the color range will start at the smallest contour value.

Fil:
Superb. Thank you for the detailed list of changes!

PR #950, opened 2022-06-23
clip supports band scales

### LINKED ISSUE #949: Band scales shift the clip option
As reported in https://talk.observablehq.com/t/plot-clip-flag-is-chopping-first-row-text/6730, having bars and the {clip: true} option results in improper clipping.

The reason is that when x (or y) is a band scale, the whole <g> element is shifted by x.bandwidth()/2.

See https://observablehq.com/@observablehq/plot-band-clip-949 and a fix in #950

Fil:
closes #949


mbostock (approved):
I tweaked some of the argument passing to consolidate logic.

PR #953, opened 2022-06-25
Change .append() to .appendChild()

jkenzer:
**Reason for change**
While using Plot with external frameworks that have their own Dom implementations, it's possible the framework's Dom implementation doesn't support the newer Dom.append() method. This pull requests addresses the instance where plot uses Dom.append() to add a legend or a caption by using the Dom.appendChild() method.

**Testing with Salesforce's Lightning Web Component (LWC) framework**
I found this issue while trying to use Plot inside of Salesforce.com to create customized visualizations that Salesforce isn't capable of out of the box. I was able to successfully use Plot version 0.5.0 with no issues other than when I tried to add either a legend or a caption.

To test this PR, I built a dist after running tests and uploaded dist/plot.umd.js to Salesforce and added both a legend and caption to my visualizations. Here is a sample screenshot

image

This contribution addresses the topic raised in [this community post](https://talk.observablehq.com/t/using-plot-externally-unable-to-add-legend/6710).

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -222,11 +222,18 @@ export function plot(options = {}) {
   if (caption != null || legends.length > 0) {
     figure = document.createElement("figure");
     figure.style.maxWidth = "initial";
-    figure.append(...legends, svg);
+    if (legends.length > 0) {
+      figure.appendChild(...legends);
+    }
+    figure.appendChild(svg);
     if (caption != null) {
       const figcaption = document.createElement("figcaption");
-      figcaption.append(caption);
-      figure.append(figcaption);
+      if (!isObject(caption)) {
+        figcaption.innerHTML = caption;
+      } else {
+        figcaption.appendChild(caption);
+      }
+      figure.appendChild(figcaption);
     }
   }
 
--- a/src/plot.js
+++ b/src/plot.js
@@ -222,17 +222,11 @@ export function plot(options = {}) {
   if (caption != null || legends.length > 0) {
     figure = document.createElement("figure");
     figure.style.maxWidth = "initial";
-    if (legends.length > 0) {
-      figure.appendChild(...legends);
-    }
+    if (legends.length > 0) figure.appendChild(...legends);
     figure.appendChild(svg);
     if (caption != null) {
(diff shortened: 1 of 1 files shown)


src/plot.js
```diff
@@ -222,11 +222,12 @@ export function plot(options = {}) {
   if (caption != null || legends.length > 0) {
     figure = document.createElement("figure");
     figure.style.maxWidth = "initial";
-    figure.append(...legends, svg);
+    if (legends.length > 0) figure.appendChild(...legends);
```
mbostock, line 225: appendChild unlike append only takes a single argument, so this will need to be a for loop now.

src/plot.js
```diff
@@ -222,11 +222,12 @@ export function plot(options = {}) {
...
+    figure.appendChild(svg);
     if (caption != null) {
       const figcaption = document.createElement("figcaption");
-      figcaption.append(caption);
-      figure.append(figcaption);
+      typeof caption === "object" ? figcaption.appendChild(caption) : (figcaption.innerHTML = caption);
```
mbostock, line 229: We’ll need a more precise test for an element or text node here. I think we can crib from hypertext literal for the exact logic. And then in the non-element case, we’ll need to create a text node. (We shouldn’t set innerHTML since that would allow the caption to contain HTML and potentially cause encoding issues.)

mbostock (changes requested):
Thanks for this contribution! This is close but we’ll need some tweaks to maintain the functionality of append. I can help with this if you’d like—let me know.

src/plot.js
jkenzer: Got it. Easy change. I think I was assuming that there would only ever be one legend in the array. I'll make this a for of loop.

src/plot.js
jkenzer: Gotcha about the innerHTML. I'm looking at the code now from the hypertext literal. I might need assistance here but let me see what I come up with.

Thanks for the feedback.

mbostock:
Merging into a branch so I that I can finish the remaining steps.

PR #954, opened 2022-06-26
adopt density.contours

mbostock:
Ref. https://github.com/d3/d3-contour/pull/58


Fil (approved):

Fil (changes requested):
I'm having second thoughts on the use of d3.ticks here. To use nice ticks on a quantity that doesn't mean much in absolute terms (value per pixel square) reduces the possibilities for no clear reason.

I think we would be better off respecting the exact number of thresholds the user asked for. Something like `d3.range(thresholds).map(i => max * (i + 1) / (thresholds + 1))` would feel appropriate.

Fil (approved):

mbostock:
I’m not sure I agree with this last change, though: I think the values _are_ still meaningful (points per square pixel) and that the exact number of contours is rarely if ever required. For example, if we wanted to display the threshold value on the contour lines like you do in [Contour Labels](https://observablehq.com/@fil/contour-labels-canvas), we’d want to have nice round values for readability.

(I also think we should implement the ability to specify **thresholds** as an array of numbers.)

Fil:
OK. When using *ticks* it's always possible to get the number of ticks we want if we adjust the **weight** option (by iterative guesswork), so in that sense it is not impossible, just a bit difficult (or rather, you need to set up a slider and find a correct value). Maybe that could be an additional option(?).

In any case, agree that we should have **thresholds** as an array of numbers.

In my early iterations I wondered if the color scale should be a threshold scale, but I think that depends on the number of thresholds; if it is high, you'll prefer a continuous scale.
Also, you can manually set the color scale to type:"quantize", but then you have a spurious value for 0
https://observablehq.com/@observablehq/density-contours-colors

mbostock:
I’ll leave it with your suggested behavior for now; we can revisit if needed.

PR #959, opened 2022-06-29
Prettier

duaneatat:
Enables prettier and reformats the whole codebase in one go 🚀 

This also adds a `.git-blame-ignore-revs` file, and if you run `git config blame.ignoreRevsFile .git-blame-ignore-revs` locally git blame will ignore the formatting changes in this PR. GitHub will also use this file for git blame: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


duaneatat:
12,253 lines become 67,636 😅

Fil:
I've read a bunch of them closely (not all 😓), and it doesn't seem more difficult to read.

Two exceptions:
- some of the comments, which sometimes were made about two variables and now are aligned to one variable only; or which were made at the end of an if() and now are inside the braces. E.g.:
```diff
-    if (isScaleOptions(scale)) { // e.g., ignore {color: "red"}
+   if (isScaleOptions(scale)) {
+     // e.g., ignore {color: "red"}
```  
- the nested _selection_.append  in mark creation are now "flattened".

(I don't think there's anything we can do to fix the latter; and for the comments, we'll fix them if it's problematic.)

Note that plot.umd.min.js goes up from 131023 to 131135 bytes (+111 bytes); not an issue, but I'm curious to know what changed.

Question: how can we ensure that prettier always runs? should it be a precommit hook? a yarn script? a test? or just an editor setting?

duaneatat:
> Question: how can we ensure that prettier always runs? should it be a precommit hook? a yarn script? a test? or just an editor setting?

We should configure our editors to run prettier on save -- and I just added a github action that will check prettier formatting for all PRs and will block merging if there are any formatting violations

Fil (approved):
lgtm! Please don't merge until Mike has a chance to review it.

FWIW I installed the "formatting toggle" extension in VSCode to allow format on save in this project.

mbostock:
Hmm, this was a lot more change than I was expecting, and I have to admit I find Prettier to be much “uglier” than it was before. This is not an objection to this PR, but I do find Prettier depressing… it’s like taking something beautiful, unique, and with character and turning into generic flavorless mush. But even so I acknowledge that it makes it easier for others to contribute to the code base since it removes the need to discuss the most trivial stylistic elements of code which would otherwise be a continuous source of ongoing friction.

I can’t even view the diff on my iPad (the page hangs) but I’ll review this soon on desktop.

duaneatat:
@mbostock should we table this for now?

mythmon:
I had a hard time believing that Prettier inflated the line count by over 50,000 lines, so I went digging. It turns out that the vast majority of those extra lines come from one file, `test/data/movies.json`:

```
$ git diff (git merge-base HEAD origin/main) HEAD --stat -- test/**/*.json | tail -n 1
 test/data/d3-survey-2015.json     |  9361 +++----------
 test/data/movies.json             | 60823 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----
 test/data/plot-documentation.json |  1076 +-
 3 files changed, 60316 insertions(+), 10944 deletions(-)
```

If I revert the changes to the three JSON data files in the test directory, the diff stat looks a lot more reasonable. It is still increasing the number of lines by a significant amount, but the overall diff is much smaller.

```
$ git diff (git merge-base HEAD origin/main) $REVERT_DATA_HASH --stat | tail -n 1
 330 files changed, 7685 insertions(+), 3333 deletions(-)
```

Switching Prettier to use line lengths of 120 (which I think are a bit extreme) closes that gap a little bit:

```
$ git diff (git merge-base HEAD origin/main) $PRETTIER_120_HASH --stat | tail -n 1
 330 files changed, 5680 insertions(+), 3167 deletions(-)
```

mbostock:
We definitely shouldn’t run Prettier on the test data (and JSON).

duaneatat:
> We definitely shouldn’t run Prettier on the test data (and JSON).

It's not running on the test data, I just missed the json

.git-blame-ignore-revs
```diff
@@ -0,0 +1,12 @@
...
+# details:
+# https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
+#
+# Github will use this file to ignore these commits for blame.
+#
+# Run this to configure git locally to also ignore these commits:
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Ran prettier on the entire codebase
+7dd26612601bbbcfccb81fdf939b7b1a301f1366
```
mbostock, line 12: Is this going to work after this PR gets squashed and merged?

mbostock: Or we just omit this part for now and/or update it after merging to be the squashed commit. We can ignore this entire PR.

mbostock:
How do we feel of doing Prettier at 120 characters per line? If that makes Prettier less intrusive I am in favor. 😄

.git-blame-ignore-revs
duaneatat: That works too!

duaneatat:
> How do we feel of doing Prettier at 120 characters per line? If that makes Prettier less intrusive I am in favor. 😄

Let's do it -- apparently it can end up with longer lines in cases where we might not expect it but I bet we'll be ok with it. Not at my computer for the day but I'll change it once I'm back

mbostock (approved):

Fil:
I have rebased the PR by manually applying a patch obtained by
> git diff 1df1d85818f082779d902d1588326880dc332e6c 772d416410c09d098fc3621cb0ca6b687eb1aef4

from which I removed all the prettified files.

Then I ran `prettier --write`, and appended that commit's id to the `.git-blame-ignore-revs` file.

(For reference I pushed the old tip of this branch (772d416410c09d098fc3621cb0ca6b687eb1aef4) to the prettier-old branch, that we can discard.)

Fil:
Note: this needs to be merged from the command line since we want to keep the three commits separate.

Issue #6, opened 2020-10-29
Custom styles for axis and ticks.


mbostock:
What if we want to change the color and font of axes?

morozovkirill:
It would be nice to set different font size for ticks on different axes. Now it's impossible even through class name selector. May be name the styles like "tickX"/"tickY"?

Fil:
Related: #126 ?

irenedelatorre:
It would be nice to add a class in the axis group (axis-x, axis-y). This is something that d3.js does, and it lets you style the axis as the user wants / needs. Plot already includes this information in the aria label.

PR #976, opened 2022-07-07
Added an optional css class to the x/y axes

dgwyer:
This is my first PR! 🙂

Related. https://github.com/observablehq/plot/issues/6

This PR introduces optional CSS classes for the `x` and `y` axes.

```
Plot.plot({
    x: {
      axisClass: true
    },
    y: {
      axisClass: true
});
```

If `axisClass` is _not_ defined or is `false`, then no CSS classes get added to the plot. I kind of like the current clean tree of attributes that Plot creates by default. So I considered maintaining this and only adding CSS classes for axes if specifically opted in.

If `axisClass` has been defined then it inserts the following classes. Note: Only the classes for the y-axis are shown below, but it's exactly the same for the x-axis if CSS classes are enabled.

```
<g class="y-axis" aria-label="y-axis" aria-description="↑ Close" transform="translate(40,0)" fill="none" text-anchor="end" font-variant="tabular-nums">
  <g class="tick" opacity="1" transform="translate(0,358.4133255882662)"><line stroke="currentColor" x2="-6"></line><line stroke="currentColor" x2="580" stroke-opacity="0.1"></line><text fill="currentColor" x="-9" dy="0.32em">60</text></g>

  ...

  <text class="y-axis-label" fill="currentColor" font-variant="normal" transform="translate(-40,20)" dy="-1em" text-anchor="start">↑ Close</text>
</g>
```

FYI - I've not added any docs for this PR yet. If it looks like it will be accepted then I'll update it of course.

### DIFF
--- a/src/axis.js
+++ b/src/axis.js
@@ -18,6 +18,7 @@ export class AxisX {
     label,
     labelAnchor,
     labelOffset,
+    axisClass,
     line,
     tickRotate,
     ariaLabel,
@@ -34,6 +35,7 @@ export class AxisX {
     this.label = string(label);
     this.labelAnchor = maybeKeyword(labelAnchor, "labelAnchor", ["center", "left", "right"]);
     this.labelOffset = number(labelOffset);
+    this.axisClass = boolean(axisClass);
     this.line = boolean(line);
     this.tickRotate = number(tickRotate);
     this.ariaLabel = string(ariaLabel);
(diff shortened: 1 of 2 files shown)


Fil:
Thanks for the contribution. I'm closing because we're probably going to revisit axes to change them to marks (see #1000), and we're also thinking about a generic className option that one could use on any mark (see #253, which would have to be rewritten as a "common option"). These two large generic changes would end up solving this.

PR #977, opened 2022-07-08
esbuild loader

mbostock:
This seems significantly faster than using ts-node.

A consequence is that we have to change the imports to drop the extension when using TypeScript. (We could import as .ts, but Rollup’s TypeScript plugin didn’t seem to support that.) Is that okay? Not sure what the trade-off is.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,6 +1,6 @@
 {
-  "require": ["module-alias/register.js", "ts-node/register"],
-  "loader": "ts-node/esm",
+  "require": ["module-alias/register.js"],
+  "loader": "@esbuild-kit/esm-loader",
   "extensions": ["js", "ts"],
   "spec": [
     "test/**/*-test.*",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   },
   "sideEffects": false,
   "devDependencies": {
+    "@esbuild-kit/esm-loader": "^2.4.0",
     "@rollup/plugin-commonjs": "22",
     "@rollup/plugin-json": "4",
     "@rollup/plugin-node-resolve": "13",
@@ -59,7 +60,6 @@
     "module-alias": "2",
     "rollup": "2",
     "rollup-plugin-terser": "7",
-    "ts-node": "^10.8.0",
     "tslib": "^2.4.0",
     "typescript": "^4.6.4",
     "typescript-module-alias": "^1.0.2",
--- a/src/axes.js
+++ b/src/axes.js
@@ -1,6 +1,6 @@
 import {extent} from "d3";
 import {AxisX, AxisY} from "./axis.js";
-import {formatDefault} from "./format.js";
+import {formatDefault} from "./format";
 import {isOrdinalScale, isTemporalScale, scaleOrder} from "./scales.js";
 import {position, registry} from "./scales/index.js";
 
--- a/src/axis.js
+++ b/src/axis.js
@@ -1,7 +1,7 @@
 import {axisTop, axisBottom, axisRight, axisLeft, format, utcFormat} from "d3";
 import {create} from "./context.js";
-import {formatIsoDate} from "./format.js";
-import {radians} from "./math.js";
+import {formatIsoDate} from "./format";
+import {radians} from "./math";
 import {boolean, take, number, string, keyword, maybeKeyword, constant, isTemporal} from "./options.js";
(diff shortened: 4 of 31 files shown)


duaneatat:
Unfortunately this will break the ES module that TypeScript emits -- the compiled js will be missing the extensions as well. See https://github.com/microsoft/TypeScript/issues/16577#issuecomment-703190339 and https://github.com/microsoft/TypeScript/issues/42151. TypeScript recommends always using the .js extension for this reason. We'll want to get esbuild-loader to resolve .js files to .ts if the .js file is missing (this is usually done with a custom resolver). I'll check to see if this is doable today

mbostock:
> Unfortunately this will break the ES module that TypeScript emits

I want understand exactly _how_ this breaks. Referring to:

https://exploringjs.com/impatient-js/ch_modules.html#module-specifiers

The ES modules that `tsc` generates are not consumed directly in the browser, and hence we don’t have to worry about the lack of extension there. And it’s already the case that these generated ES modules contain bare module specifiers that are not supported in browsers (such as the import of `"d3"` from [dist/axes.js](https://cdn.jsdelivr.net/npm/@observablehq/plot@0.5.2/dist/axes.js)), at least without import maps (which seem to be stalled anyway).

The generated ES modules _may_ be consumed by webpack and Rollup and in theory other bundlers depending on how they are configured, but these bundlers _tend_ to support file extension resolution and thus appear to continue working in my testing.

The generated ES modules _may_ be consumed in Node (though very rarely, as when server-side rendering but without the use of a typical front-end framework, _i.e._, without webpack and Babel). This breaks. I wrote a little test application to verify:

```
% node index.mjs 
node:internal/errors:465
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/mbostock/Development/observablehq-multi/plot/dist/format' imported from /Users/mbostock/Development/observablehq-multi/plot/dist/index.js
    at new NodeError (node:internal/errors:372:5)
    at finalizeResolution (node:internal/modules/esm/resolve:437:11)
    at moduleResolve (node:internal/modules/esm/resolve:1009:10)
    at defaultResolve (node:internal/modules/esm/resolve:1218:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

If it’s going to save us ~10 seconds on every test run and this is the only remaining hurdle, I definitely think it’s worth investigating how we could get this to work. It should at least be possible to post-process the generated ES modules to add the `.js` extension where needed. But I think it would be better to find a way to continue using the `.js` extension to import TypeScript files, since this seems to be the TypeScript-recommended convention.

duaneatat:
> The generated ES modules _may_ be consumed in Node

This is how I tested the compiled js.

> But I think it would be better to find a way to continue using the .js extension to import TypeScript files, since this seems to be the TypeScript-recommended convention.

I agree -- I can do some more investigation into this today 🙏

mbostock:
Okay, I learned how esbuild-kit/esm-loader worked, and wrote a custom loader that does the parts we need. The key limitation that was missing from esm-loader is that they only did the magic .js → .ts conversion when importing from a TypeScript file, but for Plot, we need to do that from JavaScript as well. Now we can use esbuild but there are no changes to the source or the generated ES modules.

mbostock:
Hmm, this works for me locally, but now it’s failing in CI. Does this work for you locally, @duaneatat?

duaneatat:
> The key limitation that was missing from esm-loader is that they only did the magic .js → .ts conversion when importing from a TypeScript file

Ah, Vite had the same limitation, and I had to write a custom resolver here:

https://github.com/observablehq/plot/blob/940ecafee6a2f61811db036c15833204d3bd86d6/vite.config.js#L22-L28

duaneatat (approved):
🚀

Fil:
Thank you!

before:
```
✨  Done in 32.28s.
yarn test  58,64s user 3,09s system 190% cpu 32,424 total
```

after:
```
✨  Done in 22.46s.
yarn test  39,20s user 2,30s system 183% cpu 22,602 total
```

PR #980, opened 2022-07-11
Document "mode" for group transform

kentr:
As far as I can tell, the `group` transform [accepts `"mode"` as an aggregation method](https://github.com/observablehq/plot/blob/08d3311f15dd628285c132997dee57b864e6aa58/src/transforms/group.js#L218).

This PR adds `"mode"` to the `README` in the list of aggregation methods for the `group` transform.

I'm not a statistics whiz, so I copied the `"mode"` entry from elsewhere with the assumption that it works the same.


Fil:
It does! thanks for catching this

Fil (approved):

PR #987, opened 2022-07-13
grid decoration mark

mbostock:
Alternative to #985.

I snuck the _axes_ state into the existing _context_ argument rather than adding yet another positional argument to _mark_.render. This allows the mark to access the _axis_.ticks option (whose default value is populated automatically by Plot). I could probably sneak this state somewhere else if you prefer, say in _dimensions_?

I chose to make this a decoration mark (with no data), but in theory we could make this a mark with automatic data based on the tick values, which would allow variable color, stroke-width, and everything else… which could be cool.

Also, maybe we could find a way to remove the grid functionality from the current axis, and instead have the existing _scale_.**grid** option translate into an implicit Plot.gridX or Plot.gridY mark being added?


mbostock:
Superseded by #1197.

PR #989, opened 2022-07-14
select{First,Last}[XY]

### LINKED ISSUE #988: Select last defined value (for some given channel)?
E.g., https://observablehq.com/@tophtucker/generations-on-google-trends

```js
Plot.text(
  data.filter((d) => d.generation !== "boomer"),
  Plot.select(
    {
      y: function* (I, Y) {
        for (let i = I.length - 1; i >= 0; --i) {
          const j = I[i];
          if (!isNaN(Y[j])) {
            yield j;
            return;
          }
        }
      }
    },
    Plot.windowY({
      x: "date",
      y: "value",
      k: 5,
      reduce: "mean",
      fill: "generation",
      text: "generation",
      textAnchor: "start",
      dx: 5
    })
  )
)
```

mbostock:
Fixes #988.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2026,9 +2026,25 @@ Plot.select({fill: "max"}, {x: "date", y: "city", fill: "temperature", z: "city"
 
 Selects the first point of each series according to input order.
 
+#### Plot.selectFirstX(*options*)
+
+Selects the first point of each series according to input order whose *x* value is defined.
+
+#### Plot.selectFirstY(*options*)
+
+Selects the first point of each series according to input order whose *y* value is defined.
+
 #### Plot.selectLast(*options*)
 
 Selects the last point of each series according to input order.
+
+#### Plot.selectLastX(*options*)
+
+Selects the last point of each series according to input order whose *x* value is defined.
+
+#### Plot.selectLastY(*options*)
+
+Selects the last point of each series according to input order whose *y* value is defined.
 
 #### Plot.selectMinX(*options*)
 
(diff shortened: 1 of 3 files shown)


tophtucker (approved):
❤️ 

so window leaves undefined points rather than changing the length of the data array? e.g. as you change the k of a Plot.windowY, the extent of the x-axis won't change, it'll just get empty bits at the ends. those points have x values but not y values, right?

when i was struggling earlier, i think i actually tried Plot.selectLastX to see if it'd work. but it'd actually be Plot.selectLastY in my case, right? i think i guessed X because it's the last along the x-axis — but no, it's selectLastY because Y is what has to be _defined_.

to position a text label in x and y at the end of the line, i really need the last point where _both_ x and y are defined, right? but one should always imply the other, unless you have weird gappy stuff.

i'm trying to think of when i wouldn't want regular selectLast to have this behavior. maybe something like: i have multi-line chart of stock prices, and wanna put a dot for the current price, but if a stock was delisted a year ago i don't want a dot at the end of that line because that wouldn't be current; its current price is truly undefined.

Fil:
Very much needed. But let me suggest a more generic alternative in #991.

Fil (commented):
I'm still not sold on this idea.

It feels like the beginning of an intractable DSL (selectMaxXYRFillOpacity?).

With #993 (and #996 (?)) the original motivation [⌾](https://github.com/observablehq/plot/issues/988) for this change has been solved.

<del>Also, I forgot to mention this alternative:
>     Plot.selectMaxX({ x: "date", y: "unemployment", filter: "unemployment" })
</del>

Fil:
You're right, the alternative only works on the original data, not in the OP.

PR #990, opened 2022-07-14
extended window transform

mbostock:
If the **extend** option is true, then the window is truncated as necessary to extend output values to completely cover the domain. For example, if **k** is 24 and **anchor** is _middle_, then the initial 11 values have effective window sizes of 13, 14, 15, … 23, and likewise the last 12 values have effective window sizes of 23, 22, 21, … 12.


Fil (approved):
It will solve many problems, in particular the "disappearing label" when a series is smaller than the window size.

A caveat is that the base for statistics becomes inconsistent; typically this will exhibit more noise on the "extended" part, in particular if the window is anchored at the start or end. Maybe this needs to be mentioned somewhere in the documentation?

In most cases however this is a worthy trade-off—I'd even want us to consider this as the default.

PR #991, opened 2022-07-14
Plot.select…({valid: ["x", "y", "r"]}) 

Fil:
will select from the values that are not filtered

this needs documentation and tests, and "valid" is not the right verb.

Example: https://observablehq.com/@observablehq/select-a-valid-point-991

alternative to #989

### DIFF
--- a/src/transforms/select.js
+++ b/src/transforms/select.js
@@ -1,6 +1,7 @@
 import {greatest, group, least} from "d3";
 import {maybeZ, valueof} from "../options.js";
 import {basic} from "./basic.js";
+import {defined} from "../defined.js";
 
 export function select(selector, options = {}) {
   // If specified selector is a string or function, it’s a selector without an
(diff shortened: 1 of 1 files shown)


tophtucker:
this is a lil over my head now 😅 

what would i pass as _valid_ in my window example?

Fil:
with this API it would be `selectMaxX({valid: "y"})`. Though you could do `selectLast({valid: ["x", "y"]})` if what you want is the last point in the series that has a valid x and y.

mbostock:
I think it is probably too much to expect people to reason about which channels need to have defined values (or “valid”) for a data point to be considered. It would be much better to respect the [mark filter](https://github.com/observablehq/plot/blob/3ef622f9d20e62c15f3f839e116d5be0a7e37a12/src/plot.js#L294-L303) which considers all of a mark’s channels and ensure consistency with rendering. However, we don’t know what the downstream mark is when applying the select transform, and hence the select transform can’t tell what options represent channels (or even access those channel values without computing them redundantly). This might be fixable by making the select transform an initializer, but I suspect that would introduce other problems and make it harder to understand. So, I think it’s probably better to stick with the proposal of #989 and hope that autocomplete guides people towards the appropriate select transform. Alternatively maybe the select transform could uniformly require the *x* and *y* channel values to be defined to consider selection (if those options are present)… but I’m not wild about redundantly computing those channels, especially implicitly.

Fil:
> redundantly computing those channels
this might get better if valueof would (optionally?) do memoization?

Agree with the comment, this was a dead end.

Issue #147, opened 2021-02-24
Dual-scale charts (bad idea)?


mbostock:
Ref. https://blog.datawrapper.de/dualaxis/

Fil:
Discussion: https://github.com/observablehq/plot/discussions/626

Examples: https://observablehq.com/@fil/dual-axis-bar-line-chart

Fil:
It would be nice to have a few concrete examples to reason about.

For example, if we have two scales, one might be linear and the other log. Or one might be ordinal and the other temporal? (The only thing they share, in fact, is a common range.)

How would we indicate which scale is associated to which mark? Would a mark be able to use the two scales at once (for example to draw an arrow from (x1, y1) scaled with _xa_ and _ya_ to (x2, y2) scaled with _xb_ and _yb_)?

(Using a normalisation map transform might help to solve the problem above, maybe?)

And there is the question of the dual _axis_: each axis would have to know what scale it represents, on which side it is, etc.

papakpmartin:
I'm not sure if I'll answer your questions very well, but in the use case I had yesterday I was displaying two measurements (on Y axes) over time (X axis), for example, flow and conductivity mentioned above. This was extremely useful to identify certain effects as we modified the system.

Concerning Y scales, maybe I'm being foolish, but why not let either scale be whatever one wants? For example, if I'm exploring effects on a system, it could be really valuable to have one scale be linear and the other log because the nature of those two measured items might be that different, but their relationship over time might still be worth seeing.

I _don't_ think one should be able to use both scales (like an arrow from one scale into the other). That doesn't make sense to me; but each mark having it's own scale makes pretty intuitive sense, I think.

As far as the each axis knowing question, I'd thought that the `y: { axis: "right"...` idea seems intuitive... maybe that can be moved inside a `mark` for these? Like maybe...

```js
Plot.lineY(output_draw_data, {
  x: "timestamp",
  y: "value",
  stroke: "red",
  curve: "step-after",
  axis: "right",
 }),
```

I'm just riffing there... I know it's not that easy. But maybe calling out an `axis` there and then having `y.axis` optionally take an an object like...

```js
Plot.plot({
  y: {
    "left":  { type: "log", nice: true, line: true, color: green },
    "right": { nice: false, line: true, color: orange }
  }
}
```

...would be clear and easy enough?

papakpmartin:
[...]

PR #992, opened 2022-07-14
Fil/grid

Fil:
* upgrade the grid option to a grid mark
* fix some cases (band scales)
* slightly changes the logic for the fx/fy grid option
* document & test

Note that empty-x has changed visually; it's a case where grid:true is set as a top-level option, but x was defined as {axis:null}. It is now expected (as suggested by @awiederkehr) that we draw the x grid in that case.

includes #987


src/plot.js
```diff
@@ -22,6 +22,12 @@ export function plot(options = {}) {
   // Flatten any nested marks.
   const marks = options.marks === undefined ? [] : options.marks.flat(Infinity).map(markify);
 
+  // Prepend grid marks if passed a grid option
+  const gridx = options.x?.grid ?? options.grid;
```
mbostock, line 26: We can’t use the nullish coalescing operator here because defaults should only be inherited in the case of undefined, not null.

src/plot.js
```diff
@@ -171,32 +177,62 @@ export function plot(options = {}) {
...
+          .data(fyDomain.map(d => fy(d) + ty))
+          .enter()
+            .append("line")
+            .attr("y1", d => d)
+            .attr("y2", d => d)
+            .attr("x1", dimensions.marginLeft)
+            .attr("x2", dimensions.width - dimensions.marginRight)
+            .attr("stroke", "currentColor")
+            .attr("stroke-opacity", 0.1);
+      }
```
mbostock, line 193: Having a (non-configurable) alternate path for the grid on the facet axes seems unfortunate. Like, if Plot supports grid on _fx_ and _fy_, then I’d want to be able to use the Plot.grid mark (or Plot.gridFX? Plot.facetGrid? Plot.frameGrid?) to control how that grid is rendered. I guess that’s not possible because the facet grid needs to extend _across_ facet frames, and thus can’t live in the same space as other marks?

I wonder if there’s a more configurable way that we could do this. One option would be to have a “frame” grid that is repeated for each frame (changing the visual appearance of facet grids). Another option would be to have an array of “facet marks” (name TBD) that are rendered separately below (or above?) the facet frames in a single pass, perhaps where _x_ = _fx_ and _y_ = _fy_.

I think eventually it’d be nice to figure out how to replace the built-in axes with a custom axis mark, for example…

src/plot.js
```diff
@@ -452,3 +488,81 @@ class FacetMap2 extends FacetMap {
...
   }
 }
+
+// The Grid class is defined here so we can prepend it per the grid option
+// see grid.js for the helpers
+const defaults = {
+  ariaLabel: "grid",
+  fill: "none",
+  stroke: "currentColor",
+  strokeOpacity: 0.1
```
mbostock, line 498: Aside: something I ran into with the Delaunay mesh mark, but I kind of wanted `Plot.grid({stroke: "red"})` to reset the stroke opacity to 1.0 rather than using the default 0.1. I.e., if the stroke is undefined, then the default stroke is _currentColor_ and the default stroke opacity is 0.1, but if the stroke is defined, then the default stroke opacity resets to 1.0.

Fil: Yes I also struggled with this — it’s not obvious visually if it’s a stroke width or opacity that makes the line faint, and you have to either inspect or refer to the documentation. I like the suggested change.

src/plot.js
Fil: I'm not sure about the use case for the facet grid (it was there so I tried to keep it, and felt it was better to have it outside of axes). I like the idea of having a mark where x would be fx (and y fy); maybe some marks would work out of the box (like Plot.cell) if called in this particular context.

> replace the built-in axes with a custom axis mark

Yes, it seems like a natural fast-follow to this PR. It would also help a bit with #147.

PR #993, opened 2022-07-14
non-strict window by default

mbostock:
I think I agree this should be the default? Also the calling the option **strict** feels slightly better than **extend**? (Though maybe **fixed** would also work? I couldn’t find much precedent though the R [runner package](https://cran.r-project.org/web/packages/runner/vignettes/apply_any_r_function.html) uses **na_pad** and defaults to non-strict).

Also the window transform was assuming that the input index is a typed array, but at least in the case of the map transform and a _z_ channel, it may be a regular array. We still want to use subarray if available because it’s faster, and it’s probably not worth changing the map transform to generate Uint32Array, so I added a workaround in the window transform; I can take a different approach if you prefer.


Fil (approved):
👌

mbostock:
Oh, another possibility is that the **strict** option could allow NaN input data to be ignored, too. For example if you have one NaN value within the window currently, the mean value is always NaN because reduceSum requires the data to be fully defined (`nans === 0`):

https://github.com/observablehq/plot/blob/3ef622f9d20e62c15f3f839e116d5be0a7e37a12/src/transforms/window.js#L119

We’d probably want a separate option for that, though… so maybe **strict** isn’t a great name?

PR #994, opened 2022-07-14
mark.channels is an object

mbostock:
Broken out of #979. We’re already using a channel object for _mark_.initialize (since #801) so it makes sense to use an object here, too.

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -15,23 +15,18 @@ export function Channel(data, {scale, type, value, filter, hint}) {
   };
 }
 
-export function Channels(channelDescriptors, data) {
-  const channels = {};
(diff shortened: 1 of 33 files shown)


mbostock:
Maybe the _channels_ argument to the Mark constructor should be an object, too? We might want to add some accommodation for backwards-compatibility if we change that, though.

mbostock:
Okay I did the above thing, and I feel good about it being more consistent. It does mean we’re slightly more lax (we won’t detect when a mark declares a channel that duplicates one already declared by the built-in `styles`), but I think that’s fine. I made this backwards-compatible just in case there’s a custom Mark out there. We might remove the backwards-compatibility eventually but this feels like the graceful approach.

PR #997, opened 2022-07-15
generalize non-strict reducers

mbostock:
Well, this basically re-implements #990 from scratch again, but I think it makes sense. Several non-strict reducers (difference, ratio, first, and last) still need to be implemented: these will need to find the first _defined_ and last _defined_ values within the current window.


src/transforms/window.js
```diff
@@ -66,48 +64,38 @@ function maybeReduce(reduce = "mean") {
...
-        T[I[i + s]] = nans === 0 ? f(C.subarray(i, i + k)) : skipNaN ? f(C.subarray(i, i + k).filter(d => !isNaN(d))) : NaN;
+        T[I[i + s]] = nans === 0 ? f(C.subarray(i, i + k)) : NaN;
         if (isNaN(C[i])) --nans;
       }
     }
+  }) : ({
+    map(I, S, T) {
+      const C = Float64Array.from(I, i => S[i] === null ? NaN : S[i]);
+      for (let i = -s; i < 0; ++i) {
+        T[I[i + s]] = f(C.subarray(0, Math.max(0, i + k)));
```
mbostock, line 87: If we assume that s <= k (which should be true), then we can remove the Math.max here, right?

```suggestion
        T[I[i + s]] = f(C.subarray(0, i + k));
```

Fil:
Ah, and another thing that we should change is that _mode_ shouldn't coerce values to numbers.

mbostock:
I fixed the _mode_ problem in the latest commit.

src/transforms/window.js
```diff
@@ -120,66 +136,144 @@ function reduceSum(k, s) {
...
+}
+
+function valid(x) {
+  return x !== null && !isNaN(x);
+}
+
+function first(S, I, i, k) {
+  let j = i;
+  do; while (!valid(S[I[j]]) && ++j < i + k);
+  return S[I[j]];
```
mbostock, line 200: At this point _j_ may be outside the window. I will fix.

src/transforms/window.js
```diff
@@ -120,66 +136,144 @@ function reduceSum(k, s) {
...
+          T[I[i + s]] = sum / count;
+          if (b !== null && !isNaN(b = +b)) sum -= b, --count;
+        }
+      }
+    };
+  }
+}
+
+function valid(x) {
+  return x !== null && !isNaN(x);
```
mbostock, line 194: This check is appropriate if we are going to coerce to a number (for the _ratio_ and _difference_ reducers), but if we’re not (for the _first_ and _last_ reducers), we want to use `defined` and `Number.isNaN` instead of `isNaN`. I will fix.

PR #1001, opened 2022-07-19
ts defs

Fil:
bugs fixed:
* `Plot.valueof(undefined, () => {})` crashed
* `Plot.map("foo")` crashed
* `Plot.selectMinX` potentially inserted `undefined` in the facet index
* `mid(x1, x2)` was wrongly typed (it expected a `data` argument)


duaneatat:
will take a look asap! let's merge into main though, the ts-testing branch is just a temporary branch

Fil:
rebased on main

Fil:
I'm going to reboot this. I want the names and the autocomplete to match the API documentation exactly.

For a low-level example I want to type an index into a channel as `index` and `channel`, not as `number` and `any[]`. I want to remove `nullish`  and be explicit about the difference between `null` and `undefined`.

For a high level example, a Dot mark options object will be `DotOptions` and explicitly list all the fields that make sense for dots. I just need to figure out how to make that an extension of a common `MarkOptions` object.

PR #1005, opened 2022-07-23
ts api

Fil:
rebooting #1001 with a "typescript as documentation" approach.


src/options.ts
```diff
@@ -6,37 +26,44 @@ const TypedArray = Object.getPrototypeOf(Uint8Array);
...
+  value: ChannelOption | ChannelTransformAccessor | null | undefined,
+  arrayType: ArrayType
+) {
+  return typeof value === "string"
     ? map(data, field(value), arrayType)
-    : type === "function"
+    : typeof value === "function"
     ? map(data, value, arrayType)
-    : type === "number" || value instanceof Date || type === "boolean"
+    : typeof value === "number" || value instanceof Date || typeof value === "boolean"
```
enjoylife, line 41: I'm seeing warnings where the union of the type is under specified.  
```
Warning:(41, 7) Invalid 'typeof' check: 'value' cannot have type 'number'
Warning:(41, 36) Invalid 'instanceof' check: 'value' has type that is not related to 'Date'
Warning:(41, 61) Invalid 'typeof' check: 'value' cannot have type 'boolean'
```
A naive fix.. just append said types to the args list. e.g. `value: ... | number | Date | boolean,`.

Or maybe instead they should be added to the already defined types, e.g. `ChannelOption`, if not too broad 🤷 .

Fil: Thanks for flagging! I don't get these warnings myself, not in VSCode nor in `yarn test:typecheck`. How come?

Fil: Are you using the --watch option?

mbostock:
Probably superseded by #1320 which goes the d.ts route. Close?

PR #1008, opened 2022-07-26
ts options & data

Fil:
### data.ts

Documents the types of data that are expected by Marks, and the way they are handled internally (via valueof, arrayify, map…). Facets, groups &c are represented by a `Series`, an index into the columnar representation of the data.

### options.ts

Plot.**valueof** and Plot.**column** get their JSDoc from the README

Type safety issues:
- the **field** accessor might crash if the datum is not a generic object; it's a choice not to type-guard this (for performance)
- the **first** and **second** accessors might crash if the datum is not an array; same

Duck-typing was interesting to type; I tried two methods:

```js
export function isOptions(option: any): boolean {
  return isObject(option) && typeof (option as {transform: null}).transform !== "function";
}

export function isDomainSort(sort: any): boolean {
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return isObject(sort) && sort!.value === undefined && sort!.channel === undefined;
}
```
maybe the second version is better, esp. if we disabled `@typescript-eslint/no-non-null-assertion` on the whole project?

I'm still hesitating between the myriad ways one can specify an object. Let me know which one you prefer ( {[key: string]: any} over Record<<……>>)?

(Also a reminder to merge in cli, since we need the mv separate from the subsequent changes, rather than a file deletion+addition.)

(first part of #1005)

todo:
- [x] add a test for the crash of valueof(undefined, accessor function)


mbostock:
Can we find a way to not duplicate the contents of the README? I really don’t want the documentation to be duplicated—it will almost immediately get out of sync and be burdensome to keep in sync. I think either the JSDoc should be automatically extracted from the README or, perhaps more likely, _vice versa_.

duaneatat:
@mbostock I'm working on that this week 👍

Fil:
superseded by #1024

PR #1010, opened 2022-07-27
Some progress

duaneatat:
I wanted to send this to you before I got to bed, but it's not completed.

This is getting us in the direction of what @mythmon was talking about, where we can incorporate the type of the Datum as a generic throughout. I'm still confused in a lot of places as to what the code is supposed to do, so some of these changes are probably incorrect. But hopefully you can get something out of this anyway 🙏 

You can see that now if you call `valueof` on data with a certain shape you get the autocompletion for an accessor function for example:
image

### DIFF
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,13 +1,13 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
-export type Datum = any; // A Datum… who knows what it is?
+type ObjectDatum = Record<string, unknown>;
+export type Datum = ObjectDatum | Value;
+export type DatumKeys<T> = T extends ObjectDatum ? keyof T : never;
 
 /**
  * The marks's data contains the data for the mark; typically an array
  * of objects or values, but can also be defined as an iterable compatible
  * with Array.from.
  */
-export type Data = ArrayLike<Datum> | Iterable<Datum>;
+export type Data<T extends Datum> = ArrayLike<T> | Iterable<T>;
 
 /**
  * An array or typed array constructor, or any class that implements Array.from
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,6 +1,6 @@
-type ObjectDatum = Record<string, unknown>;
-export type Datum = ObjectDatum | Value;
-export type DatumKeys<T> = T extends ObjectDatum ? keyof T : never;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Datum = Record<string, Value> | Value;
+export type DatumKeys<T> = T extends Record<string, unknown> ? keyof T : never;
 
 /**
  * The marks's data contains the data for the mark; typically an array
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,6 +1,6 @@
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Datum = Record<string, Value> | Value;
-export type DatumKeys<T> = T extends Record<string, unknown> ? keyof T : never;
+export type ObjectDatum = Record<string, Value>;
+export type Datum = ObjectDatum | Value;
+export type DatumKeys<T> = T extends ObjectDatum ? keyof T : never;
 
 /**
  * The marks's data contains the data for the mark; typically an array
@@ -24,7 +24,7 @@ export type DataArray = ArrayLike<Datum>;
  * Channels are arrays of values
  */
 export type Value = number | string | Date | boolean | null | undefined;
-export type Channel = ArrayLike<Value>;
+export type Channel = Value[];
 export type TextChannel = string[];
 export type NumberChannel = number[] | Float32Array | Float64Array;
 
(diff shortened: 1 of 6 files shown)


src/data.ts
```diff
@@ -1,13 +1,13 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
-export type Datum = any; // A Datum… who knows what it is?
+type ObjectDatum = Record<string, unknown>;
+export type Datum = ObjectDatum | Value;
```
Fil, line 2: Does this reflect that Plot really accepts _any_ type of Datum? It's the role of an accessor to return a Value from that the Datum.

For one relatively legit example, here's an array of classes and functions and I want to plot their names and number of arguments.

`Plot.dotX([...Object.values(Plot)], { x: "length", y: "name" }).plot()
`

src/options.ts
```diff
@@ -27,58 +27,67 @@ Plot.valueof is not guaranteed to return a new array. When a transform method is
...
+  if (data == null) {
+    return data;
+  } else if (typeof value === "string") {
+    return map(data, field(value), arrayType);
+  } else if (typeof value === "function") {
+    return map(data, value, arrayType);
+  } else if (typeof value === "number" || value instanceof Date || typeof value === "boolean") {
+    return map(data, constant(value), arrayType);
+  } else if (isTransform(value)) {
+    return arrayify(value.transform(data), arrayType); // Not sure what's wrong here
```
Fil, line 66: this part is tricky, because a Transform method must return an ArrayLike or Iterable _of Value_ and not of _Datum_ (we are in the valueof function, so we expect transform+arrayify to be a ValueArray).

src/options.ts
```diff
@@ -27,58 +27,67 @@ Plot.valueof is not guaranteed to return a new array. When a transform method is
...
+export type ValueAccessor<T extends Datum> =
+  | DatumKeys<T>
+  | AccessorFunction<T>
   | number
   | Date
   | boolean
-  | TransformMethod
-  | ValueArray
+  | TransformMethod<T>
+  // | ValueArray // This didn't seem right, how would we use ValueArray as an accessor?
```
Fil, line 85: It's when you pass a "column" as the value accessor.

Typically: `Plot.text({length:1}, {x: [1], y: [2], text: ["Hello"]})`

src/options.ts
```diff
@@ -150,24 +169,36 @@ export function keyword(input: string | null | undefined, name: string, allowed:
...
+  data: Data<T> | null | undefined,
   type?: ArrayType
-): TypedArray | Array<Datum> | null | undefined {
+): TypedArray | Array<T> | null | undefined {
   return data == null
     ? data
     : type === undefined
     ? data instanceof Array || data instanceof TypedArray
-      ? (data as ValueArray)
+      ? (data as T[]) // is this change from ValueArray to T[] correct?
```
Fil, line 201: I don't think so, since a ValueArray can be a typedarray?

Fil:
How do you think we can include unit tests for types? Do we only have to create a types-test.ts file? We could create all the scenarios that are expected by each of the definitions of polymorphic functions.

src/options.ts
```diff
@@ -326,16 +372,15 @@ export type getColumn = {transform: () => ValueArray; label?: string};
...
 }
 
-export function labelof(value: any, defaultValue?: string) {
-  return typeof value === "string"
-    ? value
-    : value && (value as {label?: string}).label !== undefined
-    ? (value as {label?: string}).label
-    : defaultValue;
+type MaybeLabel = string | {label: string} | null | undefined;
+export function labelof(value: MaybeLabel, defaultValue?: string) {
```
Fil, line 382: labelof must accept any ValueAccessor; it just checks if there is a label property (which the user might assign on a typed array if they want!)

Fil:
I'm not sure how this can work, because some value accessors are legitimately _not_ DatumKeys—even if the datum is a {key:value} object. For example in `Plot.dotX([{value: 10}], {fill: "red", x: "value"})` the fill accessor is legal.

mythmon:
Thanks for pushing this forward Duane! Hopefully this is actually possible with Plot's API.

Another thing I mentioned Monday was that having explicit return types is useful. I see you added some here as well. I wanted to re-iterate that point. As we start adding more complex types, explicit return values will be become even more helpful, since we can say things like not only is it some output type, but that it is an output that matches the input generic, for example.

> I'm not sure how this can work, because some value accessors are legitimately not DatumKeys—even if the datum is a {key:value} object. For example in Plot.dotX([{value: 10}], {fill: "red", x: "value"}) the fill accessor is legal.

This is what I was worried about in the Slack thread. We could tell TS that `string` is also a valid type for the value accessor, but then it won't provide any specific guidance for field names, making this basically useless. Are colors the only place where we allow arbitrary strings as input? Could we enable field names for other channels?

Another idea is that we could try and define the valid colors that could be passed. This might just overwhelm the auto complete, like `PXX` did, but it could be worth exploring.

PR #1012, opened 2022-07-27
fixes for duane/ts-options

Fil:
@duaneatat Some changes to your PR so that it validates, but when I add these files into #1005, everything breaks.

The reason, I suspect, is that some value accessors are legitimately _not_ DatumKeys, even if the datum is a {key:value} object. For example, in `Plot.dotX([{value: 10}], {fill: "red", x: "value"})` the fill accessor is legal.

### DIFF
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,16 +1,14 @@
-type ObjectDatum = Record<string, unknown>;
-export type Datum = ObjectDatum | Value;
-export type DatumKeys<T> = T extends ObjectDatum ? keyof T : never;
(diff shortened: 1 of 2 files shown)


duaneatat:
Ah, we should add string & {} to allow arbitrary strings (or enumerate the colors) to ValueAccessor

Fil:
I've added some of the changes you made, as well as unit tests, to the fil/ts-options branch (#1008).

Fil:
obsolete

PR #1014, opened 2022-07-27
Warn users when facet data doesn't match mark data

### LINKED ISSUE #1009: Warn users when facet data doesn't match mark data
Provide a visual warning to users when they specify facet data that doesn't match the mark data. This is especially tricky in the case when the facet and mark data both use the same `.filter()` function, for example  ([notebook here](https://observablehq.com/d/a869f84e397d44b7)):
```js
Plot.plot({
  facet: {
    data: aapl.filter(d => d.Close > 1),
    y: d => d3.utcFormat("%Y")(d.Date)
  },
  marks: [
    Plot.lineY(aapl.filter(d => d.Close > 1), {x: "Date", y: "Close"})
  ]
})
```

The logic I would imagine is:

If there is facet data and there is mark data AND the mark data is not strictly equal to `===` the facet data, provide a warning like _Facet data and mark data must be strictly equal for accurate faceting_

mkfreeman:
Closes #1009. Provides a warning when facet data doesn't strictly match mark data. 

Note, I'm unsure if the logic or wording are exactly what we want. 

Logic to trigger the warning:

```js
if(
  facet?.data && 
  mark?.data  && 
  mark.data.length === facet.data.length && 
  mark.data !== facet.data && 
  facet.data.length > 1) {}
```

Warning text: _Facet data must strictly equal mark data for appropriate faceting. Make sure you aren't filtering your facet data in the Plot command_


src/plot.js
```diff
@@ -84,6 +84,17 @@ export function plot(options = {}) {
   // Initialize the marks’ state.
   for (const mark of marks) {
     if (stateByMark.has(mark)) throw new Error("duplicate mark; each mark must be unique");
+    if (
+      facet?.data &&
+      mark?.data &&
+      mark.data.length === facet.data.length &&
+      mark.data !== facet.data &&
+      facet.data.length > 1
+    ) {
```
mbostock, line 93: Here `facet.data` can be an iterable rather than an array (see the call to `arrayify` above), and so `facet.data.length` can be undefined. You can achieve the same result by using `facetIndex.length` instead.

Likewise`mark.data` can be an iterable. This is not converted to an array during the call to _mark_.initialize, but that arrayify’d data is also passed through the mark transform, so we don’t actually store it anywhere. I think this means we’ll need to write a helper function to get the length of the data. We could call `arrayify(mark.data)?.length` here, but that would create a copy of the data in order to perform the test, and it’d be faster not to.

`mark` should never be nullish here, so you don’t need the optional chaining operator to refer to `mark?.data`.

You should use `facets` to check for whether there is faceting as is done elsewhere in this file, rather than `facet?.data`. If `facets` is truthy, then we’re using faceting. (Strictly speaking, it’s possible for faceting to be enabled but for `facet.data` to be truthy, since if you specified `facet: {data: false}` it would be equivalent to `facet: {data: []}` which means that faceting is enabled, but that there are zero facets. That’s because `Array.from(false)` returns `[]`. We could throw an error if there are no facets, I suppose, above, but the main point is that we should be consistent about how we test for faceting, and the best way to be consistent is to check _derived_ state such as `facets` rather than the input options directly, since there are so many defaults and equivalent ways to specify input options.)

Lastly and perhaps most importantly, this isn’t checking the `mark.facet` option: this test should only be applied if the mark.facet option is undefined (or perhaps `"auto"`). Setting this option is how you disable the warning (by removing the ambiguity). The warning message should explicitly instruct you to set that option, for example: “If you intend for this mark to be faceted, set the facet option to true; if you intend for this mark to not be faceted, set the facet option to false.” The warning message should also include the type of the mark so that it’s easier identify, since warning messages don’t have line numbers. I also don’t think the message should refer specifically to filtering as that seems overly specific for what we are testing against. I can provide an alternate suggestion for the wording but I want to focus on getting the logic right first.

mbostock (changes requested):
Thanks, Mike. I’m trying to guide you on how to express the logic—let me know if I can clarify anything by meeting/chatting directly.

Fil (changes requested):
Please change the test/example also, because in this case it is not easily fixable: first, you would have to specify `facet: true` on the three marks, and second you'd be disappointed by the result :)

src/plot.js
mkfreeman: Wow, thanks for taking the time to walk me through this is such a detailed way. Here's a first pass, but I feel like the check on the `mark.data.size` isn't sufficient:

```js
if (
      facets && // if truthy, faceting is used
      mark.facet === "auto" && // if the user didn't specify a facet
      mark.data &&
      (mark.data.length === facetIndex?.length || 
        mark.data.size === facetIndex?.length)&& // for iterables
      mark.data !== facet.data &&
      facetIndex.length > 1
    )

```

Unsure if there's another check to do without fully copying the `arrayify` function.

test/plots/index.js
```diff
@@ -1,3 +1,4 @@
+export {default as facet_warning} from "./facet_warning.js";
```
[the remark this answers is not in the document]
mkfreeman: Will do!

src/plot.js
mbostock: Unfortunately there’s no easy test for the length of an iterable: in the general case, you actually have to iterate over it. And it’s even further complicated because Array.from doesn’t just allow iterables, it allows “array-likes” like `{length: 10}`. You can look at the [language specification for Array.from](https://tc39.es/ecma262/#sec-array.from) to get a sense of the complexity involved…

We could write an helper that takes data and tests to see if it has the expected length, something like:

```js
function hasLength(data, length) {
  if (data instanceof Array || data instanceof TypedArray) return data.length === length;
  if (data instanceof Map || data instanceof Set) return data.size === length;
  if (typeof data[Symbol.iterator] === "function") {
    let n = 0;
    for (const x of data) {
      if (++n > length) {
        return false;
      }
    }
    return n === length;
  }
  return data.length === length;
}
```

But I think maybe it’s better to just say `arrayify(mark.data).length === facetIndex.length` as the last condition, so that it’s only done if all of the other warning conditions are true.

src/plot.js
```diff
@@ -97,6 +97,18 @@ export function plot(options = {}) {
         ? facetsExclude || (facetsExclude = facetsIndex.map((f) => Uint32Array.from(difference(facetIndex, f))))
         : undefined;
     const {data, facets, channels} = mark.initialize(markFacets, facetChannels);
+    if (
+      facets && // if truthy, faceting is used
+      mark.facet === "auto" && // if the user didn't specify a facet
+      mark.data &&
+      (mark.data.length === facetIndex?.length || mark.data.size === facetIndex?.length) && // for Maps
+      mark.data !== facet?.data &&
+      facetIndex?.length > 1
```
mbostock, line 106: Based on previous comments, here’s my take on the logic:

```suggestion
      facets && // faceting is present
      facetIndex.length > 1 && // more than one facet (non-trivial faceting)
      mark.facet === "auto" && // no explicit mark facet option
      mark.data !== facet.data && // mark is implicitly not faceted (different data)
      arrayify(mark.data)?.length === facetIndex.length // mark data seems parallel to facet data
```

src/plot.js
```diff
@@ -97,6 +97,18 @@ export function plot(options = {}) {
...
+    ) {
+      warn(
+        `Facet data must strictly equal ${mark.ariaLabel} mark data for the appropriate faceting. This can be cause by using a filter function in the facet. If you intend for this mark to be faceted, set the facet option to true; otherwise set the facet option to false to disable this warning`
```
mbostock, line 109:
```suggestion
        `Warning: the ${mark.ariaLabel} mark appears to use faceted data, but isn’t faceted. The mark data has the same length as the facet data and the mark facet option is "auto", but the mark data and facet data are distinct. If this mark should be faceted, set the mark facet option to true; otherwise, suppress this warning by setting the mark facet option to false.`
```

mbostock (commented):
Please remove the extra facet_warning.svg snapshot, too?

Fil (changes requested):
The facet-warning example doesn't "work" in the sense that, if you add facet: true, you still get a funny-looking chart. We should have a better example, even if it's only used as a test plot.
Capture d’écran 2022-07-28 à 10 37 41

src/plot.js
mkfreeman: Makes sense, thanks again for clarifying!

Fil (approved):

src/plot.js
```diff
@@ -106,8 +106,9 @@ export function plot(options = {}) {
     if (
       facetIndex?.length > 1 && // non-trivial faceting
       mark.facet === "auto" && // no explicit mark facet option
+      mark.data &&
       mark.data !== facet.data && // mark not implicitly faceted (different data)
-      (data && data.length) === facetDataLength // mark data seems parallel to facet data
+      arrayify(mark.data).length === facetDataLength // mark data seems parallel to facet data
```
mbostock, line 111: This needs to be `arrayify(mark.data)?.length` since mark.data can be nullish… see my previous suggestion.

mbostock: (And then you don’t need the `mark.data` truthy check.)

src/plot.js
```diff
@@ -59,11 +59,13 @@ export function plot(options = {}) {
   let facetChannels; // e.g. {fx: {value}, fy: {value}}
   let facetsIndex; // nested array of facet indexes [[0, 1, 3, …], [2, 5, …], …]
   let facetsExclude; // lazily-constructed opposite of facetsIndex
+  let facetDataLength;
   if (facet !== undefined) {
     const {x, y} = facet;
     if (x != null || y != null) {
       const facetData = arrayify(facet.data);
       if (facetData == null) throw new Error("missing facet data");
+      facetDataLength = facetData.length;
```
mbostock, line 68: Might as well expand the scope of the `facetData` variable instead of introducing a new `facetDataLength`.

src/plot.js
```diff
@@ -99,6 +101,18 @@ export function plot(options = {}) {
...
+
+    // Warn for the common pitfall of wanting to facet mapped data.
+    if (
+      facetIndex?.length > 1 && // non-trivial faceting
+      mark.facet === "auto" && // no explicit mark facet option
+      mark.data !== facet.data && // mark not implicitly faceted (different data)
+      arrayify(mark.data)?.length === facetDataLength // mark data seems parallel to facet data
+    ) {
+      warn(
+        `Warning: did you want to facet the ${mark.ariaLabel} mark? The mark data and facet data are distinct, but they have the same length. If this mark should be faceted, set the mark facet option to true; otherwise, suppress this warning by setting the mark facet option to false.`
```
mbostock, line 113: Can we revert to my wording please? I don’t like the use of a question, and this no longer mentions that facet "auto" is in use.

src/plot.js
mkfreeman: To be clear, you're just suggesting that we use `facetData.length`? Not sure why @Fil had added this variable.

src/plot.js
mkfreeman: Happy to, I had just merged in @Fil's PR on top of your suggestion, happy to revert.

mbostock (approved):

Fil:
A small issue with the wording is that it mentions `facet: "auto"` when (most probably) the user has not specified a facet option.

mbostock:
True, but that is the documented default. We could change this warning to only trigger when the `facet` option is not specified at all.

PR #1015, opened 2022-07-28
facet warning logic & wording

Fil:
* Compare the lengths of the arrayified versions of facet.data and mark.data.
* Remove spurious svg
* Wording based on suggestion by @mbostock (but simpler) (?).


mkfreeman (approved):
Looks good to me!

src/plot.js
```diff
@@ -97,19 +99,20 @@ export function plot(options = {}) {
...
+    // Warn for the common pitfall of wanting to facet mapped data.
     if (
-      facets && // faceting is present
-      facetIndex.length > 1 && // more than one facet (non-trivial faceting)
+      facetIndex?.length > 1 && // non-trivial faceting
       mark.facet === "auto" && // no explicit mark facet option
-      mark.data !== facet.data && // mark is implicitly not faceted (different data)
-      arrayify(mark.data)?.length === facetIndex.length // mark data seems parallel to facet data
+      mark.data !== facet.data && // mark not implicitly faceted (different data)
+      (data && data.length) === facetDataLength // mark data seems parallel to facet data
```
mbostock, line 110: This is now testing the cardinality of the possibly-transformed data, rather than the original mark data. I think we should be looking at the original mark data as I had in my suggestion (since we’re using that same data for the equality test with “auto” facet). Why switch to the transformed data here?

Fil: Sorry, my bad! Will fix in #1014

PR #1016, opened 2022-07-28
Add script to inject JSDoc into README.md

duaneatat:
This adds a script that will inject the JSDoc into the README.md file for any symbols exported from src/index.js.

Add to the README:

```html
<!-- jsdoc foo -->
<!-- jsdocEnd -->
```

where `foo` is anything exported via src/index.js. Note that when using aliases, (e.g. `export {basic as transform}`), you reference the _alias_ in the html comment. `yarn readme:update` will get the JSDoc for foo and inject it between those html comments. If injecting multiple JSDocs consecutively, `jsdocEnd` is only required after the last one. E.g. this will work:

```html
<!-- jsdoc valueof -->
<!-- jsdoc transform -->
<!-- jsdoc column -->
<!-- jsdocEnd -->
```

Right now, this script will also generate the h4 titles based on the actual parameters found in the code. This is nice, because if you add or rename a parameter, the README will require an update (and will be caught by the github action). That does require, though, that we rename parameters in the code to match what we want to appear in the readme (as I did in for `basic` in basic.js).

If you both like this approach I will go through and add the annotations for all of the other exported symbols 🙏


duaneatat:
I added `yarn readme:check` to the Node.js CI github action but I can put it into a separate action if preferable. You can see that the first commit in this PR (intentionally) was out of sync and so CI failed: https://github.com/observablehq/plot/runs/7563275291?check_suite_focus=true
image

Fil:
Love it!  I don't mind marking the parameters by hand if it can make the code a bit lighter. We're not changing parameter names at random, and it will allow more control—like the the brackets ([*param*]) for optional arguments (although I guess the script could also treat those automagically).

duaneatat:
> I don't mind marking the parameters by hand if it can make the code a bit lighter. We're not changing parameter names at random, and it will allow more control—like the the brackets ([param]) for optional arguments (although I guess the script could also treat those automagically).

Makes sense. My take was that we will want to rename these parameters anyway, so that editors hints (like signature help) will align with the documentation. And yep, we can detect if parameters are optional

src/options.js
```diff
@@ -5,7 +5,20 @@ import {color, descending, quantile} from "d3";
 const TypedArray = Object.getPrototypeOf(Uint8Array);
 const objectToString = Object.prototype.toString;
 
-// This allows transforms to behave equivalently to channels.
+/**
+ * Given an iterable *data* and some *value* accessor, returns an array (a column) of the specified *arrayType* with the corresponding value of each element of the data. The *value* accessor may be one of the following types:
```
mbostock, line 9: Thoughts on hard wrapping? I like to not use hard wrapping Markdown, but I like hard wrapping TypeScript/JavaScript. It’s not a big deal to have the hard wrapping in Markdown if it’s going to be auto-generated, though (and we could also have the script remove hard wrapping, perhaps).

Fil: We've discussed it earlier today with Duane, and arrived at that very conclusion. Unformatted descriptions looks too weird in ts files, and since the copy/paste/format operation is going to be automated it's the best choice.

duaneatat: Yep! My plan is to wrap at 120 characters (could do 100 or 80 if preferred). I didn't plan to remove those hard wraps when injecting into the README though, but that would be doable if we really wanted that (though I would need to double-check the markdown rules around hard-wrapping).

Another thing I considered was to do some pseudo-two-way-binding between the README and the JSDoc:detect (via git) whether the changes changes made in the branch were to jsdoc sections of the README and if so reverse the direction and update the JSDoc from the README. Decided that's a bit much though

src/format.ts
```diff
@@ -8,11 +8,11 @@ const numberFormat = memoize1<Intl.NumberFormat>(
 );
 const monthFormat = memoize1<Intl.DateTimeFormat>(
   (locale: string | string[] | undefined, month: "numeric" | "2-digit" | "long" | "short" | "narrow" | undefined) =>
-    new Intl.DateTimeFormat(locale, {timeZone: "UTC", month})
+    new Intl.DateTimeFormat(locale, {timeZone: "UTC", ...(month && {month})})
```
Fil, line 11: shouldn't this be a separate PR?

duaneatat: Yep, I meant to push to duane/testing - will remove!

.github/workflows/node.js.yml
```diff
@@ -27,3 +27,4 @@ jobs:
           echo ::add-matcher::.github/eslint.json
           yarn run eslint . --format=compact
       - run: yarn test
+      - run: yarn readme:check
```
Fil, line 30: could we run readme:check as one of the tests? this way we won't forget locally then be reminded by CI?

src/transforms/tree.js
```diff
@@ -3,28 +3,42 @@ import {ascendingDefined} from "../defined.js";
...
-  delimiter, // how the path is separated
-  frameAnchor,
-  treeLayout = tree,
-  treeSort,
-  treeSeparation,
-  treeAnchor,
-  ...options
-} = {}) {
+/**
+ * Based on the tree options described above, populates the **x** and **y** channels with the positions for each node. The following defaults are also applied: the default **frameAnchor** inherits the **treeAnchor**. This transform is intended to be used with [dot](https://github.com/observablehq/plot/blob/main/README.md#dot), [text](https://github.com/observablehq/plot/blob/main/README.md#text), and other point-based marks. This transform is rarely used directly; see the [Plot.tree compound mark](https://github.com/observablehq/plot/blob/main/README.md#plottreedata-options).
```
Fil, line 7: Here's an example where the reference ("the tree options described above") fails to work when viewed outside of the context of README.md (it doesn't have to be addressed in this PR, of course!).

src/scales.js
```diff
@@ -471,6 +471,384 @@ export function coerceDate(x) {
     : new Date(x);
 }
 
+/**
+ * You can also create a standalone scale with Plot.**scale**(*options*). The *options* object must define at least one scale; see [Scale options](https://github.com/observablehq/plot/blob/main/README.md#scale-options) for how to define a scale. For example, here is a linear color scale with the default domain of [0, 1] and default scheme *turbo*:
```
Fil, line 475: Another loose reference ("also")

src/options.js
Fil: Is this all that remains to do?

src/format.ts
Fil: same with weekday below

src/options.js
duaneatat: This is done in #1034

.github/workflows/node.js.yml
duaneatat: good call -- done!

src/transforms/tree.js
duaneatat: good catch 🙏

src/scales.js
```diff
@@ -471,6 +471,384 @@ export function coerceDate(x) {
...
+ *
+ * To reuse a scale across plots, pass the corresponding scale object into another plot specification:
+ *
+ * ```js
+ * const plot1 = Plot.plot(…);
+ * const plot2 = Plot.plot({…, color: plot1.scale("color")});
+ * ```
+ *
+ * For convenience, scale objects expose a *scale*.**apply**(*input*) method which returns the scale’s output for the given *input* value. When applicable, scale objects also expose a *scale*.**invert**(*output*) method which returns the corresponding input value from the scale’s domain for the given *output* value.
+ *
```
Fil, line 493: Maybe we could stop here for the jsdoc, and leave the rest in README. (The position options etc are mostly meant in the context of writing scale descriptions as top-level options of the plot.)

duaneatat: that makes sense, I'll make that change today

duaneatat: done!

src/options.js
```diff
@@ -5,7 +5,20 @@ import {color, descending, quantile} from "d3";
...
+ *
+ * * a string - corresponding to the field accessor (`d => d[value]`)
+ * * an accessor function - called as *arrayType*.from(*data*, *value*)
+ * * a number, Date, or boolean — resulting in an array uniformly filled with the *value*
+ * * an object with a transform method — called as *value*.transform(*data*)
+ * * an array of values - returning the same
+ * * null or undefined - returning the same
+ *
+ * If *arrayType* is specified, it must be Array or a similar class that implements the [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) interface such as a typed array. When *type* is Array or a typed array class, the return value of valueof will be an instance of the same (or null or undefined). If *type* is not specified, valueof may return either an array or a typed array (or null or undefined).
```
Fil, line 18: Here we have one *type* that should be *arrayType*, but I feel we should do the opposite, and rename the argument to *type* instead of *arrayType*?

README.md
```diff
@@ -1921,22 +2096,31 @@ Plot.map({y: "cumsum"}, {y: d3.randomNormal()})
 
 Groups on the first channel of *z*, *fill*, or *stroke*, if any, and then for each channel declared in the specified *outputs* object, applies the corresponding map method. Each channel in *outputs* must have a corresponding input channel in *options*.
 
-#### Plot.mapX(*map*, *options*)
+
+<!-- jsdoc mapX -->
+
+#### Plot.mapX(*mapping*, *options*)
```
Fil, line 2102: here also we have renamed an argument of the API, can we do the opposite?

README.md
```diff
@@ -2093,53 +2299,73 @@ In addition to the **y1** and **y2** output channels, Plot.stackY computes a **y
 
 If two arguments are passed to the stack transform functions below, the stack-specific options (**offset**, **order**, and **reverse**) are pulled exclusively from the first *options* argument, while any channels (*e.g.*, **x**, **y**, and **z**) are pulled from second *options* argument. Options from the second argument that are not consumed by the stack transform will be passed through. Using two arguments is sometimes necessary is disambiguate the option recipient when chaining transforms.
 
-#### Plot.stackY([*stack*, ]*options*)
+<!-- jsdoc stackY -->
+
+#### Plot.stackY(*stackOptions*, *options*)
```
Fil, line 2304: renamed argument; and let's make a note that the optional argument is not shown with brackets (but the example below reminds the user that it's quite possible to use only one argument).

README.md
```diff
@@ -2306,31 +2550,41 @@ The **anchor** option may one of *middle*, *right*, and *left* for dodgeX, and o
 
 The dodge layout is highly dependent on the input data order: the circles placed first will be closest to the dodge anchor. When using the dodge layout with circles of varying radius, the data is sorted by descending radius by default; you can disable this behavior by setting the **sort** or **reverse** option.
 
-#### Plot.dodgeY([*dodgeOptions*, ]*options*)
+<!-- jsdoc dodgeY -->
+
+#### Plot.dodgeY(*dodgeOptions*, *options*)
```
Fil, line 2555: A note that we're losing the brackets signalling an optional argument.

README.md
duaneatat: In this case I couldn't because using the name `map` would shadow the `map` function that is called internally. I wanted to avoid `m` as well as aliasing `map` in the map file. Maybe there is a better name than `mapping`?

README.md
```diff
@@ -2433,22 +2695,28 @@ Plot.formatIsoDate(new Date("2020-01-01T00:00.000Z")) // "2020-01-01"
 
 Given a *date*, returns the shortest equivalent ISO 8601 UTC string. If the given *date* is not valid, returns `"Invalid Date"`.
 
-#### Plot.formatWeekday(*locale*, *format*)
+<!-- jsdoc formatWeekday -->
+
+#### Plot.formatWeekday(*locale*, *weekday*)
```
Fil, line 2700: change of argument name

Fil:
I've added a few comments.

One thing that would be nice to change is to replace `https://github.com/observablehq/plot/blob/main/README.md#` with `#` in the produced README.md file, so that it can be read outside of github.

README.md
duaneatat: Same here, we use `format` internally in the function. Perhaps I can rename `const format` to something else so we can use `format` instead of `weekday`?

README.md
duaneatat: same here, the code uses `stackOptions` as the param name to avoid shadowing the `stack` function.

README.md
Fil: Ah yes, it makes sense—this map function is exposed, and it would probably be awkward to rename it internally to (say) mapT, the export mapT as map. *mapping* is probably fine.

README.md
Fil: yes, I'd favor that in this case

src/options.js
duaneatat: done!

duaneatat:
> I've added a few comments.
> 
> One thing that would be nice to change is to replace `https://github.com/observablehq/plot/blob/main/README.md#` with `#` in the produced README.md file, so that it can be read outside of github.

Good call -- will do!

README.md
Fil: hmmm, OK… but this also changes the link's name (produced automatically by github) from #plotstackystack-options to #plotstackystackoptions-options; it's referenced below.

README.md
Fil: another possibility would be *output* (singular), to underscore that it is just one element of the more generic outputs of the map transform

Fil: note that it also has a link (in the Plot.normalize section) that we'll have to change

README.md
```diff
@@ -1921,22 +2096,31 @@ Plot.map({y: "cumsum"}, {y: d3.randomNormal()})
...
+
+<!-- jsdoc mapX -->
+
+#### Plot.mapX(*mapping*, *options*)
 
 ```js
 Plot.mapX("cumsum", {x: d3.randomNormal()})
 ```
 
 Equivalent to Plot.map({x: *map*, x1: *map*, x2: *map*}, *options*), but ignores any of **x**, **x1**, and **x2** not present in *options*.
```
Fil, line 2108: would need to change *map* to *mapping* (also below in the mapY section)

README.md
```diff
@@ -2093,53 +2299,73 @@ In addition to the **y1** and **y2** output channels, Plot.stackY computes a **y
...
-#### Plot.stackY([*stack*, ]*options*)
+<!-- jsdoc stackY -->
+
+#### Plot.stackY(*stackOptions*, *options*)
 
 ```js
 Plot.stackY({x: "year", y: "revenue", z: "format", fill: "group"})
 ```
 
 Creates new channels **y1** and **y2**, obtained by stacking the original **y** channel for data points that share a common **x** (and possibly **z**) value. A new **y** channel is also returned, which lazily computes the middle value of **y1** and **y2**. The input **y** channel defaults to a constant 1, resulting in a count of the data points. The stack options (*offset*, *order*, and *reverse*) may be specified as part of the *options* object, if the only argument, or as a separate *stack* options argument.
```
Fil, line 2310: would need to change *stack* to *stackOptions*

Fil:
I don't like the changes in arguments' names that cascade through the documentation; if we decide for the more conservative approach of not changing any name, it means that we need to break the "shadowing" problem. Maybe worth it?

Fil (approved):

Fil:
superseded by #1034

PR #1017, opened 2022-07-28
interpolate leq geq

Fil:
when interpolating, set the "future" values (time1) to the interpolated values and currentTime

(fixes driving-animation and gapminder)


mbostock:
Paired and merged as fc2206a7.

PR #1019, opened 2022-07-29
time & transforms

Fil:
Save. (Not working, but getting closer.)

The tests are broken, they cover:
- gapminder-box.js : transforms (almost working, but some demultiplication of marks appears)
- gapminder-box-facet.js : faceted transforms (broken rects)
- gampinder-dodge.js : initializers (broken: it should not pile up all the dots from different time frames)

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -72,9 +72,22 @@ export function plot(options = {}) {
     }
   }
 
+  // Aggregate and sort time channels.
+  const timeMarks = new Map();
+  for (const mark of marks) {
+    if (mark.timeChannel) {
+      timeMarks.set(mark, valueof(mark.data, mark.timeChannel.time.value));
+    }
+  }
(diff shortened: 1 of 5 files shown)


mbostock (commented):
Thanks for exploring this. I think even though you got the facets work in the other PR #1018 we should probably jump ahead to something like this approach. I think the farther we go down the filter-on-render approach, the more we have to unwind when we switch to the facet approach that will be needed to support transforms.

Rather than doing this in mark.initialize, I think we’ll have to do this up a higher level in Plot.plot, before mark.initialize is called (so that at that point, the facets are already split by time). This way the mark doesn’t need to understand how the facets are derived, which better isolates the logic and makes it backwards compatible and maybe we’ll think of other faceting logic in the future, too.

Fil:
continuing at #1018

PR #1026, opened 2022-08-09
faster maps

Fil:
it's faster to allocate a typed array with a known size than through Array.from — when dealing with millions of points, this can boost the performance quite a bit.

(results of a pairing session with @yurivish)

saving as draft for now (we'll want to make the test on the scale type less ad-hoc, probably?)

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -28,7 +28,10 @@ export function valueObject(channels, scales) {
   return Object.fromEntries(
     Object.entries(channels).map(([name, {scale: scaleName, value}]) => {
       const scale = scales[scaleName];
-      return [name, scale === undefined ? value : map(value, scale)];
+      return [
+        name,
+        scale === undefined ? value : map(value, scale, ["x", "y", "r"].includes(scaleName) ? Float64Array : undefined)
+      ];
     })
   );
 }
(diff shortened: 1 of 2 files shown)


src/channel.js
```diff
@@ -28,7 +28,10 @@ export function valueObject(channels, scales) {
   return Object.fromEntries(
     Object.entries(channels).map(([name, {scale: scaleName, value}]) => {
       const scale = scales[scaleName];
-      return [name, scale === undefined ? value : map(value, scale)];
+      return [
+        name,
+        scale === undefined ? value : map(value, scale, ["x", "y", "r"].includes(scaleName) ? Float64Array : undefined)
```
Fil, line 33: TBD: this is possibly wrong, since in the case of numbers we want to wrap the scale function and return NaN when the result is null? Or maybe it never happens?

src/options.js
```diff
@@ -88,13 +88,19 @@ export function arrayify(data, type) {
 // An optimization of type.from(values, f): if the given values are already an
 // instanceof the desired array type, the faster values.map method is used.
 export function map(values, f, type = Array) {
-  return values instanceof type ? values.map(f) : type.from(values, f);
+  if (values instanceof type) return values.map(f);
```
mbostock, line 91: Do we know that `values` is an array here? It could be that we are using `map` for non-arrays like iterables and {length} objects too.

PR #1037, opened 2022-08-26
output d.ts files

duaneatat:
This outputs the d.ts files. Once we've fully converted Plot to typescript we will add a "types" entry to package.json that references them, but not until then.

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -8,11 +8,11 @@ const numberFormat = memoize1<Intl.NumberFormat>(
 );
 const monthFormat = memoize1<Intl.DateTimeFormat>(
   (locale: string | string[] | undefined, month: "numeric" | "2-digit" | "long" | "short" | "narrow" | undefined) =>
-    new Intl.DateTimeFormat(locale, {timeZone: "UTC", month})
+    new Intl.DateTimeFormat(locale, {timeZone: "UTC", ...(month && {month})})
 );
 const weekdayFormat = memoize1<Intl.DateTimeFormat>(
   (locale: string | string[] | undefined, weekday: "long" | "short" | "narrow" | undefined) =>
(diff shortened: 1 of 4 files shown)


src/format.ts
```diff
@@ -8,11 +8,11 @@ const numberFormat = memoize1<Intl.NumberFormat>(
 );
 const monthFormat = memoize1<Intl.DateTimeFormat>(
   (locale: string | string[] | undefined, month: "numeric" | "2-digit" | "long" | "short" | "narrow" | undefined) =>
-    new Intl.DateTimeFormat(locale, {timeZone: "UTC", month})
+    new Intl.DateTimeFormat(locale, {timeZone: "UTC", ...(month && {month})})
```
mbostock, line 11: Looks like an unrelated change? Or at least the context is not clear why this is part of this PR.

mbostock (changes requested):
I think you’ll want to add the generated types to the `files` field in package.json or they won’t get published to npm.

https://github.com/observablehq/plot/blob/8065e59af569be9e0c37ec6686e895fa4e9a4471/package.json#L24-L27

src/format.ts
duaneatat: I meant to mention this -- this was a necessary step after adding `"declaration": true,` to tsconfig.json. The type for `DateTimeFormatOptions` allows `month` to be optional, but not `month` to be present and set to `undefined`. So we need to only set month if it's defined. We hadn't run into this yet because this is a declaration emit errors, and those are produced as part of producing declaration files themselves and won't be encountered otherwise.

mbostock:
Seems like this is still missing the change to `files` in package.json?

duaneatat:
> Seems like this is still missing the change to `files` in package.json?

Gah, I forgot to push! 😶

mbostock (approved):

Issue #1038, opened 2022-08-29
Bar plot with log scale disappears


tschaub:
I was surprised to see the marks (`barX`) on my plot disappear when I added `type: "log"` to my scale.  It took a bit of fiddling to discover that I had to add an explicit start (`x1`) value.  I'm wondering if it would be possible to make it so the bar starts at the start of the domain in cases where only one value (`x`) is given.

I wrote up the details here: https://observablehq.com/@tschaub/bar-plot-with-log-scale

If there aren't changes to the behavior that make sense, I would be happy to add a note to the scale or bar docs about using log scales with bars.

mbostock:
I prefer the current behavior because it makes the pitfall of using bars with a log scale clear. A note on the docs would be welcome though.

PR #1039, opened 2022-08-30
Note about bar plots with log scales

tschaub:
This adds a hint that bars must include an explicit start and end so that the domain doesn't include zero when using log scales.

See #1038.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,8 @@ wish to know. A *sqrt* transform exaggerates differences between small values
 at the expense of large values; it is a special case of the *pow* transform
 which has a configurable *scale*.**exponent** (0.5 for *sqrt*). A *log*
 transform is suitable for comparing orders of magnitude and can only be used
-when the domain does not include zero. The base defaults to 10 and can be
+when the domain does not include zero (for bar plots this means explicit
+non-zero start and end values are required). The base defaults to 10 and can be
 specified with the *scale*.**base** option; note that this only affects the
 axis ticks and not the scale’s behavior. A *symlog* transform is more
 elaborate, but works well with wide-range values that include zero; it can be


Fil:
It's not only about bars—other marks such as rect and area also have an implicit base value = 0.

An alternative would be to write:
> A *log* transform is suitable for comparing orders of magnitude and can only be used when the domain does not include zero—for marks such as areas or bars, which implicitly extend from a zero baseline, this means specifying non-zero start and end values.

On a more philosophical note I'm struggling to know what to recommend in that case, because in general, bar charts typically encode quantities that can be added (and stacked), and addition only works well with a linear scale.

Back to your notebook, I like what you did in the last example with `x: {type: 'log', domain: [1e3, 1e7], clamp: true}`; using a symlog scale would also work well.

Here are two other approaches I've tried:

1) transforming the *values* with a log function:
```js
Plot.plot({
  marks: [
    Plot.barX(data, Plot.mapX(X => X.map(d => Math.log10(d)), {x: "count", y: "letter"}))
  ],
  x: {tickFormat: d => 10**d},
  grid: true
})
```

2) using *offset* to add 1 to every start and end values:
```js
Plot.plot({
  marks: [
    Plot.barX(data, {
      x: "count",
      y: "letter",
      offset: (index, start, end, z) => {
        for (const facet of index) {
          for (const series of facet) {
            for (const i of series) start[i] = 1 + start[i];
          }
        }
      }
    })
  ],
  x: { type: "log" },
  grid: true
})
```

Note that the offset strategy allows to stack values, but makes it even harder to make sense of the resulting chart, because you basically have to do a reverse mental calculation.

Fil:
Revising this, an alternative might be to change the [bar](https://github.com/observablehq/plot#plotbarydata-options) section, where it has:

"this is the typical configuration for a vertical bar chart with bars aligned at y = 0"

we could add " (note that this supposes that the *y* scale is linear, or at the very least can represent 0)." 

This would be correct in a sense, but alas I'm not sure it would help anyone figure out what they're doing wrong… I'll close the suggestion for now, because I don't see how we can address this properly—maybe it's more something for a tutorial. More suggestions are still welcome!

PR #1042, opened 2022-09-06
v0.6.0 changelog

duaneatat:
This is just the changelog for v0.6.0. I haven't created the tag or updated package.json yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Observable Plot - Changelog
 
+## 0.6.0
+
+[Released September 6, 2022.](https://github.com/observablehq/plot/releases/tag/v0.6.0)
+
+[breaking] The **strict** option for Plot.windowX and Plot.windowY will now output `undefined` if any window value is undefined.
+
+Parts of the README have been incorporated throughout the codebase as JSDoc comments. This allows IDEs to display the documentation as tooltips.
+
 ## 0.5.2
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Observable Plot - Changelog
 
+## 0.6.0
+
+[Released September 6, 2022.](https://github.com/observablehq/plot/releases/tag/v0.6.0)
```
Fil, line 5:
```suggestion
*Not yet released. These are forthcoming changes in the main branch.*
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+
+[breaking] The **strict** option for Plot.windowX and Plot.windowY will now output `undefined` if any window value is undefined.
```
Fil, line 7:
```suggestion
[breaking] Plot.windowX and Plot.windowY now return an aggregate value even when the window contains undefined values, for example at the beginning or at the end of a series. Use the new **strict: true** option to opt-in to the older behavior and return undefined in those cases.
```

Fil (changes requested):

CHANGELOG.md
Fil: (this is how we did it last time, so we can merge it sooner)

Fil (approved):

PR #1055, opened 2022-09-23
Basic JSDoc types - Plot.plot and all Marks

duaneatat:
This PR uses a `types.ts` file for authoring the types (this is just for convenience, since I find it easier to write typescript types than use the JSDoc syntax), and we import those types as jsdoc tags where appropriate.

This PR covers Plot.plot and all of the marks.

Known issues:
- Some types are too broad, e.g. tickX doesn't accept x1 and y1, but those appear in the completions

TODO:
- [ ] Plot.image : move the description of options from the `### Image` section to the Plot.image jsdocs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1245,8 +1245,6 @@ The given *options* are passed through to these underlying marks, with the excep
 
 #### Plot.boxX(*data*, *options*)
 
-
-
 ```js
 Plot.boxX(simpsons.map(d => d.imdb_rating))
 ```
@@ -2981,8 +2979,6 @@ passed two node arguments, the child and the parent.
 <!-- jsdoc tree -->
 
 #### Plot.tree(*data*, *options*)
-
-
 
 A convenience compound mark for rendering a tree diagram, including a
 [link](#link) to
--- a/README.md
+++ b/README.md
@@ -2348,7 +2348,6 @@ Plot.rect(athletes, Plot.bin({fillOpacity: "count"}, {x: "weight", y: "height"})
 Bins on *x* and *y*. Also groups on the first channel of *z*, *fill*, or
 *stroke*, if any.
 
-
 <!-- jsdoc binX -->
 
 #### Plot.binX(*outputs*, *options*)
@@ -2359,7 +2358,6 @@ Plot.rectY(athletes, Plot.binX({y: "count"}, {x: "weight"}))
 
 Bins on *x*. Also groups on *y* and the first channel of *z*, *fill*, or
 *stroke*, if any.
-
 
 <!-- jsdoc binY -->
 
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -1245,6 +1245,8 @@ The given *options* are passed through to these underlying marks, with the excep
 
 #### Plot.boxX(*data*, *options*)
 
+
+
```
duaneatat, line 1249: I plan to trim these new lines in the jsdoc-to-readme script

test/plots/index.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <head>
-  <meta charset="utf-8">
```
duaneatat, line 3: Odd, not sure why this change was necessary. Will take a look tomorrow

Fil: I've had the same experience on a different branch. Could it be that the github action has changed?

Fil: put this tiny change in a separate PR #1059

duaneatat: thank you! I'll merge main in once that's in

Fil:
I think I'd like to have all the types imports at the top of each file, as I was trying to do with the "import types" in #1005.


Maybe for later: I've tried to add a BarYOptions to the `barY(data, options)` mark, by creating
```
export interface BarYOptions extends MarkOptions {
  offset: null | "expand" | "center" | "wiggle" | function…;
  order?: "sum" | "value" | …;
}
```

Of course, it would have to be extended to cover all the stacking options, which means I'd want to write something like `BarYOptions = MarkOptions & StackYOptions;`. Not sure if this can be done with "type" or "interface", though.

Fil:
I've just noticed that Plot.window(X/Y) and Plot.normalize(X/Y) have no jsdocs at all.

Fil (approved):
lgtm!

src/marks/area.js
```diff
@@ -85,8 +92,10 @@ export class Area extends Mark {
  * [Plot.areaX](https://github.com/observablehq/plot/blob/main/README.md#plotareaxdata-options)
  * is used in the vertical orientation where the baseline and topline share *y*
  * values.
+ * @param {Data} data
+ * @param {Options} options
  */
-export function area(data, options) {
+export function area(data, options = undefined) {
```
duaneatat, line 98: In order to have editors know that options is optional, we had to add these default values for options throughout. Without this, the typescript language server will consider options required:

image


Adding ? to the @param also didn't seem to help:

image

Only adding a default value will produce the ? in the signature help:
image

duaneatat: @mbostock pointed out that this can be done by using the proper JSDoc syntax for optional parameters (`@param {Options=} options`) 🙏

mbostock:
Probably superseded by #1320 which goes the d.ts route. Close?

Fil:
I'm good with closing, but let's keep the branch around for a while. There are a few things that I'll want to refer to, for example the to replace the D3 type:

```
/**
 * A symbol object with a draw function
 * @link https://github.com/d3/d3-shape/blob/main/README.md#custom-symbol-types
 */
type SymbolObject = {draw: (context: CanvasPath, size: number) => void};
```

mbostock:
Superseded by #1343.

PR #1058, opened 2022-09-27
autoHeight factors in the auto and manual margins. 

### LINKED ISSUE #1045: auto height for band scale consumes margins
When using Plot.cell with a small y domain and large marginTop and/or marginBottom the auto height calculations don't add to the margins as one might expect, causing a y scale with a collapsed range:
https://observablehq.com/d/223d6091d4190e75

Screen Shot 2022-09-09 at 4 50 50 PM

Fil:
See https://observablehq.com/@observablehq/plot-band-scale-bug-1045

fixes #1045

Note: this is the least disruptive I could do; but it becomes a bit complicated to understand, as we need to compute what the default margins would have been, and add the difference.

(includes #1059, in order to pass CI)

See also #837

### DIFF
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -44,8 +44,8 @@ export function Dimensions(
   };
 }
 
-function autoHeight({y, fy, fx}, geometry, add) {
+function autoHeight({y, fy, fx}, geometry) {
   const nfy = fy ? fy.scale.domain().length : 1;
   const ny = y ? (isOrdinalScale(y) ? y.scale.domain().length : Math.max(7, 17 / nfy)) : geometry ? 17 : 1;
-  return !!(y || fy || geometry) * Math.max(1, Math.min(60, ny * nfy)) * 20 + !!fx * 30 + 60 + Math.max(0, add);
+  return !!(y || fy || geometry) * Math.max(1, Math.min(60, ny * nfy)) * 20 + !!fx * 30 + 60;
 }
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -27,11 +27,7 @@ export function Dimensions(
     marginLeft = margin !== undefined
       ? margin
       : Math.max((yAxis === "left" ? 40 : 0) + facetMarginLeft, xAxis || fxAxis ? 20 : 0.5 - offset),
-    height = autoHeight(
-      scales,
-      geometry || isProjection(projection),
-      marginTop - marginTopAuto + marginBottom - marginBottomAuto
-    )
+    height = autoHeight(scales, geometry || isProjection(projection)) + Math.max(0, marginTop - marginTopAuto + marginBottom - marginBottomAuto)
   } = {}
 ) {
   return {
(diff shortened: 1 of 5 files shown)


enjalot (approved):

src/dimensions.js
```diff
@@ -6,27 +6,25 @@ export function Dimensions(
...
       : Math.max((yAxis === "right" ? 40 : 0) + facetMarginRight, xAxis || fxAxis ? 20 : 0.5 + offset),
-    marginBottom = margin !== undefined
-      ? margin
-      : Math.max((xAxis === "bottom" ? 30 : 0) + facetMarginBottom, yAxis || fyAxis ? 20 : 0.5 + offset),
+    marginBottom = margin !== undefined ? margin : marginBottomAuto,
     marginLeft = margin !== undefined
       ? margin
-      : Math.max((yAxis === "left" ? 40 : 0) + facetMarginLeft, xAxis || fxAxis ? 20 : 0.5 - offset)
+      : Math.max((yAxis === "left" ? 40 : 0) + facetMarginLeft, xAxis || fxAxis ? 20 : 0.5 - offset),
+    height = autoHeight(scales, marginTop - marginTopAuto + marginBottom - marginBottomAuto)
```
mbostock, line 27: Instead of passing `add` in, why not

```suggestion
    height = autoHeight(scales) + Math.max(0, marginTop - marginTopAuto + marginBottom - marginBottomAuto)
```

mbostock (changes requested):

Fil:
🙏 it's easier to read this way!

src/dimensions.js
```diff
@@ -6,27 +6,25 @@ export function Dimensions(
...
-    height = autoHeight(scales),
     facet: {
       margin: facetMargin,
       marginTop: facetMarginTop = facetMargin !== undefined ? facetMargin : fxAxis === "top" ? 30 : 0,
       marginRight: facetMarginRight = facetMargin !== undefined ? facetMargin : fyAxis === "right" ? 40 : 0,
       marginBottom: facetMarginBottom = facetMargin !== undefined ? facetMargin : fxAxis === "bottom" ? 30 : 0,
       marginLeft: facetMarginLeft = facetMargin !== undefined ? facetMargin : fyAxis === "left" ? 40 : 0
     } = {},
+    marginTopAuto = Math.max((xAxis === "top" ? 30 : 0) + facetMarginTop, yAxis || fyAxis ? 20 : 0.5 - offset),
+    marginBottomAuto = Math.max((xAxis === "bottom" ? 30 : 0) + facetMarginBottom, yAxis || fyAxis ? 20 : 0.5 + offset),
```
mbostock, line 17: We shouldn’t implement it this way; it allows a user to change the behavior of Plot by specifying these undocumented marginTopAuto and marginBottomAuto options. Instead we’ll have to restructure the code so that these are local variables to the function.

mbostock (approved):

PR #1060, opened 2022-09-27
spaces are not numbers

### LINKED ISSUE #1056: a space is interpreted as a number
`Plot.plot({ marks: [Plot.barX([{ x: 200, y: " " }], { x: "x", y: "y" })] })` wrongly recognizes the space as a number, then warns about y being numbers.

Fil:
closes #1056

(includes #1059)

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -330,8 +330,10 @@ export function isTemporalString(values) {
 // coercion because we want to ignore false positives on e.g. empty strings.
 export function isNumericString(values) {
   for (const value of values) {
-    if (value == null || (typeof value === "string" && /^\s*$/.test(value))) continue;
-    return typeof value === "string" && !isNaN(value);
+    if (value == null) continue;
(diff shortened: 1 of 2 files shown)


src/options.js
```diff
@@ -331,7 +331,7 @@ export function isTemporalString(values) {
 export function isNumericString(values) {
   for (const value of values) {
     if (value == null || value === "") continue;
```
mbostock, line 333: Given that we already ignore the empty string, I wonder if we shouldn’t ignore all whitespace strings, too?

```suggestion
    if (value == null || (typeof value === "string" && /^\s*$/.test(value))) continue;
```

Then you won’t need to trim below, and it would mean that we continue checking the other values if we encounter a whitespace string (rather than returning false).

mbostock (changes requested):

mbostock (approved):
Added tests and simplified.

PR #1061, opened 2022-09-27
fix frameanchor bottom

### LINKED ISSUE #1043: frameAnchor: bottom is not properly aligned
untitled (14)

```
Plot.plot({
  marginLeft: 75,
  x: { insetRight: 10 },
  marks: [
    Plot.frame(),
    Plot.barX(
      penguins,
      Plot.groupY(
        { x: "count" },
        {
          y: "species",
          fill: "sex"
        }
      )
    ),
    Plot.text(["Annotation\nat the bottom-right"], {
      frameAnchor: "bottom-right"
    })
  ]
})
```

demo https://observablehq.com/@observablehq/bottom-right-misaligned-1043

Fil:
closes #1043


test/plots/penguin-annotated.js
```diff
@@ -0,0 +1,31 @@
...
+            y: "species",
+            fill: "sex",
+            title: "sex",
+            sort: {y: "x", reverse: true}
+          }
+        )
+      ),
+      Plot.text(["Count of penguins\ngrouped by species\n and colored by sex"], {
+        frameAnchor: "bottom-right",
+        dx: -3,
```
enjalot, line 25: looking at the image, wouldn't it be nicer if dx was a positive number to avoid the overlap with the bar?
i guess that's a bit fiddly, could we right align the text?

enjalot (approved):

test/plots/penguin-annotated.js
Fil: hmmm, it doesn't look like this on my computer.

In fact, it's worse than this: when I see the image on the vite server, it's perfect. When I look at it in the Finder, the right alignment is broken. And on github everything is broken… Why is that???

vite server

Finder view

Github View

enjalot: when i download the svg and remove ```.plot tspan {
      white-space: pre;
    }```
it fixes the text anchor end issue.
i'm not seeing line-height or dx on the text marks in the svg though.

Fil: line-height becomes the various tspan elements' *y* attributes; and dx is factored in the *x* attribute of the text element.

Apparently the faulty views are caused by the pretty formatting of the svg with beautify.html in https://github.com/observablehq/plot/blob/86789ebeb2b87f26350083dd18ed84ae4098ba0b/test/plot.js#L19 — combined with the white-space: pre; of the text elements, it creates these fake spaces.

Fil: This side-effect of testing will be fixed after #1062.

Fil: Now it's looking good. Thanks @enjalot for catching this!
image

mbostock:
This PR broke the alignment of labels on scatterplots.

Here is a frameAnchor of _top_:

Screen Shot 2022-11-23 at 1 48 30 PM

The frameAnchor of _bottom_ is supposed to be symmetric, with the baseline of the text touching the anchor:

Screen Shot 2022-11-23 at 1 48 54 PM

This PR shifts the _bottom_ anchor by 1.29 lines, resulting in dots that are significantly farther away than intended by the anchor:

Screen Shot 2022-11-23 at 1 49 40 PM

Perhaps the problem in #1043 is specific to multiline text. I will investigate but we should revert this and apply a different fix.

Issue #169, opened 2021-02-24
Should the group transform special-case date keys?


mbostock:
A line’s stroke or z cannot be specified as a Date, because each Date is a distinct object, and thus every datum is put in a distinct group. (I think no is the answer is because the SameValueZero algorithm is strong precedent.)

PR #1068, opened 2022-10-03
make facet indexes rectangular if they overlap

Fil:
Yet another approach to the overlapping facet issue.

Instead of reading the value of X for index i as X[i], we read it as X[i % X.length]. This allows to create "long" channels as rectangular representations of _n * m_ values (_n_ being the length of the data, and _m_ the number of facets), where default channels are a "line" of _n_ values. There is no need for a reindexation plan, since the addressing system doesn't need extra information to read the proper value when the index is "out of range".

Besides this change, only the transforms that need to care about overlapping facets need to be modified: stack, map, and dodge.

**Performance.** This works only if we are allowed to create a Uint32Array of length _n*m_ — in some very specific cases (such as facets that overlap by pairs), it can break earlier than a subtler reindexing plan. Since it doesn't need to expand any existing channels, it should be faster and consume less memory in general.

(In the case of cumulative facets where each data point creates its own facet (which do not exist for now but will be created by time facets), this means we get a cap at something like 32k keyframes if we use one of the expanding transforms, such as stack or dodge).

**Backward compatibility.** Note that this will change the output if you used a column that had a length inferior to the data it covers. However, relying on this would have been a hack, which afaik was never used in the wild nor in our examples and tests. Furthermore, Plot's [introduction of columnar channels](https://observablehq.com/@observablehq/plot-marks) explicitly mentions that a columnar channel is _“an array of values of the same length as the data”._ I'm not sure how to document this.


src/transforms/bin.js
```diff
@@ -365,3 +365,9 @@ function binfilter([{x0, x1}, set]) {
 function binempty() {
   return new Uint32Array(0);
 }
+
+function union(facets) {
+  const U = new Set();
+  for (const f of facets) for (const i of f) U.add(i);
+  return U;
+}
```
Fil, line 373: todo: optimize as in fc378275c1965dcd7e7227215efee3d83ee51f4c

Fil:
Closing in favor of #0169

PR #1069, opened 2022-10-04
Facet expand

Fil:
Expand all channels in parallel to the facet reindexation.

~~Alternatives: #1068, #1070.~~

TODO:
- [x] create a central registry of channels, with their definitions (from [facets.js#L56](https://github.com/observablehq/plot/pull/1069/files#diff-540ac050c947f485398ca75ea672a4d4a23c59e708346f8a5dda03408b195ab7R56)) and associated scales (from [plot.js#L756](https://github.com/observablehq/plot/blob/f7aa3b23bcadd4911c216b1a82ee5618c92b27c9/src/plot.js#L756)).
- [x] put the "other" loop upstream
- ~~[x] offer an option to expand more (unlisted) channels—or to list more channels(?)~~ we'll expose registerChannel later

PR #1070, opened 2022-10-04
add a definition to the {transform} representation of a value

Fil:
a mark can read that definition and decide that it wants to run the transform, or just use the definition. This allows the mark to decide, for example, if a color is a channel or a constant color. If a channel, then reindexation is applied lazily.

This makes #1069 lighter, in the sense that channels are expanded lazily, and the mark still gets to decide how it wants to process the original channel definition.


Fil:
Let's stay on #1069 so as not to introduce a new property.

PR #1082, opened 2022-10-11
Add support for margin to ordinal color legend.

dgwyer:
Fixes https://github.com/observablehq/plot/discussions/1080#discussioncomment-3843129

### DIFF
--- a/src/legends/swatches.js
+++ b/src/legends/swatches.js
@@ -1,10 +1,10 @@
-import {path} from "d3";
-import {inferFontVariant} from "../axes.js";
-import {maybeAutoTickFormat} from "../axis.js";
-import {Context, create} from "../context.js";
-import {isNoneish, maybeColorChannel, maybeNumberChannel} from "../options.js";
-import {isOrdinalScale, isThresholdScale} from "../scales.js";
-import {applyInlineStyles, impliedString, maybeClassName} from "../style.js";
+import { path } from "d3";
+import { inferFontVariant } from "../axes.js";
+import { maybeAutoTickFormat } from "../axis.js";
+import { Context, create } from "../context.js";
+import { isNoneish, maybeColorChannel, maybeNumberChannel } from "../options.js";
+import { isOrdinalScale, isThresholdScale } from "../scales.js";
+import { applyInlineStyles, impliedString, maybeClassName } from "../style.js";
 
 function maybeScale(scale, key) {
   if (key == null) return key;
@@ -88,6 +88,9 @@ function legendItems(scale, options = {}, swatch, swatchStyle) {
     swatchSize = 15,
     swatchWidth = swatchSize,
     swatchHeight = swatchSize,
+    marginTop = 0,
+    marginRight = 0,
+    marginBottom = 0,
     marginLeft = 0,
     className,
     style,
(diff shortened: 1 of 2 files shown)


src/legends/swatches.js
```diff
@@ -1,10 +1,10 @@
...
-import {isNoneish, maybeColorChannel, maybeNumberChannel} from "../options.js";
-import {isOrdinalScale, isThresholdScale} from "../scales.js";
-import {applyInlineStyles, impliedString, maybeClassName} from "../style.js";
+import { path } from "d3";
+import { inferFontVariant } from "../axes.js";
+import { maybeAutoTickFormat } from "../axis.js";
+import { Context, create } from "../context.js";
+import { isNoneish, maybeColorChannel, maybeNumberChannel } from "../options.js";
+import { isOrdinalScale, isThresholdScale } from "../scales.js";
+import { applyInlineStyles, impliedString, maybeClassName } from "../style.js";
```
Fil, line 7:
```suggestion
import {path} from "d3";
import {inferFontVariant} from "../axes.js";
import {maybeAutoTickFormat} from "../axis.js";
import {Context, create} from "../context.js";
import {isNoneish, maybeColorChannel, maybeNumberChannel} from "../options.js";
import {isOrdinalScale, isThresholdScale} from "../scales.js";
import {applyInlineStyles, impliedString, maybeClassName} from "../style.js";
```

please let's stick to the prettier preferences of the project

src/legends/swatches.js
```diff
@@ -170,18 +173,22 @@ function legendItems(scale, options = {}, swatch, swatchStyle) {
...
-          : `
-          width: ${width}px;`
-      }
+          : `margin-top: ${+marginTop}px;`
+        }${marginRight === undefined
+          ? ""
+          : `margin-right: ${+marginRight}px;`
+        }${marginBottom === undefined
+          ? `margin-bottom: 0.5em;`
+          : `margin-bottom: ${+marginBottom}px;`
```
Fil, line 184: this mix seems like a problem, not sure how to unify that properly. Should we default to the equivalent number of pixels, on line 93?

Fil (changes requested):

src/legends/swatches.js
dgwyer: Fixed now. I've added scripts to package.json so we can easily run prettier manually.

e.g.

```
npm run prettier
```

```
npm run prettier-fix
```

We should also maybe take a look at [`lint-staged`](https://www.npmjs.com/package/lint-staged) to automate Prettier when committing changes locally?

src/legends/swatches.js
dgwyer: I've updated it to default to `8px` which should be the equivalent. What do you think?

src/legends/swatches.js
Fil: Your editor should pick up the prettier configuration from the repo; everything is there already since https://github.com/observablehq/plot/commit/87a6bcc2024dc37d5d55ee2ccd845c981a075e44 ; there is no need for additional tooling :) Also please keep concerns separate, this PR should only be about margin support.

dgwyer:
Not sure what went on with the previous commit (32e26ed) but hopefully it's ok now.

Fil:
For me it seems like 5px instead of 8px would make it visually unchanged in the unit tests.

Fil:
I'm sorry I went back to the original idea, in order to minimize churn. Please add your review (or thumbs up or down) on #1090. Thank you for the contribution!

PR #1084, opened 2022-10-11
Plot.image rotate option

### LINKED ISSUE #1083: image should have a `rotate` option just like dot
It would be really handy for me if `image` marks had the `rotate` option just like `dot` marks do.

My particular use case is making a diamond chart, which requires rotating the whole plot, which necessitates rotating any of the elements that appear within.

Fil:
Demo: https://observablehq.com/@observablehq/plot-image-rotate-1084

closes #1083

todo:
- [ ] unit test
- [ ] dx, dy— & anticipate what will happen if dx and dy become variable (as in #909)

### DIFF
--- a/src/marks/image.js
+++ b/src/marks/image.js
@@ -40,12 +40,13 @@ function maybePathChannel(value) {
 
 export class Image extends Mark {
   constructor(data, options = {}) {
-    let {x, y, width, height, src, preserveAspectRatio, crossOrigin, frameAnchor} = options;
+    let {x, y, width, height, rotate, src, preserveAspectRatio, crossOrigin, frameAnchor} = options;
     if (width === undefined && height !== undefined) width = height;
     else if (height === undefined && width !== undefined) height = width;
     const [vs, cs] = maybePathChannel(src);
     const [vw, cw] = maybeNumberChannel(width, 16);
     const [vh, ch] = maybeNumberChannel(height, 16);
+    const [vrotate, crotate] = maybeNumberChannel(rotate, 0);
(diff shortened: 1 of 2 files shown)


src/marks/image.js
```diff
@@ -104,6 +107,19 @@ export class Image extends Mark {
...
+            R || this.rotate
+              ? X && Y
+                ? (i) => `${X[i]}px ${Y[i]}px`
+                : X
+                ? (i) => `${X[i]}px ${cy}px`
+                : Y
+                ? (i) => `${cx}px ${Y[i]}px`
+                : `${cx}px ${cy}px`
+              : null
+          )
```
Fil, line 122: unfortunate complexity… am I missing a better way?

mbostock: Can you set use transform-origin="center" instead?

mbostock: I tried it and it doesn’t worker. ☹️ So I guess this is the only option… unless you wrap the image in a G element or something, but I bet that’s worse.

llimllib:
It looks great to me, thanks for the quick response! Though I have to say I don't understand the `transform-origin` bit at all

llimllib (approved):

mbostock (commented):
I think we want to combine the x, y, and transform-origin attributes into a single transform attribute, similar to what we do for dot and vector. Also this could use the `template` helper I created for #1189.

mbostock:
This looks useful.

Fil:
superseded by #1426

PR #1085, opened 2022-10-11
mark-level facets

### LINKED ISSUE #1081: New syntax for declaring facets from a mark
Broken out of #1041 and https://github.com/observablehq/plot/commit/0c2fadffd17cb77a81de33b1998ec20d2afe8858.

Instead of this:

```js
Plot.plot({
  facets: {
    data: penuins,
    x: "sex",
    y: "island"
  },
  marks: [
    Plot.dot(penguins, {
      x: "culmen_length_mm",
      y: "culmen_depth_mm"
    })
  ]
})
```

You could say:

```js
Plot.plot({
  marks: [
    Plot.dot(penguins, {
      x: "culmen_length_mm",
      y: "culmen_depth_mm",
      fx: "sex",
      fy: "island"
    })
  ]
})
```

Or in shorthand, instead of this:

```js
Plot.dot(penguins, {
  x: "culmen_length_mm",
  y: "culmen_depth_mm"
}).plot({
  facets: {
    data: penuins,
    x: "sex",
    y: "island"
  }
})
```

You could say:

```js
Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm", fx: "sex", fy: "island"}).plot()
```

Fil:
Closes #1081

The internal changes are pretty substantial, but hopefully easy to read.

Note that one unit test changes because the order of the index returned by the "exclude" option is not guaranteed (it was arbitrary anyways).


src/facet.js
```diff
@@ -0,0 +1,36 @@
...
+    let I = index;
+    if (vx) I = I.filter((i) => facetKeyEquals(vx[i], x));
+    if (vy) I = I.filter((i) => facetKeyEquals(vy[i], y));
+    return I;
+  });
+}
+
+// test if a value equals a facet key
+export function facetKeyEquals(a, b) {
+  return a instanceof Date && b instanceof Date ? +a === +b : a === b;
```
mbostock, line 26: This will need to do the same thing that InternMap does (valueOf).

mbostock:
I’m trying to work through two issues here:

- The _fx_ and _fy_ channels and scales should only be instantiated once. We currently instantiate them twice: first when reading the mark facet state, then again along with the non-facet channels and scales. (I’ve tried, but it seems to have broken sorting of facet scale domains, and I’m not sure how to fix it yet.)
- The _fx_ and _fy_ channels should respect the corresponding scale’s _transform_ option, if any (via applyScaleTransforms). We do this for normal channels but not for the facet channels since we’ve introduced a new path.

Fil:
"fy scale transform" is indeed broken. Example in https://observablehq.com/@observablehq/testing-facets-for-plot-1085

I wasn't able to reproduce the "date domain for fy" issue though.

mbostock:
I expected Date domains to work since we now use InternMap, but I think the order of the domain would have been wrong before in some cases. Thanks for testing.

Fil:
I think we should merge this and open a new issue for the "fy scale transform" bug (which is independent)?

Fil (commented):
Can't approve my own PR, but lgtm. 👍

src/plot.js
```diff
@@ -486,73 +530,133 @@ function nolabel(axis) {
...
 
-// Unlike facetGroups, which returns groups in order of input data, this returns
-// keys in order of the associated scale’s domains.
-function facetKeys({fx, fy}) {
-  return fx && fy ? cross(fx.domain(), fy.domain()) : fx ? fx.domain() : fy.domain();
+// Returns an array of {x?, y?} objects representing the facet domain.
+function Facets(channelsByScale, options) {
+  const {fx, fy} = Scales(channelsByScale, options);
+  let fxDomain = fx?.scale.domain(); // TODO .sort(ascending)?
+  let fyDomain = fy?.scale.domain(); // TODO .sort(ascending)?
```
mbostock, line 537: Do we need to sort here or can I remove these comments?

src/plot.js
```diff
@@ -486,73 +530,133 @@ function nolabel(axis) {
...
+  const {x, y} = facet;
+  if (x == null && y == null) return;
+  const data = arrayify(facet.data);
+  if (data == null) throw new Error(`missing facet data`);
+  // TODO Call applyScaleTransforms on the fx and fy channels.
+  const fx = x != null ? Channel(data, {value: x, scale: "fx"}) : undefined;
+  const fy = y != null ? Channel(data, {value: y, scale: "fy"}) : undefined;
+  const groups = facetGroups(data, fx, fy);
+  // If the top-level faceting is non-trivial, track the corresponding data
+  // length, in order to compare it for the warning above.
```
mbostock, line 613: What does “non-trivial” faceting mean? That there’s more than one facet?

src/plot.js
```diff
@@ -486,73 +530,133 @@ function nolabel(axis) {
...
+
+// Facet filter, by mark; for now only the "eq" filter is provided.
+function filterFacets(facets, {fx, fy}) {
+  const X = fx?.value;
+  const Y = fy?.value;
+  const index = range(X || Y);
+  const gx = X && group(index, (i) => X[i]);
+  const gy = Y && group(index, (i) => Y[i]);
+  return X && Y
+    ? facets.map(({x, y}) => arrayify(intersection(gx.get(x) ?? [], gy.get(y) ?? [])))
```
mbostock, line 658: I must be tired but I’m not following why we need to use intersection here… I feel like there’s some redundancy between this calculation at the nested groups we construct in facetGroups. I’m not following why we have to do both “grouping” and “filtering” now for facets.

src/plot.js
Fil: yes

src/plot.js
Fil: Yes. I changed it back to using the nested groups.

Explanation: this change was a remnant from the work on cumulative facets filter. When the filter is not "eq" but, say, "lte", we need to iterate over the whole cross product of gx x gy. But that's for a future iteration.

src/plot.js
Fil: edited the comment

src/plot.js
Fil: removed

mbostock (changes requested):
I feel the lack of support for fx/fy scale transforms is a blocker. (I’m assuming that this is a regression from previous behavior, but I also think that this should work as advertised even if it’s admittedly rarely needed.)

Fil:
I don't think it's a regression, just an old bug. Tests in https://observablehq.com/@observablehq/testing-facets-for-plot-1085 indicate that it is not working in any branch, nor in 0.6.0. Am I missing something?

mbostock (approved):
🎉

Issue #493, opened 2021-08-10
Communicate information about filtered data points


yurivish:
It would be useful if exploratory plots came with a visual indicator of “discarded data”. 

This would improve Plot's capacity for exploratory data analysis by enabling users to become aware of anomalous values that violate their assumptions about the data. 

For example, I changed a scale from log to symlog and discovered a bunch of negative values where I wasn’t expecting any. 

The data was supposed to be strictly positive and the negative values indicated a processing error, but since the default log scale filtered those data points out automatically I only noticed because I went out of my way to do additional spot checks. 

Plot could have made it evident immediately, e.g. with a legend saying something like “100 datapoints not shown”. Even more useful (maybe) would be being able to see a "data pipeline" and how many points are filtered out at each stage.

@fil observes that some filters use the discarding as a basic mechanism to do their work as intended, so there are subtle questions about what to communicate for this to be a useful signal. 

For the exploratory use case I think it makes sense for this to be on by default, since spot-checking every individual assumption manually can get onerous (e.g. checking for null/undefined, zeros where there shouldn’t be any, negative numbers where there shouldn’t be any, values outside of the x/y/color domain, NaN, etc.)

Fil:
The *scale*.unknown option can be used to this effect — [examples](https://github.com/observablehq/plot/issues/446#issuecomment-927687059).

Fil:
This would now happen, I guess, in the default filter https://github.com/observablehq/plot/blob/9d9ba917b5eb3f58e35b81e6555f3565ecffe8eb/src/plot.js#L291 . However with each warning we need to indicate a way to fix the situation, and in this case I wouldn't know what to say, in particular because in many charts some data is ignored on purpose.

eagereyes:
As an additional twist on this, it would be great if we could provide informative error messages for two seemingly common cases:
- somebody gets the capitalization of a key wrong, e.g., `city` instead of `City`
- somebody misspells the name of a key, e.g., `delivert` instead of `delivery`

Likely candidates for capitalization errors could be found by comparing the key provided to all the keys in the input object in a way that ignores case (i.e., converting both to lowercase before comparing).

Misspellings are more complex than that, possibly using Levenshtein distance and a threshold (or finding the closest match and suggesting that).

The latter is an expensive operation, but it would only have to be run when there's an error (or a presumed error), and it would mostly delay the error message, not interfere with normal Plot operation.

mstade:
[...]

PR #1091, opened 2022-10-17
show a warning for missing/invalid values in channels.

Fil:
For discussion, as a way to address #493.

Q: should this be opt-in or opt-out?

Capture d’écran 2022-10-17 à 16 14 59

closes #invalid

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -661,7 +661,7 @@ export function plot(options = {}) {
 
 export class Mark {
   constructor(data, channels = {}, options = {}, defaults) {
-    const {facet = "auto", sort, dx, dy, clip, channels: extraChannels} = options;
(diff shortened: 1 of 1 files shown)


Fil:
word-cloud.js is an example where we deliberately use an empty value (the text mark filters out `""`) to choose the words we display:

`text: (d) => (d.length > 1 ? `${d[0]} (${d.length})` : ""),`

Having to opt-out by setting `ignoreMissing: true` in this case would be unfortunate. Maybe we need two pass filters: first, for missing/invalid values (only checking undefined and NaN), and warning about it; second, the mark's normal filter. However this doesn't really work for the case of the dot mark which rejects negative values.

Capture d’écran 2022-10-17 à 16 21 38

PR #1098, opened 2022-10-22
Add className option for marks

### LINKED ISSUE #1002: className option for marks
className should be optional and have no default value, so you have to decide what names you want to use and there are no implicit classes

Related: #253 #6 #126 #490

(Note: for the general plot there is already a top-level className; we'd initially want this only for marks.)

RLesser:
see https://github.com/observablehq/plot/issues/1002 (and related discussions).

fixes #1002 

I believe with this implementation className will be an option on all marks.

### Open questions
* Other capabilities that should be tested?
* Should the two maybeClassName functions be combined in a cleaner way?


RLesser:
Looks like the jobs are failing due to a checkout issue? lmk if I need to construct my pull request differently.

Fil (changes requested):
I would change the logic a bit, instead of introducing a new function maybeClassNameOptional (which is a bit redundant), I would instead add a second parameter to `maybeClassName(name, provide)` and have plot, ramp, and swatches, call it with provide = true.

espinielli:
This seems interesting.
My use case would be for a `Plot.image` marks where I could define a border via CSS to the their class of using something like:

```
.my_image {
    outline: 6px solid white;
}
```

Looking forward to some progress ;-)

RLesser:
@Fil updated with new logic. Tests pass, but something about the prettier action is failing due to "A branch or tag with the name 'add-className-option-for-marks' could not be found". Not sure why that is.

RLesser:
@Fil, I think there might be an issue with users who aren't @Fil or @mbostock getting their pull requests past the prettier check.

I'm getting a `Error: A branch or tag with the name 'add-className-option-for-marks' could not be found` for that check.

I looked back thru past runs of that action and found that the two most recent times that other's have put up pull requests similar things have happened:
Screen Shot 2023-01-16 at 1 42 01 PM
Screen Shot 2023-01-16 at 1 42 20 PM

Do you think it might be a permission issue? Is there some other way I should be creating the branch or pull request?

mbostock:
The Prettier issues are almost certainly a problem with our GitHub workflow, likely these lines:

https://github.com/observablehq/plot/blob/a946fd9b6b0d85b7657a5b330bcd5ed1f32b7c47/.github/workflows/prettier.yml#L14-L18

I’m not sure why we have a separate workflow for Prettier given that ES Lint is already run as part of the main workflow here:

https://github.com/observablehq/plot/blob/a946fd9b6b0d85b7657a5b330bcd5ed1f32b7c47/.github/workflows/node.js.yml#L27-L28

We should investigate consolidating the Prettier checks into the main workflow.

jcolot:
In the meantime, an easy way to do this is by using the new render option, an example here: https://observablehq.com/@jcolot/observable-plot-adding-classes-to-marks

RLesser:
@Fil do you view this as still worth getting on? If so I can clean up the conflicts and get it ready for review.

Fil:
Yes, I still think it would be a useful feature. Thanks!

mbostock (changes requested):
I’m going to tweak how this is implemented, but the functionality looks right. 👍

mbostock (approved):

PR #1100, opened 2022-10-25
Ignore falsy marks

oluckyman:
Skips not only `null` and `undefined`, but any falsy mark. It simplifies usage of conditional marks and makes behaviour similar to React (React does not throw on falsy components, but Plot does).

Before:
```
marks: [
  condition ? Plot.dot(...) : null
]
```

After:
```
marks: [
  condition && Plot.dot(...)
]
```

### DIFF
--- a/test/plots/empty.js
+++ b/test/plots/empty.js
@@ -11,6 +11,8 @@ export default async function () {
       Plot.frame(),
       undefined,
       null,
+      false,
+      "",
       () => null,
       () => undefined,
       () => svg`<circle cx=50% cy=50% r=5 fill=green>`
--- a/test/plots/empty.js
+++ b/test/plots/empty.js
@@ -13,6 +13,7 @@ export default async function () {
       null,
       false,
       "",
+      0,
       () => null,
       () => undefined,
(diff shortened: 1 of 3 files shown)


test/plots/empty.js
```diff
@@ -11,6 +11,8 @@ export default async function () {
       Plot.frame(),
       undefined,
       null,
+      false,
+      "",
```
Fil, line 15:
```suggestion
      false,
      "",
      0,
```

Fil: I can see how respecting _false_ could be useful; not sure about considering `""` and `0` (Plot tends to be very deliberately "strict").

oluckyman:
> I can see how respecting false could be useful; not sure about considering "" and 0 (Plot tends to be very deliberately "strict").

I cannot imagine how `""` or `0` could end up in the marks array by mistake, so that throwing an error would be preferable.

And "falsy" has a strict definition in JS, so the decision is what to do with falsy marks. If ignore, then ignore all of them.

mbostock:
I would prefer to retain the current stricter (nullish) behavior.

Issue #1109, opened 2022-11-22
Arbitrary clip paths


mbostock:
It’d be nice to support arbitrary clip paths. I can implement one by wrapping a mark like so:

```js
function clip(mark, renderClip) {
  return {
    initialize(facets, channels) {
      return mark.initialize(facets, channels);
    },
    filter(index, channels, values) {
      return mark.filter(index, channels, values);
    },
    render(facet, scales, values, dimensions, context) {
      const fragment = document.createDocumentFragment();
      const svg = fragment.appendChild(mark.render(facet, scales, values, dimensions, context));
      const clipPath = fragment.appendChild(renderClip(facet, scales, values, dimensions, context));
      svg.setAttribute("clip-path", "url(#clip)");
      clipPath.setAttribute("id", "clip");
      return fragment;
    }
  };
}
```

That should probably extend Mark, though? And it should generate a unique identifier rather than using “clip”.

Then I could have a function that creates a clipPath element like so:

```js
function renderClip(facet, scales) {
  const clipPath = htl.svg`<clipPath>`;
  for (const {a, b} of abDisplayabilityCoordinates.filter(d => d.displayable)) {
    clipPath.appendChild(htl.svg`<rect
      x=${scales.x(a - 0.0025)}
      y=${scales.y(b + 0.0025)}
      width=${scales.x(a + 0.0025) - scales.x(a - 0.0025) + 1}
      height=${scales.y(b - 0.0025) - scales.y(b + 0.0025) + 1}
    />`);
  }
  return clipPath;
}
```

Ref. https://observablehq.com/@mjbo/oklab-named-colors-wheel

mbostock:
Maybe duplicate of #181.

espinielli:
I have tried to add `clip-path` to `Plot.image()` but it seems not handled by `applyIndirectStyles()` (that is what my browsing of the code brings me to...)

I would like to clip an image to a geo path:

```javascript
    Plot.image(italy, {
      x: (d) => d.properties.lon,
      y: (d) => d.properties.lat,

      width: (d) => d.properties.width,
      height: (d) => d.properties.height,
      preserveAspectRatio: "none",
      src: (d) => d.properties.flag,
      clipPath: (d) => `url(#iso-${d.id})`,
      title: (d) => d.id
    })
```

I have some tinkering going on here: https://observablehq.com/d/a8fe9e54cf07cb7a

Any thoughts?

Fil:
[...]

Issue #630, opened 2021-12-22
Create plot themes


mkfreeman:
To support different styles, we could create a set of plot themes (e.g., `dark`, `light`, etc.). As a comparison, the `ggthemes` [package](https://mran.microsoft.com/snapshot/2017-02-04/web/packages/ggthemes/vignettes/ggthemes.html) has default styles to emulate popular styles like the Wallstreet Journal, FiveThirtyEight, etc. 

Perhaps a simple approach would be to define a set of styles in `Plot.themes` that could be set as the `style` property in a Plot:

```js
Plot.plot({
    style: Plot.themes.dark
})
```

Where `Plot.themes.dark` has a set of key/value pairs describing the style: `{background: "black", color: "white",...}`

Fil:
related: https://github.com/observablehq/plot/issues/114

awiederkehr:
Would love to help work on this however I can!

mbostock:
The new axis mark #1197 will help with this, for example you could have a ggplot2 theme https://github.com/observablehq/plot/pull/1197#issuecomment-1364573933.

Fil:
Here's a handy response from Mike on the user forum:
https://talk.observablehq.com/t/set-global-plot-defaults-for-observable-plot/7797/3

martindaniel4:
As a Plot user, that feature would be very useful! Here is also another R package that was relevant to style graphs from a former Airbnb DS - https://github.com/ricardo-bion/ggtech

tophtucker:
I like Mike’s nice `style: Plot.themes.dark` suggestion. And it could be prototyped as an ongoing notebook.

It gets trickier when you want to set a default theme across many plots. One can also imagine wanting to set default categorical/sequential/diverging color schemes, etc.

Maybe there could be something like 

```
Plot.defaults({
  style: Plot.themes.dark,
  color: {scheme: {sequential: "blues"}}
})
```

which would return a new instance of Plot with those defaults? There’d be various difficulties. What’s that object syntax? (Note that I abused the color scheme option.) To name your version Plot in a notebook, would you have to do something like `Plot = (await require("@observablehq/plot")).defaults({style: Plot.themes.dark})`, or could you take advantage of the built-in Plot?

- https://ggplot2.tidyverse.org/reference/theme_get.html
- https://vega.github.io/vega-lite/docs/config.html

mbostock:
[...]

PR #1111, opened 2022-11-23
geometries and projections

### LINKED ISSUE #1043: frameAnchor: bottom is not properly aligned

mbostock:
Screen Shot 2022-11-23 at 12 21 03 PM

```js
Plot.plot({
  width: 960,
  height: 600,
  projection: "albers-usa",
  color: {
    scheme: "purd",
    domain: [1, 10],
    type: "quantize",
    n: 9,
    unknown: "#ccc",
    legend: true,
    label: "Unemployment (%)"
  },
  marks: [
    Plot.geometry(counties, {fill: d => unemployment.get(d.id), title: d => d.properties.name}),
    Plot.geometry(statemesh, {stroke: "white"})
  ]
})
```

TODO:

- [x] Pass _x_ and _y_ channels through the projection (_e.g._, for Plot.text labels in lat/long).
- [x] Disable the _x_ and _y_ scales when a non-default projection is used
- [x] Make it work with initializers
- [x] More named projections
- [x] Configurable rotation
- [x] Variable-size point geometries using the _r_ scale
- [x] Test graticule and sphere marks
- [x] Test the null projection
- [x] Documentation

Potential future enhancements (but not now):

- Combine (linear) projection with _x_ and _y_ scales (for geometry in abstract coordinates)?
- Shorthand for GeoJSON properties (_e.g._, `"name"` becomes `d => d.properties["name"]`)?
- Automatically fit a projection to the geometry?

Fixes #1043.


src/scales/quantitative.js
```diff
@@ -206,7 +207,7 @@ export function ScaleQuantize(
     if (min instanceof Date) thresholds = thresholds.map((x) => new Date(x)); // preserve date types
   }
   if (order(arrayify(domain)) < 0) thresholds.reverse(); // preserve descending domain
-  return ScaleThreshold(key, channels, {domain: thresholds, range, reverse});
+  return ScaleThreshold(key, channels, {domain: thresholds, range, reverse, unknown});
```
mbostock, line 210: This fixes an (unfiled) bug where the quantize scale didn’t respect the **unknown** option.

src/marks/text.js
```diff
@@ -82,13 +82,14 @@ export class Text extends Mark {
...
   render(index, scales, channels, dimensions, context) {
+    const {x, y} = scales;
     const {x: X, y: Y, rotate: R, text: T, fontSize: FS} = channels;
     const {rotate} = this;
     const [cx, cy] = applyFrameAnchor(this, dimensions);
     return create("svg:g", context)
       .call(applyIndirectStyles, this, scales, dimensions)
       .call(applyIndirectTextStyles, this, T, dimensions)
-      .call(applyTransform, this, scales)
+      .call(applyTransform, this, {x: X && x, y: Y && y})
```
mbostock, line 92: This fixes #1043.

Fil:
Would it make sense to call it Plot.geo rather than Plot.geometries? Shorter and more evocative of maps.

mbostock:
> Would it make sense to call it Plot.geo rather than Plot.geometries? Shorter and more evocative of maps.

That’d be fine, especially since it’s inherently tied to GeoJSON.

mbostock:
Example using normal marks (dot, text) with a projection:

Screen Shot 2022-11-24 at 10 05 14 AM

```js
Plot.plot({
  width: 960,
  height: 600,
  projection: "albers-usa",
  marks: [
    Plot.geometry(states, {fill: "#ccc"}),
    Plot.geometry(statemesh, {stroke: "white"}),
    Plot.dot(capitals, {x: "longitude", y: "latitude", fill: "currentColor"}),
    Plot.text(capitals, {
      x: "longitude",
      y: "latitude",
      frameAnchor: "bottom",
      text: (d) => `${d.capital}\n${d.state}`,
      dy: -6
    })
  ]
})
```

Fil:
One thing that comes to mind is we'll want to expose the projection definition so we can share it between different maps, and make it easier to augment the current map. Maybe this could follow the same pattern as scales: chart.scale("projection") would expose the projection's parameters, and apply and invert functions.

mbostock:
> One thing that comes to mind is we'll want to expose the projection definition so we can share it between different maps

Yeah, we could do that. I’m not sure if we should call the projection a scale but they are certainly similar. For now I don’t think this is particularly urgent because there is little internal configuration of the projection (the only thing that affects it is the width and height of the chart); as long as you specify the same projection type you’ll get the same behavior across charts.

Fil:
I've tried to port all my maps from the [Plot.carto 0.5 notebook](https://observablehq.com/@fil/plot-carto-0-5) to this branch, and took notes. **It works generally very well, and is a painless transition**: see https://observablehq.com/@fil/plot-geometries-1111

Only two things didn't work:

- **pointRadius**. We should support the *r* channel (or constant) as setting the path's pointRadius. I'll add a PR for this feature today.

- **Clipping.** Some projections need an option to "clip to sphere" (e.g. armadillo). _(Not urgent, but probably not difficult.)_


Beyond these, porting the maps was for the most part a no-brainer. One issue was slowing me down consistently though: **scaling**.

In general adjusting the scale to fit the map to the frame is hard—and sometimes it might be impossible to compute in advance (for example when the frame's size depends on the number of facets);
  *  the question might be "what to fit in the frame" (Plot.Carto defaults to the first feature or features, but we could imagine having an explicit option. We would also need to respect the padding, margins, etc, as well as frameAnchor to position the map in the direction that has some freedom)
  * We need to make it easier for D3 projections—they (generally) expect a 960*600 frame, and Plot defaults to 640*400; I want to call projection: {type: d3.geoBertin1953(), scale:"auto"} or something.
  * projection{type:"identity"} should accept scale/translate

_Usage_

(As I already mentioned), repeatedly having to type Plot.**geometry** was tripping me up. I hesitated with Plot._geometries_ (since we sometimes have one feature, or several features to draw). Having Plot.**geo** instead would be a no-brainer.

I was very happy with the fact that sending a single feature as data, instead of an array, worked. But I'm not sure if this should be done on all the marks? i.e. Plot.mark(Object, options)?

Using the _x_ and _y_ channels for longitude and latitude in Plot.dot is a bit disconcerting at first (in particular if your map is not "horizontal"). However it's better than any alternative I've considered, in particular because it makes transitioning a chart for "rectangular" to "projected" a one-line business.


_Ideas for future iterations:_

**More projections.** The default list of projections is a bit restricted. I'm not advocating to embed all of d3-geo-projections and d3-geo-polygon… but I'd like it to be easier to extend… maybe with some kind of plugin that would register projection names.  (Maybe related to theming-extending color schemes #630.)

**Exposing a projection as a scale.** (As already mentioned,) we'll have to think of a way to expose/share a scale's definition. If we push people to define projections with D3, it might be hard to create the object that allows to recreate the same projection somewhere else. If we create a small DSL for projections, it should cover more than the handful that is known to d3-geo (see also https://vega.github.io/vega-lite/docs/projection.html#projection-types).

**More graticule options.** Add graticule (step) options in Plot.graticule.

_Random remarks_

**Default dimensions.** Why do `Plot.geometry(land).plot()` and `Plot.geometry(land).plot({ projection: "identity" })` differ on the default size?  The technical answer is that the first is not setting x and y, but one could argue that as soon as we have the Plot.geometry mark, a projection is implied.

Fil:
Addendum: I like the idea of string accessors for properties, but I'm not sure how we would disambiguate between "Population" as an accessor for d => d.properties.Population and "id" for d => d.id.

Fil:
There is a funny issue with faceting: the positioning of the projection seems to be computed relative to the global frame. As a consequence when we try to use facets, the first map is in the middle of the global frame, and subsequent facets extend from there. (I'll try to fix.)
Capture d’écran 2022-11-25 à 17 49 12

Fil:
Capture d’écran 2022-11-25 à 21 06 56

:-)

mbostock:
> Why do Plot.geometry(land).plot() and Plot.geometry(land).plot({projection: "identity"}) differ on the default size?

This is a bug; they should be the same. We’re not consistent in how we check whether a projection is defined. In autoHeight, we check whether the **projection** option is non-nullish:

https://github.com/observablehq/plot/blob/119e3e8aeb480d4810ed8a056c1c13170a458b2e/src/dimensions.js#L49-L50

But in maybeProjection, we convert the _identity_ projection to be equivalent to null (and undefined, the default).

https://github.com/observablehq/plot/blob/119e3e8aeb480d4810ed8a056c1c13170a458b2e/src/projection.js#L21-L22

It would be better if autoHeight used the validated **projection** option so that we’re sure to be consistent, rather than checking the raw option value in multiple places. I think that means changing how we initialize the context to avoid a circular dependency.

> Plot.geometry(land).plot({ _projection: "identity" })

The reason this doesn’t give you 640×400 is because the projection is null and that’s what autoHeight uses (shown above). What we probably should do instead is bind the _geometry_ channel to the _projection_ scale (?) or otherwise check to see if a mark uses a _geometry_ channel. And if so, change the default height to 400. That way it won’t depend on whether a projection is used. (And a dot mark or any other that defines _x_ and _y_ channels would produce the same result.)

mbostock:
> We should support the r channel (or constant) as setting the path's pointRadius

Yes, I support this. It should be pretty easy to do, assuming we handle it similarly to dot’s _r_ channel.

> Some projections need an option to "clip to sphere"

It’s a bummer that D3 projections don’t handle this internally… But we would need to add a dependency on [polygon-clipping](https://observablehq.com/@fil/hello-polygon-clipping), presumably. https://github.com/d3/d3-geo/issues/266 Also related #1109. Maybe we could support this with a **clip** option on the geometry mark, maybe with `"sphere"` being the only supported value to start (other the default null for no clipping).

mbostock:
> projection: {type: d3.geoBertin1953(), scale:"auto"}

I want to avoid mutating a passed-in projection, and to minimize our assumptions about what methods must be implemented on projection implementations. Currently the only method that needs to be implemented by a custom projection is _projection_.stream.

> projection: {type:"identity"} should accept scale/translate

My expectation was that you’d pass in a D3 (or custom) projection in this case, say using d3.geoIdentity, rather than Plot providing another API for it.

Fil:
The trouble with the geoBertin example, is that the user can't a priori scale/translate it to the facet's inner dimensions, because they don't necessarily know in advance how many facets there will be. We need a way to rescale based on the subdimensions.

However this doesn't have to be a full "fitExtent" call: we could consider that a projection given as a stream function works in a reference viewport, where the coordinate system would be expected to match [960, 500]. One option is to wrap the projection stream, another to add a linear transform attribute to a *g* container. (Note that the latter approach would scale the text sizes and stroke widths too, which might be a good idea, but might also be a bit confusing.)

mbostock:
Added clip-to-sphere in latest commit.

Screen Shot 2022-11-25 at 4 54 49 PM

```js
Plot.plot({
  width: 960,
  height: 500,
  projection: geoArmadillo().precision(0.2),
  marks: [
    Plot.sphere(),
    Plot.geometry(land, {clip: "sphere", fill: "currentColor"}),
    Plot.graticule({clip: "sphere", strokeOpacity: 0.2})
  ]
})
```

mbostock:
> the user can't a priori scale/translate it to the facet's inner dimensions, because they don't necessarily know in advance how many facets there will be

In the latest commit I’ve allowed the **projection** option to be specified as a function which is passed the dimensions along with any additional projection options (typically **rotate** and **precision**, but I pass through any options). This should make it easier to configure a projection to fit the frame (in conjunction with #1114). Possibly this function could be passed the collected geometries to allow automatic fitting… but I think I’d prefer to keep that for a future enhancement to keep things simple now.

mbostock:
> I was very happy with the fact that sending a single feature as data, instead of an array, worked. But I'm not sure if this should be done on all the marks? i.e. Plot.mark(Object, options)?

I think this works well for the geometry mark because GeoJSON is easily duck typed. I don’t think we should do it in general. (I didn’t even do it for the Plot.Geometry class; only for the Plot.geometry convenience constructor.)

mbostock:
> Why do Plot.geometry(land).plot() and Plot.geometry(land).plot({projection: "identity"}) differ on the default size?

Fixed in the latest commit d93b812.

mbostock:
> (As I already mentioned), repeatedly having to type Plot.geometry was tripping me up. I hesitated with Plot.geometries (since we sometimes have one feature, or several features to draw). Having Plot.geo instead would be a no-brainer.

I’m still on the fence with this one. If we rename the mark to “geo”, does that mean the channel would need to be named “geo”, too, or would it stay “geometry”? I don’t find the singular name to be confusing because all marks in Plot have singular names.

> More projections. The default list of projections is a bit restricted. I'm not advocating to embed all of d3-geo-projections and d3-geo-polygon… but I'd like it to be easier to extend… maybe with some kind of plugin that would register projection names. (Maybe related to theming-extending color schemes https://github.com/observablehq/plot/issues/630.)

I’ll try to add all the ones in d3-geo, at least. I don’t think we should introduce a dependency on d3-geo-projection or d3-geo-polygon; folks can construct their own projection in that case.

> I like the idea of string accessors for properties, but I'm not sure how we would disambiguate between "Population" as an accessor for d => d.properties.Population and "id" for d => d.id.

I think we would just special-case `"id"` since it’s part of the GeoJSON specification. But I’m not sure we should do this and I’d prefer to keep it out of this first version.

> More graticule options. Add graticule (step) options in Plot.graticule.

I think people can just call d3.geoGraticule for now if they want more options, but yes, we could in the future do some convenience options for Plot.graticule.

mbostock:
Screen Shot 2022-11-25 at 8 55 48 PM

```js
Plot.plot({
  width: 960,
  height: 600,
  projection: "albers-usa",
  color: {
    scheme: "blues"
  },
  marks: [
    Plot.density(walmarts, {x: "longitude", y: "latitude", bandwidth: 12, fill: "density"}),
    Plot.dot(walmarts, {x: "longitude", y: "latitude", r: 1, fill: "currentColor"}),
    Plot.geometry(statemesh, {strokeOpacity: 0.3})
  ]
})
```

Fil:
What should we do with the marks that don't make sense or don't (yet) work with projections?

In the future we can imagine that this system will support a projection: "radial" which will transform bar charts into pie charts etc. But, in the meantime, shouldn't we throw an error?

I've checked all the marks and some already work perfectly with projections (Plot.image, yes!!). Others could be fixed relatively easily, but I'm not sure we should do it without a clear view of the consequences if we ever go "radial". Any mark that expects band scales doesn't work, and should throw an error.

| mark | remark | keep, fix, throw ? |
|---|---|---|
| image | - |  👍 |
| line | a line linearly connecting the projected coordinates | 👍 |
| vector | - |  👍 |
| delaunay & voronoi | planar voronoi (sigh…) | 👍 |
| density | - | 👍 |
| area | pixel coordinates | FIX? |
| areaX, areaY | pixel coordinates | FIX? |
| arrow | pixel coordinates | FIX? |
| link | pixel coordinates | FIX? |
| ruleX, ruleY |pixel coordinates | ❌ |
| barX, barY | needs band scale | ❌ |
| cell | needs band scales | ❌ |
| rect | fills the whole frame | ❌  |
| tickX, tickY | | ❌ |

README.md
```diff
@@ -311,6 +311,27 @@ Plot automatically generates axes for position scales. You can configure these a
...
+* *equirectangular* - the equirectangular, or *plate carrée*, projection
+* *orthographic* - the orthographic projection
+* *stereographic* - the stereographic projection
+* *mercator* - the Mercator projection
+* *equal-earth* - the [Equal Earth projection](https://en.wikipedia.org/wiki/Equal_Earth_projection) by Šavrič *et al.*
+* *natural-earth* - the [Natural Earth projection](https://en.wikipedia.org/wiki/Natural_Earth_projection) by Patterson
+* *albers-usa* - a composite Albers conic equal-area projection suitable for the United States
+* *identity* - the identity projection (for pre-projected geometry)
+
+In addition to the named projections above, the **projection** option may also be specified as a [D3 projection](https://github.com/d3/d3-geo/blob/main/README.md#projections), or any custom projection that implements [*projection*.stream](https://github.com/d3/d3-geo/blob/main/README.md#projection_stream).
```
Fil, line 327:
```suggestion
In addition to the named projections above, the **projection** option may also be specified as a [D3 projection](https://github.com/d3/d3-geo/blob/main/README.md#projections), or any custom projection that implements [*projection*.stream](https://github.com/d3/d3-geo/blob/main/README.md#projection_stream), or a function that receives a configuration object ({width, height, ...options}) and returns a projection.
```

README.md
```diff
@@ -311,6 +311,27 @@ Plot automatically generates axes for position scales. You can configure these a
...
+
+The top-level **projection** option applies a (often geographic) projection in place of *x* and *y* scales. It is typically used in conjunction with [geometry marks](#geometry) to produce a map, but can be used with any mark that supports *x* and *y*, such as [dot](#dot) and [text](#text). The following built-in named projections are supported:
+
+* *equirectangular* - the equirectangular, or *plate carrée*, projection
+* *orthographic* - the orthographic projection
+* *stereographic* - the stereographic projection
+* *mercator* - the Mercator projection
+* *equal-earth* - the [Equal Earth projection](https://en.wikipedia.org/wiki/Equal_Earth_projection) by Šavrič *et al.*
+* *natural-earth* - the [Natural Earth projection](https://en.wikipedia.org/wiki/Natural_Earth_projection) by Patterson
+* *albers-usa* - a composite Albers conic equal-area projection suitable for the United States
```
Fil, line 324:
```suggestion
* *natural-earth* - the [Natural Earth projection](https://en.wikipedia.org/wiki/Natural_Earth_projection) by Patterson
* *azimuthal-equal-area* - the azimuthal equal-area projection
* *azimuthal-equidistant* - the azimuthal equidistant projection
* *conic-conformal* - the conic conformal projection
* *conic-equal-area* - the conic equal-area projection
* *conic-equidistant* - the conic equidistant projection
* *gnomonic* - the gnomonic projection
* *transverse-mercator* - the transverse Mercator projection
* *albers* - the Albers’ equal-area conic projection
* *albers-usa* - a composite Albers conic equal-area projection suitable for the United States
```

TBH I'm not sure we should support (or document) all these projections in v1; for example, the generic conic projections are pretty useless unless we can parametrize the parallels.

Fil:
The list of outstanding questions is shrinking :)

- [ ] throw an error on unsupported marks
- [ ] geo or geometry? (I stand by _geo_: note that it accepts both geometries and features)
- [ ] the “property” accessor. I’m now sold on the idea of special casing “id” and defaulting all other (non-color…) string values to d => d.properties?[field]. But, it can wait.

README.md
mbostock: You can specify parallels like so:

```js
projection: {
  type: "conic-equal-area",
  parallels: [30, 60]
}
```

The supported options are shown here:

https://github.com/observablehq/plot/blob/4b561a4080e170254e13dbd92d616d3db8d7ddbf/src/projection.js#L95-L98

mbostock:
I implemented the error on marks requiring band scales in d0e41cb.

We could throw a similar error for marks that define channels bound to either the _x_ or _y_ scale, but don’t provide exactly the _x_ and _y_ channels.

mbostock:
Added another error for non-point marks in cf8476e.

mbostock:
So, the only thing left is whether we call it Plot.geo or Plot.geometry? 😅

Fil (approved):
👍

I'm with geo, but it's your call ultimately :)

mbostock:
Do you have an opinion on the channel name being “geo” or “geometry”?

Fil:
Most of the time we won't have to use it, so I don't mind if it's longer to type or look up. Having a geo mark and a geometry channel might help disambiguate?

mbostock:
What do you think of calling it “geojson” (for both the mark and the channel)? Is that better or worse than “geometry”? E.g.,

```js
Plot.geojson(statemesh, {strokeOpacity: 0.3})
```

Fil:
yes, geojson would be fine.

espinielli:
Amazing development!

Fil:
I love how https://github.com/observablehq/plot/pull/1111/commits/f63a8eb804f05961b5d519364d7b219825c200d9 skips invisible points (e.g. on the other side of the orthographic projection).

PR #1113, opened 2022-11-24
preserve dates for mean and median

### LINKED ISSUE #1112: The mean and median reducers convert dates to numbers
These should probably produce dates if the input is dates (like min and max do).

mbostock:
Fixes #1112.

### DIFF
--- a/src/transforms/group.js
+++ b/src/transforms/group.js
@@ -28,7 +28,8 @@ import {
   labelof,
   range,
   second,
-  percentile
+  percentile,
+  isTemporal
 } from "../options.js";
 import {basic} from "./basic.js";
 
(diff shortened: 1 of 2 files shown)


src/transforms/group.js
```diff
@@ -319,6 +320,15 @@ function reduceAccessor(f) {
...
 }
 
+function reduceMaybeTemporalAccessor(f) {
+  return {
+    reduce(I, X) {
+      const x = f(I, (i) => X[i]);
+      return isTemporal(X) ? new Date(x) : x;
+    }
+  };
+}
```
Fil, line 330: small optimization:
```suggestion
function reduceMaybeTemporalAccessor(f) {
  return {
    scope: "data",
    reduce(I, X, context) {
      if (arguments.length === 2) return isTemporal(X);
      const x = f(I, (i) => X[i]);
      return context ? new Date(x) : x;
    }
  };
}
```

if we do this, we'll also have to initialize the context in https://github.com/observablehq/plot/blob/21dc1c70f4de1ec2a63ae49fce6f04089bed3d5f/src/channel.js#L66

mbostock: I’d prefer to keep this simple for now.

Fil (approved):
Needs a test plot, but I believe that will be with one of the maps (#1111)?

mbostock:
I added a few unit tests.

PR #1114, opened 2022-11-25
geo fixes

Fil:
- r for pointRadius
- dimensions for facets (Note that this changes the test plots by 1px (due to margins being 0.5-offset).

(The two commits are not fully independent, but pointRadius can be cherry-picked if necessary.)

(We could also introduce a projection: {padding(Left|Right…)} option to allow a bit of breathing room between the frame and the map; the code is simple but I think this has to be postponed to a later phase with other projection settings.)


src/dimensions.js
```diff
@@ -18,16 +18,28 @@ export function Dimensions(
     margin,
     marginTop = margin !== undefined
       ? margin
-      : Math.max((xAxis === "top" ? 30 : 0) + facetMarginTop, yAxis || fyAxis ? 20 : 0.5 - offset),
+      : Math.max(
+          (projection == null && xAxis === "top" ? 30 : 0) + facetMarginTop,
```
mbostock, line 22: I think rather than checking `projection == null` here, we should ensure that xAxis is null if the projection is not null. I’ll look into this…

mbostock: Actually, that should already be the case, since in inferScaleType we prevent the _x_ and _y_ scale from being instantiated if there is a projection, and if there is no _x_ or _y_ scale, there is no corresponding axis. So I don’t think this code is doing anything and can be safely removed (and the tests pass, so…).

PR #1115, opened 2022-11-26
all marks can clip to sphere

Fil:
walmartsVoronoi


src/style.js
```diff
@@ -314,19 +315,35 @@ export function applyIndirectStyles(selection, mark, scales, dimensions) {
...
+      const {projection} = context;
+      if (projection == null) break;
+      const path = geoPath(projection);
+      const id = getClipId();
+      selection
+        .attr("clip-path", `url(#${id})`)
+        .append("clipPath")
+        .attr("id", id)
+        .append("path")
+        .attr("d", path({type: "Sphere"}));
```
mbostock, line 344: This makes me wonder if you could say `clip: {type: "Sphere"}` to pass in arbitrary geometry. But I like starting simple like this for now.

src/style.js
```diff
@@ -314,19 +315,35 @@ export function applyIndirectStyles(selection, mark, scales, dimensions) {
...
+        .append("rect")
+        .attr("x", marginLeft - (x?.bandwidth ? x.bandwidth() / 2 : 0))
+        .attr("y", marginTop - (y?.bandwidth ? y.bandwidth() / 2 : 0))
+        .attr("width", width - marginRight - marginLeft)
+        .attr("height", height - marginTop - marginBottom);
+      break;
+    }
+    case "sphere": {
+      const {projection} = context;
+      if (projection == null) break;
```
mbostock, line 336: I’ll change this to throw an error.

mbostock (approved):
Nice, I was just thinking we should do this! 👏

PR #1129, opened 2022-11-30
Remove margins from the custom projection API

### LINKED ISSUE #1124: Should the projection definition receive the margins?
Instead of passing {width, height, marginLeft, marginRight, marginTop, marginBottom}, shouldn't we pass 

({width: width-marginLeft-marginRight, height: height-marginTop-marginBottom}),

then somehow add marginLeft, marginTop to every mark? It would simplify the work for the user, and leave less room for error, so it would help ensure that the map's extent/viewport fits the frame.

Framing it it as a question, as I'm not even sure that this is possible without mutating the projection's translate, which would be a no-go.

Fil:
We don't want users to have to worry about the layout when they define a custom projection: sending the {with, height} of the frame is simpler. This means we have to wrap the projection stream with an identity/translate projection.

fixes #1124


mbostock:
There are also global top-level **inset** options which should apply to the projection, too (like they do for **x** and **y**); I will add those.

PR #1132, opened 2022-12-01
default geo clip

### LINKED ISSUE #1131: Plot.geo should default to clip: "frame"?
Suggestion by @Fil.

mbostock:
Fixes #1131.


mbostock:
Not sure I agree with this change now: it means with the default inset: 0, the default stroked Plot.sphere will get clipped by the frame (as shown in the tests fixed in 1bc1edb).

Fil:
Good point.

To state the obvious (as a reminder for future me), since the sphere's strokeWidth defaults to 1.5, it bleeds out of the frame by 0.75px. So, if we clip it to the frame, it is ugly; plus, the apparent strokeWidth is half of what the user expects. However if we don't clip, projections might bleed out of the frame—not a huge problem per se—, but also extend over to neighboring facets, which is much more annoying, and the source of this issue.

I'm playing with:
`    inset: globalInset = 1,`

and it seems ok, event with a pointy projection like geoBerhaus.

Capture d’écran 2022-12-01 à 13 18 16

Mercator is not _on_ the frame any more, but _inside_ it, which is probably OK too (since we wouldn't materialize the frame visually, only as a clipping path):
Capture d’écran 2022-12-01 à 13 19 53

Fil:
superseded by #1151

PR #1139, opened 2022-12-01
projection support for x1/y1 x2/y2 marks

### LINKED ISSUE #1126: Plot.link, arrow, rect projection support
Probably the only thing missing is to do X1, X2, Y1, Y2 in https://github.com/observablehq/plot/blob/695248578d442ad084220b50f9dc239f0edbd032/src/projection.js#L113 ?

(not urgent)

Fil:
The rect mark might indicate a bounding box, but also an area to brush, clip, where an inset is located, all that sort of things. We define a rect by its diagonal, as extending horizontally and vertically from the projection of <x1,y1> to the projection of <x2,y2>. Arrows and links likewise connect the two projected points—in other words, they do not follow the geodesic (or "sphere" curve—use Plot.geo if that is the goal)

rect arrow

closes #1126


mbostock (changes requested):
Minor nits on the changes to valueObject and applyProjection. Also, looks like you need to delete the now-unused projectionParty.svg output. (As fun as that name is!) If I have time I will try to help… but not sure if I’ll get to it today.

mbostock (approved):

Issue #1141, opened 2022-12-01
normalize-winding option?


Fil:
https://observablehq.com/@fil/rewind as an option
Otherwise people will often be tripped up, with no easy solution.

Fil:
Maybe the right approach would be to create a warning, when the fit-to-domain approach returns the sphere, that "some of the polygons in the domain appear to cover the whole world. If this is deliberate, add an explicit rewind: false; otherwise add rewind: true to fix the geometries"?

Fil:
I'm closing in on the solution with https://observablehq.com/@fil/rewind

As I see it, Plot’s _rewind: true_ option would augment the projection’s stream with a rewind transform and the simple=true approach.

Fil:
rewind: true seems wrong, since there are several possibilities out there (we just want to implement two of them, but should keep that door open).

So maybe the **rewind** option should be a keyword:
* null (default): no rewind, the strict and faster option
* "normalize": rewind polygons that need rewinding; the forgiving but slower option
* "per-rfc-7946", or "rewind all" <= we wouldn't support that initially

Polygons that need rewinding are only checked on polygon coordinates (in Polygon and MultiPolygon geometries).

Our current heuristic in https://observablehq.com/@fil/rewind is unambiguous for polygons which have a hole. For simple polygons, which only have an outer ring, there is a choice to make: for example to leave them alone, or to check whether it would cover more than 50% of the total area of the sphere (50% being an arbitrary threshold, which could even be an option ranging between 50% and 99%).

Closing for now, as it is easy enough to use the function and rewind the geometries before passing them to Plot. But if it comes up too often we'll see. It's not a lot of code.

PR #1142, opened 2022-12-02
Ignore polygonStart when computing the bounds of the projected domain

Fil:
**Context.** Plot follows D3 and TopoJSON’s convention for polygon winding order. A polygon that describes the outline of an island should be walked as south->west->north->east->south (SWNE).

However, users sometimes have files that follow the opposite direction. The same island will be walked in the SENW order. This might happen because the shape was drawn in a geojson editor which doesn't care about winding order, or because another piece of code follows the convention adopted by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946#section-3.1.6), which is the opposite of D3’s (see [d3-geo](https://github.com/d3/d3-geo/blob/main/README.md)). 

For Plot (and D3), this describes the exterior of the island, and not its interior. This PR doesn't change that (see #1141 for discussion of a rewind helper). 

**What this PR changes** is the "fit to domain" approach when the domain has the "wrong" winding order. When we fit the projection, we consider all polygons as outlines only instead of areas. As a consequence, the map is centered on our island, where previously it was fitting to the whole sphere. As a bonus, since we bypass polygon interpolation the fitSize operation is now faster.

The test case and example is a simplified geometry of the Hawaiian islands in the "wrong" winding order. Before, they would result in fit-to-sphere map; now the fit is correct. (However each island is still painted "out", if you have a fill option, which explains the weird colors.)

*before*
hawaii before

*after*
hawaii after


mbostock:
I don’t think we should change this. The before case is technically correct and it’s what we want for cases like ocean polygons. The after is still broken so you’d have to fix the winding order anyway.

Fil:
In defense of this PR:
- it's a recurring situation and it's hard for users to understand what's wrong and how to fix it
- for the ocean polygon I'd argue you'd use `domain: {type:"Sphere"}`
- the result is only broken if you try to fill the polygon; for stroke, it's fine.

If you're not convinced, I'll try to come up with a working (opt-in) solution for #1141. In either case it's not a blocker.

mbostock:
I like the idea of an option that controls this behavior (i.e., automatically reversing the winding order to ensure that polygons are less than one hemisphere in area), and that we should be consistent when rendering and computing the automatic domain.

Fil:
The approach outlined in https://observablehq.com/@fil/rewind seems better to address #1141.

PR #1144, opened 2022-12-02
document the bundle curve

Fil:
Unless there was a reason not to document it? Couldn't find nor remember one.


mbostock (approved):
I think I intentionally omitted it because I wasn’t sure there was a valid use. But it can’t hurt to document it if we support it. And it’s not like the _bundle_ curve is deprecated or anything.

Fil:
I've used it [here](https://observablehq.com/@recifs/a-trend-line-for-counts-over-time) when I wanted to draw "some kind of approximate trend" (at the bottom of the notebook).

PR #1150, opened 2022-12-05
projection.clip: sphere, frame, angle

### LINKED ISSUE #1131: Plot.geo should default to clip: "frame"?

### LINKED ISSUE #1137: Support clip-to-frame (clipRectangle) and clipAngle projection options
* clipAngle works on the input stream, and only for spherical data. Maybe limit this option to the named azimuthal projections.
* clipRectangle works on the output stream, after rescaling (since we'll want to clip to the frame), and before simplification (#1143). It could be applied to any projection (with a special case for the null projection?).

Fil:
Introduces two projection options:

### projection.**clipAngle**

a number, passed directly to the projection factory. This also forces fitting the projection to the frame with its restricted sphere path.

### projection.**clip**

same keywords as the mark’s clip option (frame, sphere and null), but interpreted with a different technique. ~~This option applies to all the paths generated by the geo marks (geo, graticule, and sphere). [EDIT: it's more complicated than that]~~

It defaults to "frame", which uses the [clipRectangle](https://github.com/d3/d3-geo/blob/main/src/clip/rectangle.js) post-clip stream; this approach two benefits:
1. the user doesn’t have to add an explicit clip:"frame" option to the geo marks to avoid bleeding in the margins and between facets
2. the computed paths do not extend (invisibly) as much, which seems to make the SVG lighter, and in particular consume less memory when a map zoomed to a large scale has a "heavy" geo mark than spans the whole globe

The second option is "sphere", which can be used to add a clip-path to all the geo marks of the plot (an exception is made for the sphere mark itself).

Use null to disable.

Since this changes almost all the test maps, I'm taking this as an opportunity to switch most of them from 50m to 110m geometries, in order to make the tests a bit faster, and the repo a bit smaller.

closes #1137
closes #1131
supersedes #1132


Fil:
This is trying to do too much. I'll prepare an alternative minimalist version.

Fil:
superseded by https://github.com/observablehq/plot/pull/1151

PR #1151, opened 2022-12-05
projection.clip option

### LINKED ISSUE #1131: Plot.geo should default to clip: "frame"?

### LINKED ISSUE #1137: Support clip-to-frame (clipRectangle) and clipAngle projection options

Fil:
Introduces two projection clipping methods:

### New option: projection.**clipAngle**

a number, passed directly to the projection factory. This also forces fitting the projection to the frame with its restricted sphere path.

### Always clip-to-frame

Systematically applies the [clipRectangle](https://github.com/d3/d3-geo/blob/main/src/clip/rectangle.js) post-clip stream transform. Benefits:
1. the user doesn’t have to add an explicit clip:"frame" option to the geo marks to avoid bleeding in the margins and between facets
2. the computed paths do not extend (invisibly) as much, which seems to make the SVG lighter, and in particular consume less memory when a map zoomed to a large scale has a "heavy" geo mark than spans the whole globe

Note that this is a lighter, less configurable version of #1150:
- the projection is always clipped to the frame’s extent.
- not trying to add a clip-to-sphere projection option.

The only drawback I've seen so far is that you can't have a "bleed edges" projection, where the interpolation of a polygon along the sphere is hidden by the frame. However, this can be approximated with negative margins, or a white stroke Plot.frame if you need facets (see examples in tests).

Since this changes almost all the test maps, I'm taking this as an opportunity to switch most of them from 50m to 110m geometries, in order to make the tests a bit faster, and the repo a bit smaller.

closes #1137
closes #1131
supersedes #1132 and #1150


mbostock:
> It can definitely be an option on the projection and it also makes total sense to use projection clipping rather than svg clipping when it is set as an option on the mark.

If it’s a mark option, it should be two _different_ options, though (or two different values of the same _mark_.**clip** option). Meaning I don’t think we should change the existing _mark.clip `"frame"` option to behave differently when you have a projection, since the result is not the same. (SVG clipping affects strokes and dot symbols that extend outside the frame, but projection clipping does not.)

Fil:
I want it on by default: overlapping facets just look too bad. I'll give it another shot.

Fil:
lgtm!

mbostock (approved):

PR #1153, opened 2022-12-05
projection.domain accepts an iterable

### LINKED ISSUE #1148: Accept an iterable as the projection domain
Since _data_ can be provided as an iterable, it would make sense to accept this also when setting the domain.

Fil:
closes #1148


mbostock (commented):
I would prefer to leave this out to start. Every shorthand we add is ambiguity and complexity to maintain, and it doesn’t feel that hard to create a FeatureCollection or GeometryCollection by hand as needed. Maybe if we see evidence of people trying this and getting frustrated that it doesn’t work (other than you I mean 😅) we go back and add it.

PR #1154, opened 2022-12-05
support shorthand projection: {domain: "sphere"}

Fil:
As a complement to #1153 (suggested in #1148). Might come in handy, but I'm not sure we want too many special cases.


mbostock (changes requested):
Rather than first looking at `domain.type`, we should check whether `typeof domain === "string"` and then treat it as a keyword, using `maybeKeyword` so that you get a nice error message if you typo `"sphere"`.

PR #1156, opened 2022-12-05
curve: "projection"

### LINKED ISSUE #1146: variable color for geo lines?
(Suggested by @mbostock)

either by having a variable color for LineString (seems hard), or an option for Plot.line (or maybe a curve?) that follows spherical line interpolation.

> Plot.line(beagle.coordinates, {stroke: (d, i) => i, z: null, strokeWidth: 2})
untitled (71)

Fil:
closes #1146

beagle


mbostock (changes requested):
I think we can avoid creating an entirely new Mark class for this. Basically, I would check whether the `curve` is `"geodesic"` in the constructor, and if so, don’t bind the _x_ and _y_ channels to their respective scales. Then use the projection and geoPath instead of the line shape during render.

In addition to being less code, this should be more maintainable; less likely that the two implementations could drift. And of course you could test `instanceof Line` and you’d get the expected result. (It’s highly surprising if calling Plot.line _doesn’t_ return an instance of Plot.Line.)

I think I would also prefer the simpler name `"sphere"`. I know that `"geodesic"` is the mathematically-precise term, but wouldn’t `"sphere"` be easier to remember? And I don’t think we’ll have multiple forms of spherical interpolation.

src/marks/line.js
```diff
@@ -26,25 +26,49 @@ const defaults = {
 export class Line extends Mark {
   constructor(data, options = {}) {
     const {x, y, z, curve, tension} = options;
+
+    // When the line mark is used with a "sphere" curve and a projection
+    // (supposedly spherical), keep x and y as numbers.
+    const sphere = curve === "sphere";
```
mbostock, line 32: You’ll need to coerce `curve` to a string and toLowerCase etc.

src/marks/line.js
```diff
@@ -26,25 +26,49 @@ const defaults = {
...
+    let shape;
+    if (this.sphere) {
+      const {projection} = context;
+      if (projection === undefined) throw new Error("the sphere curve requires a projection");
+      const path = geoPath(projection);
+      shape = (I) =>
+        path({
+          type: "LineString",
+          coordinates: Array.from(I, (i) => [+X[i], +Y[i]]).filter(([x, y]) => !isNaN(x + y))
+        });
```
mbostock, line 63: We could do this faster by invoking the projection stream directly (calling _stream_.lineStart, _stream_.point, _stream_.lineEnd) rather than going through geoPath.

Also, this will coerce nulls to zero rather than skipping them. We might want to use coerceNumbers here, but we should at least be consistent.

mbostock (changes requested):

src/marks/line.js
mbostock: groupIndex already does filtering for undefined values, so we can check if `i` is -1 here instead. I’ll push some commits.

mbostock (approved):

Fil:
The use case for fc75219c22430d0bce405f7af5f9f00c6ac3cfcf were the x/y coordinates passed as strings in the [beagle.js example](https://github.com/observablehq/plot/blob/d70fd8bc90753f1992539026384530d3bdcb7b8f/test/plots/beagle.js#L9). Now solved by coerceNumbers in sphereLine.

src/marks/line.js
```diff
@@ -23,28 +24,40 @@ const defaults = {
...
-    const {x, y, z, curve, tension} = options;
+    const {x, y, z} = options;
+    const curve = LineCurve(options);
     super(
       data,
       {
-        x: {value: x, scale: "x"},
-        y: {value: y, scale: "y"},
+        x: {value: x, scale: curve === curveProjection ? undefined : "x"}, // unscaled if projected
+        y: {value: y, scale: curve === curveProjection ? undefined : "y"}, // unscaled if projected
```
Fil, line 45: See https://github.com/observablehq/plot/issues/1165

PR #1158, opened 2022-12-05
the dot or geo mark is empty if given a constant negative r

### LINKED ISSUE #1145: Tiny inconsistency with negative radius
With Plot.geo, if you have a variable channel for _r_, geometries with a negative radius are filtered out. This consistent with Plot.dot, and while it might be suprising if the geometry isn't of a Point or MultiPoint type, it makes sense (for example, the geometry might be a FeatureCollection that contains Point geometries…).

However if you have a constant negative value for r (typically, r: -1), Plot.dot will discard all dots, but Plot.geo will not discard the geometries, and will call pointRadius(-1).

Fil:
0 is accepted, as it doesn't result in an error for dots and images, and may be useful for geo marks.

fixes #1145

### DIFF
--- a/src/marks/geo.js
+++ b/src/marks/geo.js
@@ -1,6 +1,6 @@
 import {geoGraticule10, geoPath} from "d3";
 import {create} from "../context.js";
-import {positive} from "../defined.js";
+import {negative, positive} from "../defined.js";
 import {identity, maybeNumberChannel} from "../options.js";
 import {Mark} from "../plot.js";
 import {applyChannelStyles, applyDirectStyles, applyIndirectStyles, applyTransform} from "../style.js";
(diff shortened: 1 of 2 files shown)


src/marks/geo.js
```diff
@@ -34,6 +34,7 @@ export class Geo extends Mark {
     const {geometry: G, r: R} = channels;
     const path = geoPath(context.projection);
     const {r} = this;
+    if (this.r != null && negative(this.r)) return;
```
mbostock, line 37: This is equivalent:

```js
    if (negative(r)) return;
```

but per the top comment, I think we want something like this:

```js
    if (negative(r)) index = [];
```

mbostock (changes requested):
I think this should have the same behavior as if the channel returned entirely negative values, i.e., when `index` is the empty array. This should produce an empty G element for the accessibility tree etc.

mbostock (approved):
A small test would be nice, but not required. 🙏

PR #1160, opened 2022-12-05
rename null projection to "identity"

### LINKED ISSUE #1155: inconsistent definition of identity between scales and projections
For regular scales (e.g. x, y, color), "identity" is meant to opt out of any scaling and pass raw values. For projections, this is what _null_ does, and identity means a planar-rescalable projection.

Suggestion to rename the _null_ projection to "identity", and the identity projection to "planar". Another idea would be "planar-up" (for the current "reflect-y") and "planar-down" (for the current "identity").

Fil:
the null projection becomes "identity", and the identity projection becomes "planar"

we still have to check if the returned value is null after each step, so we accept undefined/null as a synonym of the identity projection.

closes #1155


src/projection.js
```diff
@@ -98,11 +98,11 @@ export function Projection(
 }
 
 export function hasProjection({projection} = {}) {
-  if (projection == null) return false;
+  if (isIdentityProjection(projection)) return false;
   if (typeof projection.stream === "function") return true; // d3 projection
   if (isObject(projection)) ({type: projection} = projection);
   if (typeof projection !== "function") projection = namedProjection(projection);
-  return projection != null;
+  return !isIdentityProjection(projection);
```
mbostock, line 105: This doesn’t work because here `projection` will be `identity` (the symbol) rather than `"identity"` (the string). You can test with `projection: {type: "identity"}`.

src/projection.js
```diff
@@ -199,3 +199,7 @@ export function applyProjection(values, projection) {
     stream.point(x[i], y[i]);
   }
 }
+
+function isIdentityProjection(projection) {
+  return projection == null || projection === "identity";
```
mbostock, line 204: Either

```suggestion
  return projection == null || `${projection}`.toLowerCase() === "identity";
```

Or

```suggestion
  return projection == null || /^identity$/i.test(projection);
```

mbostock (changes requested):

mbostock:
Superseded by #1161.

PR #1161, opened 2022-12-06
rename null to “identity” and “identity” to “planar”

### LINKED ISSUE #1155: inconsistent definition of identity between scales and projections

mbostock:
Supersedes #1160. Same tests. Only a trivial change is needed to src/projection.js since we already have different code paths for null vs. the identity projection.

Fixes #1155.


mbostock:
I’m not sure “planar” is the best name here, since [planar projections](https://en.wikipedia.org/wiki/Planar_projection) are already a thing. This projection only supports translate and scale. It might be better described as a [rigid body transformation](https://en.wikipedia.org/wiki/Rigid_transformation) or a (special case of) [affine transformation](https://en.wikipedia.org/wiki/Affine_transformation).

mbostock:
Or maybe `projection: "transform"`?

Fil:
transform sounds very generic; maybe just "plane"?

mbostock:
I also thought about `projection: "fit"` since the only reason to use this over `projection: null` would be to fit to a given domain. But I think I’m leaning towards either `projection: "affine"` or leaving it as-is.

Fil:
I like "fit" a lot!! I also like "plane" and "planar", but not so much "affine" as it would over-promise.

mbostock:
I don’t feel confident enough to change things here, so we’re going to leave it as-is. I think it’ll be fine.

PR #1162, opened 2022-12-06
projection autoheight

### LINKED ISSUE #1136: autoHeight + projections

Fil:
closes #1136


Fil:
it works! Can uses the projection's ratio (e.g. 1 for orthographic, 0.5 for equirectangular); measures the path if a domain is given. Defaults to the golden ratio for any other projection.

mbostock (commented):
This extends the public interface of what we’re calling a projection to include _projection_.ratio (rather than just the existing _projection_.stream). Do we intend to support user-supplied projections with auto-height by having them supply _projection_.ratio? Or do we only want to support this for the built-in projections? I think I probably wouldn’t use this mechanism as the public interface… especially since we could potentially infer the ratio by projecting the outline via *projection*.stream and d3.geoPath.bounds. But I also think it’s not a high priority to support this for custom projections. Given the time pressure I’d prefer to only do this for our internal projections to start.

Fil:
I've tried to hide the .ratio property in #1163, with a Symbol. (In the future we might consider opening it, but adding domain: sphere should be fine in most (all?) cases. Just a weeny bit slower.)

mbostock:
I need to think about this one more.

Fil:
I made the same mistake to have autoHeight depend on the actual user-given margins and width, when it should depend on the defaults (otherwise changing the width will also change the default height).

If we want to make everything fit exactly with user specified dimensions (and domain), this starts to look much more like the dataAspectRatio option (discussed in #837). It could be done as a follow-up in a comprehensive manner, maybe as an opt-in `height: "fit"` option.

In the meantime I'll put up a much simpler version of this PR.

PR #1166, opened 2022-12-07
simpler projection autoHeight

### LINKED ISSUE #1136: autoHeight + projections

Fil:
the height for projections (including the null projection when Plot.geo is used) is computed for a target frame with an aspect ratio that defaults to the golden ratio. When using facets, the computation takes into account the number of rows and columns, with similar limits to what total size is acceptable (1260px).

The projection's preferred aspect ratio is adapted for a few named projections. For example "equal-earth" and "equirectangular" are wider than tall, "mercator" and a few azimuthal projections default to a square.

closes #1136
supersedes #1162


src/dimensions.js
```diff
@@ -74,8 +80,19 @@ export function Dimensions(scales, geometry, axes, options = {}) {
...
+  const ny = y ? (isOrdinalScale(y) ? y.scale.domain().length : Math.max(7, 17 / nfy)) : 1;
+  return !!(y || fy) * Math.max(1, Math.min(60, ny * nfy)) * 20 + !!fx * 30 + 60;
+}
+
+function autoHeightProjection({fy, fx}, ratio, width, marginY) {
+  if (ratio > 0) {
+    return Math.round(
+      Math.max(
+        140,
+        Math.min(1260, (width * (fx ? fx.scale.bandwidth() : 1) * ratio) / (fy ? fy.scale.bandwidth() : 1))
```
mbostock, line 94: I think this depends on us not having set the (automatic) range of the fx/fy scales here. It’d be better to not depend on the order of operations and instead look at the scale’s padding.

src/projection.js
```diff
@@ -104,51 +108,40 @@ export function Projection(
...
-}
-
-const pi = Math.PI;
-const tau = 2 * pi;
-
 function namedProjection(projection) {
   switch (`${projection}`.toLowerCase()) {
     case "albers-usa":
-      return scaleProjection(geoAlbersUsa, 0.7463, 0.4673);
+      return {projection: scaleProjection(geoAlbersUsa, 0.7463, 0.4673), ratio: 610 / 975};
```
Fil, line 114: refactor to compute the ratio from the x and y scalings

src/dimensions.js
Fil: fixed

src/dimensions.js
```diff
@@ -74,8 +80,24 @@ export function Dimensions(scales, geometry, axes, options = {}) {
...
+}
+
+function autoHeightProjection({fy, fx}, ratio, width, marginY) {
+  if (ratio > 0) {
+    return Math.round(
+      Math.max(
+        140,
+        Math.min(
+          1260,
+          ((fy ? 1.1 * fy.scale.domain().length - 0.1 : 1) / (fx ? 1.1 * fx.scale.domain().length - 0.1 : 1)) *
```
mbostock, line 96: Is this not dependent on the padding? (I wonder if the 0.1 here corresponds to the default padding.)

Fil:
Yes, it works for me if you replace 640 by width.

(This will be different from the usual autoheight which is independent of width, but there's obviously a good reason to do it it, since the two directions are linked.)

src/dimensions.js
```diff
@@ -80,24 +79,23 @@ export function Dimensions(scales, geometry, axes, options = {}) {
...
-        140,
-        Math.min(
-          1260,
-          ((fy ? 1.1 * fy.scale.domain().length - 0.1 : 1) / (fx ? 1.1 * fx.scale.domain().length - 0.1 : 1)) *
-            width *
-            ratio
-        )
-      ) + marginY
-    );
+  // If a projection is specified, use it’s natural aspect ratio (if known).
```
Fil, line 90:
```suggestion
  // If a projection is specified, use its natural aspect ratio (if known).
```

src/projection.js
```diff
@@ -232,3 +231,15 @@ function applyProjection(cx, cy, values, projection) {
...
+// When a named projection is specified, we can use its natural aspect ratio to
+// determine a good value for the projection’s height based on the desired
+// width. When we don’t have a way to know, the golden ratio is our best guess.
+export function projectionAspectRatio(projection, geometry) {
+  if (isObject(projection)) projection = projection.type;
+  if (typeof projection === "string") {
+    const {aspectRatio} = namedProjection(projection);
+    if (aspectRatio) return aspectRatio;
+  }
+  if (geometry) return golden - 1;
```
mbostock, line 244: I think this isn’t quite right… we also want to use `golden - 1` if a projection has been specified even if there is no geometry (as when using a custom projection).

Fil: good catch; should be 
```suggestion
  if (projection !== undefined || geometry) return golden - 1;
```

(since the null projection is a projection)

src/projection.js
```diff
@@ -232,3 +231,17 @@ function applyProjection(cx, cy, values, projection) {
     stream.point(x[i], y[i]);
   }
 }
+
+// When a named projection is specified, we can use its natural aspect ratio to
+// determine a good value for the projection’s height based on the desired
+// width. When we don’t have a way to know, the golden ratio is our best guess.
+export function projectionAspectRatio(projection, geometry) {
+  if (isObject(projection)) projection = projection.type;
```
mbostock, line 239: This is still not quite right because in Projection(…) we check typeof projection.stream. We need to make sure that this code interprets the projection option in _exactly_ the same way so that we don’t have differing definitions on whether a projection has been specified.

(Although, some difference is unavoidable because it’s possible for the projection initializer to be a function that then returns a null projection; we would still consider that a non-null projection for the purposes of automatic height. But that’s such an edge case I’m not worried about it.)

mbostock (approved):

Fil:
🎉 that was the last one!

PR #1177, opened 2022-12-09
apply fx/fy transform option

### LINKED ISSUE #1176: fx/fy scale transform option is not called
See example in https://observablehq.com/@observablehq/testing-facets-for-plot-1085

Fil:
closes #1176

Notes:
* is there a way to make sure transform is only called for the fx fy scale?
* I spent more time looking up the non-standard country codes from the olympians dataset than fixing the bug.

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -69,12 +69,12 @@ export function plot(options = {}) {
       if (facetData == null) throw new Error("missing facet data");
       facetChannels = {};
       if (x != null) {
-        const fx = Channel(facetData, {value: x, scale: "fx"});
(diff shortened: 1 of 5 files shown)


src/plot.js
```diff
@@ -69,12 +69,12 @@ export function plot(options = {}) {
       if (facetData == null) throw new Error("missing facet data");
       facetChannels = {};
       if (x != null) {
-        const fx = Channel(facetData, {value: x, scale: "fx"});
+        const fx = Channel(facetData, {value: x, scale: "fx", transform: options.fx?.transform});
```
mbostock, line 72: We should use applyScaleTransforms to do this. This isn’t doing exactly the same thing (e.g., the percent option).

Issue #1122, opened 2022-11-29
heatmap mark?


Fil:
I sometimes see users trying to make heatmaps with Plot.cell, which results in incorrect charts when x and y are continuous. The recommended approach (until now) was to use Plot.bin to create the small squares (or rather, rectangles), as x/y bins. However with projections, this is now incomplete, and we should have a solution for heatmaps in the projected space. Like hexbin, but based on squares (or any other tessellation?). There's probably a lot of code to share with hexbin, the most difficult part to decide might be the API?

mbostock:
I would like to better understand how continuous heatmaps are commonly expressed. Do we want a sparse sample-based approach as with the pixel mark in #1185 (which is really just an optimization of the existing rect mark + interval transform)? Or does the user want to supply a continuous function of *xy* that is them sampled automatically (in a grid) by Plot? Or maybe the user supplies dense samples in a grid (similar to ImageData) so that the *x* and *y* channels can be implicit? Or all of the above?

Fil:
matplotlib has a funky interpolation method
https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html

mbostock:
Fixed in #1196.

Issue #984, opened 2022-07-13
A raster/2d function plot mark


Fil:
https://observablehq.com/@observablehq/plot-heatmap-984

see comment in https://observablehq.com/@observablehq/javascript-replacements-for-python-data-science-tools#comment-7873045d922f909f

mbostock:
I like it. I think we should use foreignObject instead of toDataURL though (even if it currently breaks downloads on Observable; we could add a fix for that to downloading if we want to).

I wonder if using fillRect would be faster or slower than putImageData.

I think we’ll want a different option than **resolution**, but I’m not sure exactly what… For the current functionality I would rename the option to **devicePixelRatio** (or maybe **pixelRatio**) but I wonder if it should feel more like Plot.density where there’s an underlying grid of some resolution (that’s not necessarily 1:1 with pixels). But maybe starting with **pixelRatio** is fine and I’m overthinking.

yurivish:
A while ago while playing around with my own density mark (with liberal help from Fil) I ran into an Safari bug where the foreignObject would not scale properly with the SVG. 

You can see an example [here](https://observablehq.com/d/b357813c9ec98eb3) – open it in Safari and increase the font size. Performance should be better than toDataURL, but I never figured out a workaround for the bug.

Fil:
matplotlib has a funky interpolation method
https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html

mbostock:
Fixed in #1196.

PR #1185, opened 2022-12-14
pixel mark

mbostock:
This…

* Adds a pixel mark for generating raster images from _xy_-samples with _fill_ (and optionally opacity).
* Adds the imageRendering property to the image mark (and pixel mark).
* Fixes an unfiled bug with boolean color scales that contain nullish values in the domain.
* Avoids computing the strokeWidth channel for marks that don’t support stroke.

I still need to write documentation, but I think this is functionally complete. I’m not 100% sure about the implementation, since adding and subtracting the default 0.5 can introduce tiny rounding errors (that go away when rounding during rendering, but still it would be nice to avoid them). I slightly wonder if it would be better to instead implement this as a canvas-based rect mark? Most other marks might want canvas-based rendering too… and in theory we Plot could even stack canvas and SVG elements rather than using svg:image (though that would break downloading).

Related #1122 #984.


Fil:
We could probably learn a trick or two from @yurivish's unbelievably fast https://observablehq.com/@twitter/zoomable-density-scatterplot Writing the colors directly in the canvas data would probably be faster than using fillRect, since we don't need (nor want) anti-aliasing.

Fil:
I often want a discrete quantitative scale, where I could draw a bar for a given unit, and it would have a width of 1 unit (minus insets).

For example, when drawing a bar chart for yearly data, I would prefer to write: `x: {type:"linear", unit: 1}` and use Plot.barY, than having to derive x1 and x2 with Plot.map and using Plot.rectY.

This option is slightly different from _interval: 1_, in that the bars would be ordinal-discrete, and centered on the unit, instead of covering [0, 1] we would cover [-0.5, 0.5]. And the scale would have a bandwidth. But (thinking out loud) maybe it is something we can do as an extension of interval: 1) an option to center things, and 2) add a bandwidth on the (continuous) scale.

test/plots/checkerboard.js
```diff
@@ -0,0 +1,20 @@
...
+  const n = Math.ceil((400 - 50) / k);
+  return Plot.plot({
+    marks: [
+      Plot.pixel(
+        {length: n * m},
+        {
+          pixelRatio: 1 / k,
+          x: (d, i) => i % m,
+          y: (d, i) => Math.floor(i / m),
+          fill: (d, i) => ((i % m & 1) === (Math.floor(i / m) & 1) ? true : i % m & 1 ? null : undefined)
```
Fil, line 15: I don't understand the ` i % m & 1 ? null :` part?

mbostock: I wanted to test that both null and undefined product the same output (transparent).

Fil:
I think we also need to support mixBlendMode internally (with context.globalCompositeOperation), for when we have several pixels at the same location?

mbostock:
Okay, I added mixBlendMode support, but it’s pretty much unusably slow presumably because it’s not hardware optimized.

I also added opacity support (as a fallback for fillOpacity, since both map to globalAlpha).

Fil:
I've documented the pixel mark. Many of the common options do not work, but that's expected.

mbostock:
Closing in favor of #1196.

PR #1189, opened 2022-12-16
spike vector shape + centroid transforms

### LINKED ISSUE #1178: Spike map
Some way of generating this type of map would be nice:

https://observablehq.com/@d3/spike-map

mbostock:
image

```js
Plot.plot({
  width: 960,
  height: 600,
  projection: "albers-usa",
  length: {
    range: [0, 200]
  },
  marks: [
    Plot.geo(nation, {fill: "#e0e0e0"}),
    Plot.geo(statemesh, {stroke: "white"}),
    Plot.spike(counties.features, Plot.geoCentroid({stroke: "red", length: (d) => population.get(d.id)}))
  ]
})
```

TODO

- [x] Support the **rotate** option?
- [x] Support the **anchor** option?
- [x] Plot.spikeX and Plot.spikeY?

Fixes #1178. (The diffs to the tests are just the re-ordering of the aria-label and fill attributes.)

An alternate approach here would be to expose a “shape” option similar to how the dot mark allows you to specify a custom symbol. (The difference being that vector uses length and dot uses area, so you don’t have the confusion around the *r* sqrt scale.) That’d probably be more flexible and we could still do that in the future, but I thought having a builtin in spike mark would be nice—especially with the default fill and stroke.


Fil:
We need similar Plot.centroid for planar geometries (and when in doubt about you geometries, as a best guess).

A custom symbol would be nice, and hopefully shared with the other symbols (those associated with dot). This means two changes:
1. we should rotate them with a svg transform, or alternatively make d3.path smart about rotate and translate.
2. we need to pass a few configuration options (_k_ for anchor, and _width_ for the base width, and all the arrowHead stuff if we make the vector as configurable as Plot.arrow). The current API for D3's _symbol_ only has _path, size_; we can probably add an optional {options} as a third parameter?

mbostock:
Okay, I have now:

* Added a **shape** option to vector for arbitrary shapes (similar to dot’s **symbol**, but not a channel)
* Added a *spike* shape, along with vector’s default *arrow* shape
* Added a new centroid transform for planar centroids, along with the previous geoCentroid transform
* Folded the spike mark into the existing vector mark; it is now only a convenience constructor
* Added a nifty internal `template` helper for generating dynamic SVG transform attributes

I expect people will incorrectly apply the Plot.centroid transform to spherical geometry when they should be using the Plot.geoCentroid transform. I’m not sure what to do about that; I don’t think it’s possible for the centroid transform to detect that a projection is in use and throw an error or warning. And even if a projection is in use, we don’t know whether it’s a geographic/spherical projection or a planar transform anyway. But I think this is fine for now, and the difference is often imperceptible anyway, as in the spike map example.

Please take another look?

Fil:
A slight difficulty when documenting, between width and radius. Could we settle on defining just one of them?

mbostock:
> A slight difficulty when documenting, between width and radius.

You’re referring to the fact that there’s a **width** option for the vector mark, whereas a **radius** is passed to the vector mark’s custom shape? My thinking is that for the user, it’s more intuitive to specify a width (_e.g._, of the base of a spike) because that’s what you see; whereas for the shape implementor, it’s more intuitive to receive the radius (the offset from the anchor to the respect corners of the base) because that’s what you need to construct the geometry.

Fil:
Yes. I understand the intent, but it means that the documentation for the custom shape object needs to explain that radius is half the width, which seems a bit verbose.

mbostock:
Would you prefer we use **radius** everywhere? Or **r** to match dot (even though it won’t support a channel)?

Fil:
Any of these work equally. Maybe using _r_ is indeed better since it's a bit more consistent with the rest.

README.md
```diff
@@ -1774,6 +1774,16 @@ Plot.vector(wind, {x: "longitude", y: "latitude", length: "speed", rotate: "dire
...
 
+The **shape** option supports the following shapes:
+- **arrow** - default: an arrow pointing top, with an arrow head proportional to its length and to the radius
+- **spike** - an open isosceles triangle pointing top, with a base equal to the width and a height equal to the length 
+- an object with a draw function that receives the context, a length, and the radius
+
+The following options are also supported:
+- **length** - the shape’s length, a constant or variable channel
+- **width** - the shape’s width, twice the shape’s radius, that defaults to 7 pixels.
+- **anchor** - the mark’s anchor (start, middle, end), defaults to middle
```
mbostock, line 1785: The **length** and **anchor** options are already documented above, and that’s where the **shape** option should be documented too. I’ll fix in the next commit.

Fil (approved):
with the additional geometry option it looks like it's good to go

Fil:
TIL! but we need to test centroid too :)

mbostock:
Added test to compare spherical centroids (red) and projected planar centroids (blue).

Screenshot 2022-12-20 at 10 42 20 AM

mbostock:
Also for reference here is the equirectangular centroid (green):

Screenshot 2022-12-20 at 10 48 05 AM

I guess it’s okay to lose this. Even though it’s “fast”, it’s _only_ right if you’re using the equirectangular projection. Whereas the other two at least have a valid semantic meaning (either the center on the surface of the sphere, or the center on the projected plane).

Issue #1143, opened 2022-12-02
projection fixed-precision option?


Fil:
I'm not sure how we could name it, but it would be useful to add an optional quantization/line simplification step at the end of the projection stream pipeline. For instance, the test outputs could be drastically reduced in size. And any SSR map rendered with plot would download faster.

mbostock:
There’s an ancient PR in d3-path for this, I believe.

Fil:
Yes, https://github.com/d3/d3-path/pull/12. It's for generic paths, and probably more complicated than we'd want here (we know that we are in screen coordinates and are the last transform in the stream pipeline).

mbostock:
> probably more complicated than we'd want here

How is it complicated? I feel like that’s a pretty good existing solution and it’s nice that it would also work for pre-projected and planar geometry. I guess you’re saying we could do it as a point transform and abuse the interface by passing stream.point strings instead of numbers? But I think it would be best to fix this generically in d3-path since other people have wanted this for a while.

Fil:
I meant more complicated in the sense that it also has to process arcs. But I also prefer if we can land it.

mbostock:
I have updated https://github.com/d3/d3-path/pull/12 and it should be ready to go now. Once that’s released as D3 7.8, we can adopt that here and switch to d3.pathFixed instead of d3.path.

PR #1190, opened 2022-12-16
the projection’s precision also quantizes the output coordinates

Fil:
Not a fancy algorithm, just a toFixed transform before outputing the coordinates to the svg path. On a line, we remove consecutive duplicate points. The size of the test maps is shrunk by about half, which will also make them cheaper to download and share.
I didn't see a point in making a quantize precision option separated from the line sampling precision option—I don't think we would want rough lines with lots of digits, or precise lines with integer coordinates?


Fil:
closes #1143

superseded by https://github.com/d3/d3-path/pull/12

PR #1196, opened 2022-12-21
raster mark

mbostock:
This is an alternative to the pixel mark #1185; the raster mark takes a set of discrete {_x_, _y_, _fill_} samples and produces a corresponding raster grid (image) using putImageData. The canvas is then converted to a URL for use as an svg:image, which is positioned using {_x1_, _y1_, _x2_, _y2_} abstract coordinates likewise bound to the _x_ and _y_ scales.


mbostock:
I’m not sure why CI is failing; presumably it’s a non-visible difference in the result of _canvas_.toDataURL. 😞

mbostock:
Okay, here’s my third attempt… I hope you like it better. 🙏 It feels very convenient for the continuous f(x, y) case, and supports all (continuous) scale types. And it’s still convenient for the simple case of rendering an existing raster grid (e.g., volcano) with linear scales. I would like to implement a contour mark with the same type of x/y/value specification… but I’m not sure how much work I’ll do over Christmas.

test/plots/heatmap.js
```diff
@@ -0,0 +1,76 @@
...
+  const width = Math.round(580 / pixelRatio);
+  const height = Math.round(350 / pixelRatio);
+  const x = (f => i => f.invert(i % width + 0.5))(d3.scaleLinear([x1, x2], [0, width])); // prettier-ignore
+  const y = (f => i => f.invert(Math.floor(i / width) + 0.5))(d3.scaleLinear([y2, y1], [height, 0])); // prettier-ignore
+  return Plot.plot({
+    color: {
+      type: "diverging"
+    },
+    marks: [
+      Plot.raster(d3.range(width * height), {
```
Fil, line 38: would this be easier if the signature for fill was `f({x,y}, i)`?

mbostock: I don’t think this test needs to be “easy”—the easy form is shown above, and this test isn’t supposed to be representative of recommended usage. The point here is to test the explicit form where _fill_ is a function of data.

Fil (changes requested):
I really like where this is going. As mentioned in the comments there's more we could do (in terms of performance, compatibility with opacity, faceting, projections…) but they could be done in follow-ups.

Note that Safari currently doesn't support image-rendering: pixelated on svg images which makes it a bit awkward for us. Should we expect Webkit to fix itself, or look for a workaround (foreignObject+img)?

I'm adding a bit of documentation. I must say that I'm still struggling to understand/explain how the **x** and **y** channels work in conjunction with width, x1 and x2 (?). I've been trying as an exercise to create the image of the volcano with x, y, fill as channels, but not very successfully (seems like I need to specify all of x1, x2, width…?).

mbostock:
> I've been trying as an exercise to create the image of the volcano with x, y, fill as channels, but not very successfully (seems like I need to specify all of x1, x2, width…?).

Yes, you must specify all those options. Like this:

Screenshot 2023-01-02 at 11 45 32 AM

```js
Plot.plot({
  marks: [
    Plot.raster(volcano.values, {
      width: volcano.width,
      height: volcano.height,
      x1: 0,
      y1: 0,
      x2: volcano.width,
      y2: volcano.height,
      x: (_, i) => (i % volcano.width) + 0.5,
      y: (_, i) => Math.floor(i / volcano.width) + 0.5,
      fill: (d) => d
    }),
    Plot.frame()
  ]
})
```

Alternatively if you want the samples at integer locations, you have to offset the bounds by 0.5:

Screenshot 2023-01-02 at 11 45 13 AM

```js
Plot.plot({
  marks: [
    Plot.raster(volcano.values, {
      width: volcano.width,
      height: volcano.height,
      x1: -0.5,
      y1: -0.5,
      x2: volcano.width - 0.5,
      y2: volcano.height - 0.5,
      x: (_, i) => (i % volcano.width),
      y: (_, i) => Math.floor(i / volcano.width),
      fill: (d) => d
    }),
    Plot.frame()
  ]
})
```

The **x** and **y** channels specify the centroids of the pixels, so you need **x1**, **y1**, **x2**, and **y2** to specify the extent of the pixel grid; otherwise the raster mark cannot know how wide a pixel is in abstract coordinates. And you need to specify **width** and **height** because this is the size of the raster grid (in grid coordinates, *i.e.*, pixel indexes) which might be a different resolution than the samples.

mbostock:
If you would prefer that the samples are taken at integer locations by default (rather than offset by 0.5) then we could apply this patch:

```diff
% git diff -p
diff --git a/src/marks/raster.js b/src/marks/raster.js
index 71a0af1c..52864d10 100644
--- a/src/marks/raster.js
+++ b/src/marks/raster.js
@@ -19,10 +19,10 @@ export class Raster extends Mark {
       // If X and Y are not given, we assume that F is a dense array of samples
       // covering the entire grid in row-major order. These defaults allow
       // further shorthand where x and y represent grid column and row index.
-      x1 = x == null ? 0 : undefined,
-      y1 = y == null ? 0 : undefined,
-      x2 = x == null ? width : undefined,
-      y2 = y == null ? height : undefined,
+      x1 = x == null ? -0.5 : undefined,
+      y1 = y == null ? -0.5 : undefined,
+      x2 = x == null ? width - 0.5 : undefined,
+      y2 = y == null ? height - 0.5 : undefined,
       imageRendering,
       pixelRatio = 1,
       fill,
@@ -156,7 +156,6 @@ function sampleFill({fill, fillOpacity, pixelRatio = 1, ...options} = {}) {
     if (h === undefined) h = Math.round(Math.abs(y2 - y1) / pixelRatio);
     const kx = (x2 - x1) / w;
     const ky = (y1 - y2) / h;
-    (x1 += kx / 2), (y2 += ky / 2);
     let F, FO;
     if (fill) {
       F = new Array(w * h);
```

Do you have a preference?

Fil:
Integer boundaries look better, I agree.

My difficulty is in precisely articulating how the various options must be combined, or on the contrary are "not compatible". For example, if you try to replace width: volcano.width with a "finger in the air" pixelRatio in the example above, you will get banding (unless pixelRatio=580 / 87). To document this, we need to mention that there is rounding, and that the pixelRatio is imputed from the width of the frame divided by the width of the data.

src/marks/raster.js
```diff
@@ -0,0 +1,146 @@
...
+      pixelRatio,
+      width = Math.round(imageWidth / pixelRatio),
+      height = Math.round(imageHeight / pixelRatio),
+      imageRendering
+    } = this;
+    const canvas = document.createElement("canvas");
+    canvas.width = width;
+    canvas.height = height;
+    const context2d = canvas.getContext("2d");
+    const image = context2d.createImageData(width, height);
```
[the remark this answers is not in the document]
mbostock: All of **x1**, **y1**, **x2**, and **y2** are required; it’s just in the case of a simple grid they default to 0, 0, **width**, and **height** respectively. I’ve added some more validation in the constructor.

mbostock:
> My difficulty is in precisely articulating how the various options must be combined, or on the contrary are "not compatible".

It’s best to think about this as multiple coordinate systems.

First there are a discrete set of samples in abstract coordinates _x_ and _y_ with _fill_ or _fillOpacity_ (or both). These correspond to the raster mark’s data. (The samples are imputed when **fill** or **fillOpacity** is specified as a function and **data** is null.) These samples are _typically_ in an axis-aligned grid but not necessarily so; arbitrary sample positions will be more useful in the future if and when the raster mark supports different methods of interpolation.

Second there is a raster grid (a.k.a. canvas) with its own pixel coordinates. The aforementioned samples are mapped to a canvas that is **width** pixels by **height** pixels, with the origin [0, 0] in the top-left corner. The extent of the canvas in pixel coordinates [0, 0, width, height] corresponds to an abstract extent [**x1**, **y1**, **x2**, **y2**] in the same abstract coordinate system as the samples. (In some cases **x1** and **x2** can be flipped, and likewise **y1** and **y2**.) During rendering, the raster mark assigns (bins using Math.floor) each sample in *x* and *y* to a pixel.

Lastly there are screen coordinates (really Plot frame coordinates, in the range of the _x_ and _y_ scales). These are needed to place the svg:image in the correct position within the Plot frame.

If you specify an incorrect **width** in the volcano example, then you expect to see banding because you no longer have exactly one sample per pixel in the raster grid. For example, here three columns are missing samples:

Screenshot 2023-01-02 at 12 33 45 PM

```js
Plot.plot({
  marks: [
    Plot.raster(volcano.values, {
      width: volcano.width + 3,
      height: volcano.height,
      x1: 0,
      y1: 0,
      x2: volcano.width,
      y2: volcano.height,
      x: (_, i) => (i % volcano.width) + 0.5,
      y: (_, i) => Math.floor(i / volcano.width) + 0.5,
      fill: (d) => d
    }),
    Plot.frame()
  ]
})
```

(They should really be transparent though… I think that’s a regression I introduced when adding support for _fillOpacity_.) If and when the raster mark supports better interpolation—something smarter than just binning the samples into rectangular pixels—then we could e.g. fill those gaps with the closest sample, or a blend of nearby samples.

Fil:
We could throw a useful error when using a projection; currently `TypeError: u is not a function. (In 'u(a.value[0])', 'u' is undefined)` (u being the x scale).

src/marks/raster.js
```diff
@@ -0,0 +1,146 @@
...
+      // interpolation methods here, as the current approach will often lead to
+      // a sparse image when not every pixel has a corresponding sample.
+      const kx = width / imageWidth;
+      const ky = height / imageHeight;
+      for (const i of index) {
+        const xi = Math.floor((X[i] - x1) * kx);
+        if (xi < 0 || xi >= width) continue;
+        const yi = Math.floor((Y[i] - y2) * ky);
+        if (yi < 0 || yi >= height) continue;
+        const {r, g, b} = rgb(F[i]);
```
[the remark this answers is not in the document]
mbostock: I think it would be better to do the optimization earlier when the scale is applied, like how Yuri does here:

```js
// Convenience function to create a cached color interpolator that
// returns cached rgb objects, avoiding color string parsing.
cacheInterpolator = (interpolator, n = 250) =>
  d3.scaleQuantize(d3.quantize(pc => d3.rgb(interpolator(pc)), n))
```

https://observablehq.com/@twitter/density-plot@4159

But in any case I think we should do optimizations as follow-up, as you say.

Fil:
Updated the documentation.

* Users might find the pixelRatio option name surprising, in the sense that to get the sharpest image on a given screen you want to set `pixelRatio: 1 / devicePixelRatio`. 
* I'm not sure how to document the rounding, I've used “floor rounded” but maybe it's “floored”, or we can do something with the math notation ⌊x⌋?

mbostock:
> pixelRatio: 1 / devicePixelRatio

I tried inverting this, but you end up with a lot of `pixelRatio: 1 / 3` etc. in the tests, which I think is more awkward? It’s certainly more to type in the common case.

I renamed the **interpolate** option to **rasterize** because I think it has more responsibility than just interpolation—it has to render to the canvas. (I though about the name “render” but of course there’s already a _mark_.render.) I haven’t added built-in support for barycentric interpolation yet, but I think we should do that… although I think I am happy with the **rasterize** option interface so it’s not a blocker.

I also, excitingly, found a way to make the **x1**, **y1**, **x2**, **y2** options optional in the case that **x** and **y** are specified. That simplifies a few of the tests, hooray. 🥳

Fil:
Hooray for the default **x1**!

Re: **pixelRatio** I didn't mean to invert it, but to rename the option. **pixelSize** maybe, to avoid the confusion?

I'm now thinking that the rasterize/interpolate methods could be an made into an options transform (returning a volcano-type matrix). It would allow to use them to create contours, and to combine them (I'd see a use case for barycentric interpolation then blur then contours)… I'm also not sure we can coerce values like this (for categorical fills, we'll need a different strategy anyway).

Fil:
Is there a way to **inset** x when inferring **x1**, **x2** from **x**? You can see in the walmarts rasters that we'd like to include a bit more of the statemesh.

Fil:
The rasterize API seems very low-level, and I'd like to investigate if we can make the "pixel interpolater" as an options transform instead.

The other blocking(?) issue is Safari's lack of support for `image-rendering: pixelated` in SVG.

mbostock:
> The other blocking(?) issue is Safari's lack of support for image-rendering: pixelated in SVG.

I don’t think we should do anything about this other than document it as a caveat.

> The rasterize API seems very low-level, and I'd like to investigate if we can make the "pixel interpolater" as an options transform instead.

Okay, but please don’t spend too much time on it. There are a lot other things to do, too, and we can always make further improvements or introduce a new mark type in the future, and it’s good to get stuff out and see how people use it!

Fil (approved):
Do we need to mention that this doesn't yet play well with faceting?

mbostock:
> Do we need to mention that this doesn't yet play well with faceting?

Yeah maybe we should talk about this. How would faceting work? I can see how when there is a discrete set of {x, y, fill?, fillOpacity?, fx?, fy?} samples, but I’m not sure how we’d do it with the “dense” array unless the requirement was that the array have a length of width × height × facets.length, and repeats? And since the fill as f(x, y) uses the dense array approach under the hood we’d have to figure that out first, and also think about extending the signature as f(x, y, fx, fy) or some such.

Thank you for the approval. Should I work on including the barycentric, voronoi, and perhaps walk-on-spheres interpolation methods as built-in rasterizers? Seems like you’ll definitely want some of these and it feels easy to add based on your existing work.

Fil:
Yes, in the rasterize-data branch (#1210) I'm doing faceting with a facets.length * width * height array of values, and passing facets as [[0], [1], [2]…] to know which subarray we're using for that facet. In other words the rasterizer transforms the data from whatever size it has initially to a size of 1 (matrix) per facet.

I think it would be grand to have these three rasterizers:
* barycentric (extrapolated only, I think; not sure we need the weird triangles)
* nearest-neighbor (I don't think we should call it voronoi, since the implementation could be using any nn algo)
* walk-on-spheres is my favorite, and it's just an extension of the nearest-neighbor (delaunay.find is faster in that case since we can use the memoization trick).

For WoS, I think using 3 steps (with the implementation that I fixed in #1210) is enough to get a palatable mixing, a good quality/speed trade-off. You can get a sense of how we go from voronoi to WoS by adjusting the slider in https://observablehq.com/@visionscarto/spherical-wos; for me visually, the artifacts are all gone with 3 steps. I don't think we need to include blurring or adaptive averaging at this stage, but I'll definnitely research that. (I must say I love how the grain of the texture reflects the local uncertainty/variance.)

mbostock:
Started a new branch here: https://github.com/observablehq/plot/compare/mbostock/image-data...mbostock/more-rasterizers

src/marks/raster.js
```diff
@@ -0,0 +1,427 @@
...
+    if (FO) context.globalAlpha = Math.abs(FO[j]);
+    context.fill();
+    if (context.globalAlpha === 1) (context.strokeStyle = context.fillStyle), context.stroke();
+  }
+}
+
+const ex = Symbol("extrapolate");
+
+function rasterizeBarycentric(canvas, index, {color}, {fill: F, fillOpacity: FO}, {x: X, y: Y}) {
+  const random = randomLcg(42); // TODO allow configurable rng?
```
Fil, line 271:
```suggestion
  const random = randomLcg(42);
```

Not sure how it would be useful. In fact I'd even want the rng to be reset across facets (when we have facets), to give more consistency.

src/marks/raster.js
```diff
@@ -0,0 +1,427 @@
...
+    case "none":
+      return rasterizeNone;
+    case "dense":
+      return rasterizeDense;
+    case "nearest":
+      return rasterizeNearest;
+    case "barycentric":
+      return rasterizeBarycentric;
+    case "walk-on-spheres":
+      return rasterizeWalkOnSpheres;
```
Fil, line 192: We should probably call this "random-walk" rather than "walk-on-spheres". It is more commonly known, and expresses well enough what's happening (as you put it: “if you were to be airdropped somewhere (…) and proceeded to walk randomly, this map tells you the expected [value] you’d run in to”). WoS converges much faster than a basic random walk would, and makes it feasible, but this is a question of implementation.

src/marks/raster.js
```diff
@@ -0,0 +1,427 @@
...
+    }
+  }
+
+  context2d.putImageData(image, 0, 0);
+}
+
+// TODO adaptive supersampling in areas of high variance?
+// TODO configurable iterations per sample (currently 2)
+function rasterizeWalkOnSpheres(canvas, index, {color}, {fill: F, fillOpacity: FO}, {x: X, y: Y}) {
+  const random = randomLcg(42); // TODO allow configurable rng?
```
Fil, line 368:
```suggestion
  const random = randomLcg(42);
```

src/marks/raster.js
```diff
@@ -0,0 +1,428 @@
...
+  );
+  let k = 0;
+  for (let y = 0, i; y < height; y++) {
+    for (let x = 0; x < width; x++, k += 4) {
+      i = delaunay.find(x, y, i);
+      for (let j = 0, cx = x, cy = y; j < 2; ++j) {
+        const radius = Math.hypot(X[index[i]] - cx, Y[index[i]] - cy);
+        const angle = random() * 2 * Math.PI;
+        cx += Math.cos(angle) * radius;
+        cy += Math.sin(angle) * radius;
```
Fil, line 389:
```suggestion
        const dx = X[index[i]] - cx;
        const dy = Y[index[i]] - cy;
        const angle = random() * 2 * Math.PI;
        const ca = Math.cos(angle);
        const sa = Math.sin(angle);
        cx += ca * dx - sa * dy;
        cy += sa * dx + ca * dy;
```

would save 1 Math.hypot, but I'm not sure it's worth the effort—makes the code a bit more difficult to grasp.

Fil:
I think I want a shorthand for 
> Plot.raster([1, 0, 0, 1], {width: 2})
which would set fill = data, and possibly height to data.length / width — maybe even width could be optional.

Fil:
The barycentric extrapolation method fails when the number of points is too low (here, 4 points, each controlling  only a tiny corner. The black region is invalid (the _ex_ symbol).
Capture d’écran 2023-01-07 à 13 05 28

Not sure if we must fix this (and how), but it makes it harder to show a demo of the method.

Fil:
When there are lots of pixels (like on a map where we project a raster image), it would probably make sense to paint the pixels forward with the null algorithm, then call the interpolation method for pixels that haven't been touched.

Fil:
a constant fillOpacity < 1, with the _nearest_ rasterize, shows faint voronoi lines; this could probably be avoided if we set the opacity on the image rather than in the imageData.
irisChart
See https://observablehq.com/@observablehq/random-walk-dithering

Fil:
There is an error with clip: sphere
Capture d’écran 2023-01-07 à 17 29 13

Fil:
The whole image shifts by 1px (??) when we switch from "nearest" to "random-walk". This is a bit suprising.

Fil:
I don't have anything more to add!

mbostock:
Is random-walk sampling at integers (the bottom-left of each pixel) or is it offset by a half to sample the pixel centroid?

mbostock:
Looks like the latest changes introduced a regression where pixels that don’t contain samples are now opaque black instead of transparent black. We should add a test for this (but note that all of our rasterization tests are effectively disabled currently because node-canvas doesn’t produce identical output in CI).

Screenshot 2023-01-07 at 12 19 45 PM

I’m also not sure we should bootstrap random-walk rasterization with none rasterization. I know that’s faster, but it changes the semantics of random-walk rasterization if you have dense samples if we add multisampling in the future. It also felt like the random-walk was fast enough that I don’t see a big need to mix the two.

mbostock:
The regression was from previously when we switched to interpolating in data space rather than interpolating scaled values. Fixed in the latest commit.

src/marks/raster.js
```diff
@@ -0,0 +1,437 @@
...
+  context2d.putImageData(image, 0, 0);
+}
+
+function rasterizeNearest(canvas, index, {color}, {fill: F, fillOpacity: FO}, {x: X, y: Y}) {
+  const {width, height} = canvas;
+  const context = canvas.getContext("2d");
+  const voronoi = Delaunay.from(
+    index,
+    (i) => X[i],
+    (i) => Y[i]
```
Fil, line 255:
```suggestion
    (i) => X[i] + 0.5,
    (i) => Y[i] + 0.5
```

mbostock:
I’ve been experimenting with random-walk and I think we might want to implement the minimum distance threshold to terminate the random walk. Compare these two results. The second looks much better in my opinion.

64 samples per pixel with jittering, no minimum distance, and fixed steps of 1 + 2, 5.98 seconds compute:

Screenshot 2023-01-07 at 1 31 52 PM

64 samples per pixel with jittering, minimum distance of 1px, and maximum steps of 1 + 20, 24.8 seconds compute:

Screenshot 2023-01-07 at 1 30 15 PM

4 samples per pixel with jittering, no minimum distance, and fixed steps of 1 + 2, 0.44 seconds compute:

Screenshot 2023-01-07 at 1 38 36 PM

4 samples per pixel with jittering, minimum distance of 1px, and maximum steps of 1 + 20, 1.64 seconds compute:

Screenshot 2023-01-07 at 1 38 05 PM

It’s obviously slower though since you’re walking more steps per sample.

src/marks/raster.js
mbostock: I think it’s correct as-is. You can see a noticeable drift with large pixelSize if you make this change.

Fil:
To check the alignment I generated images for the vapor test with pixelSize:2 and the different rasterize options.

I finally understood that it was the nearest which was not correctly targeting the center of the pixel (see suggestion above, which corresponds to the aa-nearest-fix-0.5.png image).

The second pair of images that's interesting to compare is the walk-seeding vs walk-no-seeding; without seeding (ie when we remove the call to rasterizeNull), the result is clearly more grainy. But maybe that's the style :-) For this task  anyhow, the best rasterize is _barycentric_.

**null**

the reference

aa-null


**nearest**

first: the misaligned version
aa-nearest

then the correct version (fixed by adding 0.5):
aa-nearest-fix-0 5

**random-walk**

seeded:
aa-walk-seeding

non-seeded:
aa-walk-no-seeding.

**barycentric**

today's winner
aa-barycentric

mbostock:
I’ll investigate the alignment again. Your test images look correct so something must be wrong here. 😄

mbostock:
One question I have is whether the test is correct. It’s currently saying this

```js
        x: (d, i) => (i % 360) - 180,
        y: (d, i) => 90 - ((i / 360) | 0),
```

but usually you’d want to offset the samples by 0.5 here, too, because that puts the sample locations in the middle of the pixel rather than the corner.

```js
        x: (d, i) => (i % 360) - 180 + 0.5,
        y: (d, i) => 90 - ((i / 360) | 0) + 0.5,
```

It’s the test’s responsibility to do that (and we don’t want our rasterization methods to assume that the sample locations are always off by 0.5 because there’s no requirement that they are aligned to a grid).

Fil:
Regarding jittering and multiple samples, it seems to me that you're now mixing the colors. I understand the aesthetic appeal (the images are "smoother", almost velvety, really gorgeous), but there is a drawback. Where before I could explain how the original is created (“pixels of three distinct colors, represented in proportions that match the probability distribution”), now we have brownish in-between hues which don't exist in the color legend. In fact, both approaches seem interesting (one to create a poster, the other to teach the technique); and we should probably offer both at some point in the future. I had focused on the faster and simpler one, but I'm happy to begin with any of them.

Fil:
It's a safe bet that the true coordinates are offset in the test, but I couldn't find details on the NEO website (it says "center of the cell", but only for the "csv for excel" specific file. https://neo.gsfc.nasa.gov/blog/2013/12/23/csv-and-floating-point-geotiffs/). We can definitely add these 0.5, and I agree with you that it's not the rasterizer's job to know how your data is originally aligned.

My point about the misalignment is different; it is that all the rasterizers should return an image with the same "base". Of course since every pixel behaves slightly differently, it's hard to tell exactly what is off—but if we see that one technique seems to shift the whole image .5px or 1px to the right/bottom compared to the others, we should fix that.

mbostock:
Okay, so the problem with the alignment is that the rasterVaporEqualEarth test isn’t specifying the dimensions of the grid; it’s using the defaults, which is based on the extent if **x** and **y**, but instead we always want the raster to span the full extent of the sphere from x in [-180, 180] and y in [-90, 90].

```js
        x1: -180,
        x2: 180,
        y1: -90,
        y2: 90,
```

However, if you do that, the default frame clipRect that gets applied projects one of the y coordinates slightly outside the frame (311.50150609099563 vs. 311.5) and then the **y1** becomes NaN and everything breaks.

Fil:
Since we're using equal earth, the rasterizer doesn't see longitudes and latitudes, only projected coordinates. I'm too tired I have to drop for today, sorry. But we're very close!

mbostock:
I agree there’s something wrong with the alignment but I haven’t figured out what it is yet.

mbostock:
Okay, I figured out why rasterizeNearest isn’t lining up… and it’s the call to context.stroke!

The problem is that the data here isn’t in random order, and so if we apply a 1px stroke to all cells, then each cell will extend beyond its boundary by 0.5px, and if the cells are drawn in nonrandom order, that will introduce a bias (that looks like a 0.5px shift). Here is without the stroke:

Screenshot 2023-01-07 at 2 39 46 PM

And here is with the stroke:

Screenshot 2023-01-07 at 2 40 23 PM

So either we’d need to randomize the order with which we draw cells, or we’d need to do two passes, drawing the strokes first and then drawing the fills.

mbostock:
Last step here is to figure out faceting, or at least document that it doesn’t work yet.

mbostock:
@Fil The latest changes are in https://github.com/observablehq/plot/compare/1e22d3d...a745cc3; documentation aside, this should be ready to go! 🚢

src/plot.js
```diff
@@ -592,8 +602,7 @@ function maybeTopFacet(facet, options) {
   if (facet == null) return;
   const {x, y} = facet;
   if (x == null && y == null) return;
-  const data = arrayify(facet.data);
-  if (data == null) throw new Error(`missing facet data`);
```
Fil, line 596: It's unexpected that we'd remove this error message (in 85d81ab486). Same in maybeMarkFacet.

Fil:
I'm tempted to say "ship", but this morning I tried to plug in https://observablehq.com/@jobleonard/pseudo-blue-noise instead of randomLcg(42), and the improvement in quality is really remarkable:

## walk-on-spheres (before, after)
a-walk-lcg
a-walk-pbn

## barycentric (before, after)
a-barycentric-lcg
a-barycentric-pbn

## contours with random walk (before, after)
a-contour-lcg
a-contour-pbn

src/plot.js
mbostock: This is needed or else Plot.raster/Plot.contour will crash when faceted when used in conjunction with the sampler initializer because the data is null.

mbostock:
The pseudo blue noise looks nice but I would like to focus on documenting and releasing this and returning to the axis mark. I’d be willing to expose the built-in spatial interpolation methods as functions so that you can pass in a custom RNG.

mbostock:
The latest two commits expose the built-in spatial interpolation methods

* Plot.interpolateNone
* Plot.interpolateNearest
* Plot.interpolatorBarycentric({random})
* Plot.interpolatorRandomWalk({random, minDistance, maxSteps})

So you can now provide your own RNG that uses pseudo blue noise e.g.

src/plot.js
mbostock: Latest commit restores the error when `data` is strictly undefined which should catch the case where the user has not specified the data (or specified it incorrectly _e.g._ `facet: {data2}`), but if it’s set explicitly to null as in the case of the raster/contour mark, it’s allowed.

Issue #522, opened 2021-08-24
When faceting, draw axes in empty facets


mbostock:
In this example:

<img width="640" alt="a faceted scatterplot showing the correlation between culmen depth and length across sex and species; each facet shows the current sex and species with black dots and all other penguins as pale gray dots" src="https://user-images.githubusercontent.com/7001/130275538-ec33aaa4-73e2-4fb3-862a-8fdd47a788d6.png">

It’d be nice if *x*-axis were repeated below the Adelie + null facet since the Chinstrap + null facet is empty. Similarly if the *y*-axis were on the right, the *y*-axis could be drawn on the right edge of the Chinstrap + male facet, rather than on the far right of the plot.

For inspiration, [geofacet](https://ryanhafen.com/blog/geofacet/):

![featured_hua660489bb22633e8177750bc9f451ca8_1250648_1360x1000_fill_q90_box_smart1_2](https://user-images.githubusercontent.com/230541/130697453-1fc80baf-d824-4429-9d60-a759fb99f2f0.png)

PR #1197, opened 2022-12-23
axis mark

### LINKED ISSUE #6: Custom styles for axis and ticks.

### LINKED ISSUE #148: Repeating axes on both sides (axis: "both")?

### LINKED ISSUE #926: Control axis z-order?
E.g., below the hexbins are drawn on top of the axis, but I’d rather the axis be drawn on top

image

### LINKED ISSUE #1000: Axis mark?
Axes could be created as marks, like grids would be after #992.

This would address #6, #926, #976, and #148.

mbostock:
This is a proof-of-concept implementing axes as a mark (instead of needing to be built-in to Plot’s rendering). There’s a bunch of stuff to figure out still:

- [ ] Smarter default ticks (port `autoAxisTicks`)
- [ ] Smarter default tick format ()
- [ ] Right-oriented _y_-axis
- [ ] Top-oriented _x_-axis
- [ ] Axis labels?
- [ ] Grid lines
- [ ] Remove Plot’s built-in axis rendering and instead automatically add axis marks as needed
- [ ] More configuration options
- [ ] Allow channels (e.g., varying color based on value)

Fixes #6.
Fixes #148.
Fixes #926.
Fixes #976.
Fixes #1000.


mbostock:
Since the axes are driven by data, you can now do neat stuff like vary the _x_-position of the _y_-axis:

Screenshot 2022-12-24 at 9 19 22 AM

```js
Plot.axisY({grid: true, x: (y) => AAPL.find((d) => d.Close >= y)?.Date, insetLeft: -10})
```

mbostock:
Screenshot 2022-12-24 at 11 16 56 AM

```js
Plot.frame({fill: "#eaeaea"}),
Plot.gridY({ticks: 12, stroke: "#fff", strokeOpacity: 1, strokeWidth: 0.5}),
Plot.gridX({ticks: 12, stroke: "#fff", strokeOpacity: 1, strokeWidth: 0.5}),
Plot.gridY({stroke: "#fff", strokeOpacity: 1, strokeWidth: 1}),
Plot.gridX({stroke: "#fff", strokeOpacity: 1, strokeWidth: 1}),
Plot.axisY({anchor: "right"}),
Plot.axisX({anchor: "top"}),
Plot.axisY({anchor: "left"}),
Plot.axisX({anchor: "bottom"}),
```

mbostock:
We’ll be able to fix #522 with this, too!

Screenshot 2022-12-26 at 5 04 05 PM

mbostock:
This is fun!

Screenshot 2022-12-26 at 5 45 20 PM

mbostock:
This is fun!

Screenshot 2022-12-26 at 5 45 20 PM

mbostock:
Wrapping long tick labels with `Plot.axisX({lineWidth: 8})`:

Screenshot 2022-12-26 at 7 02 44 PM

mbostock:
The latest commit undoes part of #1062; it shouldn’t be necessary to inline text elements because the white-space: pre only applies to tspan elements within text elements; i.e., we only need to inline tspan elements (and other things _within_ text elements). And inlining text elements makes it hard to review changes and generally understand the structure of the SVG we’re generating.

test/output/aaplBollinger.svg
```diff
@@ -13,80 +13,70 @@
       white-space: pre;
     }
   </style>
-  <g aria-label="y-axis" aria-description="↑ Close" transform="translate(40,0)" fill="none" text-anchor="end" font-variant="tabular-nums">
```
Fil, line 16: The aria-description here, with the axis label, is now lost, is this expected?

src/axes.js
```diff
@@ -1,168 +1,31 @@
...
-  }
-}
-
-// Mutates axis.labelAnchor, axis.label, scale.label!
-function autoAxisLabelsX(axis, scale, channels) {
-  if (axis.labelAnchor === undefined) {
-    axis.labelAnchor = isOrdinalScale(scale) ? "center" : scaleOrder(scale) < 0 ? "left" : "right";
+export function autoScaleLabels(channels, scales, options) {
+  for (const key in scales) {
+    autoScaleLabel(key, scales[key], channels.get(key), options[key]);
```
Fil, line 9:
```suggestion
    autoScaleLabel(key, scales[key], channels.get(key), {label: options.label, ...options[key]});
```

to support the top-level label option (e.g. in the yearly-requests test).

test/output/aaplBollinger.svg
mbostock: It’s redundant with the label, right? I’m not sure how much value it adds but we could try to find a new place for it. There’s no longer a single G element for the axis since it is now multiple marks.

Fil: Maybe we don't need it, but I wanted to mention it if it wasn't intentional. BTW maybe this is an indication that nested/composite marks should return nested layers?

mbostock:
With faceting, the _x_ and _y_ axes are now drawn in a slightly different position: rather than being drawn on the outside of the main (singular, top-level) frame, they are drawn on the outside of the first or last facet frame, and possibly repeated if facets are empty. I think this is appropriate given that axes may be drawn again if there are empty facets, and we want them to appear the same. But it’s worth noting and is visible in a few of the changed test snapshots.

Fil:
Noting that having to type "x1" or "x2" for fancy tick positioning is a bit tricky, in particular because it depends on the axis' anchor:
>    Plot.axisY({anchor: "right", x2: (y) => AAPL.find((d) => d.Close >= y)?.Date})

vs
>    Plot.axisY({anchor: "left", x1: (y) => AAPL.find((d) => d.Close >= y)?.Date})

it's probably OK though (x1 is the "start" or "near end" and x2 the "opposite" or "far end").

Fil:
I wasn't able to solve any of the items on the todo list before _documentation_, and I'm not sure [70ebb1f5f](https://github.com/observablehq/plot/pull/1197/commits/70ebb1f5facd80be9027dbbf8fdd9be180d0fc26) is on the right track.

mbostock:
> it's probably OK though (x1 is the "start" or "near end" and x2 the "opposite" or "far end").

Yep, that’s what I was going for. I’ll start putting together a notebook with examples using a test build. That should help with the documentation here too. And please send questions. :)

Fil:
I've added a chart to investigate: the area marks in metroUnemploymentRidgeline are shifted up (but the axes are unchanged).

mbostock:
The issue with the metroUnemploymentRidgeline example was that it hard-codes the _y_ range, and the definition of the _y_ range changes in this PR to include margins when faceted, whereas previously with faceted plots the facet dimensions always had margins of zero. I edited the test to fix the “overdraw” of the ridgelines.

src/axes.js
mbostock: It’s slightly more complicated than this because there’s also the facet.label option which overrides the top-level label option for the _fx_ and _fy_ scales (mainly so you can turn them off with label = null).

Fil:
The issue with projectionHeightGeometry is probably similar, since it uses the null projection, which is a bit raw.

mbostock:
Looks like the _x_-axis labels are in the wrong position, _e.g._ in d3Survey2015Comfort.

Fil:
I checked all the tests visually. Here's a list of the issues that I saw (I might have missed some).

_missing facet label:_ athletesSample, athletesSportWeight, emptyFacet, footballCoverage, hexbinR, hexbinText, industryUnemploymentTrack, penguinCulmen, penguinCulmenArray, penguinCulmenMarkFacet, penguinDensityFill, penguinDensityZ, penguinMassSex, penguinMassSexSpecies, penguinSexMassCulmenSpecies, penguinSpeciesIslandRelative

_wrong facet label position or rotate:_ beckerBarley, diamondBoxPlot, downloadsOrdinal, mobyDickLetterPosition, penguinAnnotated, penguinFacetAnnotated, penguinFacetAnnotatedX, penguinIslandUnknown, penguinSex, penguinSpeciesIslandSex, simpsonsRatings, sparseCell, usPopulationStateAge
  
_misalignment_: projectionHeightGeometry

_clipped label_: stocksIndex (this one is an old bug)

mbostock:
I’ve enabled the facet axis labels, so we’re pretty close now! I see two issues with the current facet axis label position:

The first, shown in the athletesSample test, is that the facet axis label now tends to overlap with the facet axis ticks.

image

The second, shown in the anscombeQuartet test, is that the facet axis label is now centered in the “middle” facet (3), but when there is an even number of facets, it would be better to draw the axis label in the middle of the outer frame (between 2 and 3).

image

mbostock:
I’ve now fixed the first issue with the facet axis label position. The second issue is trickier: we don’t currently have a way of drawing a mark “detached” from a facet (conceptually across all facets), so we have to decide whether it’s acceptable to draw the facet axis label in the middle facet (like we do for the _x_ and _y_ axis labels when center-anchored), or whether we need to implement some new sort of facetAnchor (“outer”?) to support this style of positioning that is only needed for the facet axis labels.

Fil:
Yes I think we'll want centered labels for facet axes, sorry :)

mbostock:
Everything is now implemented except for fixing the use of named time intervals in the polylinear test! I will try to get to that tonight and then hopefully we can finally put a bow on this. 🎁

mbostock:
It’s done! _Let’s gooooo._ 🚅

Fil:
I've been following along and testing the latest build; so far so good. Is there anything in particular that we should focus on (besides documentation)?

mbostock:
I can’t think of anything particular besides documentation, but I’m self-reviewing, too.

README.md
```diff
@@ -958,6 +960,46 @@ Returns a new arrow with the given *data* and *options*.
 
 <!-- jsdocEnd arrow -->
 
+### Axis
+
+[Source](./src/marks/axis.js) · [Examples](https://observablehq.com/@observablehq/plot-axis) · Draws an axis.
+
+Plot automatically generates axes for position scales, and draws them below the other marks. Each axis is composed of up to 5 marks: a grid, and an axis mark which might contain a line, ticks, tick labels, and axis label.
+
+When you need more control, you can add axis and grid marks explicitely in the marks options. Note that Plot’s automatic axis for *x* is disabled when a mark’s aria-label property begins by `x-axis `—and likewise for *y*, *fx* and *fy*.
```
yurivish, line 969: small typo: "grid marks explicitely" →"grid marks explicitly"?

Fil: Darn, I _always_ make this typo! Thanks for spotting

mbostock:
I think axis = "both" should only draw on axis label? It’s nice to have the ticks drawn twice but I think you often don’t want a second axis label. (And you can always do that by adding an axis mark yourself, so it’s nice to favor the common one with axis = "both".)

Fil:
An open question: do we need to support *data* in the axisX(*data*, *options*) syntax, since the tick values can be passed as options.ticks, with—I think?—no loss of generality.

mbostock:
> An open question: do we need to support data in the axisX(data, options) syntax

I chose not to do this. Here are my reasons:

1. When data is specified up-front, we can compute channels the normal way rather than deferring it to the initializer. This means that specifying the data up-front has _different behavior_ than specifying an array of ticks: in the former case, it can extend the scale domain, whereas in the latter, it can’t (because initializers aren’t allowed to affect the behavior of position scales). I realized I broke this along the way, but I’ve fixed it in 61e1145 and added a test.

Screenshot 2023-01-31 at 8 34 11 AM

```js
Plot.plot({
  marks: [
    Plot.axisY(d3.ticks(0, 200, 10), {anchor: "left"}),
    Plot.lineY(AAPL, {x: "Date", y: "Close"})
  ]
})
```

2. Philosophically, I think it’s desirable to support the same syntax with axis marks that we do with other marks, especially when “under the hood” the data associated with an axis mark is this array of tick values. Yes, axis marks are special in that typically the tick values depend on the scale domains, and hence cannot be known until after the other marks are initialized; and because channels depend on data, that means typically all of an axis mark’s channels need to be deferred to an initializer too. But per (1) it’s nice that if the data is specified up-front, we can have the axis mark follow the same code path as a normal mark.*

*The fontVariant and tickFormat also depend on the scale, and so also need to be deferred to the initializer. And likewise axis label orientation. But we can at least make the axis mark behave as much like a normal mark as possible.

mbostock:
In a case like this

```js
Plot.plot({
  y: {
    grid: true
  },
  marks: [
    Plot.axisY({anchor: "left"}),
    Plot.lineY(AAPL, {x: "Date", y: "Close"})
  ]
})
```

should we still draw implicit the _y_-axis grid even though there’s an explicit _y_-axis? I kind of think that yes, we should. Though that would also mean that the implicit _y_-axis grid would use the default ticks, which could be different than those chosen by the explicit _y_-axis. That seems fine, though.

Fil (approved):
Thanks for https://github.com/observablehq/plot/commit/61e114585575a362cd17921a83cd33226a7f5cce which clarifies the issue.

We still need to iterate a bit on the documentation, but I'm happy with everything

mbostock:
Oh, there was another idea I had for shorthand: if the **grid** option is a number or an interval (or an interval name), it gets promoted to the **ticks** option of the grid mark; if the **grid** option is a color, it gets promoted to the **stroke** option of the grid mark. Then you can say `grid: 20` for more grid lines, or `grid: "red"` for a red grid, etc. Not really needed since you can do the same thing with an explicit mark but it’s always fun to have extra shorthand…

mbostock:
This seems like a bug: when there’s an explicit *x*-axis, the default anchor of the implicit *fx*-axis changes to bottom instead of top, resulting in the two axes overlapping. I’m guessing that the *fx*-axis should always default to a top anchor.

Screenshot 2023-01-31 at 12 40 19 PM

```js
Plot.plot({
  facet: {marginRight: 80},
  marks: [
    Plot.frame(),
    Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm", fx: "sex", fy: "species"}),
    Plot.axisX({color: "red"})
  ]
})
```

Fil:
Should Plot.axisX be a wrapped with Plot.marks? Currently we have to write
```js
Plot.marks(Plot.axisX(d3.ticks(0, 1, 10), { stroke: "red" })).plot(…)
```

it could be simpler to manipulate as:
```js
Plot.axisX(d3.ticks(0, 1, 10), { stroke: "red" }).plot(…)
```

mbostock:
Yes, it should use Plot.marks. That’s an oversight on my part.

yurivish:
tiny note: i just noticed that the README says "The mark.facetAnchor option controls TK. " which was introduced in this PR. I'd fix it myself but am not sure what it should say...

mbostock:
Yes, the documentation is still being written. Thanks!

PR #1204, opened 2022-12-30
fix clip: frame for band scales

### LINKED ISSUE #1202: The clip option is misaligned with band scales
E.g. below, the rightmost bar is erroneously truncated. 

untitled (89)

```js
Plot.plot({
  grid: true,
  x: {
    interval: d3.utcDay
  },
  marks: [
    Plot.ruleY([0]),
    Plot.frame(),
    Plot.barY(
      data,
      Plot.groupX(
        { y: "sum" },
        { x: "Date", y: "Count", rx: 6, insetBottom: -3, clip: "frame" }
      )
    )
  ]
})
```

where

```js
data = [
  {Date: new Date("2022-12-01"), Count: 10},
  {Date: new Date("2022-12-02"), Count: 1},
  {Date: new Date("2022-12-02"), Count: 1},
  {Date: new Date("2022-12-03"), Count: 2},
  {Date: new Date("2022-12-04"), Count: 3},
  {Date: new Date("2022-12-05"), Count: 4},
  {Date: new Date("2022-12-06"), Count: 5},
]
```

https://observablehq.com/d/022b0876b936b4cb

Ref. https://github.com/observablehq/plot/discussions/1201

mbostock:
Fixes #1202. The previous fix in #950 was only partly correct: in essence, it was trying to undo transform applied by applyTransform:

https://github.com/observablehq/plot/blob/c095660c9ae13e88b0906e6474bf2ba4e20585ab/src/style.js#L383-L389

But the behavior of applyTransform varies from mark to mark, and even within a mark depending on which channels are applied, for example for collapsed rects:

https://github.com/observablehq/plot/blob/c095660c9ae13e88b0906e6474bf2ba4e20585ab/src/marks/rect.js#L62

Rather than trying to undo the transform, this fix instead wraps the G element with another (untransformed) G element, applying the clip to the parent G element. And thus the clipped region is not affected by the mark’s transform.

To simplify the adoption of this fix, applyIndirectStyles now mutates the passed-in _selection_.node to return the parent G element when clipping is applied. This means that callers effectively get the fix “for free” (and we now don’t even need to pass _scales_ to applyIndirectStyles). I’m sure there’s a cleaner way to do this, but this seems good for now and minimizes churn.

Previously #949.


test/plots/band-clip.js
```diff
@@ -14,3 +15,23 @@ export default async function () {
...
+    {Date: new Date("2022-12-04"), Count: 3},
+    {Date: new Date("2022-12-05"), Count: 4},
+    {Date: new Date("2022-12-06"), Count: 5}
+  ];
+  return Plot.plot({
+    grid: true,
+    x: {interval: d3.utcDay},
+    marks: [
+      Plot.ruleY([0]),
+      Plot.barY(data, Plot.groupX({y: "sum"}, {x: "Date", y: "Count", rx: 6, insetBottom: -3, clip: "frame"}))
```
Fil, line 34:
```suggestion
      Plot.barY(data, Plot.groupX({y: "sum"}, {x: "Date", y: "Count", rx: 6, insetBottom: -6, clip: "frame"}))
```

To be clear I don't mind it being -3, but as an example it complicates things as it raises the unnecessary question "is this supposed to be half the rx?".

Fil (approved):

PR #1205, opened 2022-12-30
named time intervals

### LINKED ISSUE #307: Named time ticks.
If you want ticks at a specific time interval, you can currently do that as _e.g._ `ticks: d3.utcDay`. It’d be nicer if you could use a named string instead, say `ticks: "day"`. Also that would be nice because if you do `type: "time"` instead of `type: "utc"`, then `ticks: "day"` would be interpreted as `ticks: d3.timeDay`, similar to the _tickFormat_ option.

I don’t really want to add syntax that’s equivalent to d3.utcDay.every(3) though…

mbostock:
Fixes #307. This is going to conflict with #1197, so I could merge the two if desired.


Fil (approved):
I have extended this a little, as time intervals can be used as thresholds for the bin transform.

PR #1206, opened 2023-01-02
Plot.raster defaults x1, x2 to the domain of x

Fil:
Capture d’écran 2023-01-02 à 15 36 56


mbostock (changes requested):
Thanks for this suggestion! You inspired me to add a number of tests to the parent PR and I found some bugs. Per comment below, I don’t think we should add the feature proposed here, but I’m happy to discuss further! Nice to have you back from vacation. 🙏

Fil:
Thanks for the detailed answer; the error messages now gently guide the user to the solution (which is to explicitly specify x2 and y2), which solves the issue.

PR #1207, opened 2023-01-04
exploration of raster interpolate methods

Fil:
Just to see how it might work
Capture d’écran 2023-01-04 à 18 59 26
Capture d’écran 2023-01-04 à 18 59 14
Capture d’écran 2023-01-04 à 18 59 30
Capture d’écran 2023-01-04 à 18 59 21


Fil:
Raises the question of what we interpolate: is it value or color?

Fil:
* I don't think we need to check that the points are valid; invalid points have already been filtered out?

* the extrapolate code could be isolated from the image code, maybe, so we could use it to draw contours too.

PR #1209, opened 2023-01-06
Add an error message when using a given stacking order (as an array) …

### LINKED ISSUE #1208: when z is not defined, zorder crashes
this crashes:
> Plot.text(versions, Plot.stackX({x: "downloads", order: Plot.valueof(versions, "version")})).plot();

```
stack.js?t=1672927922568:329 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'map')
    at orderZDomain (stack.js?t=1672927922568:329:12)
    at stack.js?t=1672927922568:321:29
    at Text.transform (stack.js?t=1672927922568:86:26)
```

this works:
> Plot.text(versions, Plot.stackX({x: "downloads", z: [], order: Plot.valueof(versions, "version")})).plot();

see https://observablehq.com/@mbostock/npm-daily-downloads#comment-87022bafaa591b9a

Fil:
…without a z channel

closes #1208

### DIFF
--- a/src/transforms/stack.js
+++ b/src/transforms/stack.js
@@ -325,6 +325,7 @@ function orderGiven(domain) {
 // O that can be used with applyOrder to sort stacks. Note that this is a series
 // order: it will be consistent across stacks.
 function orderZDomain(Z, domain) {
+  if (!Z) throw new Error("missing channel: z");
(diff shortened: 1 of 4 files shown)


mbostock (approved):
Makes sense! Thanks for finding the cause. 🙏

PR #1210, opened 2023-01-06
A rasterizer operates on the data rather than on the canvas

Fil:
This approach could make a lot of operations easier (such as combining a walk-on-spheres with blurring, or using the same functions for contouring).

The only thing I had to drop is the use of voronoi.renderCell for nearest-neighbor.

I took a few shortcuts to evaluate the approach:
- [ ] fillOpacity, constant fill
- [ ] undefined values are not handled (value coercion creates NaN which come out black)
- [ ] facets do not fully work
- [ ] the checks for x and y are not re-entrant

### DIFF
--- a/src/marks/raster.js
+++ b/src/marks/raster.js
@@ -1,6 +1,7 @@
 import {rgb} from "d3";
 import {create} from "../context.js";
-import {map, first, second, third, isTuples} from "../options.js";
+import {first, second, third, isTuples, range, valueof} from "../options.js";
+import {Position} from "../projection.js";
 import {Mark} from "../plot.js";
 import {applyAttr, applyDirectStyles, applyIndirectStyles, applyTransform, impliedString} from "../style.js";
 import {initializer} from "../transforms/basic.js";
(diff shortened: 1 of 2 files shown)


Fil:
there is a weird artifact on ?test=spatialPenguinsSpheres, the horizontal red lines are not supposed to be there (??)
Capture d’écran 2023-01-06 à 16 58 34

mbostock:
I appreciate you taking the time to explore this option. I am inclined to keep the existing API. I like that you can easily use the Canvas2D api or even WebGL to render to the canvas, and I think we don’t need to include support for composing rasterizers yet (and you could always do that on top of the rasterize method if you wanted).

Fil:
I believe that for a WebGL render we'll need to make this mark asynchronous?

mbostock:
Superseded by #1214 [bd4d29c](https://github.com/observablehq/plot/pull/1214/commits/bd4d29ce38873f72ffbfe688b634e87d4f8f032d)?

Fil:
Yes, this is exactly what I wanted to do. It's awesome to see things fall into place!

PR #1211, opened 2023-01-06
labelAnchor

Fil:
lazily computed once the scale type is known. This means that a few defaults need to be computed "by hand".

- [x] penguinFacetAnnotatedX is wrong (the y label is filtered out), I can't see why.
- [ ] the central position for labelAnchor is "center", whereas frameAnchor, textAnchor, and now facetAnchor are "middle"; is this something we could normalize, by converting "center" to "middle" automagically?

I'm leaving three examples broken on purpose:
- [x] longLabels: the label (_name_) is in the middle of the long labels; we should reimplement **labelOffset**
- [x] downloadsOrdinal: the label (Date) is in the middle of the date labels; (in this case, the default labelOffset should be computed from the facet margins)
- [ ] yearlyRequests: the top-level `label: null` setting should remove both labels (fixed in [this suggestion](https://github.com/observablehq/plot/pull/1197#discussion_r1060541761))

TODO:
- [ ] check that it works as intended with faceting (the original code has _e.g._ marginTop - facetMarginTop)
- [ ] fix googleTrendsRidgeline & metroUnemploymentRidgeline (also broken in mbostock/axis-mark)
- [ ] fix projectionHeightGeometry (also broken in mbostock/axis-mark)


Fil:
Not 100% sure that this is doing the right thing with labelOffset. The original code does marginTop - facetMarginTop, which I don't do (or know how to do) here.

Fil:
I could not salvage this PR from the conflicts. Giving up is the best strategy in this case.

PR #1214, opened 2023-01-08
contour mark

mbostock:
The contour mark functions similarly to the raster mark, except that after generating the raster grid, it generates contour polygons using d3-contour (instead of an image). I figure we should flesh out this API to make sure it works well with the new raster mark in #1196.

TODO

- [x] fill as a function
- [x] dense grids (volcano)
- [x] non-gridded samples (ca55)


mbostock:
This implementation is feeling pretty good to me now, but I’m beginning to see what you were going after with #1210: currently there’s not an easy way for the Contour mark to leverage the nearest/barycentric/random-walk interpolation methods to convert non-gridded samples into a grid suitable for d3-contour. But clearly we would like to be able to use them e.g. to produce contours over the ca55 dataset. 

So I think what I’d like to do is change the **rasterize** function option to return a dense array of gridded values (in data space, without scaling). That’ll make things a little slower for the raster mark since it’s another copy of the data, and it means that the nearest interpolation method won’t be able to use the Canvas2D API to render, but I think the performance difference will probably be small enough that it won’t be noticeable, and it will allow the Contour mark to reuse these implementations.

Let me know if that sounds good or please take a crack at it as you like, @fil!

test/plots/function-contour.js
```diff
@@ -0,0 +1,18 @@
+import * as Plot from "@observablehq/plot";
+
+export async function functionContour() {
+  return Plot.plot({
+    color: {
+      type: "diverging"
+    },
+    marks: [
+      Plot.contour({
+        value: (x, y) => x * y * Math.sin(x) * Math.cos(y),
```
mbostock, line 10: I feel like it would be more natural to specify **fill** here, rather than **value**. Maybe we can make that work as shorthand, but I didn’t want to support **fill** and **stroke** being set to _different_ functions (multiple overlapping contours?). 🤔

src/marks/raster.js
```diff
@@ -241,52 +241,34 @@ function interpolateNone(index, width, height, X, Y, V) {
...
-  const mix2 = mixer2(V, random);
-  for (let i = 0; i < triangles.length; i += 3) {
-    const a = triangles[i];
-    const b = triangles[i + 1];
-    const c = triangles[i + 2];
-    if (V[a] === ex) V[a] = V[c] === ex ? V[b] : V[b] === ex ? V[c] : mix2(V[c], V[b]);
-    if (V[b] === ex) V[b] = V[a] === ex ? V[c] : V[c] === ex ? V[a] : mix2(V[a], V[c]);
-    if (V[c] === ex) V[c] = V[b] === ex ? V[a] : V[a] === ex ? V[b] : mix2(V[b], V[a]);
-  }
+  // To each edge point, assign the closest (non-extrapolated) value.
```
Fil, line 263: nice!!

Fil: Looks like this changes the extrapolation in the penguins barycentric image quite a bit, and it takes a moment to understand why there are blue triangles on the left and at the bottom-right corner. But it makes sense, and the previous version was already quite arbitrary (tossing a coin to decide on the color…).

I'm glad we have the unit test, because it shows how the algorithm operates, but it's not a good example (random-walk is the best for this task).

test/plots/function-contour.js
Fil: yes it would be very nice; let's just throw an error if both are functions?

Fil:
The contours for ca55 are very pretty with more blurring (e.g. blur: 3), and the barycentric interpolate doesn't have the arbitrary high frequency component of random-walk. So I'd suggest to change the test to barycentric+blur: 3, to make it a better example.

The contours are empty if we don't specify an interpolate strategy; I think this would probably be addressed by https://github.com/d3/d3-contour/pull/61: it would make contours around the samples. But I don't think that's what people will want in general (and certainly not on this dataset), and I'd suggest we either:
* throw an error or give a warning to say "the samples don't cover the whole raster, please select an interpolate strategy (e.g.: none, or barycentric)
* default to barycentric+blur3

It would be good to backport this to d3-contour at some point, to allow operations on the returned array of GeoJSON?

Fil:
I think blurring + random-walk is also a good solution to https://github.com/observablehq/plot/pull/1196#issuecomment-1374621305 (see [dd06ea05c commit log](https://github.com/observablehq/plot/pull/1214/commits/dd06ea05c72de83eb7d7eb69c974fa600c8ec04f)).
rasterPenguinsBlur

src/marks/raster.js
mbostock: Yes, I thought this was weird at first too, but then I checked the Voronoi diagram which explains it:

Screenshot 2023-01-08 at 10 39 33 AM

👍

test/plots/function-contour.js
mbostock: I will work on more shorthand tomorrow.

Fil:
* Fixed two bugs I found while playing with the contours and rasters
* Checked degenerate cases such as n=1 or n=2 points.

About the shorthand and defaults, I suggest we follow a similar convention as in [Plot.density](https://github.com/observablehq/plot#density):
* default to black stroke, no fill; use fill: _value_ if we want a colored fill, same for stroke (for the density mark, the special keyword is "density"; here we can use "value" or "threshold").
* the value of a point is specified with the _value_ option, as a channel if we have data, and as a function f(x,y) if we accept a function sampler.

src/marks/contour.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+
+    // Compute any deferred channels. X and Y are set to as many zeroes as there
+    // are geometries, in order to pass the filter.
+    const newChannels = {geometry: {value: geometries}};
+    for (const key in this.channels) {
+      const value = this.channels[key].value;
+      if (value.defer) newChannels[key] = {value: valueof(geometries, value.defer), scale: true};
+      else if (key === "x" || key === "y") newChannels[key] = {value: valueof(geometries, 0)};
```
mbostock, line 104: You can set the filter option on the channel to disable filtering. I’ll fix this.

mbostock: I was able to reproduce the filtering problem using Plot.contour instead of Plot.raster in the vapor equal-earth barycentric interpolate test. The latest commit fixes it.

mbostock: Oops, one more fix needed…

Screenshot 2023-01-09 at 11 01 46 AM

```js
Plot.contour(await vapor(), {
  fill: (d) => d,
  x: (d, i) => (i % 360) - 180 + 0.5,
  y: (d, i) => 90 - ((i / 360) | 0) + 0.5,
  blur: 0.5,
  thresholds: 20,
  stroke: "currentColor",
  strokeWidth: 0.5,
  interpolate: "barycentric",
  clip: "sphere"
})
```

Fil:
It's in a test I haven't pushed because it's very slow—I'll try to make a lighter one, but for now I've built 11dbb07 and added it to https://observablehq.com/@observablehq/global-temperatures-raster#issue11dbb0; you can see that it fails to create the contours.

Fil:
Ah no! It works—it's just that the defaults have changed 👍

mbostock:
I think there are still some issues with the concept of a “dense” grid, and in particular whether this is a linear grid in data space or in screen coordinates. For example, say we start here:

Screenshot 2023-01-09 at 11 54 21 AM

```js
Plot.plot({
  marks: [Plot.raster(await vapor(), {width: 360, height: 180, fill: (d) => d})]
})
```

We can fix the _y_ inversion y reversing the _y_ scale:

Screenshot 2023-01-09 at 11 55 27 AM

```js
Plot.plot({
  y: {reverse: true},
  marks: [Plot.raster(await vapor(), {width: 360, height: 180, fill: (d) => d})]
})
```

But what if we wanted instead to specify the _x_ domain as [-180, 180] and the _y_ domain as [90, -90]? Then we get something surprising because the defaults for x1, y1, x2, y2 and 0, 0, width, height.

Screenshot 2023-01-09 at 11 58 09 AM

```js
Plot.plot({
  x: {domain: [-180, 180]},
  y: {domain: [90, -90]},
  marks: [Plot.raster(await vapor(), {width: 360, height: 180, fill: (d) => d})]
})
```

We can fix that by specifying these options explicitly, I suppose.

Screenshot 2023-01-09 at 11 57 49 AM

```js
Plot.plot({
  x: {domain: [-180, 180]},
  y: {domain: [90, -90]},
  marks: [Plot.raster(await vapor(), {x1: -180, y1: -90, y2: 90, x2: 180, width: 360, height: 180, fill: (d) => d})]
})
```

But notice what happens when we change the _y_ scale type of _sqrt_: the raster stays the same, even though the axis has changed.

Screenshot 2023-01-09 at 11 59 06 AM

```js
Plot.plot({
  y: {type: "sqrt"},
  marks: [Plot.raster(await vapor(), {width: 360, height: 180, fill: (d) => d})]
})
```

In other words, when you use the dense grid shorthand, you have to provide a dense grid in _scaled_ (projected) coordinates, which makes it rather difficult to use. I get more what I expect if I specify _x_ and _y_ explicitly (non-optimized).

Screenshot 2023-01-09 at 12 00 40 PM

```js
Plot.plot({
  y: {type: "sqrt"},
  marks: [
    Plot.raster(await vapor(), {
      width: 360,
      height: 180,
      x: (d, i) => (i % 360) - 180 + 0.5,
      y: (d, i) => 90 - ((i / 360) | 0) + 0.5,
      fill: (d) => d,
      interpolate: "nearest"
    })
  ]
})
```

Also note that if I try to use the _interpolate_ option with a dense grid, it crashes because the _x_ and _y_ channels are missing.

```
TypeError: Cannot read properties of undefined (reading '362')
```
```js
Plot.plot({
  y: {type: "sqrt"},
  marks: [
    Plot.raster(await vapor(), {
      width: 360,
      height: 180,
      fill: (d) => d,
      interpolate: "nearest"
    })
  ]
})
```

mbostock:
I think the challenge with “dense grid in abstract rather than projected coordinates” is that it doesn’t work as well for evaluating continuous functions. When you have a continuous function, we want to compute the grid in projected (scaled) coordinates and then invert it back into abstract (data) coordinates for the best results. Maybe we need to separate the code path for the continuous function case somehow; we could still provide shorthand for the dense grid case but not worry about optimizing it so much.

mbostock:
Okay, I think this is looking really good now! There’s still some optimization that’s likely possible when you specify a dense grid of values, but only in the case where _x_ and _y_ are linear scales. If we figure out how to make that faster in the future that’s gravy but it shouldn’t change the API! From my side I’m ready to merge this and the parent branch into main and start working on documentation.

Fil (approved):
🌐 🎉 👍 🚢

PR #1215, opened 2023-01-10
fix empty facet bug

### LINKED ISSUE #1212: crash with empty facets
```
{
  const data = [];
  return Plot.plot({
    facet: { data, x: "TYPE" },
    fx: { domain: ["a", "b"] },
    marks: [Plot.rectY(data, { x1: "PERIOD" })]
  });
}
```

TypeError: i is not iterable

(reported by @mootari )

Fil:
When the facet domain does not include any value from the actual contents, the facet index is undefined instead of [].

I'm not sure this is the correct fix, I have an alternative that moves a bit more things but seems better in #TKTK.

~~~diff
--- a/src/plot.js
+++ b/src/plot.js
@@ -645,8 +645,8 @@ function maybeMarkFacet(mark, topFacetState, options) {
 // Facet filter, by mark; for now only the "eq" filter is provided.
 function filterFacets(facets, {channels: {fx, fy}, groups}) {
   return fx && fy
-    ? facets.map(({x, y}) => groups.get(x)?.get(y))
+    ? facets.map(({x, y}) => groups.get(x)?.get(y) ?? [])
     : fx
-    ? facets.map(({x}) => groups.get(x))
-    : facets.map(({y}) => groups.get(y));
+    ? facets.map(({x}) => groups.get(x) ?? [])
+    : facets.map(({y}) => groups.get(y) ?? []);
 }
~~~

closes #1212

### DIFF
--- a/src/transforms/stack.js
+++ b/src/transforms/stack.js
@@ -89,7 +89,7 @@ function stack(x, y = one, ky, {offset, order, reverse}, options) {
       const Y2 = setY2(new Float64Array(n));
       const facetstacks = [];
       for (const facet of facets) {
-        const stacks = X ? Array.from(group(facet, (i) => X[i]).values()) : [facet];
+        const stacks = X ? Array.from(group(facet ?? [], (i) => X[i]).values()) : [facet];
         if (O) applyOrder(stacks, O);
         for (const stack of stacks) {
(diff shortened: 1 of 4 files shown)


mbostock:
Superseded by #1216.

PR #1219, opened 2023-01-11
pseudo blue noise

Fil:
use Job van der Zwan’s pseudo-blue noise (https://observablehq.com/@jobleonard/pseudo-blue-noise)

* the quality is so much better that it seems worth it to do a fourth WoS jump
* ~~there is a license issue, tbd with @jobleonard if we want to proceed~~
* add 5357 bytes to the build, 1181 bytes gzipped

the first commit could get in and allow users to opt in with their own noise generator; however I think we should just build this as a non-configurable default.

left: before; right: after
Capture d’écran 2023-01-11 à 11 32 02


Fil:
And for interleaved gradient noise (IGN, see https://blog.demofox.org/2022/01/01/interleaved-gradient-noise-a-different-kind-of-low-discrepancy-sequence/) you could do:

```
export function noise() {
  return function (x, y, z) {
    z *= 5.588238;
    x += z;
    y += z;
    return frac(52.9829189 * frac(0.06711056 * x + 0.00583715 * y));
  };
}
function frac(x) {
  return x - Math.floor(x);
}
```

interesting, but clearly inferior

ign

src/marks/raster.js
```diff
@@ -393,8 +394,8 @@ function blend(a, ca, b, cb, c, cc) {
 }
 
 function pick(random = Math.random) {
-  return (a, ca, b, cb, c) => {
-    const u = random();
+  return (a, ca, b, cb, c, cc, x, y, z) => {
+    const u = random(x, y, z);
```
mbostock, line 398: `z` is always zero here. The concept of a step count makes sense for the random-walk interpolation, but maybe for barycentric interpolation we should only pass the x and y.

src/marks/raster.js
```diff
@@ -375,8 +376,8 @@ function interpolateRandomWalk(index, width, height, X, Y, V) {
       if (x === 0.5) iy = ix;
       let distance; // distance to closest sample
       let step = 0; // count of steps for this walk
-      while ((distance = Math.hypot(X[index[iw]] - cx, Y[index[iw]] - cy)) > 0.5 && step < 2) {
-        const angle = random() * 2 * Math.PI;
+      while ((distance = Math.hypot(X[index[iw]] - cx, Y[index[iw]] - cy)) > 0.5 && step < 3) {
+        const angle = random(x, y, step) * 2 * Math.PI;
```
mbostock, line 380: Should we pass cx and cy here instead? Would it look better or worse?

```suggestion
        const angle = random(cx, cy, step) * 2 * Math.PI;
```

Fil: the quality looks about the same, but we must make sure they're non-negative

mbostock: I think conceptually it makes sense to use x and y because that is the sample we are generating; cx and cy is just an intermediate location. But I don’t really know what I’m talking about… 🤷

Fil: The noise generator creates a 128x128 texture with "random" numbers. Looking up the value on this texture (with x and y at the first step), neighboring points avoid using the angles that are too similar on their first jump—in a sense, they explore the space better.

Now, we also want two consecutive jumps to go in "well-different" directions—we don't want to continue in a straight line, it would defeat the purpose.

To guarantee that, we could use (cx, cy), when I try it, it seems to work as well. However, cx, cy can be anywhere on the plane, so we'd need to adjust the noise function to allow negative numbers (it's not complicated, but currently only non-negative numbers are accepted, and the solution is not a simple reflection of the positive values with Math.abs). Here instead I use z=step to look up a different area of the texture at each step.

Fil:
@JobLeonard no need to sign it in blood, but it would be lovely if you wanted to change the license on your notebook. We can continue the discussion on #1220.

PR #1220, opened 2023-01-12
blue noise

Fil:
supersedes #1219

TODO:
- [ ] shorter code?
- [ ] test more (dimensions of the noise, etc)


test/plots/index.js
```diff
@@ -284,6 +284,7 @@ export * from "./heatmap.js";
 export * from "./legend-color.js";
 export * from "./legend-opacity.js";
 export * from "./legend-symbol.js";
+export * from "./raster-blue-noise.js";
```
JobLeonard, line 287: It's not actually blue noise though, so perhaps better avoid calling it that even just in the file name. Otherwise someone will end up thinking it is.

JobLeonard:
By the way, I changed the license to ISC: https://observablehq.com/@jobleonard/pseudo-blue-noise

test/plots/raster-blue-noise.js
```diff
@@ -0,0 +1,107 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+// https://observablehq.com/@jobleonard/pseudo-blue-noise
+// CC-BY-SA
```
Fil, line 5:
```suggestion
```

PR #1224, opened 2023-01-13
treemap

mbostock:
Quick sketch of a treemap mark and transform. TODO

- [x] treemap layout
- [x] value option
- [x] tiling method option
- [ ] labels
- [ ] clipping for labels?
- [ ] aspect ratio based on chart dimensions? or on default height?
- [ ] rounding option (only makes sense when computing treemap in screen coordinates)
- [ ] option for showing internal nodes (nested or cascading treemap)

I kind of wonder if this should be an initializer so that it can access the chart dimensions (and since in any case the _x_ and _y_ scales won’t be meaningful).


Fil:
Love it!

As a random remark, if we let users specify their own tiling method, can we anticipate that some tiles might be generic polygons rather that rectangles (and maybe base this mark on Plot.geo in that case?). I'm thinking of @Kcnarf’s [voronoi treemaps](https://github.com/Kcnarf/d3-voronoi-treemap), of course.

We'll also need a strategy for faceting (maybe using frameAnchor decide where the facet with the smaller total aligns its treemap).



Link: this earlier experiment of a treemap plugin https://observablehq.com/@observablehq/plot-layouts-treemap-775 (#775).

Fil:
I'm pushing a few experiments in #1732. We'll also need a strategy for faceting; in particular, the surface area between facets should be comparable—a proportion reducer would equalize the total area to 1 on each facet.

PR #1226, opened 2023-01-16
ensure thresholds consistency across facets

Fil:
Using thresholds: 10 would create 10 different levels on each facet; this ensures they are consistent, like we do with the bin transform.


src/marks/contour.js
```diff
@@ -158,7 +158,12 @@ function contourGeometry({thresholds, interval, ...options}) {
         ? thresholds.range(...(([min, max]) => [thresholds.floor(min), max])(finiteExtent(VV)))
         : typeof thresholds === "function"
         ? thresholds(V, ...finiteExtent(VV))
-        : thresholds;
+        : typeof thresholds === "number"
+        ? ticks(...nice(...finiteExtent(VV), thresholds), thresholds)
```
mbostock, line 162: This isn’t sufficient. We’ll need these lines too (https://github.com/d3/d3-contour/pull/68):

https://github.com/d3/d3-contour/blob/0ba2dd047ce4cc2980025eff5c4fea989bf2da85/src/contours.js#L41-L42

src/marks/contour.js
```diff
@@ -158,7 +158,12 @@ function contourGeometry({thresholds, interval, ...options}) {
...
+        : isIterable(thresholds)
+        ? [...thresholds]
+        : null;
```
mbostock, line 165: Use arrayify here (with type = Array since d3-contour requires is strictly)?

```suggestion
        : arrayify(thresholds, Array);
```

mbostock (changes requested):
Thanks for catching this.

Did you consider switching to _contour_.contour directly since we are now effectively opting-out of d3-contour’s built-in ticks logic?

mbostock (approved):

PR #1227, opened 2023-01-16
combine prettier into main workflow

mbostock:
This seems to work. Example of prettier failure: https://github.com/observablehq/plot/actions/runs/3933377163


.github/workflows/node.js.yml
```diff
@@ -25,7 +25,8 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: |
           echo ::add-matcher::.github/eslint.json
-          yarn run eslint . --format=compact
+          yarn run eslint src test --format=compact
```
mbostock, line 28: I think the `eslint .` works in CI because we haven’t generated any files yet (whereas it will fail locally with errors on generated files in build and types). So, this is just being more explicit about what we’re testing so that it’s not confusing when you try to run this command locally.

Fil (approved):

PR #1231, opened 2023-01-18
Export DTS files so that dependent packages can see the types

Mitame:
I couldn't see a reason why this wasn't set, but setting it means that the types in `types/*` as exported by the module can be used by other typescript dependencies, i.e. you don't get the "this module doesn't have types" warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,6 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "types": "./types/index.d.ts"
 }


mbostock:
These types are still under development and are not ready to be exported yet.

Mitame:
Yeah, now that I've played around with it a bit more with the types on, I see what you mean.

mbostock:
TypeScript declarations were added in #1320 and will be available in the next release of Plot. 🙏

PR #1235, opened 2023-01-20
move Mark class

mbostock:
This is broken out of #1197 so that we can minimize the drift from main and make it easier to keep the axis mark branch up-to-date with changes. There are no functional changes in this PR; just moving code (and tweaking how we handle null marks, but the difference should not be externally visible).

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,9 @@
   "_moduleAliases": {
     "@observablehq/plot": "./src/index.js"
   },
-  "sideEffects": false,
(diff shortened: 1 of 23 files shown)


Fil (approved):
Out of curiosity, why the change to render: null?

mbostock:
> Why the change to render: null?

I just thought it was simpler (and faster) to filter the nulls out of the marks array, rather than promote them to Render marks whose render function does nothing.

PR #1238, opened 2023-01-26
Auto mark

tophtucker:
Adds Plot.auto, a shapeshifter mark that assumes the form of a bar, rect, line, area, or dot, depending on inferences from the data it’s passed. That is, you start from an intention about which dimensions of data you want to see, not what mark you want to see them in. In order to get you to a sensible chart even faster, it's even more opinionated than other marks. 

E.g., `Plot.auto(aapl, {x: "Volume"}).plot()` gives you a histogram, implicitly binning on x and setting y to the count reducer (corresponding to `Plot.rectY(aapl, Plot.binX({y: "count"}, {x: "Volume"})).plot()`).

**Plot.auto(data, options)**

The options are six "channels" and a mark override:
- _x_: corresponds to each mark's x channel; if specified without y, bins or groups on x and shows the count on y
- _y_: corresponds to each mark's y channel; if specified without x, bins or groups on y and shows the count on x
- _fx_: corresponds to each mark's fx channel
- _fy_: corresponds to each mark's fy channel
- _color_: corresponds to stroke (for line, rule, dot) or fill (for area, rect, bar, cell)
- _size_: corresponds to the dot's r channel; setting this always results in a dot mark
- _mark_: dot, line, area, rule, or bar; each option except dot includes x and y variants, and bar tries picks the appropriate mark from barX, barY, rectX, rectY, rect, or cell

The six "channels" take one of the following: 
- a string, which can be a field name or a reducer name
- an accessor function
- an object _{value: field, reduce: reducer}_, where field and reducer are strings or functions
 
Setting a reducer on x or y implicitly groups or bins on the other (y or x).

- [ ] Add unit tests for errors
- [x] Add more description error for `Plot.auto(aapl).plot()` than "mark is not a function"
- [ ] Pass through other channel options for e.g. bin thresholds
- [x] How to specify a constant color, either color: {constant: "red"} or maybeColorChannel?
- [ ] How to specify a z for multi-series line or area marks?
- [ ] How to limit and sort bar charts?
- [x] Provide a default sizeReduce for ordinal/ordinal case?
- [ ] Document "magical" behavior, and how it may change in a future version!


Fil:
I think this will lead us to add a new **color** option to all marks (where dot would have stroke default to color, and bar or rect would have fill default to color). Similar story for size and r, maybe.

I would also like `Plot.auto(data, options)` to be somehow guiding the user who has maybe forgotten to write .plot()?

(None of this is blocking.)

mbostock (commented):
Here are a few style nits on the code—because it’s easiest to comment on the color of the bike shed. One big-ish thing is whether we can avoid duplicating the `arrayIsPrimitive` logic/needing type inference to check whether the data is an array of primitives.

More importantly, it’d be helpful to recap whether there are significant outstanding tasks here? The list of the TODO’s in the code is different from the PR description, and in particular I wonder about these tasks:

* How to specify a constant color, either `x: {field: "red"}` or `maybeColorChannel`?
* How to specify a *z* for multi-series line or area marks?
* How to limit and sort bar charts?
* Provide a default sizeReduce for ordinal/ordinal case?

The first, at least, feels like a blocker. And maybe the second and third…

All that said, I’m so excited for this even more high-level API!! 🥳

Fil:
A challenge (also non-blocking) is to see how Plot.auto might impact the way people interact with the API. In particular, we want to make sure users will not rely on a specific outcome of plot.auto (for example if they use it to build a new system that has Plot.auto marks “deep inside”). We don't want our hands tied by the initial version. As Mike remarks, it might be enough to document visibly that it will not be considered a breaking change when the heuristic evolves over time to account for more (and refined) use cases.

tophtucker:
>I think this will lead us to add a new color option to all marks (where dot would have stroke default to color, and bar or rect would have fill default to color). Similar story for size and r, maybe.

In my testing, the alias "color" has felt pretty good for making it easier to maintain sensible encoding across marks. But I'm not sure about the alias "size" for "r". From showing a couple people, it feels like the name "size" is deceptively generic; they expect it to somehow affect the size of any mark (without knowing exactly what that would mean), when it fact it always makes a dot.

>I would also like Plot.auto(data, options) to be somehow guiding the user who has maybe forgotten to write .plot()?

I think the inspector for any mark could be more helpful, but that feels like a separate issue. (Where's your nice notebook where anything can expose a ".toInspect" method that draws a custom inspector? 😝)

>we want to make sure users will not rely on a specific outcome of plot.auto

It'd help if we had the `Plot.auto(…).source()` method we've discussed to eject to a fixed mark.

src/marks/auto.js
```diff
@@ -0,0 +1,338 @@
...
+      options = binX(transformOptions, options); // bin x, group y
+    }
+  }
+
+  return marks(
+    // x: { zero: xZero },
+    // y: { zero: yZero },
+    // color: colorValue == null && colorReduce == null ? null : { legend: true },
+    // marks: [
+    fx || fy ? frame({stroke: "#eee"}) : null,
```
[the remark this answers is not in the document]
tophtucker: I don't have a strong opinion on whether we have a frame by default. I think especially with faceted dots I find it hard to see the boundaries; for someone unfamiliar with faceting, I think the frame makes it more clear what it's doing. With bars it's less of an issue. (But I'll definitely at least update to match grid.)

image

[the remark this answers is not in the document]
Fil: I find myself adding a frame almost every time I use facets and quantitative x or y, because there's a kind of "visual continuity" that needs to be broken. (When x and y are ordinal it's ok not to have it.)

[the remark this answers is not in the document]
mbostock: Okay, let’s do it. Seems like maybe we should also file a feature request to have Plot do this by default when faceting though!

mbostock:
> It'd help if we had the `Plot.auto(…).source()` method we've discussed to eject to a fixed mark.

I feel this is out-of-scope for Plot; Plot shouldn’t be responsible for code generation.

src/marks/auto.js
```diff
@@ -0,0 +1,328 @@
...
+  let {reduce: xReduce} = x;
+  let {reduce: yReduce} = y;
+  if (xReduce === undefined)
+    xReduce = yReduce == null && xValue == null && sizeValue == null && yValue != null ? "count" : null;
+  if (yReduce === undefined)
+    yReduce = xReduce == null && yValue == null && sizeValue == null && xValue != null ? "count" : null;
+
+  const {zero: xZero} = x;
+  const {zero: yZero} = y;
+  const {value: colorValue, reduce: colorReduce} = color;
```
mbostock, line 47: Let’s support this for unambiguously specifying a constant color:

```js
Plot.auto(penguins, {x: "culmen_length_mm", color: {color: "red"}})
```

And then the shorthand string color needs to disambiguate. The above is the same as this:

```js
Plot.auto(penguins, {x: "culmen_length_mm", color: "red"})
```

Whereas

```js
Plot.auto(penguins, {x: "culmen_length_mm", color: "foo"})
```

would be shorthand for

```js
Plot.auto(penguins, {x: "culmen_length_mm", color: {value: "foo"}})
```

src/marks/auto.js
```diff
@@ -0,0 +1,338 @@
...
+      case "dot":
+        mark = dot;
+        stroke = color;
+        strokeReduce = colorReduce;
+        break;
+      case "line":
+        mark = x && y ? line : x ? lineX : lineY; // 1d line by index
+        stroke = color;
+        strokeReduce = colorReduce;
+        if (new Set(color).size > color?.length >> 1) {
```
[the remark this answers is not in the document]
mbostock:
```js
function isHighCardinality(value) {
  return new InternSet(value).size > value?.length >> 1;
}
```

src/marks/auto.js
```diff
@@ -0,0 +1,328 @@
...
+  // behavior depends on the actual values anyway, so this probably is
+  // what we want. By materializing the columns here, we also ensure
+  // that they aren’t re-computed later in Plot.plot.
+  x = valueof(data, xValue);
+  y = valueof(data, yValue);
+  color = valueof(data, colorValue);
+  size = valueof(data, sizeValue);
+  fx = valueof(data, fxValue); // TODO Should we still materialize if heuristic doesn't depend on it?
+  fy = valueof(data, fyValue);
+
```
mbostock, line 72: Something like this for the default sizeReduce in the ordinal-ordinal case?

```js
if (sizeReduce === undefined && sizeValue == null && xReduce == null && yReduce == null && isOrdinal(x) && isOrdinal(y)) {
  sizeReduce = "count";
}
```

I dunno if you also want to check that `mark === undefined || mark === "dot"`… maybe?

src/style.js
```diff
@@ -244,9 +244,13 @@ function groupAesthetics({
...
 }
 
+export function isHighCardinality(value) {
+  return new InternSet(value).size > value?.length >> 1;
+}
+
 export function groupZ(I, Z, z) {
   const G = group(I, (i) => Z[i]);
-  if (z === undefined && G.size > I.length >> 1) {
+  if (z === undefined && isHighCardinality(I.map((i) => Z[i]))) {
```
mbostock, line 253: We should leave this line as-is; reusing the isHighCardinality function here is extra work (because we have to create an InternSet when we’ve already done it with `G`).

mbostock (approved):
Needs documentation, and will need to be merged with the axis mark changes now in main (meaning all the test snapshots will need to be regenerated), but this looks great! We can followup with some of the anticipated improvements in future! 🎉

src/marks/auto.js
```diff
@@ -0,0 +1,269 @@
...
+import {cell} from "./cell.js";
+import {frame} from "./frame.js";
+import {bin, binX, binY} from "../transforms/bin.js";
+import {group, groupX, groupY} from "../transforms/group.js";
+import {marks} from "../mark.js";
+import {ascending} from "d3";
+
+export function auto(data, {x, y, fx, fy, color, size, mark} = {}) {
+  // Shorthand: array of primitives should result in a histogram
+  if (x === undefined && y === undefined) x = identity;
```
tophtucker, line 18: Last blocker, IMO: I think this line for the "shorthand" is now _too_ permissive. We probably want to check that the first element is a primitive, or else throw an error if no options are specified? Currently `Plot.auto(aapl).plot()` looks like this:

image

mbostock: I suggest we drop this shorthand entirely for now. Maybe Plot.auto returns null or throws an error? I dunno.

tophtucker: Yeah probably. If I just remove the line it looks like this:

image

Which I think looks broken. So I added an error:

image

mbostock:
Filed #1252 to track auto mark fast follows.

PR #1244, opened 2023-02-02
fix regression

Fil:
where scales defined by a domain or a type (with default domain) where considered absent.

Capture d’écran 2023-02-02 à 12 22 43

bug introduced in 93b1a7782958ed16a1e28b778d1e9266c55e944c, which is bit too radical in eliminating x based only on its type.


src/plot.js
```diff
@@ -608,7 +608,8 @@ function hasAxis(marks, k) {
   return marks.some((m) => m.ariaLabel?.startsWith(prefix));
 }
 
-function hasScale(marks, k) {
+function hasScale(k, {domain, type}, marks) {
+  if (domain !== undefined || type != undefined) return true;
```
mbostock, line 612:
```suggestion
  if (domain != null || type != null) return true;
```

mbostock (approved):

PR #1247, opened 2023-02-02
per-channel scale override, and “auto” scale

### LINKED ISSUE #1246: Opt-out of the color scale on a channel-by-channel basis
Currently we have some special logic that checks to see if the color channels are literal colors:

https://github.com/observablehq/plot/blob/b97ba31a50062317116547bd0d3114023172aa63/src/scales.js#L411-L417

This works well in the simple case where you have one (or more) marks that wants to specify literal colors.

But what if you want to have some marks that use a color scale, and other marks such as grid lines that want to specify literal colors? It would be nice if you could opt-out of the color scale by providing literal colors on a channel-by-channel basis. Essentially, instead of adopting an _identity_ color scale if all color channels are literal colors, we would instead ignore the `scale: "color"` declaration if the provided channel values are already literal colors.

mbostock:
Fixes #1246. Consider this example from usCongressAgeColorExplicit where you want the _stroke_ to go through the color scale, but for _fill_ you want to specify a literal color:

```js
Plot.dot(
  data,
  Plot.stackY2({
    x: (d) => 2021 - d.birth,
    stroke: "gender",
    fill: (d) => (d.gender === "F" ? "rgb(132, 165, 157)" : "#f6bd60"),
    title: "full_name"
  })
),
```

Before:

Screenshot 2023-02-02 at 1 01 11 PM

After:

Screenshot 2023-02-02 at 1 02 30 PM


mbostock (commented):
I’m maybe being too aggressive here by removing the checks on the scale domain—do you think we still want to default to the _identity_ scale if the provided domain is colors or symbols? Feels like an edge case but may be worth doing for backwards-compatibility. Sadly we don’t have any tests for that behavior!

Fil:
It will help users a lot, I think, in particular for the case you're showing where you want _fill_ to be a scaled channel, and _stroke_ to be either black or white.

🤔 However this might open to ambiguities in certain cases? Suppose you have two teams "team red" & "team cat". The first mark (or channel) represents both teams—so it will use a categorical color scale, but the second mark (or channel) only represents the "red" team, for some reason (say, a filter). Will team red be represented by the categorical color for the first mark, and the color "red" for the second mark? How might we let users opt-in (or out) explicitly?

An alternative (thinking out loud) would be a helper Plot.color(_x_) which would return "something" that would always be scaled as _x_, independently of the color scale. Maybe a {color:"red"} object.

mbostock:
> However this might open to ambiguities in certain cases?

Yes. I don’t think ambiguity in itself is a concern, but the problem here is that there is no way to resolve the ambiguity—there’s no way to disambiguate. In the current release of Plot you can do that by setting the type (or scheme or range) of the color scale (to _identity_ if you want literal colors, or to _linear_ or _ordinal_ to encode abstract data). There’s no equivalent syntax in this PR for saying which one you want.

I think we might want a way to override whether or not a channel is bound to a scale, or perhaps even which scale it is bound to. The latter might allow us to support dual-axis charts #147, too (not that I’m particularly excited about such a thing, but it might be sometimes useful).

mbostock:
What if there were a **scale** option that let you specify the association of channel name with scale name? And, if you don’t specify what scale is associated with a given channel, the mark can specify a default (the current behavior). This way you can explicitly enable or disable a scale on a channel-by-channel basis.

```js
Plot.dot(
  data,
  Plot.stackY2({
    x: (d) => 2021 - d.birth,
    stroke: "gender",
    fill: (d) => (d.gender === "F" ? "rgb(132, 165, 157)" : "#f6bd60"),
    title: "full_name",
    scale: {stroke: "color", fill: null}
  })
)
```

We could also support true (automatically chose the default scale for this channel) and false (equivalent to null).

Fil:
Sounds good! Very explicit. It would be nice also to opt out of projection (e.g. to position a label somewhere on the screen without having longitude/latitude coordinates).

mbostock:
The latest commit adds a mark **scale** option to override the association of channels and scales. I think there’s still some more work to do here, though…

* If you reference a scale that doesn’t exist, it should throw an error?
* Should we use `fill: {value: (d) => …, scale: null}` instead of `fill: value, scale: {fill: null}`?

mbostock:
@Fil that seems appealing, but I expect there are other places where we call Plot.valueof with a channel definition and we expect it to work. Those places will now need to check for a `{value}` object, too. I think it might be more appropriate to parse the `{value, scale}` object earlier in the Mark constructor. I will look into that.

mbostock:
Okay, I think this is good now.

The _fill_, _stroke_, and _symbol_ channels are now declared with the _auto_ scale (rather than the _color_ and _symbol_ scales, respectively); this gives an indication that the scale should be inferred automatically based on the channel name and provided values, _i.e._ if literal values are provided, then no scale will be applied.

Furthermore, you can now specify any channel as a {value, scale} object to indicate which scale should be bound to the channel, overriding the mark’s default. For example you can set the scale to null to opt-out of the scale, or set it to _color_ to opt-in.

src/mark.js
```diff
@@ -41,11 +41,20 @@ export class Mark {
...
-        if (value != null) return true;
-        if (optional) return false;
-        throw new Error(`missing channel value: ${name}`);
-      })
+      Object.entries(channels)
+        .map(([name, channel]) => {
+          const {value} = channel;
+          if (isOptions(value)) {
+            channel = {...channel, value: value.value};
+            if (value.scale !== undefined) channel.scale = value.scale;
```
Fil, line 49: Shouldn't we guard against bad scale values?

For example, we want to throw an error if the user passes
`stroke: {value: "field", scale: "x"}`
or 
`strokeOpacity: {value: Math.random, scale: "x"}`

I was thinking we should test that `channel.scale === "auto"`, and that value.scale is one of ["color", null] if name is stroke or color, and one of ["symbol", null] if name is symbol. And error in all other cases?

Fil (changes requested):
It seems that we should validate user input more strictly.

I've added a bit of documentation, I hope it helps but it's not super clear (which is why I put it in a different branch).

Also we'll need a few unit tests, and as I was fiddling with the hexbin-symbol.js test, I got errors (an Object appeared in the legend, and the symbols didn't respect the "value"). Maybe something doesn't work well with binning?

src/mark.js
mbostock: You’re saying we should restrict which scales we allow to be bound to which channels? Why? Is that likely? What if someone invents a creative use for doing something we didn’t anticipate?

I do think it’d be reasonable to enforce that the scale, if non-nullish, is a valid scale name (and not, say `scale: "foo"`). But I’m not sure we should do any validation beyond that.

Fil: Yes. I don't think the symbol or color channel can be assigned to *x*, even in a creative way—since their ranges are not in the same space? x, y, r and length are numbers, so there might be cases where you'd want to mix those, maybe—but most probably it's going to be a mistake, or a misunderstanding. (I also tried true, false… these will be fixed if we make sure that scale is nullish or a key of scales.)

Fil:
Here are two examples that create scales where they should opt out; possibly only due to the hexbin initializer (I'll dig more into it tomorrow).
https://observablehq.com/d/093e2a39e16bb27c

mbostock:
I’ve fixed the issues with initializers that you reported. Thanks @Fil!

Fil (approved):
To answer my own question, this also allows a mark to ignore a projection and give x, y as pixel values. You just need to opt out of both scales:
```js
Plot.plot({
  projection: "mercator",
  marks: [
    Plot.graticule(),
    Plot.text(["hello"], {x: {value: [200], scale: null}, y: {value: [100], scale: null}})
  ]
})
```
This will be particularly useful for legends.

All that remains is the README.

Issue #1254, opened 2023-02-04
Spurious facet anchors


Fil:
there are two spurious axes on this faceted chart:
![spurious](https://user-images.githubusercontent.com/7001/216758339-edb2ed70-55cf-4fbc-b47c-c4886220bd37.png)
```
Plot.auto(penguins, {
  x: "body_mass_g",
  y: "culmen_length_mm",
  fx: "island",
  fy: "species"
}).plot()
```

(noted by @tophtucker )

PR #1255, opened 2023-02-04
avoid spurious facet anchor

Fil:
fixes https://github.com/observablehq/plot/issues/1254

When a facet is empty, we could return undefined and repeat the test on f.empty after the `??`. However it felt more logical to return false, as we have “settled the matter”, so to speak.
https://github.com/observablehq/plot/blob/a7f76ef66fc7f7e6d6727e935ffdd370e69013b8/src/plot.js#L257

### DIFF
--- a/src/facet.js
+++ b/src/facet.js
@@ -117,42 +117,38 @@ function facetAnchorRight(facets, {x: X}, {x}) {
   return X ? X.indexOf(x) === X.length - 1 : true;
 }
 
-function facetAnchorTopEmpty(facets, {y: Y}, {x, y}) {
+function facetAnchorTopEmpty(facets, {y: Y}, {x, y, empty}) {
+  if (empty) return false;
   const i = Y?.indexOf(y);
-  if (i > 0) {
-    const y = Y[i - 1];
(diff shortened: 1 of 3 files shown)


src/facet.js
```diff
@@ -155,6 +147,10 @@ function and(a, b) {
   };
 }
 
+function isEmptyFacet(facets, {x, y}) {
```
mbostock, line 150: This code de-deplication isn’t related to this PR; it would be easier to review if you just made the minimal change necessary to fix the bug (adding the early return if (empty)).

mbostock (approved):
Thanks!

test/plots/autoplot.js
```diff
@@ -210,5 +210,5 @@ export async function autoDotFacet2() {
     fx: "island",
     fy: "species",
     color: "sex"
-  }).plot();
+  }).plot({marginRight: 70, x: {ticks: 5}});
```
mbostock, line 213: This is another unrelated change (and also we should be using facet.marginRight, and I think it’s good as previously stated to have the Plot.auto tests be as short has possible since the focus is on speed and not looking good).

PR #1261, opened 2023-02-06
Auto mark passes through other options to mark

tophtucker:


mbostock (changes requested):
Yea, this is what I was imagining. But I think we need to give this some more thought. For example, what happens if you specify the **fill** or **stroke** option? Should those be ignored? Should they only be ignored if you specify the **color** option? Or should they take precedence over **color**? Etc.

I think first we should survey the standard mark options, as well as the options that are potentially supported by all the mark types that are eligible for use by the auto mark (which is arguably all of ’em) and see if there are any weirdnesses.

The other big question is whether these should go into the `transformOptions` _i.e._ before the transform is invoked, so that _e.g._ you can specify the number of thresholds you want for the bin transform, rather than doing it after the transform and only passing it directly to the mark.

tophtucker:
Stale, too naive an approach, I’m closing; tracked by https://github.com/observablehq/plot/issues/1631.

PR #1262, opened 2023-02-07
default frame for facets

### LINKED ISSUE #1251: Generate a default (subtle) frame when faceting
Ref. https://github.com/observablehq/plot/pull/1238#discussion_r1091025411

image

Could be disabled by setting _facet_.**frame** to false, similar to the top-level **line** option.

mbostock:
Fixes #1251. I think we need some additional smarts, though…

- [ ] Don’t draw an implicit frame if there’s an explicit frame
- [ ] Don’t draw an implicit frame if the *x* or *y* scale is missing? (_e.g._, not athletesSampleFacet)
- [ ] Don’t draw an implicit frame if there’s a *projection*?

Needs more thought. 🤔


tophtucker:
I agree with Fil that it feels like it depends on the mark, so it might be harder to do in Plot generally than in Plot.auto. (I feel like I add frames only for dots and lines.)

This is too subtle to base any heuristic on, but I also feel like the frame matters more when you have a quantitative scale that doesn't start at zero (because people spot and parse 0 very easily), or isn't nice (because the ticks neatly bracketing the domain help you see the "gestalt").

Issue #1263, opened 2023-02-08
Finalize the design of the render API


Fil:
The render API still has a few quirks that we might want to address:

### Scales

The _scales_ argument is an object of D3 scale functions. This is inconsistent with how we expose the scales elsewhere. Currently internal marks need the scales for a few operations:
- to scale values out of band (for example, on X and Y for scaleProjection in the geo mark, and on color for the raster/contour/density mark); this can be replaced by _scale_.apply
- _isCollapsed_ tests the materialized domain (exposed as _scale_.domain)
- _applyTransform_ needs the materialized bandwidth (in the bar, tick marks…), it's exposed as _scale_.bandwidth
- axes want to call the scale.ticks function _which is not exposed_
- …more which I haven't located yet?…

If we passed the same objects that _plot_.scale("x") returns (_i.e._ scale descriptors), we need to pass the "ticks" function, or to recreate it from the descriptor.

### Facets

The _index_ passed to render is decorated with the facet information : the values of _fx_ and _fy_ for the current facet, as well as a facet index _fi_. (_fi_ is used internally for the raster mark.) This needs to be documented. Example [here](https://observablehq.com/@observablehq/autoplot-matrix).

### svg, parent element

Passing the svg, and the parent element, could help to build marks that need to use these (for instance, to add defs to the svg). They could be referenced in the _context_ argument. For example, #1304 experiments with passing the svg in the context. (This was done in 0.6.7.)

### Event listeners

A mark wants to listen to events (pointermove…), and send events (setting a value and sending a CustomEvent(input)). Needs an "official" api support rather than the mark fishing for the figure element. (Done in 0.6.7.)

### Update

Farther in the future, we might want to have a strategy to update a “layer”; this could happen by passing the previously returned node.

### Documentation

Once this is is decided and done (one way or another), a documentation notebook with examples will help advanced users who want to extend Plot.

(for an earlier version of this issue, see #501)

mbostock:
(Clarified the issue title to make clear that we already have a render API, but that the design isn’t finalized and has the described quirks and outstanding questions.)

mbostock:
Closing as we have done most of the work here, though there may be follow-up feature enhancements.

PR #1265, opened 2023-02-09
pass the "exposed scales" to the render function

Fil:
This cleans up the render API (#1263), as the render function now receives the same scale object as is exposed through the _plot_.scale("x") API.

It means a bit of extra work, as axes have to compute their own tickFormat and ticks, and a special case for the x and y identity scales. But it's worth it, in the name of consistency.

The difference that I still need to reduce is the scale’s label, which we censor when exposing the scale… I think this could change (we could expose it)… alternatively, we could try and pass the label through a separate argument.

*todo*
- [ ] decide whether we expose the scale's label or censor it

It would be a breaking change for, _e.g._, https://observablehq.com/@observablehq/plot-calendar, which reads _x_.bandwidth() and would need to use _x_.bandwidth instead (number).

### DIFF
--- a/src/axes.js
+++ b/src/axes.js
@@ -1,11 +1,6 @@
 import {format, utcFormat} from "d3";
 import {formatIsoDate} from "./format.js";
 import {constant, isTemporal, string} from "./options.js";
-import {isOrdinalScale} from "./scales.js";
-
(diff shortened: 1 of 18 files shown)


Fil:
superseded by #1810

Issue #768, opened 2022-02-19
Smarter formatting for year channels?


mbostock:
In a case like this, it’d be nice to avoid the commas for the year axis:

![untitled-249](https://user-images.githubusercontent.com/230541/154818415-71568432-db1e-4602-9945-86a2fafe9211.png)

```js
Plot.plot({
  width,
  color: {legend: true},
  marks: [
    Plot.rectY(overview, Plot.binX({y: "sum"}, {x: "Year", y: "Value", fill: "Type", interval: 1}))
  ]
})
```

Of course, you can do it with `x: {tickFormat: ""}`, but could Plot figure this out automatically?

Similarly when you do something like `title: "Year"`, it’s a bummer that the automatic formatting for numbers shows a comma. I think we could maybe track a hint that looks for (case-insensitive) “year” and avoids the comma.

Fil:
related #355

Fil:
In fact this is one of the first questions that came up in a workshop last Friday. But it was happening in Inputs.table, we hadn't started plotting the data yet. The offending field was called "annee" (French for year).

arky:
I think this is most common problem/gotcha for those starting with plot. This handling could reduce friction for new users who are not familiar with JavaScript data handling features. 

I had posted exactly this question in forums: 
https://talk.observablehq.com/t/handling-date-column-during-file-import/6333

https://observablehq.com/@arky/disasters-in-south-eastern-asia-1900-2021

eagereyes:
Here's [a related paper](https://research.tableau.com/sites/default/files/TR-2018-01.pdf). I wonder if there are implementations of something like this out there in open-source land, it's such a common issue.

Fil:
Question: a possible strategy would be for the default format to depend on the domain. If the domain is contained in, say, [1500, 2200], set the default formatter to be `d => '${d}'` rather than `Intl.NumberFormat`. This heuristic might be slightly surprising in the odd case, but it would fix the very common issue of years being poorly formatted, while still giving nice numbers by default in general.

mbostock:
> If the domain is contained in, say, [1500, 2200], set the default formatter to be d => '${d}' rather than Intl.NumberFormat.

We should also ensure that all channel values are integers (or at least sample the first ~40? channel values). If we see a fractional value such as 1500.1231041240 we should use the default number format rather than assuming it is a year.

mbostock:
[...]

PR #1268, opened 2023-02-10
First pass at #768, works for linear-integer

Fil:
TODO
* should trigger on the interval option being integer
* band-integer (e.g., yearly-request)
* find a nicer way to pass the information up
* test for the [-10, 10] cases—maybe more what about 10-30?


mbostock (commented):
What if we did this by introducing a new _year_ scale type? Then we could change the inferScaleType function to choose _year_ over _linear_ if the data matches. We already have a bunch of inference in inferScaleType (rather than mixing inference into Scales), and using a new scale type gives an explicit way for the user to opt-in or opt-out of this behavior.

I guess we’d need _year-point_ and _year-band_ for the ordinal case? Or maybe a separate year option? I dunno. I’ll need to think about that one.

mbostock:
Can we do this without coupling it to #1265? I’m looking for smaller wins.

PR #1271, opened 2023-02-15
document gridY, gridFx, gridFy

### LINKED ISSUE #1270: The grid constructors are documented as “…same…”
We should copy the definition from Plot.gridX and adapt it appropriately rather than saying “…same…”, at least.

https://github.com/observablehq/plot/blob/main/README.md#grid

Fil:
closes #1270


mbostock (approved):
Thank you!

PR #1273, opened 2023-02-15
only default to bar for x or y is zeroed

### LINKED ISSUE #1272: The auto mark should not default to bars when the reduce is not sum or count (e.g., don’t stack means)

mbostock:
Fixes #1272.

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -97,19 +97,13 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
     mark =
       sizeValue != null || sizeReduce != null
         ? "dot"
-        : colorReduce != null // e.g., heatmap
+        : xZero || yZero || colorReduce != null // histogram or heatmap
         ? "bar"
         : x && y
         ? isContinuous(x) && isContinuous(y) && (xReduce != null || yReduce != null || isMonotonic(x) || isMonotonic(y))
           ? "line"
-          : (isContinuous(x) && xZero) || (isContinuous(y) && yZero)
-          ? "bar"
           : "dot"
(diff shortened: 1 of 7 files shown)


Fil (approved):
lgtm!

BTW I'm not sure #1238 has documented the reducer's **zero** option? I can't see it.

mbostock:
I think this is _better_, but I think we could still refine this logic. Like, it makes sense to me that this is a line:

Screenshot 2023-02-15 at 9 16 00 AM

```js
Plot.auto(weather, {x: "date", y: {value: "temp_max", reduce: "mean"}}).plot()
```

But I’m not convinced that we should switch to a bar when you tell it that zero is meaningful:

Screenshot 2023-02-15 at 9 16 48 AM

```js
Plot.auto(weather, {x: "date", y: {value: "temp_max", reduce: "mean", zero: true}}).plot()
```

I think a line + zero rule would be equally appropriate (and less arbitrarily different):

Screenshot 2023-02-15 at 9 17 42 AM

```js
Plot.auto(weather, {x: "date", y: {value: "temp_max", reduce: "mean", zero: true}, mark: "line"}).plot()
```

Or perhaps even an area:

Screenshot 2023-02-15 at 9 18 53 AM

```js
Plot.auto(weather, {x: "date", y: {value: "temp_max", reduce: "mean"}, mark: "area"}).plot()
```

On the other hand, I think it would be surprising if histograms rendered as lines or areas by default, but I can’t articulate why.

Screenshot 2023-02-15 at 9 20 21 AM

```js
Plot.auto(weather, {x: "temp_max", mark: "line"}).plot()
```

PR #1276, opened 2023-02-15
fix bars with maybe-ordinal reducers

mbostock:
Fixes another bug where we assumed that reducers produce quantitative values. In some cases, _e.g._ with the _mode_ reducer, reducing preserves the input value type. For example:

```js
Plot.auto(penguins, {x: "island", y: {value: "species", reduce: "mode"}, mark: "bar"}).plot()
```

Before, barY is wrongly chosen, resulting in a degenerate plot:

Screenshot 2023-02-15 at 1 07 58 PM

After, cell is correctly chosen, since both _x_ and _y_ are ordinal:

Screenshot 2023-02-15 at 1 08 53 PM

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -51,7 +51,7 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
   // TODO Limit and sort for bar charts (e.g. alphabet)?
   // TODO Look at Plot warnings and see how many we can prevent
   // TODO Default to something other than turbo for continuous? Like:
-  //      scheme: (colorValue && isContinuous(color)) || colorReduce ? "ylgnbu" : undefined
+  //      scheme: (colorValue && !isOrdinal(color)) || colorReduce ? "ylgnbu" : undefined
 
   // To apply heuristics based on the data types (values), realize the columns.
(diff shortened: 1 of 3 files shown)


src/marks/auto.js
```diff
@@ -100,9 +100,9 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
         : xZero || yZero || colorReduce != null // histogram or heatmap
         ? "bar"
         : x && y
-        ? isContinuous(x) && isContinuous(y) && (xReduce != null || yReduce != null || isMonotonic(x) || isMonotonic(y))
-          ? "line"
-          : "dot"
+        ? isOrdinal(x) || isOrdinal(y) || (xReduce == null && yReduce == null && !isMonotonic(x) && !isMonotonic(y))
+          ? "dot"
+          : "line"
```
mbostock, line 105: (This is an unrelated change where I just wanted to get rid of isContinuous by replacing it with !isOrdinal. Sorry for the churn but hopefully it’s clear that this is not changing behavior!)

tophtucker (approved):
I love how this makes us think about the type signature of reducers like this! 😄

Issue #1252, opened 2023-02-03
Auto mark fast follows


mbostock:
Fast follows from #1238:

- [ ] Pass through other channel options, _e.g._ thresholds for binning
- [ ] Pass through other mark options, _e.g._ strokeWidth
- [ ] How to specify _z_ for multi-series line or area marks?
- [ ] How to limit and sort bar charts?

tophtucker:
Maybe lower-priority, but not to be forgotten: 
- [ ] "Shorthand" (make it work for passing arrays of primitives)
- [ ] Better default unconfigured state for `Plot.auto(aapl).plot()` than an error ([notes](https://observablehq.com/d/bde38bbf32f27fec))

tophtucker:
It’s a little hard to say why, but this should be lines, not rects:

<img width="893" alt="image" src="https://user-images.githubusercontent.com/841829/216683850-ce052aad-3710-41dc-adf4-9a705e83f7b4.png">

Is it because date_of_birth is continuous, albeit not monotonic, and reduce is something other than count? Implicitly stacking means feels wrong.

tophtucker:
An explicit size reducer should stop us from setting an implicit y reducer:

<img width="575" alt="image" src="https://user-images.githubusercontent.com/841829/216684494-f5688a5a-ab6e-412b-b560-e6e1c9b3666d.png">

mbostock:
> reduce is something other than count

Yes, we should only implicitly stack when the reduce is count, not when it’s something else. You can always add counts but adding means/medians/etc. often doesn’t make sense.

Fil:
_sum_ is also additive :-) I'm adding it to the TODO.

tophtucker:
Setting size to an ordinal dimension shows nothing. That makes sense, since `Plot.dot(data, {x: "culmen_depth_mm", r: "sex"}).plot()` shows nothing either, and I'm not sure what it'd mean… maybe size by the value's index in the domain? But it feels broken when I'm trying every combination and every other one works. Could also throw an error.

<img width="691" alt="image" src="https://user-images.githubusercontent.com/841829/217330422-085aabe3-45ee-4f4f-a930-a04312b8bfe9.png">

mbostock:
> Setting size to an ordinal dimension shows nothing.

Related #493; the data here is being filtered because the _size_ channel implicitly coerces to numbers producing NaN.

tophtucker:
Should we throw an error when a column isn't found in the data?

[...]

PR #1277, opened 2023-02-15
auto bin options

mbostock:
Related #1261 #1252, allows _x_- and _y_-specific options to be passed through when binning. This is primarily useful for control over thresholds. Given the ambiguities raised in https://github.com/observablehq/plot/pull/1261#pullrequestreview-1285786800, I’m only doing this for the bin transform; for example you still can’t use scale overrides added in #1247. (I think that’s okay because scale overrides are fairly esoteric, and you can switch to a concrete mark type if needed.)

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -29,16 +29,12 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
   if (isOptions(fx)) ({value: fx} = makeOptions(fx));
   if (isOptions(fy)) ({value: fy} = makeOptions(fy));
 
-  const {value: xValue} = x;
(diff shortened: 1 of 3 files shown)


mbostock:
Also note that you can’t use top-level options for binning, _e.g._,

```js
Plot.auto(weather, {x: "date", thresholds: d3.utcMonth, y: {value: "temp_max", reduce: "mean"}}).plot()
```

That would be nice to add as part of #1261 but we’ll have to add the aforementioned ambiguities.

tophtucker (approved):
👍 Seems like a fine cutoff for now. (I hadn't thought about how bin has options but group doesn't! A telling asymmetry!!)

PR #1279, opened 2023-02-16
minimal support for bigint

### LINKED ISSUE #1278: Support bigints?
Can we just change this:
```diff
function coerceNumber(x) {
-  return x == null ? NaN : +x;
+  return x == null ? NaN : Number(x);
}
```
(doing so does not break a single test, but there might be implications I haven't seen?)

Ref. https://github.com/observablehq/plot/discussions/1099; the use case is to facilitate passing database results to Plot.

Fil:
closes #1278

@llimllib see if you can break this branch by passing bigints to unsuspected places (a build is available at https://observablehq.com/@observablehq/plot-bigint-1278 )… note that we're supporting bigints only as channel values, not as options (_e.g._ we don't try to support `inset: 1n` or `margin: 10n`).

As I was researching all the places where we coerce values to a number, I realized that this line is probably redundant, as we enter inferLogDomain only with already-coerced channels:
https://github.com/observablehq/plot/blob/d8378f088fa39d611f3219992ad5e452418765b6/src/scales/quantitative.js#L293

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
   "plugins": ["@typescript-eslint"],
   "parser": "@typescript-eslint/parser",
   "env": {
-    "es6": true,
+    "es2020": true,
     "node": true,
     "browser": true
   },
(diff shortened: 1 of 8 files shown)


llimllib:
I'll try to take a hack at it this afternoon 👍

mbostock (approved):
I suspect there are some further tweaks we can make to coercion but this seems like a good incremental step forward!

PR #1282, opened 2023-02-17
avoid a crash on facet: exclude on a non-faceted mark

### LINKED ISSUE #1281: Throw a better error when facet: "exclude" is used without faceting?
This current throws TypeError: Cannot read properties of undefined (reading 'facetsIndex').

```js
Plot.plot({
  marks: [
    Plot.dot(penguins, {x: "culmen_depth_mm", y: "culmen_length_mm", fill: "#ccc", facet: "exclude"}),
    Plot.dot(penguins, {x: "culmen_depth_mm", y: "culmen_length_mm", fx: "sex", fy: "species"})
  ]
})
```

Since no top-level or mark-level faceting is declared, `facet: "exclude"` should probably throw a specific helpful error instead of crashing internally.

Fil:
closes #1281

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -100,8 +100,7 @@ export function plot(options = {}) {
     for (const mark of marks) {
       if (mark.facet === "exclude") {
         const facetState = facetStateByMark.get(mark);
-        if (facetState === undefined) throw new Error(`the facet: "exclude" option is restricted to top-level facets`);
-        facetState.facetsIndex = facetExclude(facetState.facetsIndex);
+        if (facetState !== undefined) facetState.facetsIndex = facetExclude(facetState.facetsIndex);
       }
     }
   }
--- a/src/plot.js
+++ b/src/plot.js
@@ -100,6 +100,7 @@ export function plot(options = {}) {
     for (const mark of marks) {
       if (mark.facet === "exclude") {
         const facetState = facetStateByMark.get(mark);
+        if (facetState === undefined) throw new Error(`the facet: "exclude" option is restricted to top-level facets`);
         facetState.facetsIndex = facetExclude(facetState.facetsIndex);
       }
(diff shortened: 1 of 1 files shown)


src/plot.js
```diff
@@ -100,6 +100,7 @@ export function plot(options = {}) {
     for (const mark of marks) {
       if (mark.facet === "exclude") {
         const facetState = facetStateByMark.get(mark);
+        if (facetState === undefined) throw new Error(`the facet: "exclude" option is restricted to top-level facets`);
```
mbostock, line 103: This isn’t accurate—you can also use `facet: "exclude"` if you specify mark-level faceting, e.g.

```js
Plot.dot(penguins, {facet: "exclude", fx: "sex", fy: "species", x: "culmen_depth_mm", y: "culmen_length_mm", r: 2, fill: "#ddd"}),
```

mbostock (changes requested):

Fil:
We should be consistent when the **facet** option is activated (ie not undefined) on a non-faceted mark: either ignore all cases, or error all cases.

Currently we ignore all cases but crash on "exclude". This can be fixed by changing that line to:
>         if (facetState !== undefined) facetState.facetsIndex = facetExclude(facetState.facetsIndex);

I'm not sure erroring on all cases would bring much, but it's an alternative.

src/plot.js
```diff
@@ -100,6 +100,7 @@ export function plot(options = {}) {
...
         facetState.facetsIndex = facetExclude(facetState.facetsIndex);
```
Fil, line 104:
```suggestion
        if (facetState !== undefined) facetState.facetsIndex = facetExclude(facetState.facetsIndex);
```

mbostock (approved):

PR #1283, opened 2023-02-17
textOverflow

### LINKED ISSUE #394: Truncate long tick labels with ellipses so they don’t get cropped.
In a plot like this, it would be nice if the axis labels has something like text-overflow: ellipsis so that they don’t get cropped by the edge of the chart.

Screen Shot 2021-05-12 at 9 55 53 AM

Fil:
Adds a **textOverflow** option to the text mark (and by way of consequence to the axes), which respects the **lineWidth** option (with the same text metrics as the multiline text, including monospace). But, instead of wrapping the text over several lines, clips it to the given length.

If the option is specified as "clip", it's a "white clip", if it's specified as "ellipsis", an ellipsis is added at the end of the clipped string.

The option tries to be a little smart in that it does not replace the last character by an ellipsis, and it trims whitespace.

Unless there is an explicit title channel, any text that is clipped that way receives a title with the full text. Texts that are not clipped do not receive a title (it's a bit opinionated, but easy to override by specifying the title channel).

Regarding axes, only the tick labels are clipped: the option is _not_ passed to the axis label.

closes #394

Capture d’écran 2023-02-17 à 15 09 10

Build available at https://observablehq.com/@fil/plot-text-overflow-1283 if you want to play with it.


mbostock:
This is great! Very excited to see this at last.

I wonder what you think about supporting an ellipsis in the middle, like macOS Finder does? Could that be the default? Or how would you specify whether you want the ellipsis at the start, middle, or end? I see some discussion around a CSS textOverflowMiddle property, but it seems like it never made it into browsers.

Screenshot 2023-02-17 at 10 57 11 AM

src/marks/text.js
```diff
@@ -100,7 +103,7 @@ export class Text extends Mark {
           .enter()
           .append("text")
           .call(applyDirectStyles, this)
-          .call(applyMultilineText, this, T)
+          .call(applyMultilineText, this, T, channels.title)
```
mbostock, line 106: Maybe pass `channels` here directly and have applyMultilineText extract both _text_ and _title_ itself? Alternatively, you could pull out `channels.title` above where we destructure the other channels (though we’ll need a different name since `T` is already taken…). (You could probably pass `channels` to applyIndirectTextStyles, too, if you want to keep them consistent.)

Fil:
Yes I think we could have a **clipAnchor** option that defaults to "end"? In terms of implementation, it should be as simple as (in theory):
- "start": reverse the string, clip, reverse
- "middle": interlace the first part of the string with the second part flipped, clip, deinterlace

Fil:
I implemented the different overflow strategies and extended them to work with accents and emojis. Tricky! but it seems to work.

README.md
```diff
@@ -2039,7 +2039,14 @@ The following text-specific constant options are also supported:
...
+If a **lineWidth** is specified, input text values will be wrapped as needed to fit while preserving existing newlines. The line wrapping implementation is rudimentary; for non-ASCII, non-U.S. English text, or for when a different font is used, you may get better results by hard-wrapping the text yourself (by supplying newlines in the input). If the **monospace** option is truthy, the default **fontFamily** changes to “ui-monospace, monospace”, and the **lineWidth** option is interpreted as characters (ch) rather than ems.
+
+If the **textOverflow** option is specified, only one line will be shown: if the text is longer than the given line width, it is clipped to that length, a title with the complete text is provided unless the mark already has a title channel. The following textOverflow options are supported:
+* **clip** - the text is clipped when the lineWidth is reached; extra spaces are trimed
```
mbostock, line 2045: There should be a blank line before the start of lists, and since these are _values_ rather than option names, they should be in italics rather than in bold (one asterisk instead of two).

mbostock (commented):
To recap our in-person discussion this morning, we agreed you would:

1. Add *clip-end* and *ellipsis-end* aliases that are equivalent to *clip* and *ellipsis*, respectively. (The *ellipsis-end* alias already appears to be implemented, but it’d be nice if the Text mark constructor implemented the alias explicitly instead of it relying on the switch statement/default casing in `overflow`. I think this means that *clip* should be promoted to *clip-end* in the constructor, and *ellipsis* promoted to *ellipsis-end*?)

2. Rewrite the proposed implementation to use the existing strategy of detecting surrogate pairs during iteration (`first >= 0xd800`) rather than converting the input lines into arrays of characters (`[...input]`). The existing strategy should be more performant and there probably aren’t too many places where we have to deal with multi-code point characters explicitly (I hope).

Let me know when you’ve done that and I’ll take another look. Thanks! 🙏

Fil:
I see two issues with the existing text metrics code, which are a bit difficult to disentangle from this PR:

* Emojis are counted as double chars in monospace, which doesn't seem right. We're going to clip them much too soon.
* The default to "e" doesn't really work in two common cases: the long dash (em-dash "—"), and emojis. I'm not sure how to address this. Defaulting unknown to 100 (% of 1em) solves it for these two cases but there are probably other characters for which it would not be a good idea?

My changes address these two issues (otherwise the clipping code doesn't work well with emojis) … but I'm not sure that's the right call.

Random note: I lost quite a bit of time trying to understand why a certain emoji is not clipped correctly. Here's the answer (see the added invisible char after the cross). This seems like an indication that I'm missing something?
`[..."🐤🐥🍎✖️🦜🎃🦜"].join("-") // "🐤-🐥-🍎-✖-️-🦜-🎃-🦜"`

mbostock:
If you think the em dash is common, then we should add it to `defaultWidthMap`; that’s what it’s for. The fallback to the default “e” is for anything uncommon (or “less” common) because it would be prohibitive to include an entry in the map for every possible Unicode character.

https://github.com/observablehq/plot/blob/228f62e551a9beb1780ba06ae0473cf7ec2f08d7/src/marks/text.js#L286

We should fix the `monospaceWidth` implementation so that it counts surrogate pairs as a single character, too. It should be the same as `defaultWidth` assuming that `defaultWidthMap` contained 1 (or 100, or whatever) for every character.

src/marks/text.js
```diff
@@ -406,7 +406,9 @@ function defaultWidth(text, start, end) {
 }
 
 function monospaceWidth(text, start, end) {
-  return 100 * (end - start);
+  let sum = end - start;
+  for (let i = start; i < end; ++i) sum -= isSurrogatePair(text[i], text[i + 1]);
```
mbostock, line 410: If a surrogate pair starts at _i_ (_i.e._, [_i_, _i_ + 1] is a surrogate pair) then we shouldn’t test whether another surrogate pair starts at _i_ + 1 (_i.e._, [_i_ + 1, _i_ + 2]); we should skip an index. (Hence the `i += …` in `defaultWidth`.)

Fil: oh but monospace emoji _are_ very much larger than the other chars! I'm sending what I have, and will review tomorrow with fresh eyes.

mbostock: We shouldn’t assume that all surrogate pairs are emojis. If you want to add testing specifically for emojis, I suppose we could do that (testing for the [Unicode emoticons block](https://en.wikipedia.org/wiki/Emoticons_(Unicode_block))), but my thinking is that we shouldn’t assume anything about the characters, at least for the monospace case—we should interpret monospace as “all characters have equal width” even if they don’t in practice.

Fil:
I've reverted all changes to the metrics. Many character ranges are poorly measured (emoji, and non-US), but that's OK. The way to frame this for now is that the user has to adapt the lineWidth to their particular case (for example, if you're working with emojis, you need to give a smaller lineWidth; with arabic script, a larger one).

Fixing this fully will be for another PR when time comes, but it's not simple: some emojis are composed of many code points (👩‍❤️‍💋‍👩,  👁️‍🗨️, and there are many modifiers for 🧑🏾👨🏻👧🏼👦🏽🧒🏿🙆🏻‍♂️, sometimes several fitz modifiers in a single emoji like 🫱🏻‍🫲🏼).

This doesn't matter much (or at all) for the line wrapping algorithm (which is not meant to wrap lines made of many emojis, and only cuts on spaces). but for the clipping algorithm it's important that we don't cut a complex emoji in two parts. But I think we'll have to solve this independently of the clipping algorithm. I've reached a version where at least it's not resulting in broken characters �, but  👩‍❤️‍💋‍👩 might be clipped as 👩‍❤ (which is just a partial rendering of the composed glyph, and is not great, but acceptable). I've left these examples in the unit test.

src/marks/text.js
```diff
@@ -377,21 +397,65 @@ const defaultWidthMap = {
...
+  return dropped <= (insert ? 1 : 0) ? text : head.join("").trimEnd() + insert + tail.join("").trimStart();
+}
+
+// The size of the glyph at the given index. Supports the zero width joiner for
+// composed emojis. TODO: support more modifiers (e.g. fitz).
+function glyphLength(text, i) {
+  let a;
+  let j = 0;
+  do {
+    a = text.charCodeAt(i + j++);
```
mbostock, line 456: The mixed use of `j++` and `++j` is likely wrong here; doesn’t this mean we’re reading the same character we previously read on line 458 in the case of a surrogate pair?

Fil: It's deliberate and I think, correct—this only happens when the codePoint is the zero width joiner. But I should probably set up unit tests.

mbostock: Hmm, I’ll investigate further as I’m not convinced. 🤔 It means that we’re reading the same character twice.

Also, smaller nit, but I think this function should return the index of the next character/glyph so that the calling code can say `i = readGlyph(text, i)` (slightly simpler) instead of `i += glyphLength(text, i)`.

README.md
```diff
@@ -2040,6 +2041,14 @@ The following text-specific constant options are also supported:
 
 If a **lineWidth** is specified, input text values will be wrapped as needed to fit while preserving existing newlines. The line wrapping implementation is rudimentary; for non-ASCII, non-U.S. English text, or for when a different font is used, you may get better results by hard-wrapping the text yourself (by supplying newlines in the input). If the **monospace** option is truthy, the default **fontFamily** changes to “ui-monospace, monospace”, and the **lineWidth** option is interpreted as characters (ch) rather than ems.
 
+If the **textOverflow** option is specified, only one line will be shown: if the text is longer than the given line width, it is clipped to that length, a title with the complete text is provided unless the mark already has a title channel. The following textOverflow values are supported:
```
mbostock, line 2044: This statement makes me wonder if the **textOverflow** behavior should apply after splitting into lines (after wrapping, potentially, or given hard-wrapped input with newlines already present), and that we should need have a separate **whiteSpace** option to control whether or not we wrap (independently of overflow). We support a mix of _pre_, _pre-wrap_, and _nowrap_ currently.

src/marks/text.js
Fil: yes; I also wanted to explore making a generator that yields each glyph in turn, but it started to impact the line wrapping code.

src/marks/text.js
```diff
@@ -114,12 +127,14 @@ export class Text extends Mark {
...
   if (!T) return;
+  if (textOverflow && !isFinite(lineWidth)) throw new Error(`the textOverflow option requires a specified lineWidth`);
+  const measure = monospace ? monospaceWidth : defaultWidth;
   const linesof = isFinite(lineWidth)
-    ? monospace
-      ? (t) => lineWrap(t, lineWidth, monospaceWidth)
-      : (t) => lineWrap(t, lineWidth * 100, defaultWidth)
+    ? textOverflow
+      ? (t) => [overflow(t, lineWidth * 100, measure, textOverflow)]
+      : (t) => lineWrap(t, lineWidth * 100, measure)
```
mbostock, line 137: I’d like to investigate promoting this into separate functions (similar to what we do for facetAnchor) so there is less case branching during evaluation.

src/marks/text.js
```diff
@@ -377,21 +397,86 @@ const defaultWidthMap = {
...
+  const head = [];
+  const tail = [];
+  const lengths = [];
+  let dropped = 0;
+  let w = 0; // width consumed by selected chars
+  for (let i = 0, j; i < text.length; i = j) {
+    j = readCharacter(text, i);
+    const char = text.slice(i, j);
+    const l = widthof(text, i, j);
+    if (w < width * p) {
```
mbostock, line 442: At the point that we test `w` here, I think we want to already have added `l` to it. I.e., we want to test whether adding the next character will exceed the allowed maximum line width. I believe that currently the code allows one too many characters before clipping. And we want `w <= width * p` instead of `w < width * p`. And we don’t want to pre-emptively subtract `widthof(insert, 0, insert.length)` from `width` (because it we can fit the line in the given length, it doesn’t matter how long the insertion text is).

test/plots/text-overflow.js
```diff
@@ -0,0 +1,138 @@
...
+          frameAnchor: textOverflow.endsWith("-start") ? "left" : textOverflow.endsWith("-middle") ? "middle" : "right"
+        })
+      ),
+      Plot.text(names, {
+        text: names,
+        y: names,
+        fx: () => "monospace",
+        monospace: true,
+        textOverflow: "ellipsis-end",
+        lineWidth: 13,
```
mbostock, line 68:
```suggestion
        lineWidth: 13.5,
```

Giving a fractional width here is useful for testing `<` _vs._ `<=` (per previous comment).

Fil:
[0f4342a](https://github.com/observablehq/plot/pull/1283/commits/0f4342abaaed0922ab6f170b34a8708a0cb4e8a2) reverses the meaning of end and start (and changes the default orientation) of the clipping. Was this intentional?

mbostock:
No, it wasn’t intentional to reverse the meaning.

README.md
mbostock: I’ve now structured the code this way, such that **textOverflow** can now apply to multiple lines if you insert newlines manually. We’d still need to add a **whiteSpace** option to unlock more of this functionality (and maybe another option to specify the maximum number of lines). I don’t think we should add any more functionality to this PR, though.

mbostock (approved):
👏👏👏

src/marks/text.js
```diff
@@ -375,23 +396,126 @@ const defaultWidthMap = {
...
+// as emoji skin color modifiers. https://unicode.org/reports/tr29/
+export function readCharacter(text, i) {
+  reSurrogatePair.lastIndex = i++; // consume first character
+  if (reSurrogatePair.test(text)) ++i; // consume surrogate pair
+  reCombiner.lastIndex = i;
+  if (reCombiner.test(text)) i = reCombiner.lastIndex; // consume combiners
+  reZeroWidthJoiner.lastIndex = i;
+  if (reZeroWidthJoiner.test(text)) return readCharacter(text, i + 1); // consume zero-width join and what follows
+  return i;
+}
```
Fil, line 516:
```suggestion
const reSurrogatePair = /[\uD800-\uDBFF][\uDC00-\uDFFF]/y;
const reCombiner = /[\p{Combining_Mark}\p{Emoji_Modifier}]+/uy;
const rePictographic = /\p{Extended_Pictographic}/uy;

// Reads a single “character” element from the given text starting at the given
// index, returning the index after the read character. Ideally, this implements
// the Unicode text segmentation algorithm and understands grapheme cluster
// boundaries, etc., but in practice this is only smart enough to detect UTF-16
// surrogate pairs, combining marks, and zero-width joiner (zwj) sequences such
// as emoji skin color modifiers. https://unicode.org/reports/tr29/
export function readCharacter(text, i) {
  i++; // consume first character
  if (text.charCodeAt(i) >= 0xd800 && ((reSurrogatePair.lastIndex = i - 1), reSurrogatePair.test(text))) ++i; // consume surrogate pair
  if (text.charCodeAt(i) > 0x7f) {
    reCombiner.lastIndex = i;
    if (reCombiner.test(text)) i = reCombiner.lastIndex; // consume combiners
  }
  if (text.charCodeAt(i) === 0x200d) return readCharacter(text, i + 1); // consume zero-width join and what follows
  return i;
}
```

this seems quite a bit faster on the three browsers I've tested:
https://observablehq.com/@observablehq/benchmark-readchar-1283

mbostock: Thanks. I want to try a few additional tweaks using charCodeAt and I’ll push a commit if it works. (Not a blocker though; I expect we could do this performance improvement as a followup.)

mbostock: Tidied up in 954ee22.

PR #1284, opened 2023-02-17
autoSpec

tophtucker:
Factors out the parts of the auto mark that fill in its options. Produces an object that you can pass as a fully-specified Plot.auto options object.

As an optimization, there are two versions:
- the internal _spec_ function optionally returns the materialized columns, so that Plot.auto doesn't have to re-materialize them in order to determine the implementation of the marks
- the exported _autoSpec_ function, which calls _spec_ but doesn't return the materialized columns

So, calling Plot.auto will still only materialize columns once; calling Plot.auto(data, Plot.autoSpec(data, options)) will materialize them twice.

Optional to-dos, I'm not sure they're necessary for now:
- [ ] Investigate making all undefineds into nulls
- [ ] More tests. I have a couple just for autoSpec but am still mostly relying on the usual mark snapshots
- [ ] Bring zero and z assignments up from auto to autoSpec?

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -93,7 +93,7 @@ export function autoSpec(data, options) {
     color: {
       value: colorValue ?? null,
       reduce: colorReduce ?? null,
-      ...(colorColor !== undefined && {color: colorColor}) // TODO ?? null
+      ...(colorColor !== undefined && {color: colorColor})
     },
     size: {
       value: sizeValue ?? null,
@@ -184,11 +184,11 @@ export function auto(data, options) {
   let markOptions = {
     fx,
     fy,
-    x: X ?? undefined,
-    y: Y ?? undefined,
+    x: X ?? undefined, // treat null x as undefined for implicit stack
+    y: Y ?? undefined, // treat null y as undefined for implicit stack
     [colorMode]: C ?? colorColor,
     z: Z,
-    r: S ?? undefined
+    r: S ?? undefined // treat null size as undefined for default constant radius
   };
   let transform;
   let transformOptions = {[colorMode]: colorReduce ?? undefined, r: sizeReduce ?? undefined};
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -12,7 +12,11 @@ import {bin, binX, binY} from "../transforms/bin.js";
 import {group, groupX, groupY} from "../transforms/group.js";
 import {marks} from "../mark.js";
 
-export function autoSpec(data, {x, y, color, size, fx, fy, mark} = {}) {
+export function autoSpec(data, options) {
+  return spec(data, options);
+}
+
+function spec(data, {x, y, color, size, fx, fy, mark} = {}, {materialize} = {}) {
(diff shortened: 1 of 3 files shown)


Fil:
I'd suggest we make the tests much more stringent, by comparing their serialized version to a template (which can be in a separate file on disk, or in the unit test page).

For example instead of testing 

```
it("Plot.autoSpec makes a histogram from a quantitative dimension", () => {
  const data = [{value: 1}, {value: 1}, {value: 38}];
  const A = Plot.autoSpec(data, {x: "value"});
  assert.strictEqual(A.y.reduce, "count");
  assert.strictEqual(A.mark, "bar");
});
```

we do:
```
it("Plot.autoSpec makes a histogram from a quantitative dimension", () => {
  const data = [{value: 1}, {value: 1}, {value: 38}];
  assert.strictEqual(JSON.stringify(Plot.autoSpec(data, {x: "value"})), `{ my big fat expected spec }`);
});
```

mbostock:
Why not use assert.deepStrictEqual?

src/marks/auto.js
```diff
@@ -91,7 +86,7 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
   // Determine the default mark type.
   if (mark === undefined) {
     mark =
-      sizeValue != null || sizeReduce != null
+      size != null || sizeReduce != null
```
mbostock, line 89: What is this change doing?

src/marks/auto.js
```diff
@@ -104,6 +99,49 @@ export function auto(data, {x, y, color, size, fx, fy, mark} = {}) {
...
+export function auto(data, initialOptions) {
+  let {
+    x: {value: x, reduce: xReduce, zero: xZero, ...xOptions},
+    y: {value: y, reduce: yReduce, zero: yZero, ...yOptions},
+    color: {value: color, color: colorColor, reduce: colorReduce},
+    size: {value: size, reduce: sizeReduce},
+    fx,
+    fy,
+    mark
+  } = spec(data, initialOptions, {materialize: true});
```
mbostock, line 142: A different way we can do this (without requiring an internal option and separate function) is to materialize the values as arrays, and then pass them into autoSpec; this way valueof does nothing (it returns the existing arrays). I’ll take a crack at this and push a commit if it works.

src/marks/auto.js
tophtucker: At one point in refactoring, I didn't have a reference to sizeValue here, and it seemed inconsistent to refer to x and y (not xValue and yValue) below, but sizeValue (not size) here, so I changed it. Upon further reflection, if sizeValue is undefined / null, then size will be undefined / null, so it doesn't matter. I'll revert this.

mbostock: I think it’ll be clearer if we store the materialized columns as a separate array, rather than redefining local variables/function arguments. I’ll incorporate into this into a commit I’m about to push.

tophtucker (commented):
TIL "??=". Thanks for simplifying back to one function and for tackling the "return something that's defined one way or the other" thing!

PR #1288, opened 2023-02-21
close the loophole 

Fil:
loophole that would allow to pass x in the axisFy mark, or y in the axisFx mark

see https://github.com/observablehq/plot/pull/1287


Fil:
Oh you're right! I was confused because I was trying to reach the empty facets (ref. #1287) and thought about a weird trick—but proper usage is not weird. Thanks!

PR #1298, opened 2023-02-28
include scale name in error

mbostock:
Includes the scale name in errors for easier debugging.

I also rewrote the `isOrdered` helper to iterate directly, instead of materializing all the pairs using d3.pairs and then testing with _array_.every. I think it’s cleaner this way, and slightly faster, not that it matters.

### DIFF
--- a/src/scales/ordinal.js
+++ b/src/scales/ordinal.js
@@ -52,7 +52,9 @@ export function ScaleOrdinal(key, channels, {type, interval, domain, range, sche
       }
     }
   }
-  if (unknown === scaleImplicit) throw new Error("implicit unknown is not supported");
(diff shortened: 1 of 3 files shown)


tophtucker (approved):
👍 Nice! Better!

(isMonotonic in auto.js could probably be rewritten as isOrdered(domain, orderof(domain)) if we want one fewer helper?)

mbostock:
> (isMonotonic in auto.js could probably be rewritten as isOrdered(domain, orderof(domain)) if we want one fewer helper?)

I thought about that, and no, the behavior is actually slightly different. The one for the threshold scale is more strict because it doesn’t allow non-comparable values either. At any rate I don’t think it’s worth the effort to DRY yet.

PR #1299, opened 2023-03-01
more support for bigint

### LINKED ISSUE #1297: The stack transform should support BigInt
As a follow-up on #1279, we want this to work:
Capture d’écran 2023-02-28 à 19 20 06

(it currently works if we use y2 instead of y, and skip the stack transform)

cc: @pstuffa

Fil:
I'm not sure this is the right way to do it, but I can't find another way… (we certainly don't want to create a normal array, peek at the values, then convert to typed array).

closes #1297

### DIFF
--- a/README.md
+++ b/README.md
@@ -3151,7 +3151,7 @@ Given an iterable *data* and some *value* accessor, returns an array (a column) 
 * an array of values - returning the same
 * null or undefined - returning the same
 
-If *type* is specified, it must be Array or a similar class that implements the [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) interface such as a typed array. When *type* is Array or a typed array class, the return value of valueof will be an instance of the same (or null or undefined). If *type* is not specified, valueof may return either an array or a typed array (or null or undefined).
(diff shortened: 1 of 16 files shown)


mbostock:
I think what we do instead is split the current map function into map and typedMap, where map always returns an Array, and typedMap always returns a given TypedArray. Then, in the typedMap case we compose the given accessor with Number to safely coerce rather than relying on the implicit version which errors. (We could also short-circuit the additional number coercion if the input array is already a TypedArray.)

Fil:
[0d2e2d3](https://github.com/observablehq/plot/pull/1299/commits/0d2e2d3bafd9fda550d05a9c34e6b4b3825359f7) keeps a single entry point (map). Is it enough?

mbostock:
> a single entry point

My thinking was we can easily look for all places that are passing a TypedArray to map and change those to typedMap. That way we don’t need the `type === Array` test (which also isn’t strictly accurate, since you can technically subclass Array or TypedArray, so we’d have to use `isTypedArray` instead).

Fil:
See #1305 for an exploration of typedMap vs. map.

mbostock (approved):

Fil:
I'll add more unit tests tomorrow and merge.

mbostock:
I added some more tests. 👍 Feel free to add more. I feel confident in this change.

I spent some time thinking we could bypass the coerceNumber call for the integer array types (_e.g._, Uint8Array) since these types have no way of representing NaN, and hence you can’t avoid mapping null to 0. But I was forgetting that the other reason we want coerceNumber is to avoid the error coercing a BigInt to a number, which was the title of this PR, so I threw the work away. 😅

PR #1300, opened 2023-03-01
document facet: super

### LINKED ISSUE #1293: document facet: super

Fil:
closes #1293

while its main usage is internal, facet: super can be used to add a radius legend or a title


mbostock:
Actually, I think “top frame” is a little confusing here, since we don’t explain what that means. I’m tempted to say “across all facets” but we use exactly that to describe repeating a mark across facets. 🤔 Maybe something like “draw this mark in a single frame that covers all facets”?

mbostock (approved):

PR #1304, opened 2023-03-01
Tooltips!

### LINKED ISSUE #4: Tooltips (hover to read values).
It’d be nice to have a tooltip that shows the value nearest the mouse.

mbostock:
This is a first pass at the foundation for tooltips (hover-based assistance for reading values out of plots). I’m focusing to start on the semantics with the expectation that we will devise better shorthand syntax for declaring tooltips in the future (_e.g._, adding tooltip: true to an existing mark rather than having to repeat everything on a separate tooltip mark).

The goal in this prototype is to demonstrate that we can use a _pointermove_ event listener to determine the closest data point to the pointer based on the scaled position channels (a subset of _x_, _y_, _fx_, and _fy_).

There are a few tricky things here. First, the tooltip mark needs to know the owner SVG element so that it can register an event listener at the SVG level, rather than within each facet, and without obscuring other pointer-based interaction such as text selection; this is done by adding a context.ownerSVGElement property. Second, the tooltip mark needs to “multiplex”: like other marks, it gets rendered separately for each facet, but we only want to show a single tooltip across facets. This is done using a mark-scoped WeakMap, but perhaps with additional support from Plot we could make this more elegant. In a sense the tooltip mark isn’t a mark because it has no (conventional) visual representation, but it does have channels that are accessed by the event listener.

To-do:

- [x] Get local coordinates of the pointer
- [x] Register one pointermove listener per plot
- [x] Find the closest point in screen (scaled) coordinates
- [x] Find the closest point across all facets
- [x] Limit the search radius
- [x] Hide the tooltip on pointerdown (pre-click)
- [x] Display unscaled values in a tooltip
- [x] Handle x or y not existing; respect frameAnchor
- [x] Handle fx or fy not existing (for the entire plot)
- [x] Handle faceting being disabled for this mark (facet: null)
- [x] Display all channels, not just position
- [x] Custom channels for showing arbitrary values
- [x] Remove the red dot for testing purposes
- [x] Render the tooltip as SVG or HTML, anchored to a point
- [x] Draw a little connection to the anchor point
- [x] Allow text in the tooltip to be selected
- [x] Draw the tooltip on top of the facet axis labels
- [x] Add a lil’ drop shadow
- [x] Fix the _x_ label when the automatic label is dropped (_e.g._, “Date”)
- [x] Fix tooltips for dodge initializer
- [x] ~~Automatically~~ Click to switch between “sticky” and “free” mode
- [x] Decide whether the tooltip appears immediately ~~or after a delay?~~
- [x] Implement four corner orientations
- [x] Option for specifying the preferred orientation (when it fits)?
- [x] Orient the tooltip relative to the anchor automatically based on position
- [x] Minimize changes in tooltip orientation
- [x] Configurable text styles
- [x] Limit the maximum width of the tooltip, clipping long text
- [x] Show title when tooltip value is truncated
- [x] Tooltips for line, searching first on x and then on y, and vice versa
- [x] Fix competing event listeners when multiple charts are on the page
- [x] Display **x1**-**x2** extent for binX (and similar for binY)
- [x] Display **y2**-**y1** length for stackY( and similar for stackX)
- [x] Rename **corner** option to **anchor** in case we want sides in future
- [x] Fix oscillation when tooltip doesn’t fit with any orientation
- [x] Ensure that the tooltip is always rendered last (Promise.resolve)?
- [x] Fix mobile Safari?
- [ ] Reorder the channels (custom, then standard)
- [ ] Ability to suppress known channels (*e.g.*, **x** or **y**)
- [ ] Control formatting of values in tooltip
- [ ] Shorthand from existing marks #1525

Optional:

- For link and arrow, treat {x1, y1} and {x2, y2} as distinct points?
- Option for controlling where a tooltip appears on a rect/area/etc.?
- Respect the **r** channel for hovering large circles?
- Side anchors (not just corners: *top*, *right*, *bottom*, *left*)?
- Rename to “tip”?
- Configurable colors (maybe even channels?)
- Handle multiple dots in the same position (e.g., click to cycle)?
- Dark mode (media query or checking computed background color)?

Above, the idea of *peruse* mode is when you’re moving the pointer around and you want the tooltip to update instantly, whereas *focus* mode is when the pointer seems to be pointing at a particular point of interest, and you want to be able to move the pointer into the tooltip to select text. I’m thinking that you start in *focus* mode (where the tooltip is stable), but during the first 250ms (or whatever), you can either move the pointer into the tooltip, or you move it elsewhere, and in the latter case you switch to *peruse* mode and the tooltip will not receive pointer events.

Fixes #4 (eventually).


src/plot.js
```diff
@@ -258,8 +266,11 @@ export function plot(options = {}) {
           let index = null;
           if (indexes) {
             index = indexes[facetStateByMark.has(mark) ? f.i : 0];
+            const {fx, fy} = index;
             index = mark.filter(index, channels, values);
             if (index.length === 0) continue;
+            index.fx = fx; // TODO do we need to expose the facet values?
+            index.fy = fy; // TODO do we need to expose the facet values?
```
Fil, line 273: I was confused by this, because the https://observablehq.com/@observablehq/autoplot-matrix example relies on the fx, fy properties being attached to the index (as mentioned in #1263).

The explanation of why it works in that example, is because the custom SubPlot mark does not filter, these properties were left as-is. If calling mark.filter removes them, it seems like a bug, so it's nice that we add them back. Alternatively, we could require mark.filter to be more cautious? (i.e. move this part into  https://github.com/observablehq/plot/blob/fd1837179bb7a19f83e306c0a293b2caaf9154a2/src/mark.js#L86)

mbostock:
Another point I forgot to mention: the tooltip mark will need a way to access the unscaled channel values, too. Currently marks only receive the scaled values during render.

src/plot.js
mbostock: I think it’s okay to always have these properties available. I don’t see any reason to hide them, at least, and it’s good to be consistent.

Fil: Oh I definitely want them—what I meant is that I thought they were already there (always), when in fact they could be removed if the mark was filtering the index.

src/channel.js
```diff
@@ -21,13 +21,18 @@ export function Channels(channels, data) {
...
+  const values = Object.fromEntries(
     Object.entries(channels).map(([name, {scale: scaleName, value}]) => {
       const scale = scaleName == null ? null : scales[scaleName];
+      raw[name] = value;
       return [name, scale == null ? value : map(value, scale)];
     })
   );
+  values.raw = raw;
+  return values;
 }
```
Fil, line 36: Here we'll need to take into account that the scaled values can diverge from their original position when going through an initializer. They might even not be based on the same index or have the same cardinality as the unscaled channels.

For example, if you try:
```
Plot.tooltip(athletes, Plot.hexbin(
  {fill:"count"},
  {x: "weight", y: "height", fx: "sex", fy: "date_of_birth"}
))
```

the raw channels are:
`{x: Array(242), y: Array(242), fill: Float64Array(242), fx: Array(11538), fy: Array(11538)}`

x and y contain scaled values, so they're not "raw"; and fy and fy are not parallel to the new index.

Just flagging—I have no real idea on how to solve this atm.

mbostock: I think the best we can do is that the channels represent the transformed data. In the example you gave, it should be equivalent to precomputing the hexagonal bins prior to calling Plot.plot and passing them in. Therefore I think that the _fx_ and _fy_ channels are the ones that are “wrong” as they don’t match the (transformed, aggregated) data and index. (We could try to devise a formal mechanism to allow transforms to describe aggregations, but I think that will be tricky and it doesn’t generalize that well, for example with weighted rolling window transforms.)

Since the hexbin transform operates in screen space, this implies that there wouldn’t be an automatic way to show the abstract _x_ and _y_ values of the hexbin that you are hovering over. But that _x_ and _y_ merely represent the centroid of the hexbin anyway, so it’s not particularly meaningful (especially if _x_ or _y_ is not linear), and if you’re binning on _x_ and _y_, it implies you don’t care about the precise value of _x_ and _y_ anyway. More likely what you want in that case is for a tooltip to show other details, _e.g._, the count of athletes in that particular hexagon, say, or maybe the distribution of their nationalities or sports.

mbostock: I dropped the _fx_ and _fy_ channels in 0486d24. (We could impute _fx_ and _fy_ channels if we wanted, but that would just be slower so it seems cleaner this way.)

dennismphil:
Thank you for the work on this. Ran into an issue where it was difficult to build a on:pointermove interaction.   
  
With this design, on hover in a histogram, is it possible to change the opacity of all the bars except the one on which is mouse pointer is over?

mbostock:
This has been rebased against the latest main, and now requires even fewer core changes—just the new ownerSVGElement property on the context. 😌 Remaining work centers around the contents and appearance of the tooltip, the various anchor modes, and shorthand syntax for deriving tooltips from an existing mark.

src/channel.d.ts
```diff
@@ -61,7 +61,7 @@ export type ChannelName =
  * An object literal of channel definitions. This is also used to represent
  * materialized channel states after mark initialization.
  */
-export type Channels = {[key in ChannelName]?: Channel};
+export type Channels = Record<string, Channel>;
```
mbostock, line 64: I’ve changed this here to allow specifying custom channels for tooltips, but we might want to make this change in more places… and we might even want to change the ChannelName type to include string so that it’s always allowed, but use the never hack to still let VS Code autocomplete known channel names. Although, that makes it less likely we will detect typos.

mbostock:
https://user-images.githubusercontent.com/230541/236108871-bc9da770-78e4-4da1-a07a-8ec7c219f4e6.mov

mbostock:
@Fil Can you do an initial review here? I want to know if we have sufficiently captured the requirements for the first version or if there are unanticipated requirements that could delay this landing.

mbostock:
I think we’ll need to figure out some hint so that the tooltip can be smarter with **x** and **y** when **x1** and **x2** or **y1** and **y2** are available. Sometimes we want to express an extent (for example with binX, it’d be nice to see the upper and lower bound of the bin in **x** rather than the midpoint) and other times we want a length (for example with stackY, we want to see the difference between **y2** and **y1**, not the midpoint **y**). The latter—fixing the tooltip with the stack transform—is especially important because it is misleading to show the midpoint, especially in the streamgraph case. I wonder whether the stack transform can populate a hint that **y1** and **y2** represent lengths (say using the channel.source hint, and suppressing **y2**)?

mbostock:
https://user-images.githubusercontent.com/230541/236587624-d67e2b6d-b1bb-4402-8cbd-bb481ac6a2cd.mov

mbostock:
Given the _numerous_ requests we have received related to this feature, and knowing that there is a lot of demand for customization here, I do wonder whether the initial implementation of this feature could be decoupled a bit (a bit like how axes were initially built-in, but then later we redesigned them to be explicit marks affording greater flexibility).

In particular I’d like to try implementing the (non-interactive) display of the tooltip as a standard mark. I imagine using it to annotate a specific point in the plot (statically and manually), or perhaps multiple points at the same time (say the extrema).

If we had such a mark, perhaps we could have a separate “point” interaction that picks a specific point of interest using pointer events. And then lastly some way to tie that pointing interaction with the tooltip display mark (having Plot re-render the mark automatically in response to the interaction saying that the focused point has changed).

A lot of details to figure out, but since we’re going to need to do some of this for interaction anyway, it might be worth experimenting even if it means delaying the initial release of tooltips?

mbostock:
Closing in favor of #1527, which does almost everything this PR does, but is capable of much more, too.

PR #1305, opened 2023-03-02
split map and typedMap

Fil:
Addressing [this comment](https://github.com/observablehq/plot/pull/1299#issuecomment-1450307659).

One tiny additional thing we could do is make Float64Array the default, since we only ever use this specific type when calling typedMap. But, I don't know, it feels a bit noisy.

### DIFF
--- a/src/marks/contour.js
+++ b/src/marks/contour.js
@@ -1,7 +1,7 @@
-import {blur2, contours, geoPath, map, max, min, nice, range, ticks, thresholdSturges} from "d3";
+import {blur2, contours, geoPath, max, min, nice, range, ticks, thresholdSturges} from "d3";
 import {Channels} from "../channel.js";
 import {create} from "../context.js";
-import {labelof, identity, arrayify} from "../options.js";
+import {labelof, identity, arrayify, map} from "../options.js";
 import {Position} from "../projection.js";
 import {applyChannelStyles, applyDirectStyles, applyIndirectStyles, applyTransform, styles} from "../style.js";
 import {initializer} from "../transforms/basic.js";
@@ -156,7 +156,9 @@ function contourGeometry({thresholds, interval, ...options}) {
     const {contour} = contours().size([w, h]).smooth(this.smooth);
     const contourData = [];
     const contourFacets = [];
-    for (const V of VV) contourFacets.push(range(contourData.length, contourData.push(...T.map((t) => contour(V, t)))));
+    for (const V of VV) {
+      contourFacets.push(range(contourData.length, contourData.push(...map(T, (t) => contour(V, t)))));
+    }
 
     // Rescale the contour multipolygon from grid to screen coordinates.
     for (const {coordinates} of contourData) {
@@ -188,7 +190,7 @@ function contourGeometry({thresholds, interval, ...options}) {
 function maybeTicks(thresholds, V, min, max) {
   if (typeof thresholds?.range === "function") return thresholds.range(thresholds.floor(min), max);
   if (typeof thresholds === "function") thresholds = thresholds(V, min, max);
-  if (typeof thresholds !== "number") return arrayify(thresholds, Array);
+  if (typeof thresholds !== "number") return arrayify(thresholds);
   const tz = ticks(...nice(min, max, thresholds), thresholds);
   while (tz[tz.length - 1] >= max) tz.pop();
   while (tz[1] < min) tz.shift();
--- a/src/marks/density.js
+++ b/src/marks/density.js
@@ -1,6 +1,6 @@
 import {contourDensity, create, geoPath} from "d3";
 import {Mark} from "../mark.js";
-import {coerceNumbers, isTypedArray, maybeTuple, maybeZ} from "../options.js";
+import {coerceNumbers, maybeTuple, maybeZ, TypedArray} from "../options.js";
 import {Position} from "../projection.js";
 import {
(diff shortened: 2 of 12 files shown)


mbostock (commented):
Yes I think based on our usage we should remove the `type` argument to `typedMap` entirely and always use float64. We could call it `floatMap` or `numberMap` if you prefer.

Fil:
Does this mean we retire the *type* argument from the public API of Plot.valueof?

Internally we use it once, in https://github.com/observablehq/plot/blob/fd1837179bb7a19f83e306c0a293b2caaf9154a2/src/transforms/stack.js#L84

and we miss an opportunity to use it in https://github.com/observablehq/plot/blob/fd1837179bb7a19f83e306c0a293b2caaf9154a2/src/transforms/dodge.js#L76

mbostock:
No, I don’t think we should retire it if it’s part of the public API.

mbostock:
I want to think about this one a little more and push a few commits. In a case like this

```js
floatMap(X2, (x2, i) => Math.abs(x2 - X1[i]))
```

we are doing too much work because we are coercing the result of the accessor even though we already know that it can only return a number.

src/options.js
```diff
@@ -9,12 +9,16 @@ const objectToString = Object.prototype.toString;
 /** @jsdoc valueof */
 export function valueof(data, value, type) {
   const valueType = typeof value;
+  const access =
+    type && Object.getPrototypeOf(type) === TypedArray
```
mbostock, line 13: This should handle subclasses of typed arrays, too. For example, given:

```js
TypedArray = Object.getPrototypeOf(Uint8Array)
class MyUint8Array extends Uint8Array {}
```

Then

```
MyUint8Array instanceof TypedArray // false
Uint8Array instanceof TypedArray // false
Object.getPrototypeOf(MyUint8Array) === TypedArray // false
Object.getPrototypeOf(Uint8Array) === TypedArray // true
```

The fix here is pretty simple, I think:

```suggestion
    isTypedArray(type?.prototype)
```

Fil: makes sense, but do we need the function call? : `type?.prototype instanceof TypedArray`

mbostock: Short functions like this typically get inlined anyway so it shouldn’t matter. And I figured we already had the helper.

mbostock:
In 0e74a1e I reverted the addition of floatMap, and instead isolated the implicit number coercion to valueof. Well, and arrayify, but I actually think that’s a bit of mistake since it’s only needed when arrayify is used to create a typed array, and that only happens via valueof. So I’d like to follow this up with another slight change. This change avoids unnecessary coercion when we call map and we know that the provided accessor is already number-safe.

src/marks/contour.js
```diff
@@ -1,7 +1,7 @@
-import {blur2, contours, geoPath, map, max, min, nice, range, ticks, thresholdSturges} from "d3";
+import {blur2, contours, geoPath, max, min, nice, range, ticks, thresholdSturges} from "d3";
 import {Channels} from "../channel.js";
 import {create} from "../context.js";
-import {labelof, identity, arrayify} from "../options.js";
+import {labelof, identity, arrayify, map} from "../options.js";
```
mbostock, line 4: Oops, this fixes the calls here: 😳 

https://github.com/observablehq/plot/blob/1186b5b19c04e8a52cfec399743ad43fb44ab25b/src/marks/contour.js#L125-L126

src/marks/density.js
```diff
@@ -129,7 +128,7 @@ function densityInitializer(options, fillDensity, strokeDensity) {
     // If explicit thresholds were not specified, find the maximum density of
     // all grids and use this to compute thresholds.
     let T = thresholds;
-    if (!isTypedArray(T)) {
+    if (!(T instanceof TypedArray)) {
```
mbostock, line 131: (Taking your advice on using instanceof directly… though I think the performance will be the same and maybe it’s slightly more readable to have the helper. I’m happy either way.)

src/marks/link.js
```diff
@@ -2,7 +2,7 @@ import {geoPath, pathRound as path} from "d3";
 import {create} from "../context.js";
 import {curveAuto, PathCurve} from "../curve.js";
 import {Mark} from "../mark.js";
-import {coerceNumbers} from "../scales.js";
+import {coerceNumbers} from "../options.js";
```
mbostock, line 5: I moved the coercion routines up to options.js since coerceNumbers is now needed for valueof. I wanted to keep coerceDates together, but I left coerceSymbols where it was since that has more dependencies…

src/options.js
```diff
@@ -3,21 +3,43 @@ import {color, descending, range as rangei, quantile} from "d3";
...
+}
+
+// When valueof is asked to produce a typed array (i.e., numbers) we implicitly
+// apply null-safe type coercion.
+function maybedTypedArrayify(data, type) {
+  return type === undefined
+    ? arrayify(data) // preserve undefined type
+    : data instanceof type
+    ? data
+    : type.prototype instanceof TypedArray && !(data instanceof TypedArray)
```
mbostock, line 36: Bonus: if we already have a typed array, we can skip the coercion (even if we have to change the array type)!

mbostock (approved):
LGTM. What do you think @Fil?

Fil:
Overall it's a good cleanup. I'm still a bit nervous with all the changes, but hopefully the tests have good coverage.

PR #1306, opened 2023-03-02
auto-label functions

mbostock:
This an idea that I think we should not pursue, but was somewhat interesting. What if we inferred a channel label from the name of the associated value function? Then in cases like this, we can get a nice automatic label (“decade” for _fy_):

```js
{
  const athletes = await d3.csv("data/athletes.csv", d3.autoType);
  const decade = (d) => `${Math.floor(d.date_of_birth.getUTCFullYear() / 10) * 10}`;
  return Plot.plot({
    grid: true,
    marks: [
      Plot.dot(athletes, {x: "weight", y: "height", fx: "sex", fy: decade}),
      Plot.tooltip(athletes, {x: "weight", y: "height", fx: "sex", fy: decade})
    ]
  });
}
```

(Note that ECMAScript formalizes how arrow functions are named; see https://stackoverflow.com/a/37488652.)

Unfortunately, there are a few cases where we get undesirable names this way:

1. Internal functions such as `bin` need to be anonymized.
2. Channel declarations such as `fy: (d) => …` inherit the name of the channel.
3. Code minimization might introduce nonsense names.

As such, I don’t think we should pursue this. But it was an interesting experiment.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -21,7 +21,7 @@ export function valueof(data, value, type) {
 }
 
 export const field = (name) => (d) => d[name];
-export const indexOf = (d, i) => i;
+export const indexOf = anonymous((d, i) => i);
 /** @jsdoc identity */
 export const identity = {transform: (d) => d};
 export const zero = () => 0;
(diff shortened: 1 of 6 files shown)


src/options.js
```diff
@@ -217,7 +217,17 @@ export function maybeColumn(source) {
 }
 
 export function labelof(value, defaultValue) {
-  return typeof value === "string" ? value : value && value.label !== undefined ? value.label : defaultValue;
+  return typeof value === "string"
+    ? value
+    : typeof value === "function" && value.name && !/^f?[xy]\d*$/.test(value.name)
```
mbostock, line 222: Here I’m trying to avoid inheriting a channel name (_e.g._, `fy: (d) => …`). But this doesn’t work for the _fill_ channel, for example.

src/options.js
```diff
@@ -217,7 +217,17 @@ export function maybeColumn(source) {
...
+    ? value.name
+    : value && value.label !== undefined
+    ? value.label
+    : defaultValue;
+}
+
+export function anonymous(f) {
```
mbostock, line 229: This is a trick to avoid _e.g._ `const bin = (d) => …` inheriting the name `bin`.

PR #1310, opened 2023-03-02
automatic opt-out of opacity scale

### LINKED ISSUE #1308: Automatically opt-out of opacity scale if channel values are in [0, 1]
Similar to what we do for _color_ and _symbol_ scales, we could opt-out of the _opacity_ scale if all values are numbers in the range [0, 1]. This is slightly more likely to have a false positive than with _color_ and _symbol_ scales, but is might still be a good idea. Ref. #1307.

### LINKED ISSUE #1309: Per-channel scale overrides cannot be used in conjunction with bin and group transforms
For example, consider this snippet:

```js
Plot.rectX(penguins, Plot.binX({fillOpacity: "proportion"}, {x: "body_mass_g", thresholds: 20}))
```

If whether or not the **fillOpacity** channel is bound to the _opacity_ scale is inferred automatically by the values (per #1308), how would you tell Plot that you want (or don’t want) the _opacity_ scale? The following syntax errors:

```js
Plot.rectX(penguins, Plot.binX({fillOpacity: {value: "proportion", scale: null}}, {x: "body_mass_g", thresholds: 20}))
```

mbostock:
Fixes #1308; adds automatic opt-out of opacity scale.
Fixes #1309; adds per-channel scale overrides for bin & group transform outputs.


mbostock:
What do you think, @fil? If you don’t think automatic opt-out of the _opacity_ scale is a good idea, I could extract the fix for #1309 into a separate PR. I have mixed feelings about this because it feels like it’s more likely to get a false positive (false negative?) where numbers that happen to be in the interval [0, 1] are erroneously interpreted as literal opacities.

Fil (approved):
It makes sense to me that we would consider values in [0, 1] as literal opacities—the risks and benefits are similar to what we do for colors, and it would be more consistent.

I'm going to change the markov-chain example back to "default", because it's what I had in mind when I wrote it — even though it ended up scaled up by 1/0.8 :-)

net:
Could we get this for just "opacity" (in addition to "strokeOpacity" and "fillOpacity")?

Fil:
good catch, @net! this is fixed now in main (issue #1570, merged PR #1571).

net:
@Fil awesome!

PR #1318, opened 2023-03-07
color scale override for Plot.raster

### LINKED ISSUE #1317: Per-channel scale override fails with Plot.raster
This should show the high-vapor-value pixels in red.

```
Plot.plot({
  color: {scheme: "blues", legend: true},
  x: {transform: (x) => x - 180},
  y: {transform: (y) => 90 - y},
  marks: [
    Plot.raster(await vapor(), {
      width: 360,
      height: 180
    }),
    Plot.raster(await vapor(), {
      width: 360,
      height: 180,
      fill: {value: (d) => (d > 4 ? "red" : null), scale: null}
    })
  ]
});
```

Fil:
Because of https://github.com/observablehq/plot/pull/1196/commits/f028a08e558bd6b0fc48b1d9955b6387d1bdb247 color scale overrides didn't work for Plot.raster

Capture d’écran 2023-03-07 à 14 22 10

closes #1317


src/marks/raster.js
```diff
@@ -95,10 +95,11 @@ export class Raster extends AbstractRaster {
   }
   // Ignore the color scale, so the fill channel is returned unscaled.
   scale(channels, {color, ...scales}, context) {
+    this.recolor = channels.fill?.scale;
     return super.scale(channels, scales, context);
   }
   render(index, scales, channels, dimensions, context) {
-    const color = scales.color ?? ((x) => x);
+    const color = this.recolor && scales.color ? scales.color : (x) => x;
```
mbostock, line 102: I think you want something like this:

```suggestion
    const color = scales[this.channels.fill?.scale] ?? (x) => x;
```

mbostock (changes requested):

src/marks/raster.js
Fil: hadn't thought about that, but it makes sense!

Fil: this.channels.fill.scale comes in as "auto" and not "color", so I had to adapt a bit

mbostock (commented):
Maybe it wasn’t clear, but I’m trying to avoid needing a side-effect in the raster.scale method. Why can’t we just figure out the appropriate color scale in raster.render instead?

PR #1320, opened 2023-03-08
TypeScript declarations!

mbostock:
TODO

- [x] Complete TypeScript declarations for public APIs
- [x] Port snapshot tests to TypeScript (to test TypeScript declarations)
- [x] Enable ESLint on TypeScript declarations
- [x] Enable [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks)?
- [x] Re-enable `yarn test:tsc`

Possible future enhancements:

- Documentation for all TypeScript declarations
- Consolidate test data loading logic so that it can be type-safe across tests


dennismphil:
Absolute lifesaver. Thank you!! 

I have been maintaining these manually the ones I use and it is quite a lot of work.

src/plot.d.ts
```diff
@@ -0,0 +1,75 @@
+import type {ChannelValue} from "./channel.js";
+import type {Data, Markish} from "./mark.js";
+import type {ProjectionFactory, ProjectionImplementation, ProjectionName, ProjectionOptions} from "./projection.js";
+import type {ScaleLegendOptions, Scale, ScaleOptions} from "./scales.js";
+
+export interface PlotOptions {
```
dennismphil, line 6: Suggestion: Use generics here to represent the shape of Data

```suggestion
export interface PlotOptions<T> {
  ...
  filter?: (d: T) => boolean;
  text?: (d: T) => string;
  ... 

}
```

src/dimensions.d.ts
```diff
@@ -0,0 +1,14 @@
+export interface Dimensions {
+  width: number;
+  height: number;
+  marginTop: number;
+  marginRight: number;
+  marginBottom: number;
+  marginLeft: number;
```
dennismphil, line 7: missing `margin`?

src/mark.d.ts
```diff
@@ -0,0 +1,79 @@
...
+import type {Context} from "./context.js";
+import type {Facet, FacetAnchor} from "./facet.js";
+import type {plot} from "./plot.js";
+
+export type Data = Iterable<any> | {length: number};
+
+export type RenderFunction = (
+  index: number[],
+  scales: {[name: string]: (value: any) => any},
+  values: {[name: string]: any[]},
```
dennismphil, line 12: channels?

src/plot.d.ts
mbostock: Different marks often have different data so it’s not always appropriate to do template the data at the Plot level. We might do this for the MarkOptions.

src/mark.d.ts
mbostock: Yes these are derived from the mark’s declared channels.

mbostock:
There is a _lot_ more work to do regarding documentation here (probably dozens of hours). But the types here should be ready to review, and I think we probably shouldn’t block this PR on finishing the documentation—just on the types themselves.

Fil (approved):
Awesome work! Can't wait to start adding jsdoc everywhere.

And a medal for removing 500+ lines (+3,300 −3,841) 🏅

src/reducer.d.ts
```diff
@@ -0,0 +1,32 @@
...
+  | "deviation"
+  | "min"
+  | "min-index"
+  | "max"
+  | "max-index"
+  | "mean"
+  | "median"
+  | "variance"
+  | "mode"
+  | `p${number}${number}` // percentile
```
duaneatat, line 18: This allows something like "p1.51.5" where 1.5 and 1.5 each satisfy `number`. If we want to just allow numbers between 00 and 99 we can change number to Digit and define Digit as `type Digit = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9;`

src/transforms/bin.js
```diff
@@ -344,7 +341,7 @@ function isInterval(t) {
   return t ? typeof t.range === "function" : false;
 }
 
-function Bin(EX, EY) {
+function bing(EX, EY) {
```
duaneatat, line 344: I know we can't use `bin`, but I'm curious why you chose `bing`

src/transforms/group.js
```diff
@@ -34,31 +34,27 @@ import {
 } from "../options.js";
 import {basic} from "./basic.js";
 
-/** @jsdoc groupZ */
+// Group on {z, fill, stroke}.
```
duaneatat, line 37: should these use jsdoc blocks? 
```suggestion
/** Group on {z, fill, stroke}. */
```

src/transforms/window.d.ts
```diff
@@ -0,0 +1,39 @@
...
+  | "median"
+  | "min"
+  | "mode"
+  | "sum"
+  | "variance"
+  | "difference"
+  | "ratio"
+  | "first"
+  | "last"
+  | `p${number}${number}`;
```
duaneatat, line 17: same comment as above, should be something like `p${Digit}${Digit}`

duaneatat (approved):
This is great, so nice to finally have a working full-breadth implementation!

src/reducer.d.ts
Fil: This avoids listing the 100 pXX combination. Ideally we'd have a way to give two examples (say, p05 and p95), cover p00...p99, but without expanding to p00…p99 in the autocomplete.

This works for me in VSCode:

```diff
-  | `p${number}${number}` // percentile
+  | "p05"
+  | "p95"
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | (`p${Digit}${Digit}` & String) // percentile
```

mbostock: You shouldn’t use `String` as a type here (hence the ESLint warning). You need to use `Record<never, never>` instead. (And that’s only because of https://github.com/microsoft/TypeScript/issues/29729.)

src/transforms/group.js
mbostock: This is an internal-only comment, so I think `//` is more appropriate. Public comments live exclusively in the TypeScript declarations.

src/reducer.d.ts
Fil: Good call, I've edited my suggestion. And it works!
Capture d’écran 2023-03-13 à 15 58 12

mbostock: > This avoids listing the 100 pXX combination.

We don’t allow `p100` at runtime. I think that’s fine; you can use `max` instead.

https://github.com/observablehq/plot/blob/5830098d729edb56bde0b60f281c2e744919d5fa/src/transforms/group.js#L244

mbostock: Making Duane’s fix causes VS Code to suggest every possible percentile… 😂 I might suppress these using the Record hack.

Screenshot 2023-03-13 at 8 08 26 AM

src/transforms/group.js
duaneatat: Got it!

src/reducer.d.ts
duaneatat: Yep! This is what @Fil was suggesting above, to manually include a few common percentiles and then suppress the rest

PR #1322, opened 2023-03-08
quantile scales support for empty/invalid channels

### LINKED ISSUE #1323: The quantile scale errors if the channel is invalid/empty
The error is guarding against non-monotonous domain, but our domain in this case is [undefined… undefined], and the test returns NaN.

Fil:
When all the quantile values are NaN, ignore the "non monotonic" error and instead just proceed. The values will all be drawn in the unknown color, if set, otherwise discarded.

closes #1323


mbostock (commented):
Seems reasonable. Can you add tests (either a snapshot test of the success case, but ideally code tests of both the success and error case)?

mbostock (approved):

Issue #1088, opened 2022-10-14
Determine length of column-oriented dataframe?


Fil:
Suppose the data is given as an object of columns:

`df = {x: [1, 2, 3], y: [1, 2, 3]}`

(This is how Quarto returns dataframes, and arquero does something similar.)

To use this in a mark we can call:

`Plot.barX(data, {x: df.x, y: df.y, …)`

But there is no good way to specify *data*:
* if we specify it as {length: n} it will get materialized at some point, which is not optimal if the dataframe has millions of rows.
* if we pass `df.x` as data, it is semantically incorrect
* technically `new Array(df.x.length)` is fine, but it's a mental stretch

I wonder if we could have either: *data = n* (a number) —which would be read as `new Array(n)`—; or a special symbol that would say "use the channels' length". Another useful possibility would be for "dataframe objects" to have some sort of length property.

cc: @allisonhorst

Issue #191, opened 2021-03-04
Optimize field shorthand for columnar data?


mbostock:
We currently support field (named properties specified as strings) as channel value shorthand, e.g., `y: "foo"` is upgraded to `y: d => d["foo"]` or equivalently `y: data.map(d => d["foo"])`. For Arquero (and perhaps other column-oriented data structures), this still works because Arquero provides an object iterator; however, it’d be more efficient if Plot supported the same shorthand with Arquero’s column accessors, upgrading `y: "foo"` to ~~`y: data.column("foo")`~~ `y: data.array("foo")`, which returns an iterable over the values. ~~Although, currently this would still involve copying the column into a new array, so we actually want `y: data.column("foo").data` if possible (e.g., Arquero’s Column class could support an toArray method which returns the data if possible, and otherwise constructs a new array using the iterator).~~

mbostock:
I think we should focus on doing this specifically for Apache Arrow table instances first (since these are the most widely-used columnar data representation, including by the DuckDBClient/SQL code blocks in Observable Framework).

mbostock:
Here’s is trivial example of how to pass in columnar data. Given an Arrow table:

```js
const data = Arrow.tableFromArrays({
  id: [1, 2, 3],
  name: ["Alice", "Bob", "Charlie"],
  age: [35, 25, 45]
});
```

You can pass the columns in like so:

```js
Plot.barY({length: data.numRows}, {x: data.getChild("name"), y: data.getChild("age")}).plot()
```

The goal of this PR is to make the above equivalent to the following shorthand syntax:

```js
Plot.barY(data, {x: "name", y: "age"}).plot()
```

To do this, we will need to detect when `data` is an Apache Arrow table, and change the definition of the field shorthand to use `data.getChild(name)` instead of `Array.from(data, (d) => d[name])`. Also, we’ll want to avoid materializing the array-of-objects as `data` (the `arrayify(this.data)` call in `mark.initialize`), and instead use an efficient numeric index [0, … *numRows* - 1].

PR #1324, opened 2023-03-10
Support for columnar datasets: arquero and quarto

Fil:
Another stab at columnar datasets; this adopts something similar to arquero’s .array method to read a column as a (possibly typed) array. Columnar datasets are transformed into a {length, array: (field, type) => Array, iterator} object.

see #1088 and #191


deleted:
Hi @Fil .

Just a note to say I'm very interested in the feature. Is it still planned?

Context:
I'm writing a visualizer for large sets of timeseries data that I've gone to a great deal of trouble to transform into Apache-Arrow IPC on the backend, so it's a shame that Observable plot has to transpose it to rows for rendering (though I was very pleased to discover it will accept an arrow `Table` as a data object!)

Fil:
To be clear what this PR is trying to do is autodetect (by duck-typing) a few columnar formats and transform them into iterators that generate objects with key:value pairs. I don't think we should go this way, and I'll close this PR.

Plot supports channels passed directly as columns, saving one copy of the values (a copy will have to be made as soon as you have a scale or a data transform, anyway). I think that is what you want?

For example with apache-arrow@14, if C is a column:

```js
const C = Arrow.makeVector(Float64Array.from({ length: 100 }, d3.randomLcg(42)));
Plot.lineY({length: C.length}, {y: C}).plot()
```

PR #1326, opened 2023-03-11
time intervals coerce isostrings

### LINKED ISSUE #1182: Date strings, time scales, and `interval`
I came across the behavior shown in [this notebook](https://observablehq.com/d/ba8082640bc48b87), where plotting date strings on a `utc` scale ceases to work if the scale additionally has an `interval` property.

```js
Plot.dotX(dates).plot({ x: { type: "utc" } }) // dots are rendered

Plot.dotX(dates).plot({ x: { type: "utc", interval: d3.utcDay } }) // dots are not rendered
```

According to the docs time scales coerce channel values to dates, so I think the first chart is right, and there might be something weird happening when the `interval` option gets involved.

Fil:
The discussion below ends up with the recommendation that if we want to do this, we should coerce isostrings to dates in the channels, rather than wrap the interval object.

todo/done:
- [ ] isTimeInterval: should follow a better definition of a "time interval", compatible with our typescript def; maybe “floor(0) gives a Date”.
- [x] plot/src/transforms/interval.js (dateIntervalWeekMonth example)
- [x] scales/ordinal.js (scale transform, _e.g._ regularize dates as ordinal years)
- [ ] marks/contour.js (thresholds given as a date interval—ultra rare)

closes #1182

~~(Note: I considered doing this in d3-time, but it would mean adding d3-time-format as a dependency of d3-time.)~~


mbostock (changes requested):
I think it’s fine to do this for the built-in time intervals, but we shouldn’t do this implicit coercion if someone provides an interval implementation, including one of the D3 time interval implementations. The implicit coercion should be part of the interval implementation, and we can do that for the named time intervals we provide. (The magic of invoking the interval implementation and seeing if it returns a Date is too much, and some intervals may want to have different semantics for input coercion.)

Fil:
I had the same hesitation but went the other way:
* since we're not going to remove the duck-typing of d3 time intervals (for the bin transform), we might as well use it consistently.
* this could help people who use string dates and funky intervals—for example, transitioning from 1 month interval to 2 months interval requires only one operation (switching from "month" to d3.utcMonth.every(2)), instead of that + coercion of strings to dates.

And, it's well defined in the README: “Time intervals are intervals that are also functions that return a Date instance when called with no arguments.”

src/options.js
```diff
@@ -288,7 +288,7 @@ export function maybeInterval(interval, type) {
   if (typeof interval === "string") return (type === "time" ? maybeTimeInterval : maybeUtcInterval)(interval);
   if (typeof interval.floor !== "function") throw new Error("invalid interval; missing floor method");
   if (typeof interval.offset !== "function") throw new Error("invalid interval; missing offset method");
-  return interval;
+  return maybeIsoInterval(interval);
```
mbostock, line 291: Since `maybeIsoInterval` is only used here, I suggest we inline.

```suggestion
  return isTimeInterval(interval) ? isoInterval(interval) : interval;
```

src/time.js
```diff
@@ -41,14 +42,22 @@ const utcIntervals = new Map([
   ["sunday", utcSunday]
 ]);
 
+function isoInterval(interval) {
+  return Object.assign((t) => interval(t), {...interval, floor: (t) => interval.floor(coerceDate(t))});
```
mbostock, line 46: This feels awkward to me. Like, isTimeInterval is the way it is because that was a quick and dirty way to test for a D3 time interval. But now we have to recreate that representation even though we don’t need it to satisfy the isTimeInterval check. And this is only adding coercion for _one_ of the possible methods (floor) on the time interval, not the main function, not _interval_.offset and _interval_.range, etc. I’m not sure what fix I recommend yet.

test/plots/date-intervals.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function intervalWeekMonth() {
+  const data = await d3.csv("data/bls-industry-unemployment.csv");
+  return Plot.plot({
+    marginLeft: 80,
+    color: {type: "linear", scheme: "blues"},
+    marks: [
+      {label: "month", interval: "month" as Plot.TimeIntervalName},
+      {label: "d3.utcMonth", interval: d3.utcMonth as Plot.IntervalImplementation<Date>},
+      {label: "week", interval: "week" as Plot.TimeIntervalName},
+      {label: "d3.utcWeek", interval: d3.utcWeek as Plot.IntervalImplementation<Date>}
```
Fil, line 13: the way I handle types here is probably wrong

mbostock (changes requested):
I’d still like to try to find a different way to do this. I think in many cases, like with binning, we’ve already done the coercion by the time the interval is applied, so this coercion is redundant. Would it be possible to instead do the coercion outside the interval, before the interval is applied, when the scale interval is a time interval?

Edit: For what it’s worth, the performance of the bin1m isn’t affected by this branch, at least. So performance doesn’t appear to be an issue. Let me think some more.

src/time.js
Fil: We can give them the same treatment, but in practice don't use them on user-provided data. Their input is in all cases already coerced.

(Agree it's awkward!)

test/plots/date-intervals.ts
mbostock: You should use a const assertion instead, like so:

```js
    marks: (
      [
        {label: "month", interval: "month"},
        {label: "d3.utcMonth", interval: d3.utcMonth},
        {label: "week", interval: "week"},
        {label: "d3.utcWeek", interval: d3.utcWeek}
      ] as const
    ).map(({label, interval}) =>
      Plot.barX(data, {
        x: "date",
        filter: (d) => d.industry === "Construction",
        interval,
        fill: "unemployed",
        title: "unemployed",
        inset: 0,
        fy: () => label
      })
    )
```

src/options.js
```diff
@@ -490,3 +490,11 @@ export function named(things) {
 export function maybeNamed(things) {
   return isIterable(things) ? named(things) : things;
 }
+
+export function isTimeInterval(t) {
+  return isInterval(t) && typeof t === "function" && t() instanceof Date;
```
mbostock, line 495: This disagrees with how we’ve defined intervals in TypeScript, where they are objects rather than decorated functions:

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/interval.d.ts#L19-L22

And this means that you can use a {floor, offset, range?} object in Plot, which works, except when we try to do this time interval test.

Also, I don’t think we actually need this restriction, because D3 doesn’t require that time intervals are functions, either. It instead checks for the range method (and then blindly invokes the floor and offset methods):

https://github.com/d3/d3-time/blob/9e8dc940f38f78d7588aad68a54a25b1f0c2d97b/src/ticks.js#L38
https://github.com/d3/d3-scale/blob/2b7db622b1a224d9ea19ff15c4cc8cbb3b25f4a4/src/time.js#L58

I added that duck testing in #572, but I think it would be better to replace that duck testing with something consistent with our new TypeScript declarations and with D3. Probably we should invoke the floor method and see if it returns a Date? But that’s a little messy too, since that method requires an argument, and it’s unclear whether we should pass it a Date or a number… All D3 time intervals support coercion (as does Plot’s internal number interval), but should we expect users to implement coercion in their own intervals? I mean, isn’t the point of this PR is that they don’t? 🤔

test/plots/date-intervals.ts
```diff
@@ -0,0 +1,28 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export async function intervalWeekMonth() {
```
mbostock, line 4: Generally speaking it’s nice if the test name starts with the file name, so that the tests are easier to find. For example, this could be called `dateIntervalWeekMonth` or some such.

mbostock:
I’ve surveyed the code base to see where we might need coercion for intervals.

The (implicit) interval transform uses intervals in maybeIntervalK and maybeIntervalMidK. Both of these would benefit from an implicitly-coercing interval. (Though it would also be possible to coerce all the values prior to invoking the interval.)

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/transforms/interval.js#L31

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/transforms/interval.js#L50

A scale transform can be specified as an interval as described in #1182. This happens in applyScaleTransform when the _interval_.floor is promoted to a transform. This would benefit from an implicitly-coercing interval, too.

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/plot.js#L365-L367

The axisMark can also use the scale interval to compute ticks. This would _not_ benefit from an implicitly-coercing interval: the scale should already have coerced values at this point, and we wouldn’t want the interval to use different coercion than the scale.

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/marks/axis.js#L520-L521

Ordinal scales use intervals to compute the domain. This would _not_ benefit from an implicitly-coercing interval: we would need to do the coercion earlier on the scale’s associated channel values, so that the extent is computed correctly and so that the scale’s imputed domain matches the channel values.

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/scales/ordinal.js#L100-L101

The contour mark can use an interval to compute thresholds. The majority of the time the contour mark will be used with numeric data, but it is theoretically possible to compute contours of arbitrary values, including temporal data. So, this might benefit from implicit coercion. However, it again would not work to perform that coercion within the interval; we would need to coerce all the contour values, not just coerce when computing ticks.

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/marks/contour.js#L191

The bin mark can likewise use an interval to compute thresholds. It already performs coercion and would not benefit from additional implicit coercion inside the interval.

https://github.com/observablehq/plot/blob/820144e4a89b07f2a212f66156b4af0dc48a250b/src/transforms/bin.js#L232

To summarize the effect of implicitly-coercing intervals:

* interval transform 👍
* scale transform 👍
* ordinal scales 👎 (incorrect result)
* contour mark 👎 (incorrect result)
* axis mark 👎 (redundant with existing coercion)
* bin transform 👎 (redundant with existing coercion)

My recommendation therefore is, instead of implicitly-coercing time intervals, that we coerce values as appropriate prior to using the interval similar to what the bin transform already does. In the case of ordinal scales, we’ll also need to coerce to numbers, not just dates, so that we compute the extent correctly.

Fil:
I have revised this PR's description with the TODO.

src/transforms/interval.js
```diff
@@ -28,7 +29,9 @@ function maybeIntervalK(k, maybeInsetK, options, trivial) {
   let D1, V1;
   function transform(data) {
     if (V1 !== undefined && data === D1) return V1; // memoize
-    return (V1 = map(valueof((D1 = data), value), (v) => interval.floor(v)));
+    let V = valueof((D1 = data), value);
+    if (isTimeInterval(interval)) V = map(V, coerceDate, Float64Array);
```
mbostock, line 33:
```suggestion
    if (isTimeInterval(interval)) V = coerceDates(V);
```

I think we want the slower `coerceDates` here. This code using Float64Array is faster, but we can only use it in the bin transform because we know it’s not exposed outside the bin transform. I don’t think that’s true here; we want to pass the interval Date instances, not numbers.

src/plot.js
```diff
@@ -358,12 +369,13 @@ function applyScaleTransforms(channels, options) {
...
-    interval,
-    transform = percent ? (x) => x * 100 : maybeInterval(interval, type)?.floor
-  } = options[scale] ?? {};
+  let {type, percent, interval, transform} = options[scale] ?? {};
+  if (transform === undefined) {
+    if (percent) return (channel.value = channel.value.map((x) => x * 100)), undefined;
+    const i = maybeInterval(interval, type);
+    if (isTimeInterval(i)) return (channel.value = channel.value.map((d) => i.floor(coerceDate(d)))), undefined;
+    transform = i?.floor;
+  }
```
mbostock, line 378: I think it would be cleaner to refactor this into a helper like so:

```js
function maybeIntervalTransform(interval, type) {
  interval = maybeInterval(interval, type);
  return isTimeInterval(interval) ? (d) => interval.floor(coerceDate(d)) : interval?.floor;  
}
```

And then replace the default like so:

```js
transform = percent ? (x) => x * 100 : maybeIntervalTransform(interval, type)
```

In other words, I think it’s preferable to consolidate the code paths so that the interval is still promoted to a transform. And also we’d want to use `map` here instead of `channel.value.map` because `channel.value` could be a typed array.

mbostock: Also do we want to coerce to numbers for number intervals? Should we further assume that intervals only apply to numbers and dates and not other types? 🤔

PR #1332, opened 2023-03-14
strict range interval, nice interval

### LINKED ISSUE #1240: Time scales should support named intervals for the nice option
E.g., nice: "week" should be the same as nice: d3.utcWeek for a utc scale.

### LINKED ISSUE #1331: The axis mark requires a range interval, but doesn’t check for it
For example, if you say

```js
Plot.plot({
  x: {
    domain: [0, 10],
    interval: {
      floor: (x) => Math.floor(x),
      offset: (x, offset = 1) => x + offset
    }
  }
})
```

you get

```
TypeError: interval.range is not a function
```

here

https://github.com/observablehq/plot/blob/6b87468a7dafa6c25f10949a3124565c53b2c965/src/marks/axis.js#L521

You can fix the problem like so

Screenshot 2023-03-13 at 5 51 20 PM

```ts
Plot.plot({
  x: {
    domain: [0, 10],
    interval: <Plot.RangeInterval>{
      floor: (x) => Math.floor(x),
      offset: (x, offset = 1) => x + offset,
      range: (x1, x2) => d3.range(x1, x2)
    }
  }
})
```

but you need the type declaration to avoid a lint error, since ScaleOptions["interval"] is also erroneously an Interval instead of a RangeInterval. Though, there are some cases like with ordinal scales where a plain Interval may be allowable?

We should decide whether a range interval is required for the ScaleOptions["interval"] option. If it’s not, then the axis mark should gracefully handle the interval not implementing _interval_.range. If it is, then the axis mark should throw a nicer “invalid interval” error message.

### LINKED ISSUE #1333: scale.nice should support named intervals, too
You can pass e.g. `d3.utcDay` as the **nice** option for a _utc_ scale, but you can’t pass `"day"` and get the same result.

mbostock:
Fixes #1331.
Fixes #1240.

(Also fixes #1333 as a duplicate of #1240)

### DIFF
--- a/src/marks/axis.d.ts
+++ b/src/marks/axis.d.ts
@@ -33,9 +33,9 @@ export interface AxisXOptions extends AxisOptions, TickXOptions {}
 
 export interface AxisYOptions extends AxisOptions, TickYOptions {}
 
-export interface GridXOptions extends GridOptions, RuleXOptions {}
+export interface GridXOptions extends GridOptions, Omit<RuleXOptions, "interval"> {}
 
-export interface GridYOptions extends GridOptions, RuleYOptions {}
+export interface GridYOptions extends GridOptions, Omit<RuleYOptions, "interval"> {}
 
 export function axisY(options?: AxisYOptions): CompoundMark;
 
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -14,7 +14,7 @@ import {
   coerceDate,
   coerceNumbers,
   maybeColumn,
-  maybeInterval,
+  maybeRangeInterval,
   maybeTuple,
   maybeColorChannel,
   maybeValue,
@@ -24,6 +24,7 @@ import {
   isIterable,
   map
 } from "../options.js";
+import {maybeUtcInterval} from "../time.js";
 import {basic} from "./basic.js";
 import {
   hasOutput,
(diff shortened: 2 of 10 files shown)


src/transforms/bin.js
```diff
@@ -311,20 +313,11 @@ export function maybeThresholds(thresholds, interval, defaultThresholds = thresh
       case "auto":
         return thresholdAuto;
     }
-    const interval = maybeInterval(thresholds);
-    if (interval !== undefined) return interval;
-    throw new Error(`invalid thresholds: ${thresholds}`);
+    return maybeUtcInterval(thresholds);
```
mbostock, line 316: At this point, we’ve already checked that `typeof thresholds === "string"`, so the other code paths should not be possible.

src/marks/axis.d.ts
```diff
@@ -33,9 +33,9 @@ export interface AxisXOptions extends AxisOptions, TickXOptions {}
 
 export interface AxisYOptions extends AxisOptions, TickYOptions {}
 
-export interface GridXOptions extends GridOptions, RuleXOptions {}
+export interface GridXOptions extends GridOptions, Omit<RuleXOptions, "interval"> {}
```
mbostock, line 36: This is needed because the rule mark allows a base (non-range) interval, whereas the grid mark requires a range interval. You can widen a type when extending, but you can’t narrow it.

Fil (approved):

PR #1335, opened 2023-03-14
fix NaN ticks crashing

### LINKED ISSUE #1334: the default axis can crash on NaN ticks
OK it's pretty rare, but it's possible to crash by providing both the unknown option (so the tick goes somewhere on screen), and a NaN tick value:

Plot.plot({x: { type: "linear", ticks: [NaN, 1, 2], tickFormat: (d) => d, unknown: 1 }})

related: #493

Fil:
Guard against formatDefault returning undefined (it always returns a string except when the value is NaN)

Fixes #1334

related to #493


src/marks/text.js
```diff
@@ -137,7 +137,9 @@ function applyMultilineText(selection, mark, T, TL) {
   if (!T) return;
   const {lineAnchor, lineHeight, textOverflow, splitLines, clipLine} = mark;
   selection.each(function (i) {
-    const lines = splitLines(formatDefault(T[i])).map(clipLine);
+    const t = formatDefault(T[i]);
+    if (t == null) return;
+    const lines = splitLines(t).map(clipLine);
```
mbostock, line 142: How about this, for brevity? I don’t think we need to optimize for the case where formatDefault returns null…

```suggestion
    const lines = splitLines(formatDefault(T[i]) ?? "").map(clipLine);
```

mbostock (changes requested):
Nice catch. I’m a little surprised (I had forgotten) that the format functions can return undefined. I wonder if it wouldn’t be better to have them return the empty string so that callers can rely on them always returning a string? It looks like this would be acceptable for our usage and it would obviate a need for a workaround here.

I’m also a little surprised that we format invalid dates as “Invalid Date” when we format NaN as undefined (equivalent to the empty string). Maybe those should be more consistent, too? But that’s a separate issue.

mbostock (approved):
I experimented more with changing the format functions to return the empty string instead of undefined for the edge cases, and I changed my mind: I think it’s nice to retain the type safety of undefined in these cases. For example, our default filtering removes null and undefined, but not the empty string.

I incorporated the `??` style suggestion.

PR #1337, opened 2023-03-14
type tests

Fil:
- [x] tests salvaged from https://github.com/observablehq/plot/pull/1024
- [x] fix valueof crash


src/options.js
```diff
@@ -128,7 +128,7 @@ export function arrayify(data) {
 // An optimization of type.from(values, f): if the given values are already an
 // instanceof the desired array type, the faster values.map method is used.
 export function map(values, f, type = Array) {
-  return values instanceof type ? values.map(f) : type.from(values, f);
+  return !values ? values : values instanceof type ? values.map(f) : type.from(values, f);
```
mbostock, line 131: My expectation was that the nullish (or falsey) check should happen earlier, not here in map. At this point we should know that values is non-null and an array or array-like.

Fil: I don't know. Yes it was in valueof in #1024, but with the current code we need the test on lines 12, 14 and 16. And it felt repetitive to do:

```
  return valueType === "string"
    ? !data
      ? data
      : maybeTypedMap(data, field(value), type) // 12
    : valueType === "function"
    ? !data
      ? data
      : maybeTypedMap(data, value, type) // 14
    : valueType === "number" || value instanceof Date || valueType === "boolean"
    ? !data
      ? data
      : map(data, constant(value), type) // 16
    : value && typeof value.transform === "function"
    ? maybeTypedArrayify(value.transform(data), type)
    : maybeTypedArrayify(value, type);
```

I noticed that line 16 could be optimized as:
`new (type ?? Array)(arrayify(data).length).fill(value)
`
or maybe even:
`new (type ?? typeof value === "number" ? Float64Array : Array)(arrayify(data).length).fill(value)`

(Both versions pass all tests.)

mbostock: Can we move the `data == null` test first to avoid repeating it?

mbostock: Maybe this?

```js
  if (typeof value?.transform === "function") return maybeTypedArrayify(value.transform(data), type);
  if (data == null) return data;
  return valueType === "string"
    ? maybeTypedMap(data, field(value), type)
    : valueType === "function"
    ? maybeTypedMap(data, value, type)
    : valueType === "number" || value instanceof Date || valueType === "boolean"
    ? map(data, constant(value), type)
    : maybeTypedArrayify(value, type);
}
```

Fil: nope… if _value_ is an array, we don't care if _data_ is nullish. But from this variant we can test data==null in maybeTypedMap, which (contrary to map) is not used anywhere else.

test/types/valueof-test.js
```diff
@@ -0,0 +1,45 @@
+import assert from "assert";
+import {valueof} from "../../src/options.js";
+
+it("valueof reads arrays", () => {
```
mbostock, line 4: We already have tests for many of these cases in test/options-test.js We should consolidate.

We still don’t have test coverage for is the valueof(null, array) case you mentioned.

src/options.js
mbostock: > if value is an array, we don't care if data is nullish

That’s not a requirement, and I don’t think that was intentional historically. The only time data should be nullish is for decorations—marks that are not data-driven, such as a frame. Decoration marks never have channels associated with them. And marks without data by definition don’t have an index, so even if they did theoretically have a channel, it wouldn’t be meaningful because there would be no index by which to lookup values in the channel. So, it may have been true that valueof returned the array in the valueof(null, array) case, but I don’t think we need to maintain that behavior. And we should add tests for it in any case.

test/types/valueof-type-test.ts
mbostock, line 1: What if we instead turned test/options.js into test/options.ts. Would we still need this file or could we rely on type validation of the other tests we are already running?

Fil: Yes, we used tsd in the #1024 branch, but it's simpler to use tsc.

The "valid input data" were there to contrast with the ones below that expect ts errors, but they were not useful.

I've consolidated the tests (most of them were redundant).

mbostock (approved):

Issue #1340, opened 2023-03-15
Auto mark: no display with non-zero reducer and bar mark


tophtucker:
[See examples](https://observablehq.com/d/02b8ff51e7cf1c33). You get a bunch of zero-height bars:

<img src="https://user-images.githubusercontent.com/841829/225201341-83fed53e-7b6b-49a6-9c95-fb8d182102f8.png" width="640" />  

```
Plot.auto(olympians, {x: "date_of_birth", y: { value: "height", reduce: "mean" }, mark: "bar"}).plot()
```

The easy fix is to change the last row of this to use rectY instead of rect, on the theory that Plot.auto never has enough information to use Plot.rect effectively:

https://github.com/observablehq/plot/blob/c4892908a75c8b85abdd11475539c41d17b80962/src/marks/auto.js#L159-L174

That works in this case:

<img src="https://user-images.githubusercontent.com/841829/225201687-0f6e64f1-f6cb-4581-b9ea-7305380c138a.png" width="640" />

But it breaks the heatmap, because it gives it a zero baseline:

<img src="https://user-images.githubusercontent.com/841829/225201810-ebe3cd7b-dc19-4061-808e-bf65a27300da.png" width="640" />

```
Plot.auto(penguins, {x: "culmen_length_mm", y: "body_mass_g", color: "count"}).plot()
```

Because we're inferring whether to show the zero baseline from the mark type:

https://github.com/observablehq/plot/blob/c4892908a75c8b85abdd11475539c41d17b80962/src/marks/auto.js#L217-L220

We could change that bit. But, moreover, the heatmap example shows that Plot.auto _does_ sometimes have enough information to specify a rect mark: in the case of 2D binning. So just switching rect to rectY feels wrong.

mbostock:
The mark-based zero check is also forgetting about rule. This should have a y = 0 rule like rect does.

![untitled (25)](https://user-images.githubusercontent.com/230541/225204706-200d217f-de05-4040-8b6d-34dc96d5e6b9.png)

```js
Plot.auto(olympians, {x: "date_of_birth", y: {value: "height", reduce: "mean"}, mark: "rule"}).plot()
```

tophtucker:
@mbostock Can you think of a test case where using rect instead of rectY would break aside from the zero-baseline issue? These look identical:

```js
Plot.rect(penguins, Plot.bin({fill: "count"}, {x: "culmen_length_mm", y: "body_mass_g"})).plot()
[...]

PR #1341, opened 2023-03-15
Auto mark: zero baseline default for rules

tophtucker:
[Mike writes:](https://github.com/observablehq/plot/issues/1340#issuecomment-1469302368)

>The mark-based zero check is also forgetting about rule. This should have a y = 0 rule like rect does.  

Now it does! 

|Before|After|
|-|-|
|autoRuleZero-changed|autoRuleZero|

```js
Plot.auto(olympians, {x: "date_of_birth", y: {value: "height", reduce: "mean"}, mark: "rule"}).plot()
```

A couple notes:

barY or areaY or rectY or… ruleX? Yes; ruleY would be the one parallel to the y baseline, and can make sense without any zero baseline; ruleX is the one orthogonal to the baseline, which has to “stand on it” (like barY).

Merely adding `|| mark === ruleY` and `|| mark === ruleX` to the conditional breaks autoNullReduceContinuous and autoNullReduceDate, so I also added the `X &&` and `Y &&` checks, which seem like a good idea anyway.

|Broken|Fixed|
|--|--|
|autoNullReduceContinuous-changed|autoNullReduceContinuous|

The new test case demonstrates the existing issue that a rule can overhang its baseline… but that sounds like a whole can of worms and I wouldn't worry about it lol:

image

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -216,8 +216,8 @@ export function auto(data, options) {
 
   // If zero-ness is not specified, default based on whether the resolved mark
   // type will include a zero baseline. TODO Move this to autoSpec.
-  if (xZero === undefined) xZero = transform !== binX && (mark === barX || mark === areaX || mark === rectX);
-  if (yZero === undefined) yZero = transform !== binY && (mark === barY || mark === areaY || mark === rectY);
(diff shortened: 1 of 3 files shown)


mbostock (approved):
Nice. 👍

PR #1347, opened 2023-03-16
allow mixed-case class names

### LINKED ISSUE #1345: validClassName doesn't like camelCase class names.
https://github.com/observablehq/plot/blob/e2749e8bd3f28305f02c7aa6e7f0d9c0f99e93cb/src/style.js#L417

validClassName.test('myNewClassName') fails.

mbostock:
Fixes #1345.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -415,7 +415,7 @@ export function impliedNumber(value, impliedValue) {
 }
 
 const validClassName =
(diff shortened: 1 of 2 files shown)


mbostock:
@Fil Can you provide a reference for the regex you added to validate class names in #582? I’d like to double-check that we can’t simplify that further.

Fil:
***@***.***/valid-classname

Fil (approved):
Ha that's the topic of this PR :)
I should have read the whole thing before answering your question!

Issue #1343, opened 2023-03-15
Tracking issue for TypeScript documentation


mbostock:
- [x] [src/channel.d.ts](https://github.com/observablehq/plot/blob/main/src/channel.d.ts)
- [x] ~~[src/color.d.ts](https://github.com/observablehq/plot/blob/main/src/color.d.ts)~~
- [x] [src/context.d.ts](https://github.com/observablehq/plot/blob/main/src/context.d.ts)
- [x] [src/curve.d.ts](https://github.com/observablehq/plot/blob/main/src/curve.d.ts)
- [x] [src/dimensions.d.ts](https://github.com/observablehq/plot/blob/main/src/dimensions.d.ts)
- [x] ~~[src/facet.d.ts](https://github.com/observablehq/plot/blob/main/src/facet.d.ts)~~
- [x] [src/format.d.ts](https://github.com/observablehq/plot/blob/main/src/format.d.ts)
- [x] [src/inset.d.ts](https://github.com/observablehq/plot/blob/main/src/inset.d.ts)
- [x] ~~[src/interpolate.d.ts](https://github.com/observablehq/plot/blob/main/src/interpolate.d.ts)~~
- [x] [src/interval.d.ts](https://github.com/observablehq/plot/blob/main/src/interval.d.ts)
- [x] [src/legends.d.ts](https://github.com/observablehq/plot/blob/main/src/legends.d.ts)
- [x] [src/mark.d.ts](https://github.com/observablehq/plot/blob/main/src/mark.d.ts) [#1405](https://github.com/observablehq/plot/pull/1405)
- [x] [src/marker.d.ts](https://github.com/observablehq/plot/blob/main/src/marker.d.ts)
- [x] [src/marks/area.d.ts](https://github.com/observablehq/plot/blob/main/src/marks/area.d.ts) [#1402](https://github.com/observablehq/plot/pull/1402)
- [x] [src/marks/arrow.d.ts](https://github.com/observablehq/plot/blob/main/src/marks/arrow.d.ts) [#1398](https://github.com/observablehq/plot/pull/1398)
- [x] [src/marks/auto.d.ts](https://github.com/observablehq/plot/blob/main/src/marks/auto.d.ts) [#1385](https://github.com/observablehq/plot/pull/1385)
- [x] [src/marks/axis.d.ts](https://github.com/observablehq/plot/blob/main/src/marks/axis.d.ts) [#1403](https://github.com/observablehq/plot/pull/1403)
[...]

PR #1349, opened 2023-03-17
document window

Fil:
remarks:
* *k* is not optional
* *shift* is deprecated
* added the *strict* option to README

for #1343

### DIFF
--- a/src/transforms/window.d.ts
+++ b/src/transforms/window.d.ts
@@ -65,8 +65,8 @@ export interface WindowOptions {
    */
   anchor?: "start" | "middle" | "end";
 
-  /** The shift option is deprecated. See **anchor**. */
-  shift?: never; // was "leading" | "centered" | "trailing";
+  /** @deprecated See **anchor**. */
+  shift?: "leading" | "centered" | "trailing";
 
   /** If the **strict** option is true, the output start values or end values or
    * both (depending on the **anchor**) of each series may be undefined since
--- a/src/transforms/window.d.ts
+++ b/src/transforms/window.d.ts
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
 export type WindowReducer = WindowReducerName | WindowReducerFunction;
 
 export interface WindowOptions {
-  k?: number;
+  /**
+   * The size —number of consecutive values— in the window; includes the current
+   * value.
+   */
+  k: number;
+
+  /**
+   * The reducer returns a summary statistic from the **k** values in the
+   * current window, say to compute rolling averages, rolling minimums, or
+   * rolling maximums. The following window reducers are supported:
+   *
+   * * *min* - the minimum
+   * * *max* - the maximum
+   * * *mean* - the mean (average)
+   * * *median* - the median
+   * * *mode* - the mode (most common occurrence)
+   * * *pXX* - the percentile value, where XX is a number in [00,99]
+   * * *sum* - the sum of values
+   * * *deviation* - the standard deviation
+   * * *variance* - the variance per [Welford’s
+   *   algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm)
+   * * *difference* - the difference between the last and first window value
+   * * *ratio* - the ratio of the last and first window value
+   * * *first* - the first value
+   * * *last* - the last value
+   * * a function to be passed an array of *k* values
+   *
+   * Defaults to mean.
+   */
   reduce?: WindowReducer;
+
+  /**
+   * How to align the window:
+   *
+   * * *start* - the current value is the first element in the window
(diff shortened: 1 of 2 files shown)


src/transforms/window.d.ts
```diff
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
...
+   *   the left if the size **k** is even)
+   * * *end* - the current value is the last element in the window.
+   *
+   * Defaults to middle.
+   */
   anchor?: "start" | "middle" | "end";
-  shift?: "leading" | "centered" | "trailing"; // deprecated!
+
+  /** The shift option is deprecated. See **anchor**. */
+  shift?: never; // was "leading" | "centered" | "trailing";
```
mbostock, line 69: The **shift** option is deprecated but still supported: existing usage of the option should not cause an error. Changing the type to `never` will cause a `Type 'string' is not assignable to type 'never'` error for TypeScript users:

Screenshot 2023-03-17 at 6 35 27 AM

Instead, we should leave the types as-is, but use the JSDoc `@deprecated` tag. VS Code can then show deprecated features as striked out:

Screenshot 2023-03-17 at 6 41 44 AM

Unfortunately it doesn’t seem to strike out option declarations, though…

Fil: I understand why we wouldn't want a type error, but the paradoxal consequence is that _shift_ is now actively suggested as a proper autocomplete choice. Let's hope VS Code becomes smarter with this tag?

mbostock: It’s suggested, but striked out. So hopefully it’s clear that it’s supported but deprecated? Seems like this is functioning as desired.

Screenshot 2023-03-17 at 7 54 02 AM

Fil: All good now (after a restart of VS Code…)

src/transforms/window.d.ts
```diff
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
 export type WindowReducer = WindowReducerName | WindowReducerFunction;
 
 export interface WindowOptions {
-  k?: number;
+  /**
+   * The size —number of consecutive values— in the window; includes the current
```
mbostock, line 26:
```suggestion
   * The size (number of consecutive values) in the window; includes the current
```

src/transforms/window.d.ts
```diff
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
...
+ * rolling averages, rolling minimums, or rolling maximums. The window options
+ * can be specified as the first argument, or grouped with the *options*. For
+ * example, the following are equivalent:
+ *
+ * ```js
+ * Plot.windowX(24, {x: "Anomaly", y: "Date"});
+ * Plot.windowX({k: 24, reduce: "mean", x: "Anomaly", y: "Date"});
+ * Plot.windowX({k: 24, reduce: "mean"}, {x: "Anomaly", y: "Date"});
+ * ```
+ */
```
mbostock, line 114: Good news! It looks like if you document the first declaration of an overloaded function, the documentation is inherited by the other overloads. So we don’t need to repeat this documentation twice.

```suggestion
```

src/transforms/window.d.ts
```diff
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
...
+ * rolling averages, rolling minimums, or rolling maximums. The window options
+ * can be specified as the first argument, or grouped with the *options*. For
+ * example, the following are equivalent:
+ *
+ * ```js
+ * Plot.windowY(24, {x: "Date", y: "Anomaly"});
+ * Plot.windowY({k: 24, reduce: "mean", x: "Date", y: "Anomaly"});
+ * Plot.windowY({k: 24, reduce: "mean"}, {x: "Date", y: "Anomaly"});
+ * ```
+ */
```
mbostock, line 144:
```suggestion
```

src/transforms/window.d.ts
```diff
@@ -22,19 +22,135 @@ export type WindowReducerFunction = (values: any[]) => any;
...
+   *
+   * Defaults to middle.
+   */
   anchor?: "start" | "middle" | "end";
-  shift?: "leading" | "centered" | "trailing"; // deprecated!
+
+  /** @deprecated See **anchor**. */
+  shift?: "leading" | "centered" | "trailing";
+
+  /** If the **strict** option is true, the output start values or end values or
```
mbostock, line 71:
```suggestion
  /**
    * If the **strict** option is true, the output start values or end values or
```

mbostock (approved):
I tightened up the wording slightly, and split the paragraphs for the **strict** option to improve readability (since that was a bit long). I also clarified that the windowX and windowY transforms don’t just affect _x_ and _y_ respectively, they affect _x_, _x1_, and _x2_, and _y_, _y1_, and _y2_.

PR #1351, opened 2023-03-17
documentation for the tree mark and the treeLink, treeNode transforms

Fil:
For #1343

We could possibly add more information, like how dx defaults to +6 if the anchor is left, etc.


src/marks/tree.d.ts
```diff
@@ -6,10 +6,32 @@ import type {TextOptions} from "./text.js";
...
+   */
   dot?: boolean;
+  /**
+   * A **stroke** color to pass to the text mark, to limit label occlusion. Defaults to white.
+   */
   textStroke?: MarkOptions["stroke"];
 }
 
+/**
+ * Transforms a tabular dataset into a hierarchy according to the given **path** input channel, which is typically a slash-separated string; then executes a tree layout algorithm to compute **x** and **y** output channels; these channels can then be fed to other marks to construct a node-link diagram.
```
mbostock, line 20: Needs wrapping. (I will fix.)

src/marks/tree.d.ts
```diff
@@ -6,10 +6,32 @@ import type {TextOptions} from "./text.js";
...
+ *
+ * * **treeLayout** - a tree layout algorithm; defaults to [d3.tree](https://github.com/d3/d3-hierarchy/blob/main/README.md#tree)
+ * * **treeAnchor** - a tree layout orientation, either *left* or *right*; defaults to *left*
+ * * **treeSort** - a node comparator, or null to preserve input order
+ * * **treeSeparation** - a node separation function, or null for uniform separation
+ *
+ */
 export function tree(data?: Data, options?: TreeOptions): CompoundMark;
 
 export function cluster(data?: Data, options?: TreeOptions): CompoundMark;
```
mbostock, line 52: Missing documentation. (I have fixed.)

src/transforms/tree.d.ts
```diff
@@ -2,14 +2,106 @@ import type {ChannelValue} from "../channel.js";
...
+   * is typically slash-separated, as with UNIX-based file systems or URLs.
+   * Defaults to identity.
+   */
   path?: ChannelValue;
+  /**
+   * The path separator; defaults to forward slash (/).
+   */
   delimiter?: string;
+  /**
+   * If the **treeAnchor** is *left*, the root of the tree will be aligned with
```
mbostock, line 16: These options could use a summary sentence prior to diving into the possible ways in which the options could be specified. For example:

* treeAnchor - how to orient the tree
* treeLayout - how to position nodes in the tree
* treeSeparation - how much space to reserve between adjacent nodes
* treeSort - how to order nodes prior to computing the layout

etc.

src/transforms/tree.d.ts
```diff
@@ -2,14 +2,106 @@ import type {ChannelValue} from "../channel.js";
...
+   * _node_.sort](https://github.com/d3/d3-hierarchy/blob/main/README.md#node_sort)
+   * for more. The **treeSort** option can also be specified as a string, in
+   * which case it refers either to a named column in data, or if it starts with
+   * “node:”, a node value such as node:name.
+   */
   treeSort?: CompareFunction | {node: (node: any) => any} | string | null;
 }
 
+/**
+ *  Based on the **tree** options (**path**, **delimiter**, **treeAnchor**,
```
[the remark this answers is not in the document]
mbostock: Also, I don’t think we need to repeat the tree options here. They should be clear from autocomplete (or jumping to the TreeTransformOptions interface).

mbostock (approved):

src/marks/tree.d.ts
```diff
@@ -6,10 +6,47 @@ import type {TextOptions} from "./text.js";
...
+   */
   textStroke?: MarkOptions["stroke"];
 }
 
+/**
+ * Transforms a tabular dataset into a hierarchy according to the given **path**
+ * input channel, which typically contains slash-separated strings; then
+ * executes a tree layout algorithm, by default Reingold–Tilford’s “tidy”
+ * algorithm, to compute *x* and *y* output channels; these channels can then be
+ * fed to other marks to construct a node-link diagram.
```
mbostock, line 26: This sounds like it’s describing the tree transform, whereas here we should be describing the tree compound mark, which means this function is doing the “feeding” (to a link, dot, and text mark).

PR #1352, opened 2023-03-17
document stack

Fil:
* expand is deprecated; use normalize in tests

### DIFF
--- a/test/plots/industry-unemployment-share.ts
+++ b/test/plots/industry-unemployment-share.ts
@@ -15,7 +15,7 @@ export async function industryUnemploymentShare() {
           x: "date",
           y: "unemployed",
           fill: "industry",
-          offset: "expand",
+          offset: "normalize",
           title: "industry"
         })
       ),
--- a/test/plots/stacked-bar.ts
+++ b/test/plots/stacked-bar.ts
@@ -12,7 +12,7 @@ export async function stackedBar() {
           x: (d, i) => i,
           fill: (d, i) => i,
           insetLeft: 1,
-          offset: "expand"
+          offset: "normalize"
         })
       )
     ]
--- a/test/plots/stacked-rect.ts
+++ b/test/plots/stacked-rect.ts
@@ -12,7 +12,7 @@ export async function stackedRect() {
           x: (d, i) => i,
           fill: (d, i) => i,
           insetLeft: 1,
-          offset: "expand"
+          offset: "normalize"
         }
       )
     ]


Fil:
Thanks for the clarification. I was wanting to investigate what "x" or "y" did differently than "value", since I wasn't sure I remembered precisely. It's fine to keep them around as aliases, we just need to document them more clearly.

README.md
```diff
@@ -2570,7 +2571,7 @@ The stack transform supports diverging stacks: negative values are stacked below
 After all values have been stacked from zero, an optional **offset** can be applied to translate or scale the stacks. The following **offset** methods are supported:
 
 - null (default) - a zero baseline
-- *expand* (or *normalize*) - rescale each stack to fill [0, 1]
+- *normalize* (or *expand*) - rescale each stack to fill [0, 1]
```
mbostock, line 2574: Funny, in #289 I was thinking of *normalize* as an alias for *expand*, not that we were deprecating *expand*. But I agree we should favor one to reduce confusion. In other cases where we’ve deprecated options, we’ve removed the deprecated options from the README entirely. I think we should probably do that here, too—better to not spend the user’s attention on documenting deprecated options.

src/transforms/stack.d.ts
```diff
@@ -21,32 +21,110 @@ export type StackOrder =
   | any[]; // explicit ordinal values
 
 export interface StackOptions {
+  /**
+   * After all values have been stacked from zero, an optional **offset** can be
+   * applied to translate or scale the stacks:
+   *
+   * - null (default) - a zero baseline
+   * - *normalize* (or *expand*) - rescale each stack to fill [0, 1]
+   * - *center* (or *silhouette*) - align the centers of all stacks
```
mbostock, line 30:
```suggestion
   * - *normalize* - rescale each stack to fill [0, 1]
   * - *center* - align the centers of all stacks
```

Per above.

src/transforms/stack.d.ts
```diff
@@ -21,32 +21,110 @@ export type StackOrder =
...
+   * the default order to *inside-out*.
+   */
   offset?: StackOffset | null;
+
+  /**
+   * The order in which stacks are layered:
+   *
+   * - null (default) - input order
+   * - *value* - ascending value order (or descending with **reverse**)
+   * - *x* for stackY, or *y* for stackX - alias of *value*
```
mbostock, line 47: Better to keep the list consistently formatted like so:

```suggestion
   * - *x* - alias of *value*; for stackY only
   * - *y* - alias of *value*; for stackX only
```

src/transforms/stack.d.ts
```diff
@@ -21,32 +21,110 @@ export type StackOrder =
...
+   * - *sum* - order series by their total value
+   * - *appearance* - order series by the position of their maximum value
+   * - *inside-out* - order the earliest-appearing series on the inside
+   *   (recommended for the *wiggle* offset)
```
mbostock, line 51: It’s the default when using the _wiggle_ offset, in fact.

src/transforms/stack.d.ts
```diff
@@ -21,32 +21,110 @@ export type StackOrder =
...
+   * - a named field or function of data - order data by priority
+   * - an array enumerating all the *z* values in the desired order
+   *
+   * The **reverse** option reverses the effective order. For the *value* order,
```
mbostock, line 55: Do we need to document the **reverse** option here? Isn’t it already covered below?

Fil: 👍 I've now removed the whole paragraph—the three sentences are all explicited elsewhere.

README.md
```diff
@@ -2557,9 +2557,12 @@ The following **order** methods are supported:
 
 - null (default) - input order
 - *value* - ascending value order (or descending with **reverse**)
+- *x* - alias of *value*; for stackY only
+- *y* - alias of *value*; for stackX only
```
mbostock, line 2561: This is reversed. I can fix.

```suggestion
- *x* - alias of *value*; for stackX only
- *y* - alias of *value*; for stackY only
```

mbostock (approved):

PR #1353, opened 2023-03-17
document select

Fil:
for #1343 

note that we don't cover the case where symbol is constant, but we try to select it with `select({symbol: "min"})`; this should be red:

Capture d’écran 2023-03-17 à 18 28 18

### DIFF
--- a/src/transforms/select.d.ts
+++ b/src/transforms/select.d.ts
@@ -13,16 +13,50 @@ export interface SelectOptions {
   z?: ChannelValue;
 }
 
+/**
+ * Selects points from each series (usually defined by *stroke*, *fill* or *z*)
+ * based on a selector. The selector can be specified either as a function which
+ * receives as input the index of the series, the shorthand “first” or “last”,
+ * or as a {*key*: *value*} object with exactly one *key* being the name of a
+ * channel and the *value* being a function which receives as input the index of
+ * the series and the channel values. The *value* may alternatively be specified
+ * as the shorthand “min” and “max” which respectively select the minimum and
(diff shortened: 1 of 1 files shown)


mbostock:
> we don't cover the case where symbol is constant, but we try to select it

This is true in general; the types don’t validate that the channels referenced in the outputs actually exist. For example, you could declare `title: "mean"` as an output, but you wouldn’t get an error if you don’t declare the _title_ channel as an input.

src/transforms/select.d.ts
```diff
@@ -13,16 +13,50 @@ export interface SelectOptions {
   z?: ChannelValue;
 }
 
+/**
+ * Selects points from each series (usually defined by *stroke*, *fill* or *z*)
```
mbostock, line 17: The “usually defined by *stroke*, *fill* or *z*” feels unnecessarily imprecise. How would the user know when or not the “usual” definition applies? We should list these channels in order of precedence: *z*, *fill*, *stroke*. The README is more definite:

> The data are grouped into series using the z, fill, or stroke channel in the same fashion as the [area](https://github.com/observablehq/plot/blob/main/README.md#area) and [line](https://github.com/observablehq/plot/blob/main/README.md#line) marks.

And similarly for the area mark:

> By default, the data is assumed to represent a single series (i.e., a single value that varies over time). If the z channel is specified, data is grouped by z to form separate series. Typically z is a categorical value such as a series name. If z is not specified, it defaults to fill if a channel, or stroke if a channel.

mbostock (approved):

PR #1354, opened 2023-03-17
document normalize

Fil:
for #1343

lots of repetitions, but looks ok?

### DIFF
--- a/src/transforms/normalize.d.ts
+++ b/src/transforms/normalize.d.ts
@@ -22,12 +22,117 @@ export interface NormalizeOptions {
   basis?: NormalizeBasis;
 }
 
+/**
+ * Normalize series values relative to the given basis. For example, if the
+ * series values are [*x₀*, *x₁*, *x₂*, …] and the *first* basis is used, the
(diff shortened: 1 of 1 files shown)


src/transforms/normalize.d.ts
```diff
@@ -22,12 +22,73 @@ export interface NormalizeOptions {
   basis?: NormalizeBasis;
 }
 
+/**
+ * Normalize series values relative to the given basis. For example, if the
```
mbostock, line 26: This is another instance where we want the first sentence to emphasize what is unique about normalizeX as opposed to shared across all variations of the normalize transform: it should mention specifically which channels are going to be transformed. (Fixing as I edit.)

mbostock: Also, since the normalize transform is built on top of the map transform, we should also mention that data are first grouped into series. From Plot.map in the README:

> Groups on the first channel of z, fill, or stroke, if any, and then for each channel declared in the specified outputs object, applies the corresponding map method.

PR #1359, opened 2023-03-20
document hexbin

Fil:
for #1343 

(do we need to explain that it's an initializer?)

### DIFF
--- a/src/transforms/hexbin.d.ts
+++ b/src/transforms/hexbin.d.ts
@@ -43,6 +43,7 @@ export interface HexbinOptions {
  *   the bin
  * * an object with a *reduce* method
  *
+ * This transform defaults the *symbol* option of the **dot** mark to *hexagon*.
  * See also the **hexgrid** mark.
  */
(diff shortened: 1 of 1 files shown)


mbostock (approved):

mbostock:
> (do we need to explain that it's an initializer?)

The return type should be `Initialized<T>` instead of `Transformed<T>`; fixed.

src/transforms/hexbin.d.ts
```diff
@@ -4,8 +4,8 @@ import type {Initialized} from "./basic.js";
 /** Options for the hexbin transform. */
 export interface HexbinOptions {
   /**
-   * The distance between centers of neighboring (pointy-topped) hexagons, in
-   * pixels; defaults to 20.
+   * The horizontal distance between centers of neighboring (pointy-topped)
+   * hexagons, in pixels; defaults to 20.
```
Fil, line 8: _The distance between centers of neighboring hexagons_ was complete and correct :)

mbostock: Of course! Oops. Thanks for the correction.

PR #1360, opened 2023-03-20
document group & bin

Fil:
for #1343 

TODO:
- [ ] this is already very long (too long)
- [x] add comments on using the sort{ fy: "x", limit: 10} input option.
- [x] reducer _scope_? (as in transforms/bin)
- [x] fix types for the {reduce()} function (& create test plots with an explicit scope)
- [x] bin
- [x] ~~reducer label (? mentioned in the comments)~~ (see #1364)
- [x] tests for reducer & scale (see #1363)
- [x] test {x1,x2…} parameter for the bin reducer function
- [ ] apply changes from #1367 (reduceIndex)

### DIFF
--- a/src/transforms/bin.d.ts
+++ b/src/transforms/bin.d.ts
@@ -7,13 +7,13 @@ import type {Transformed} from "./basic.js";
 export type ThresholdsName = "freedman-diaconis" | "scott" | "sturges" | "auto";
 
 /** How to subdivide a continuous domain into discrete bins (based on data). */
-export type ThresholdsFunction = (values: any[], min: any, max: any) => any[] | RangeInterval | number;
+export type ThresholdsFunction<T = number | Date> = (values: T[], min: T, max: T) => T[] | RangeInterval<T> | number;
 
 /** How to subdivide a continuous domain into discrete bins. */
-export type Thresholds = ThresholdsName | ThresholdsFunction | RangeInterval | any[];
+export type Thresholds<T = number | Date> = ThresholdsName | ThresholdsFunction<T> | RangeInterval<T> | T[] | number;
 
 /** Options for the bin transform. */
-export interface BinOptions {
+export interface BinOptions<T = number | Date> {
   /**
    * If true or a positive number, produce a cumulative distribution; if a
    * negative number, produce a [complementary cumulative](https://en.wikipedia.org/wiki/Cumulative_distribution_function#Complementary_cumulative_distribution_function_.28tail_distribution.29)
@@ -30,7 +30,7 @@ export interface BinOptions {
    * domain is used, the start and end of the domain will be extended to align
    * with the interval.
    */
-  domain?: ((values: any[]) => [min: any, max: any]) | [min: any, max: any];
+  domain?: ((values: T[]) => [min: T, max: T]) | [min: T, max: T];
 
   /**
    * How to subdivide the domain into bins. May be one of:
@@ -44,7 +44,7 @@ export interface BinOptions {
    * * an interval; see **interval**
    * * a function that returns an array, count, or interval
    */
-  thresholds?: Thresholds;
+  thresholds?: Thresholds<T>;
 
   /**
    * How to subdivide the domain into bins; an alternative to **thresholds**.
@@ -53,7 +53,7 @@ export interface BinOptions {
    * number *n*, threshold values are consecutive multiples of *n* that span the
    * domain.
    */
-  interval?: RangeInterval;
+  interval?: RangeInterval<T>;
 }
 
 /** How to reduce binned values. */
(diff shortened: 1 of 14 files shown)


Fil:
I've just added an incomplete commit that shows an underlying type bug with ChannelReducers. Not sure how to solve this.

mbostock:
> The incomplete test is in test/plots/penguin-island-mode.ts; it fails because our types aren't certain that the reduce function is called with two arguments

That’s only true in this branch because you changed the definition of ReducerImplementation. In the main branch TypeScript seems to infer that it’s a ReducerImplementation:

Screenshot 2023-03-20 at 10 44 31 AM

But, I do think there is ambiguity and likely a bug here introduced by #1310 (to fix #1309). Take this code:

```js
Plot.groupX({y: "count", fill: {reduce: f}}, {x: "species", fill: "island"})
```

In this case, it’s hard to tell whether the **fill** output channel is being specified as

1. a reducer with a scale override, in which case *f* is a ReducerFunction (or a ReducerImplementation object)
2. a reducer implementation, in which case *f* is a ReducerImplementation["reduce"]

Currently we assume (1)

https://github.com/observablehq/plot/blob/2ac08e266d1dd9743870d2c1ebbaf56ceb992524/src/transforms/group.js#L181

and if you want a reducer implementation, you would need another level of nesting to get (2):

```js
Plot.groupX({y: "count", fill: {reduce: {reduce: f}}}, {x: "species", fill: "island"})
```

but this should be considered a bug since we didn’t intend to change the interpretation. Hmm, tricky! I suppose we could look at the length of the reduce function to disambiguate but that could be brittle.

mbostock:
I wonder if we should make the map/reducer implementations much less ambiguous by using a symbol, e.g., {[Plot.reduce]: …} instead of {reduce: …} to declare an implementation. That would be possible like so:

```ts
export declare const reduce: unique symbol;

export interface ReducerImplementation {
  [reduce](index: number[], values: any[]): any;
}
```

It wouldn’t be backwards compatible, though… 🤔

mbostock:
Yes, please don’t make the label change in this documentation PR. It’s more subtle than it looks and is now delaying adding the documentation.

mbostock (changes requested):
The changes to the Reducer types aren’t right.

A ReducerImplementation is an _implementation_; it can only be an object with a **reduce** method that takes _index_ and _values_ (plus a _context_ if there’s a **scope**, and an _extent_ in the case of the bin transform). A ReducerImplementation is never a named built-in reducer name, and a ReducerImplementation cannot declare a scale.

The only place where you can declare a scale override is in the context of a ChannelReducers object.

Also, the BinReducer types are specific to the bin transform, and therefore belong in src/transforms/bin; not sure why you moved them to src/reducer.

I think it’s okay to propose some small type changes along with documentation, but this is bigger than I was expecting. Changes to the types and lots of tests should be done separately from documentation. Also, I intentionally avoided documenting the reducer **scope** before because it’s _very_ rarely used and adds complexity. I’m not against including **scope** in the types if we can figure it out, but if we can’t, I definitely think we should cut it because we already have a huge amount of work just to do the documentation, and the types are useful as-is.

Fil:
Got it—I'm reverting all this. I was following my comment [above](https://github.com/observablehq/plot/pull/1360#issuecomment-1476386450), but it took me too far :)


Also this seems to work, but I wouldn't know how to express it as an _interface_:

```
export type ReducerImplementation =
  | {
      reduce(index: number[], values: any[]): any;
    }
  | {
      reduce(index: number[], values: any[], summary?: any): any;
      scope: "data" | "facet";
    };
```


```
export type BinReducerImplementation =
  | {
      reduce(index: number[], values: any[], extent: {x1: any; y1: any; x2: any; y2: any}): any;
    }
  | {
      reduce(index: number[], values: any[], summary?: any, extent?: {x1: any; y1: any; x2: any; y2: any}): any;
      scope: "data" | "facet";
    };

```

src/transforms/bin.d.ts
```diff
@@ -7,12 +7,45 @@ export type ThresholdsName = "freedman-diaconis" | "scott" | "sturges" | "auto";
...
 export type ThresholdsFunction = (values: any[], min: any, max: any) => any[];
 
-export type Thresholds = ThresholdsName | ThresholdsFunction | RangeInterval;
+export type Thresholds = ThresholdsName | ThresholdsFunction | RangeInterval | Iterable<any>;
 
 export interface BinOptions {
+  /**
+   * If true or a positive number, produce a cumulative distribution; if a
+   * negative number, produce a [complementary cumulative](https://en.wikipedia.org/wiki/Cumulative_distribution_function#Complementary_cumulative_distribution_function_.28tail_distribution.29)
+   * distribution; if false or zero (the default), produce a probability
```
Fil, line 16:
```suggestion
   * distribution; if false or zero (the default), produce a frequency
```

mbostock (approved):

PR #1361, opened 2023-03-20
document dodge

Fil:
for #1343 

Note: I've separated the options for dodgeX and dodgeY.

### DIFF
--- a/src/transforms/dodge.d.ts
+++ b/src/transforms/dodge.d.ts
@@ -1,20 +1,56 @@
 import type {ChannelValueSpec} from "../channel.js";
 import type {Initialized} from "./basic.js";
 
-export type DodgeAnchor = "top" | "right" | "bottom" | "left" | "middle";
+export type DodgeAnchorX = "right" | "left" | "middle";
 
-export interface DodgeOptions {
-  x?: ChannelValueSpec;
+export type DodgeAnchorY = "top" | "bottom" | "middle";
+
+export interface DodgeOptionsX {
+  /** The vertical position of the mark. */
   y?: ChannelValueSpec;
+  /** The radius of the mark. */
   r?: ChannelValueSpec;
-  anchor?: DodgeAnchor;
+  /** The side of the frame used as an anchor. */
+  anchor?: DodgeAnchorX;
+  /** A number of pixels added to the radius of the mark. */
   padding?: number;
 }
 
-export function dodgeX<T>(options?: T & DodgeOptions): Initialized<T>;
+export interface DodgeOptionsY {
+  /** The horizontal position of the mark. */
+  x?: ChannelValueSpec;
+  /** The vertical position of the mark. */
+  y?: ChannelValueSpec;
+  /** The radius of the mark. */
+  r?: ChannelValueSpec;
+  /** The side of the frame used as an anchor. */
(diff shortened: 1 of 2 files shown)


src/transforms/dodge.d.ts
```diff
@@ -1,20 +1,56 @@
...
 import type {Initialized} from "./basic.js";
 
-export type DodgeAnchor = "top" | "right" | "bottom" | "left" | "middle";
+export type DodgeAnchorX = "right" | "left" | "middle";
 
-export interface DodgeOptions {
-  x?: ChannelValueSpec;
+export type DodgeAnchorY = "top" | "bottom" | "middle";
+
+export interface DodgeOptionsX {
```
mbostock, line 8: This should be DodgeXOptions. (I will fix.)

src/transforms/dodge.d.ts
```diff
@@ -1,20 +1,56 @@
...
 }
 
-export function dodgeX<T>(options?: T & DodgeOptions): Initialized<T>;
-
-export function dodgeX<T>(dodgeOptions?: DodgeOptions | DodgeAnchor, options?: T): Initialized<T>;
+export interface DodgeOptionsY {
+  /** The horizontal position of the mark. */
+  x?: ChannelValueSpec;
+  /** The vertical position of the mark. */
+  y?: ChannelValueSpec;
```
mbostock, line 23: This is asymmetric with DodgeXOptions, which doesn’t allow _x_. I assume this is a mistaken copy-paste, so I will remove.

src/transforms/dodge.d.ts
```diff
@@ -1,20 +1,56 @@
...
+  /** The radius of the mark. */
+  r?: ChannelValueSpec;
+  /** The side of the frame used as an anchor. */
+  anchor?: DodgeAnchorY;
+  /** A number of pixels added to the radius of the mark. */
+  padding?: number;
```
mbostock, line 29: We can avoid repeating this by creating a shared DodgeOptions that both DodgeXOptions and DodgeYOptions extend.

src/transforms/dodge.d.ts
```diff
@@ -1,20 +1,56 @@
 import type {ChannelValueSpec} from "../channel.js";
 import type {Initialized} from "./basic.js";
 
-export type DodgeAnchor = "top" | "right" | "bottom" | "left" | "middle";
+export type DodgeAnchorX = "right" | "left" | "middle";
```
mbostock, line 4: This should be DodgeXAnchor etc.

src/transforms/dodge.d.ts
```diff
@@ -1,20 +1,56 @@
...
 
-export interface DodgeOptions {
-  x?: ChannelValueSpec;
+export type DodgeAnchorY = "top" | "bottom" | "middle";
+
+export interface DodgeOptionsX {
+  /** The vertical position of the mark. */
   y?: ChannelValueSpec;
+  /** The radius of the mark. */
```
mbostock, line 11: This is tricky because when a channel, as opposed to a constant radius number, the channel values typically represent areas instead of radii due to the default *r* *sqrt* scale.

Fil: the size?

mbostock (approved):

PR #1367, opened 2023-03-21
fix reducer scale override confusion

### LINKED ISSUE #1363: Reducer implementations are confused with reducer scale overrides
This should display a chart in greyscale, by opting in the scale.

Note that it works when reduce is a string (reduce: "mode").

(Seen in #1360)

```
export async function reducerGroupScaleFunction() {
  const penguins = await d3.csv<any>("data/penguins.csv", d3.autoType);
  return Plot.plot({
    color: {scheme: "greys"},
    marks: [
      Plot.barY(
        penguins,
        Plot.groupX(
          {
            y: "count",
            fill: {
              reduce: (index: number[], values: any[]) => d3.mode(index, (i) => values[i]),
              scale: true // 🌶
            }
          },
          {x: "species", fill: (d) => (d.island === "Biscoe" ? "orange" : "green"), fy: "sex"}
        )
      )
    ]
  });
}

```

mbostock:
Fixes #1363.

~~Strictly speaking this is not backwards compatible because now a channel that is expressed as {value} will no longer be interpreted as a channel value spec (with an undefined scale); but, that means we can use a consistent test for both channel value specs and channel reducer specs. I’m a little worried that this is now not consistent with Plot.auto, though. 🤔~~

~~The alternative would be to check the length of the reduce function, which is what I did first…~~

Update: This is now backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2126,9 +2126,9 @@ The following aggregation methods are supported:
 * *y1* - the lower bound of the bin’s *y* extent (when binning on *y*)
 * *y2* - the upper bound of the bin’s *y* extent (when binning on *y*)
 * a function to be passed the array of values for each bin and the extent of the bin
-* an object with a *reduce* method, and optionally a *scope*
+* an object with a *reduceIndex* method, and optionally a *scope*
 
-In the last case, the *reduce* method is repeatedly passed three arguments: the index for each bin (an array of integers), the input channel’s array of values, and the extent of the bin (an object {x1, x2, y1, y2}); it must then return the corresponding aggregate value for the bin. If the reducer object’s *scope* is “data”, then the *reduce* method is first invoked for the full data; the return value of the *reduce* method is then made available as a third argument (making the extent the fourth argument). Similarly if the *scope* is “facet”, then the *reduce* method is invoked for each facet, and the resulting reduce value is made available while reducing the facet’s bins. (This optional *scope* is used by the *proportion* and *proportion-facet* reducers.)
+In the last case, the *reduceIndex* method is repeatedly passed three arguments: the index for each bin (an array of integers), the input channel’s array of values, and the extent of the bin (an object {x1, x2, y1, y2}); it must then return the corresponding aggregate value for the bin. If the reducer object’s *scope* is “data”, then the *reduceIndex* method is first invoked for the full data; the return value of the *reduceIndex* method is then made available as a third argument (making the extent the fourth argument). Similarly if the *scope* is “facet”, then the *reduceIndex* method is invoked for each facet, and the resulting reduce value is made available while reducing the facet’s bins. (This optional *scope* is used by the *proportion* and *proportion-facet* reducers.)
 
 Most aggregation methods require binding the output channel to an input channel; for example, if you want the **y** output channel to be a *sum* (not merely a count), there should be a corresponding **y** input channel specifying which values to sum. If there is not, *sum* will be equivalent to *count*.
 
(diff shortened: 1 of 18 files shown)


mbostock:
It’s still unfortunate that TypeScript can’t infer that `values` here is `any[]` instead of `any`, i.e. that the **reduce** option is a ReducerFunction and not that **fill** is a ReducerImplementation.

```ts
Plot.barY(
  penguins,
  Plot.groupX(
    {y: "count", fill: {reduce: (values) => d3.mode(values), scale: true}},
    {x: "species", fill: (d) => (d.island === "Biscoe" ? "orange" : "green"), fy: "sex"}
  )
).plot()
```

The problem is that TypeScript interfaces are loose; you’re allowed to define additional properties, _e.g._

```js
const r = {reduce: (index, values) => d3.mode(index, (i) => values[i]), foo: true};
const s: Plot.ReducerImplementation = r;
```

You only get the strict validation when the type is strongly implied:

```js
const r: Plot.ReducerImplementation = {reduce: (index, values) => d3.mode(index, (i) => values[i]), foo: true};
```

Fundamentally, the issue here is ambiguity. We use a lot of duck typing in Plot, which is great for easy extensibility but leads to ambiguities. As we saw before, it’s easy to confuse an array (which has _array_.reduce) with a ReduceImplementation; there’s a similar potential confusion with an array and a MapImplementation (_array_.map). And #1310 introduced a confusion between a ReduceImplementation and a reducer scale override.

Here are some ways we could reduce ambiguity:

1. Use a more unique method name (e.g., applyReduce instead of reduce) for implementations.
2. Use a (guaranteed unique) symbol method name (e.g., Plot.reduce) for implementations.
3. Use a more unique property name for reducer scale overrides (e.g., reducer?).
4. Require implementations to extend a class and use instanceof tests.

Regarding backwards compatibility, we could continue to support the existing duck types but only declare the new types in our TypeScript declarations. And (3) is outright compatible because we haven’t released reducer scale overrides yet; however if we don’t use the name “reduce” for reducer scale overrides, that will be inconsistent with the **reduce** option we use in other contexts (area, line, and auto mark options; imputing scale domains from channels; window transform options). Changing those would be very disruptive. I suppose we could simply break compatibility, too, given that reducer and map implementations are rare—but I’d like to avoid that to make it as easy for folks to keep upgrading.

So, disregarding (3), how do the other options look? Let’s consider the simplest identity reducer. Today that is:

```js
const reduceIdentity = {
  reduce(I, X) {
    return take(X, I);
  }
};
```

Under (1):

```js
const reduceIdentity = {
  applyReduce(I, X) {
    return take(X, I);
  }
};
```

Under (2):

```js
const reduceIdentity = {
  [Plot.reduce](I, X) {
    return take(X, I);
  }
};
```

Under (4):

```js
const reduceIdentity = new class extends Plot.Reducer {
  reduce(I, X) {
    return take(X, I);
  }
};
```

(2) is attractive, especially with JavaScript precedent from Symbol.iterator, Symbol.asyncIterator, Symbol.toPrimitive, _etc._ But it means that you can’t mix-and-match implementations across different versions of Plot (because if you load multiple versions of Plot, each will declare its own unique symbols). And there will always be non-symbolic duck interfaces in Plot that we inherit from D3, such as intervals and projections; I don’t think we can or should expect that all duck typing will use strict symbols.

(4) has the same problems with multiple versions of Plot. And the syntax is clunky… it feels strictly worse than (2).

So I guess that leaves (1)? And (1) I think is fairly easy to implement. The additional name “apply” is a little clunky, but it’s not too bad.

Fil:
Yes, (1) seems like the way to go. In terms of words one could use: summary, **summarize**, aggregate, agg; with a more biological metaphor: condense, **digest**, extract, distill… with a picture/camera metaphor: snap, depict, **describe**. applyReduce (and variants such as reducer, indexReduce…) has the merit of not introducing a new concept.

PR #1373, opened 2023-03-23
Sketch: compile auto mark to explicit marks

tophtucker:
To satisfy my curiosity, just looking into what it would take.

The easy problem: it should have the accessor in the options, not the materialized array.

The hard problem: I wasn't thinking about this before, but obviously Plot doesn't know the name of the data that was passed in (or rather, it doesn't know the literal expression), so this couldn't really work. And the compiler can't actually execute the code. So… I don't know where this would live. It's both dynamic and needs "reflection". It'd have to be something wild like: editor asks the worker, worker calls this, this returns a function that takes the literal data expression, worker passes that back to editor, editor calls compiler with the returned function, compiler passes in the data source expression, that returns a string which we insert as a new cell…


tophtucker:
Superseded by https://github.com/observablehq/plot/pull/1504

PR #1377, opened 2023-03-23
document vector

Fil:
For #1343 

First mark to be documented with its default tuples, and X, Y variants. Once we agree on the wording, it will be used as a template for all similar marks.

I tried to make a simple sentence describing the mark with its (main) options in bold—not sure if this will work for all marks.

### DIFF
--- a/src/marks/vector.d.ts
+++ b/src/marks/vector.d.ts
@@ -3,6 +3,9 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
 
 export type VectorShapeName = "arrow" | "spike";
 
+/**
+ * A custom vector shape implementation.
+ */
 export interface VectorShapeImplementation {
   draw(context: CanvasPath, length: number, radius: number): void;
 }
@@ -10,22 +13,68 @@ export interface VectorShapeImplementation {
 export type VectorShapeSpec = VectorShapeName | VectorShapeImplementation;
 
 export interface VectorOptions extends MarkOptions {
+  /** A channel for the horizontal position. */
   x?: ChannelValueSpec;
+  /** A channel for the vertical position. */
   y?: ChannelValueSpec;
+  /**
+   * The radius channel (in pixels) describes the mark’s secondary dimension
+   * (typically the *arrow* head’s span, or the *spike*’s base).
+   */
   r?: ChannelValueSpec;
+  /**
+   * The length of the shape, typically associated with a linear *length* scale.
+   */
   length?: ChannelValueSpec;
+  /** Vector rotation, in degrees; with rotate = 0, the shape points up. */
   rotate?: ChannelValue;
+  /** Vector shape, as a name (*arrow* or *spike*), or as a custom implementation. */
   shape?: VectorShapeSpec;
+  /** Vector anchor; by default a vector is anchored at its base (*start*). */
   anchor?: "start" | "middle" | "end";
+  /**
+   * The frame anchor, for vectors without a position, or positioned on a single
+   * (*x* or *y*) dimension in a two-dimensional frame.
+   */
   frameAnchor?: FrameAnchor;
 }
 
+/**
+ * A vector of the given **length**, drawn in the **shape** of an arrow from its
+ * **anchor** at the given [**x**, **y**] position, and an orientation indicated
+ * by **rotate**. To create a vector field representing a wind map:
+ *
+ * ```js
+ * Plot.vector(wind, {x: "longitude", y: "latitude", length: "speed", rotate: "direction"})
+ * ```
+ *
+ * By default, accepts an array of coordinates pairs (tuples) as *data*.
+ */
 export function vector(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * The vector mark, but the **x** channel defaults to identity, positioning one
+ * arrow for each horizontal position passed as *data*.
+ */
 export function vectorX(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * The vector mark, but the **y** channel defaults to identity, positioning one
+ * arrow for each vertical position passed as *data*.
+ */
 export function vectorY(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * A variant of the vector mark where the **shape** defaults to *spike*, the
+ * **stroke** defaults to *currentColor*, the **strokeWidth** defaults to 1, the
+ * **fill** defaults to **stroke**, the **fillOpacity** defaults to 0.3, and the
+ * **anchor** defaults to *start*. Typically used to create a spike map:
+ *
+ * ```js
+ * Plot.spike(counties, {stroke: "red", length: "population"})
+ * ```
+ */
 export function spike(data?: Data, options?: VectorOptions): Vector;
 
+/** The vector mark */
 export class Vector extends RenderableMark {}
--- a/src/marks/vector.d.ts
+++ b/src/marks/vector.d.ts
@@ -10,7 +10,7 @@ export interface VectorShapeImplementation {
   draw(context: CanvasPath, length: number, radius: number): void;
 }
 
-export type VectorShapeSpec = VectorShapeName | VectorShapeImplementation;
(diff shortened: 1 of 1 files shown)


src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
 export function vectorY(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * A variant of the vector mark where the **shape** defaults to *spike*, the
+ * **stroke** defaults to *currentColor*, the **strokeWidth** defaults to 1, the
+ * **fill** defaults to **stroke**, the **fillOpacity** defaults to 0.3, and the
+ * **anchor** defaults to *start*. Typically used to create a spike map:
+ *
+ * ```js
+ * Plot.spike(counties, {stroke: "red", length: "population"})
```
mbostock, line 74: Since counties are typically represented as polygons, we’d more likely need Plot.geoCentroid here. And if we assume that `counties` is polygons, the `length: "population"` would need to be rewritten as `length: d => d.properties.population`. I think it would be simpler to give an example with a point geometry instead, namely cities instead of counties. And then maybe we’d define **x** and **y** to be explicit:

```suggestion
 * Plot.spike(cities, {x: "longitude", y: "latitude", length: "population"})
```

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
 export function vector(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * The vector mark, but the **x** channel defaults to identity, positioning one
+ * arrow for each horizontal position passed as *data*.
+ */
 export function vectorX(data?: Data, options?: VectorOptions): Vector;
 
+/**
+ * The vector mark, but the **y** channel defaults to identity, positioning one
```
mbostock, line 62: The vectorY constructor returns a Vector mark, so the “but” is awkward here. I think the “but” applies to the difference with the vector constructor (not the Vector mark class). So…

```suggestion
 * Like vector, but the **y** channel defaults to identity, positioning one
```

It might also be useful to mention that this convenience constructor is only useful when `data` is an array of numbers (vertical positions). Otherwise you’d just use vector.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
 export type VectorShapeName = "arrow" | "spike";
 
+/**
+ * A custom vector shape implementation.
+ */
 export interface VectorShapeImplementation {
   draw(context: CanvasPath, length: number, radius: number): void;
 }
 
 export type VectorShapeSpec = VectorShapeName | VectorShapeImplementation;
```
mbostock, line 13: This should be `VectorShape` instead of `VectorShapeSpec`. I think that’s an oversight on my part.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
   r?: ChannelValueSpec;
+  /**
+   * The length of the shape, typically associated with a linear *length* scale.
+   */
   length?: ChannelValueSpec;
+  /** Vector rotation, in degrees; with rotate = 0, the shape points up. */
   rotate?: ChannelValue;
+  /** Vector shape, as a name (*arrow* or *spike*), or as a custom implementation. */
   shape?: VectorShapeSpec;
+  /** Vector anchor; by default a vector is anchored at its base (*start*). */
```
mbostock, line 33: I’d prefer a bulleted list for enums. Maybe, assuming a default orientation:

* *start* - from [x, y] to [x, y + length]
* *middle* - from [x, y - length / 2] to [x, y + length / 2]
* *end* - from [x, y - length] to [x, y]

Not sure if that’s right by something like that.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
   x?: ChannelValueSpec;
+  /** A channel for the vertical position. */
   y?: ChannelValueSpec;
+  /**
+   * The radius channel (in pixels) describes the mark’s secondary dimension
+   * (typically the *arrow* head’s span, or the *spike*’s base).
+   */
   r?: ChannelValueSpec;
+  /**
+   * The length of the shape, typically associated with a linear *length* scale.
```
mbostock, line 26: Or a constant length in pixels.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
+ * A custom vector shape implementation.
+ */
 export interface VectorShapeImplementation {
   draw(context: CanvasPath, length: number, radius: number): void;
 }
 
 export type VectorShapeSpec = VectorShapeName | VectorShapeImplementation;
 
 export interface VectorOptions extends MarkOptions {
+  /** A channel for the horizontal position. */
```
mbostock, line 16: Perhaps obvious, but we should mention which scale (here *x*) a channel is bound to by default.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
   x?: ChannelValueSpec;
+  /** A channel for the vertical position. */
   y?: ChannelValueSpec;
+  /**
+   * The radius channel (in pixels) describes the mark’s secondary dimension
```
mbostock, line 21: The radius is not a channel for vectors; it’s a constant option.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
+   * (typically the *arrow* head’s span, or the *spike*’s base).
+   */
   r?: ChannelValueSpec;
+  /**
+   * The length of the shape, typically associated with a linear *length* scale.
+   */
   length?: ChannelValueSpec;
+  /** Vector rotation, in degrees; with rotate = 0, the shape points up. */
```
mbostock, line 29: Can be either a constant or a channel.

src/marks/vector.d.ts
```diff
@@ -3,29 +3,78 @@ import type {Data, FrameAnchor, MarkOptions, RenderableMark} from "../mark.js";
...
   rotate?: ChannelValue;
+  /** Vector shape, as a name (*arrow* or *spike*), or as a custom implementation. */
   shape?: VectorShapeSpec;
+  /** Vector anchor; by default a vector is anchored at its base (*start*). */
   anchor?: "start" | "middle" | "end";
+  /**
+   * The frame anchor, for vectors without a position, or positioned on a single
+   * (*x* or *y*) dimension in a two-dimensional frame.
```
mbostock, line 37: Tempted to enumerate the frame anchor options here, but I know we’ll have to repeat it in a few places. Might at least be worth stating that it can be one of the four corners, four edges, or the middle, as the anchor point if x and y are not specified. And also that it effectively provides a constant default for x and y, meaning that you can use it conjunction with either x or y (but that if you’ve already specified both x and y there is no reason to specify a frameAnchor).

mbostock (approved):

src/marks/vector.d.ts
```diff
@@ -1,80 +1,114 @@
...
+   * or a constant number in pixels. Defaults to 12 pixels.
    */
   length?: ChannelValueSpec;
-  /** Vector rotation, in degrees; with rotate = 0, the shape points up. */
+
+  /**
+   * The vector’s orientation (rotation angle); either a constant number in
+   * degrees clockwise, or an optional channel (with no associated scale).
+   * Defaults to 0 degrees with the vector pointing up; positive angles proceed
+   * clockwise.
```
Fil, line 45: repeats "clockwise"

PR #1382, opened 2023-03-24
expose identity reducer

mbostock:
Might as well expose this, right? TODO documentation and tests.


Fil:
Added tests and documentation. Note: it's not supported on the window transform.

Fil (approved):

PR #1385, opened 2023-03-24
document auto

tophtucker:
for https://github.com/observablehq/plot/issues/1343

### DIFF
--- a/src/marks/auto.d.ts
+++ b/src/marks/auto.d.ts
@@ -3,66 +3,189 @@ import type {CompoundMark, Data} from "../mark.js";
 import type {Reducer} from "../reducer.js";
 import type {BinOptions} from "../transforms/bin.js";
 
-// TODO: should jsdoc go here or inside interface?
 type MarkType = "dot" | "line" | "area" | "rule" | "bar";
 
 export interface AutoOptions {
   /**
-   * The horizontal spatial encoding. If **y** is set without **x**, the **x** channel
-   * will default to the *count* reducer to produce a histogram; you can
+   * The horizontal spatial encoding. If **y** is set without **x**, the **x**
+   * channel will default to the *count* reducer to produce a histogram; you can
    * override this by passing {reduce: null}.
    */
-  x?: ChannelValue | Reducer | ({value?: ChannelValue; reduce?: Reducer | null; zero?: boolean} & BinOptions);
+  x?:
+    | ChannelValue
+    | Reducer
+    | ({
+        /**
+         * A string representing a column of your data, or an accessor function.
+         */
+        value?: ChannelValue;
+
+        /**
+         * A string representing a reducer, or your own reducer function.
+         * Setting a reducer on **x** will automatically bin or group on **y**.
+         */
+        reduce?: Reducer | null;
+
+        /**
+         * Draws a vertical baseline where x is 0.
+         */
+        zero?: boolean;
+      } & BinOptions);
+
   /**
-   * The vertical spatial encoding. If *x* is set without *y*, the *y* channel
-   * will default to the **count** reducer to produce a histogram; you can
+   * The vertical spatial encoding. If **x** is set without **y**, the **y**
+   * channel will default to the *count* reducer to produce a histogram; you can
    * override this by passing {reduce: null}.
    */
-  y?: ChannelValue | Reducer | ({value?: ChannelValue; reduce?: Reducer | null; zero?: boolean} & BinOptions);
+  y?:
+    | ChannelValue
+    | Reducer
+    | ({
+        /**
+         * A string representing a column of your data, or an accessor function.
+         */
+        value?: ChannelValue;
+
+        /**
+         * A string representing a reducer, or your own reducer function.
+         * Setting a reducer on **y** will automatically bin or group on **x**.
+         */
+        reduce?: Reducer | null;
+
+        /**
+         * Draws a horizontal baseline where y is 0.
+         */
+        zero?: boolean;
+      } & BinOptions);
+
   /**
    * The color encoding, which corresponds to the stroke channel for dots,
    * lines, and rules, and the fill channel for areas and bars.
    */
-  color?: ChannelValue | Reducer | {value?: ChannelValue; reduce?: Reducer | null; color?: string};
+  color?:
+    | ChannelValue
+    | Reducer
+    | {
+        /**
+         * A string representing a column of your data, or an accessor function.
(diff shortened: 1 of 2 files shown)


src/marks/auto.d.ts
```diff
@@ -3,26 +3,74 @@ import type {CompoundMark, Data} from "../mark.js";
...
   fx: ChannelValue;
   fy: ChannelValue;
-  mark: "dot" | "line" | "area" | "rule" | "bar";
+  mark: MarkType;
 }
 
+/**
+ * Returns a fully-specified *auto* options object, with nothing left undefined.
+ * This is mostly for internal use, but it can be used to “lock down” the
+ * specification of an *auto* mark, or to inspect the decisions its heuristic is
```
Fil, line 64:
```suggestion
 * specification of an **auto** mark, or to inspect the decisions its heuristic is
```

src/marks/auto.d.ts
```diff
@@ -3,26 +3,74 @@ import type {CompoundMark, Data} from "../mark.js";
...
+   */
   fx?: ChannelValue | {value?: ChannelValue};
+  /**
+   * The vertical facet dimension.
+   */
   fy?: ChannelValue | {value?: ChannelValue};
-  mark?: "dot" | "line" | "area" | "rule" | "bar";
+  /**
+   * A type of mark; for example, the type **bar** encompasses bar, barX, and
+   * barY. Used to guide what sort of mark *auto* should use. It should be thought
```
Fil, line 42:
```suggestion
   * A type of mark; for example, the type *bar* encompasses bar, barX, and
   * barY. Used to guide what sort of mark **auto** should use. It should be thought
```

option (strings) in italics, marks in bold

src/marks/auto.d.ts
```diff
@@ -3,26 +3,74 @@ import type {CompoundMark, Data} from "../mark.js";
 import type {Reducer} from "../reducer.js";
 import type {BinOptions} from "../transforms/bin.js";
 
+// TODO: should jsdoc go here or inside interface?
+type MarkType = "dot" | "line" | "area" | "rule" | "bar";
+
 export interface AutoOptions {
+  /**
+   * The horizontal spatial encoding. If *y* is set without *x*, the *x* channel
+   * will default to the **count** reducer to produce a histogram; you can
```
Fil, line 12:
```suggestion
   * The horizontal spatial encoding. If **y** is set without **x**, the **x** channel
   * will default to the *count* reducer to produce a histogram; you can
```

channel names in boldface, string option in italics

src/marks/auto.d.ts
mbostock: I’ve been using bold to refer to named options, and italics to refer to parameters (and in some cases scale names… though maybe those should be bold). I still haven’t totally decided on what convention works best, though, so I imagine we’ll eventually take another pass to unify our typographic conventions.

src/marks/auto.d.ts
```diff
@@ -72,5 +195,11 @@ export function autoSpec(data?: Data, options?: AutoOptions): AutoSpec;
  * useful initial plot as quickly as possible through opinionated defaults, and
  * to accelerate exploratory analysis by letting you refine views with minimal
  * changes to code.
+ *
+ * For example, for a histogram of penguins binned by weight:
+ *
+ * ```js
+ * Plot.auto(penguins, {x: "weight"})
```
Fil, line 202:
```suggestion
 * Plot.auto(penguins, {x: "body_mass_g"})
```

src/marks/auto.d.ts
```diff
@@ -3,26 +3,208 @@ import type {CompoundMark, Data} from "../mark.js";
...
+  /**
+   * The vertical facet dimension.
+   */
   fy?: ChannelValue | {value?: ChannelValue};
-  mark?: "dot" | "line" | "area" | "rule" | "bar";
+
+  /**
+   * A type of mark; for example, the type *bar* encompasses bar, barX, and
+   * barY. Used to guide what sort of mark **auto** should use. It should be
+   * thought of as as an override; *auto* should usually do the right thing
```
Fil, line 126:
```suggestion
   * thought of as an override; *auto* should usually do the right thing
```

src/marks/auto.d.ts
```diff
@@ -3,26 +3,208 @@ import type {CompoundMark, Data} from "../mark.js";
...
+  color: {value: ChannelValue; reduce: Reducer | null; color: string};
+
+  /**
+   * Like auto’s **size** option, but with nothing left undefined.
+   */
   size: {value: ChannelValue; reduce: Reducer | null};
+
+  /**
+   * Like auto’s **fx** option, but with nothing left undefined.
+   */
```
mbostock: It’s better to leave these undocumented because then it will inherit the documentation from AutoOptions, which will be more informative. We can leave the “nothing left undefined” context up at the AutoSpec interface level.

src/marks/auto.d.ts
```diff
@@ -3,26 +3,208 @@ import type {CompoundMark, Data} from "../mark.js";
...
+   * channel will default to the *count* reducer to produce a histogram; you can
+   * override this by passing {reduce: null}.
+   */
+  x?:
+    | ChannelValue
+    | Reducer
+    | ({
+        /**
+         * A string representing a column of your data, or an accessor function.
+         */
```
mbostock: We don’t document these fields e.g. for ChannelValueSpec, so I’m going to fold these comments back up into the option-level documentation. And then it’ll be surfaced earlier when you look at the option, to boot.

src/marks/auto.d.ts
```diff
@@ -3,26 +3,208 @@ import type {CompoundMark, Data} from "../mark.js";
 import type {Reducer} from "../reducer.js";
 import type {BinOptions} from "../transforms/bin.js";
 
+/**
+ * One of the categories of mark; *line* includes lineX and lineY; *area*
+ * includes areaX and areaY; *rule* includes ruleX and ruleY; *bar* includes
+ * barX, barY, rectX, and rectY; and *dot* is just dot.
+ */
+type MarkType = "dot" | "line" | "area" | "rule" | "bar";
```
mbostock: I’m not 100% sure when we should break enums out like this, versus leaving them in-place within the interface. But I think in this case it feels like it should be inlined in AutoOptions. Note that we can use AutoOptions["mark"] to still refer to this type in another place (which I did for AutoSpec).

mbostock (approved):

PR #1386, opened 2023-03-25
document raster and contour

Fil:
for #1343

Note: added *interval* in the raster options, and explicitly discarded the raster interval option from the type, to add the contour interval option instead—not sure it's needed, but it felt like the right thing to do.

### DIFF
--- a/src/marks/contour.d.ts
+++ b/src/marks/contour.d.ts
@@ -12,12 +12,14 @@ export interface ContourOptions extends Omit<RasterOptions, "interval" | "imageR
    * **blur** option.
    */
   smooth?: boolean;
+
   /**
    * The value can be specified as a channel based on the sample *data*, or as a
    * function *f*(*x*, *y*) to be evaluated at each pixel if the *data* is not
    * provided.
    */
   value?: ChannelValue | RasterSampler;
+
   /**
    * How to subdivide the domain into discrete level sets; defaults to *auto*;
    * one of:
@@ -37,6 +39,7 @@ export interface ContourOptions extends Omit<RasterOptions, "interval" | "imageR
    * See also the **interval** option.
    */
   thresholds?: Thresholds;
+
   /**
    * An alternative way of specifying the contour thresholds. For example to
    * create an isoline every 15 meters on a topographic map:
@@ -90,20 +93,11 @@ export interface ContourOptions extends Omit<RasterOptions, "interval" | "imageR
  *  })
  * ```
  *
- * The resolution of the underlying raster grid may be specified with the
- * following options:
+ * The resolution of the rectangular raster image may be specified with
+ * **width** and **height**.
  *
- * * **width** - the number of pixels on each horizontal line
- * * **height** - the number of lines; a positive integer
- *
- * The raster dimensions may also be imputed from the extent of *x* and *y* and
- * a pixel size:
- *
- * * **x1** - the starting horizontal position; bound to the *x* scale
- * * **x2** - the ending horizontal position; bound to the *x* scale
- * * **y1** - the starting vertical position; bound to the *y* scale
- * * **y2** - the ending vertical position; bound to the *y* scale
- * * **pixelSize** - the screen size of a raster pixel; defaults to 1
+ * The raster dimensions may alternatively be imputed from a rectangular extent
+ * *x1*, *y1*, *x2*, *y2* and a **pixelSize**.
  */
 export function contour(options?: ContourOptions): Contour;
 export function contour(data?: Data, options?: ContourOptions): Contour;
--- a/src/marks/contour.d.ts
+++ b/src/marks/contour.d.ts
@@ -8,8 +8,8 @@ import type {RasterOptions, RasterSampler} from "./raster.js";
 export interface ContourOptions extends Omit<RasterOptions, "interval" | "imageRendering"> {
   /**
    * Whether to apply linear interpolation after marching squares when computing
-   * contour polygons (defaults to true). For smoother contours, see the
-   * **blur** option.
+   * contour polygons; defaults to true. For smoother contours, see the **blur**
+   * option.
    */
   smooth?: boolean;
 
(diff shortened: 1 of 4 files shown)


src/marks/contour.d.ts
```diff
@@ -4,15 +4,109 @@ import type {Data, RenderableMark} from "../mark.js";
...
+ * * **height** - the number of lines; a positive integer
+ *
+ * The raster dimensions may also be imputed from the extent of *x* and *y* and
+ * a pixel size:
+ *
+ * * **x1** - the starting horizontal position; bound to the *x* scale
+ * * **x2** - the ending horizontal position; bound to the *x* scale
+ * * **y1** - the starting vertical position; bound to the *y* scale
+ * * **y2** - the ending vertical position; bound to the *y* scale
+ * * **pixelSize** - the screen size of a raster pixel; defaults to 1
```
mbostock, line 106: I don’t think we should repeat the options here; that should be covered by the ContourOptions/RasterOptions interface. (This description is quite long and we should look for ways to pare it down.)

src/marks/contour.d.ts
```diff
@@ -4,15 +4,109 @@ import type {Data, RenderableMark} from "../mark.js";
...
+ */
 export function contour(options?: ContourOptions): Contour;
-
 export function contour(data?: Data, options?: ContourOptions): Contour;
```
mbostock, line 109: In other places I swapped the order of overloaded methods to put the more explicit version first. I think we should probably do that here (and anywhere else), too: it favors the more explicit form when you type `contour(|)` so that you get more explicit guidance about what arguments are expected.

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
+ * * **width** - the number of pixels on each horizontal line
+ * * **height** - the number of lines; a positive integer
+ *
+ * The raster dimensions may also be imputed from the extent of *x* and *y* and a pixel size:
+ *
+ * * **x1** - the starting horizontal position; bound to the *x* scale
+ * * **x2** - the ending horizontal position; bound to the *x* scale
+ * * **y1** - the starting vertical position; bound to the *y* scale
+ * * **y2** - the ending vertical position; bound to the *y* scale
+ * * **pixelSize** - the screen size of a raster pixel; defaults to 1
```
mbostock, line 173: Same comment re. enumeration of options.

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
+ * maximum allowable number of steps (*maxSteps*) have been taken, and then
+ * assigning the current pixel the closest sample’s value. The random walk uses
+ * the “walk on spheres” algorithm in two dimensions described by [Sawhney and
+ * Crane](https://www.cs.cmu.edu/~kmcrane/Projects/MonteCarloGeometryProcessing/index.html),
+ * SIGGRAPH 2020.
+ */
 export function interpolatorRandomWalk(options?: {
   random?: RandomSource;
   minDistance?: number;
   maxSteps?: number;
```
mbostock, line 220: Needs comments.

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
+   * - *barycentric* - apply barycentric interpolation over the Delaunay
+   *   triangulation
+   * - *random-walk* - apply a random walk from each pixel, stopping when near a
+   *   sample
+   * - a function - custom interpolation
+   */
   interpolate?: RasterInterpolate | "none" | null;
+  /**
+   * The [image-rendering
+   * attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/image-rendering);
```
mbostock, line 127: I’ve started using footnote-style links in Markdown to reduce the wrapping problems with long URLs.

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
+   */
+  interval?: Interval;
+  /**
+   * A non-negative pixel radius for smoothing; defaults to 0. Note that
+   * blurring is applied on the values (before the color scale is applied) if
+   * quantitative, and after (on the materialized pixels), if ordinal.
+   */
   blur?: number;
+  /**
+   * The spatial interpolation method, when using *data* samples. One of:
```
mbostock, line 113: Needs to mention the default.

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
+   * The screen size of a raster pixel, used to determine the height and width
+   * of the raster from the frame’s dimensions; defaults to 1
+   */
   pixelSize?: number;
+  /**
+   * The sampling interval, used to determine the height and width of the raster
+   * from the frame’s dimensions and the *x* and *y* extents.
+   */
+  interval?: Interval;
+  /**
```
mbostock, line 106: Use a blank line between options (unless using the dense form with one-line comments).

src/marks/raster.d.ts
```diff
@@ -12,43 +29,196 @@ export type RasterInterpolateFunction = (
...
 
+/** Options for the raster mark. */
 export interface RasterOptions extends Omit<MarkOptions, "fill" | "fillOpacity"> {
+  /** The horizontal coordinate of a sample. */
   x?: ChannelValueSpec;
+  /** The vertical coordinate of a sample. */
   y?: ChannelValueSpec;
+  /**
+   * The lower value of *x*, on the left edge of the raster (right edge if
+   * reversed).
```
mbostock, line 68: This and other options need the defaults described (which is a bit tedious to describe because there are a lot of different defaults at play here, but that’s a big part of the value that documentation provides).

Fil: I hope I got those right.

src/marks/raster.d.ts
```diff
@@ -1,54 +1,258 @@
...
+   * The screen size of a raster pixel, used to determine the height and width
+   * of the raster from the frame’s dimensions; defaults to 1
+   */
   pixelSize?: number;
+
+  /**
+   * The sampling interval, used to determine the height and width of the raster
+   * from the frame’s dimensions and the *x* and *y* extents.
+   */
+  interval?: Interval;
```
mbostock, line 126: The raster mark doesn’t support the **interval** option; that’s only for the contour mark.

mbostock (approved):

PR #1387, opened 2023-03-25
do not repeat degrees clockwise

Fil:


mbostock (changes requested):
I think it’s fine as-is, but if I were going to drop something, I would drop the second reference (“; positive angles proceed clockwise”) and leave the first since just saying it’s “a number” in the first sentence tells you almost nothing about what the number means.

mbostock (approved):

PR #1389, opened 2023-03-25
document hexgrid

Fil:
for #1343 

also bags a stray "layout" in README

### DIFF
--- a/src/marks/hexgrid.d.ts
+++ b/src/marks/hexgrid.d.ts
@@ -1,9 +1,18 @@
 import type {MarkOptions, RenderableMark} from "../mark.js";
 
+/** Options for the hexgrid mark. */
 export interface HexgridOptions extends MarkOptions {
+  /**
+   * The distance between centers of neighboring hexagons, in pixels; defaults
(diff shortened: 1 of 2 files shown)


src/marks/hexgrid.d.ts
```diff
@@ -1,9 +1,18 @@
...
 export interface HexgridOptions extends MarkOptions {
+  /**
+   * The distance between centers of neighboring hexagons, in pixels; defaults
+   * to 20. Should reflect the binWidth of the **hexbin** transform.
+   */
   binWidth?: number;
 }
 
+/**
+ * The hexgrid mark can be used to support marks using the **hexbin** transform.
```
mbostock, line 13: It would be nice to document the default appearance of the grid here (stroke currentColor, strokeOpacity 0.1, etc.), and that fill is unsupported. And to explain that the hexagons span the entire frame… and that it’s useful to show empty hexagons (in conjunction with the hexbin transform, which doesn’t support the filter output option and thus currently can’t render empty hexagons).

mbostock (approved):

PR #1390, opened 2023-03-25
document linearRegression

Fil:
for #1343 

_todo_
- [x] I couldn't make sense of the **reduce** option, PTAL?

### DIFF
--- a/src/marks/linearRegression.d.ts
+++ b/src/marks/linearRegression.d.ts
@@ -25,7 +25,11 @@ interface LinearRegressionOptions extends MarkOptions, BinOptions {
    */
   precision?: number;
   /**
-   * 🌶 How is this used?
+   * Reducer for when the independent variable is binned with an **interval**. For example, to create a trend line on the maximum value of the AAPL ticker by month:
+   *
+   * ```js
+   * Plot.linearRegressionY(aapl, {x: "Date", y: "Close", interval: "month", reduce: "max"})
+   * ```
    */
   reduce?: BinReducer;
 }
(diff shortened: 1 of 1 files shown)


src/marks/linearRegression.d.ts
```diff
@@ -2,23 +2,85 @@ import type {ChannelValue, ChannelValueDenseBinSpec, ChannelValueSpec} from "../
...
+   */
   ci?: number;
+  /**
+   * The distance (in pixels) between samples of the confidence band; defaults to 4.
+   */
   precision?: number;
+  /**
+   * 🌶 How is this used?
+   */
   reduce?: BinReducer;
```
mbostock, line 30: It’s from maybeDenseInterval. It’s the same idea as line and area. For example for areaY you can specify an interval along _x_, and then reduce defaults to _first_ to extract the first value in _y_ for the current interval in _x_.

src/marks/linearRegression.d.ts
```diff
@@ -2,23 +2,85 @@ import type {ChannelValue, ChannelValueDenseBinSpec, ChannelValueSpec} from "../
...
 
+/** Options for the linearRegressionX and linearRegressionY marks. */
 interface LinearRegressionOptions extends MarkOptions, BinOptions {
+  /**
+   * A channel for the dependent variable **x**.
+   */
   x?: ChannelValueSpec;
+  /**
+   * A channel for the dependent variable **y**.
+   */
```
Fil, line 13: these represent the dependent variable, and should probably go into their respective mark (x in linearRegressionX, and y in linearRegressionY)

src/marks/linearRegression.d.ts
Fil: I've documented what works. Can you open an issue with what doesn't? (It's not clear to me what's missing here.)

src/marks/linearRegression.d.ts
Fil: done

mbostock (approved):

PR #1394, opened 2023-03-26
document box

Fil:
for #1343

Say what happens with *stroke*/*z*? In my view it's technically working but hard to see if you don't add fillOpacity, so maybe it needs a bit more work.


mbostock (commented):
I think this needs a bit more explanation about the underlying group and map transforms.

mbostock (approved):

PR #1402, opened 2023-03-28
document area

Fil:
for #1343 

very similar to Plot.line (#1401), but I realized the README glosses over the variable aesthetics for this mark (I've added them here for the jsdocs).

### DIFF
--- a/src/marks/area.d.ts
+++ b/src/marks/area.d.ts
@@ -7,40 +7,39 @@ import type {StackOptions} from "../transforms/stack.js";
 /** Options for the area, areaX, and areaY marks. */
 export interface AreaOptions extends MarkOptions, StackOptions, CurveOptions {
   /**
-   * The vertical position channel for the left side, typically bound to the *x*
-   * scale. When used with the **areaY** mark, it is typically equal to **x2**.
-   * When used with **areaX** and inferred from the **x** channel, it is
-   * typically the position of lower magnitude of the values stacked from zero.
+   * The required primary horizontal position channel, typically bound to the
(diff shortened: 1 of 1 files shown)


src/marks/area.d.ts
```diff
@@ -4,30 +4,208 @@ import type {Data, MarkOptions, RenderableMark} from "../mark.js";
 import type {BinOptions, BinReducer} from "../transforms/bin.js";
 import type {StackOptions} from "../transforms/stack.js";
 
+/** Options for the area, areaX, and areaY marks. */
 export interface AreaOptions extends MarkOptions, StackOptions, CurveOptions {
+  /**
+   * The vertical position channel for the left side, typically bound to the *x*
```
mbostock, line 10: Vertical and horizontal are flipped here, yes? I’ll fix.

Fil: Ha, yes. (I'm always looking at my hands when I need to figure out left from right.)

src/marks/area.d.ts
```diff
@@ -4,30 +4,208 @@ import type {Data, MarkOptions, RenderableMark} from "../mark.js";
...
+   * scale. When used with the **areaY** mark, it is typically equal to **x2**.
```
mbostock, line 11: The areaY mark doesn’t allow you to specify **x1**/**x2**; you can only specify **x**. This is confounding the constructor options with what happens internally (where `X2` defaults to `X1` during render); but implementation details shouldn’t be part of the public documentation.

mbostock (approved):

PR #1407, opened 2023-03-30
support for fillOpacity in swatch and ramp color legends

### LINKED ISSUE #1356: color legend doesn't account for fillOpacity
Color legends show the color with opacity: 1. Thus the legend might look quite different than the plot, in case you set the fillOpacity < 1

Fil:
Note: I used **fillOpacity** and not **opacity**, for consistency with the *symbol* legends.

closes #1356

### DIFF
--- a/README.md
+++ b/README.md
@@ -653,6 +653,7 @@ Categorical and ordinal color legends are rendered as swatches, unless *options*
 * *options*.**columns** - the number of swatches per row
 * *options*.**marginLeft** - the legend’s left margin
 * *options*.**className** - a class name, that defaults to a randomly generated string scoping the styles
+* *options*.**fillOpacity** - the swatch fill opacity; defaults to 1
 * *options*.**width** - the legend’s width (in pixels)
 
 Symbol legends are rendered as swatches and support the options above in addition to the following options:
@@ -679,6 +680,7 @@ Continuous color legends are rendered as a ramp, and can be configured with the 
 * *options*.**marginRight** - the legend’s right margin
 * *options*.**marginBottom** - the legend’s bottom margin
 * *options*.**marginLeft** - the legend’s left margin
+* *options*.**fillOpacity** - the ramp’s fill opacity; defaults to 1
 
 The **style** legend option allows custom styles to override Plot’s defaults; it has the same behavior as in Plot’s top-level [layout options](#layout-options).
 
--- a/README.md
+++ b/README.md
@@ -653,7 +653,7 @@ Categorical and ordinal color legends are rendered as swatches, unless *options*
 * *options*.**columns** - the number of swatches per row
 * *options*.**marginLeft** - the legend’s left margin
 * *options*.**className** - a class name, that defaults to a randomly generated string scoping the styles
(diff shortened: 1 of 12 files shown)


mbostock (changes requested):
The reason I suggested calling it **opacity** (and not **fillOpacity**) is that we use the **color** option for the _opacity_ _ramp_ legend. Also more practically, there is no possible stroke here, so we do not need to distinguish between fill and stroke, and opacity is shorter.

Fil:
My code was really missing a few pieces :-(

Now, it's a bummer that we have to select **opacity** if there is no symbol, **fillOpacity** if the symbols are filled, and **strokeOpacity** if they are stroked. Maybe we could support opacity across the board (as well as strokeOpacity and fillOpacity)?

src/legends/swatches.js
```diff
@@ -12,14 +12,20 @@ function maybeScale(scale, key) {
...
     options,
     (selection, scale) =>
-      selection.append("svg").attr("fill", scale.scale).append("rect").attr("width", "100%").attr("height", "100%"),
+      selection
+        .append("svg")
+        .attr("fill", scale.scale)
+        .append("rect")
+        .attr("width", "100%")
+        .attr("height", "100%")
+        .attr("opacity", maybeNumberChannel(opacity)[1]),
```
mbostock, line 28: We can implement the option as fill-opacity here (for faster rendering, not that it is likely to matter in this case).

```suggestion
        .attr("fill-opacity", maybeNumberChannel(opacity)[1]),
```

README.md
```diff
@@ -679,6 +680,7 @@ Continuous color legends are rendered as a ramp, and can be configured with the
 * *options*.**marginRight** - the legend’s right margin
 * *options*.**marginBottom** - the legend’s bottom margin
 * *options*.**marginLeft** - the legend’s left margin
+* *options*.**opacity** - the ramp’s opacity
```
mbostock, line 683: Did we forget to document the color option for opacity ramp legends?

src/legends/ramp.js
```diff
@@ -117,6 +121,7 @@ export function legendRamp(color, options) {
 
     svg
       .append("g")
+      .attr("opacity", opacity)
```
mbostock, line 124: Implement as fill-opacity for performance?

```suggestion
      .attr("fill-opacity", opacity)
```

src/legends/ramp.js
```diff
@@ -137,6 +142,7 @@ export function legendRamp(color, options) {
...
```
mbostock, line 145: Likewise…

```suggestion
      .attr("fill-opacity", opacity)
```

src/legends/swatches.js
Fil: yup, why not; but then maybe also document that (by adding "fill" in README and in the jsdocs?)

mbostock: I’m not sure we need to document it because it’s an implementation detail how the swatches are rendered… but sure.

mbostock (approved):

PR #1409, opened 2023-03-30
upgrade actions

mbostock:
We’re using some old deprecated actions which use Node 12. Found [here](https://github.com/observablehq/plot/actions/runs/4567228301).

Screenshot 2023-03-30 at 12 39 20 PM

Ref. https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,9 +17,8 @@ jobs:
         node-version: [16.x]
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Fil (approved):

Fil:
Out of curiosity, does "upload artifact" do anything?

mbostock:
Yes, when a snapshot test fails it keeps the output so we can see why it failed.

Issue #1422, opened 2023-04-04
Auto mark: add more override-only mark types?


tophtucker:
So far, the only mark types that Plot.auto supports are those that it _might_ pick automatically — except area, which is a lot like a line with _zero: true_, but you can only get it by explicitly setting the mark type. We could add more marks like that. E.g. box plots and voronois would be easy. 

In these cases, you're not getting much automatic inference out of Plot.auto, except the different reduce syntax; you should almost certainly be using the explicit mark. But if we want mark types to be available in the chart cell, we'd have to add them to Plot.auto. But most of these override-only marks would take us farther outside the fuzzy intersection we've aimed for, where most channels (except _size_) apply to all the marks you could choose, and most configs lead to _some_ chart.

The other issue is that many of these are most valuable in combination with another mark, in which case maybe we should think about them more like we think about the zero rule: as a "decoration" atop the main chosen mark.

- Box plot: could only support fill or stroke; wouldn’t work on quant-quant scales.
- Voronoi: might be better as a decoration atop dot; unclear value if not for interaction
- Vector: notionally the reason “size” isn’t called “radius”, but not very useful without a rotate channel passed through, which wouldn’t apply to any other mark
- Hexbin: it’s a transformation, not a mark; benefits from having a background of Plot.hexgrid.
- Density / contour: sorta rare; _maybe_ a useful mark to infer if number of points is gigantic?
- Linear regression: would have to have x and y version, because we can't infer independent/dependent variable; might be better as a decoration atop dot; could be specified more like `y: {value: "height", regress: true}`??

Fil:
The **spike** mark (technically a **vector** with a _spike_ shape) is useful without rotate.

Issue #1330, opened 2023-03-14
The box mark could use bin/rect instead of group/bar for quantitative data


mbostock:
Ref. https://talk.observablehq.com/t/how-to-label-a-boxplot-with-dates-format-axis/7739/3

Fil:
I feel this could be addressed at a higher abstraction level (#1336)?

mbostock:
I think we could use exactly the same technique we use in Plot.auto, then have the box mark chose dynamically between bin+rect and group+bar.

Fil:
This is the test example in http://localhost:8008/?test=aaplMonthly (minus the outliers)

mbostock:
Here’s an example workaround:

![untitled (55)](https://user-images.githubusercontent.com/230541/232239057-87d9c41e-2b71-4fde-b3b6-d6088c25fe63.png)

```js
Plot.plot({
  y: {
    grid: true,
    label: "↑ Price"
  },
  x: {
    interval: 0.5,
    label: "Carats →",
    labelAnchor: "right",
    tickFormat: (x) => x.toFixed(1)
  },
  marks: [
    Plot.ruleY([0]),
    Plot.boxY(diamonds, {x: (d) => Math.floor(d.carat * 2) / 2, y: "price"})
  ]
})
```

PR #1423, opened 2023-04-04
Sketch: adding override-only mark types to auto

tophtucker:
See https://github.com/observablehq/plot/issues/1422


tophtucker:
Still a decent idea, at least for box plots, but stale and trivial, so closing for now! More in the issue: https://github.com/observablehq/plot/issues/1422

mbostock:
Seems like a quick win if we want to add box charts to the chart cell. I’m supportive! (Voronoi seems more dubious.)

tophtucker:
Just checking in on my stale PRs…
- This should wait for https://github.com/observablehq/plot/pull/1843
- This would benefit from https://github.com/observablehq/plot/issues/1330, but I don’t think that blocks it

Issue #1192, opened 2022-12-19
Expose warnings as property on DOM node


tophtucker:
It'd be great if the ⚠️ warnings were also exposed as a property on the DOM node (like the _plot_.scales property), in addition to being sent to console.warn, so that other code could read them.

I don't know if this is possible or makes sense, but it'd also be great if the warning came with information about where it originated, as if we could place an error caret at the spot in the configuration object where you should look: somewhere in the nth mark, in the x scale, etc.

Related links (mostly a note to self):
- https://github.com/observablehq/plot/issues/536
- https://github.com/observablehq/plot/issues/755
- https://github.com/observablehq/plot/blob/1df1d85818f082779d902d1588326880dc332e6c/src/warnings.ts

PR #1424, opened 2023-04-04
Auto: error on reducer without input; warn on ordinal size; expose warnings

tophtucker:
[Demo](https://observablehq.com/d/35391f601fb9181d). Sketching some additional errors / warnings based on what I've seen leading people astray.

**Error on reducer without input** — Setting a non-count reducer without an input field _error_, not warns, because it's a static error in the config, regardless of what data gets passed in. TODO: do any other reducers not need an input field?

**Warn on ordinal size** — Since this is dynamic, it’s just a warning, not an error. One issue is there’s no way to suppress it if for some reason you want this. It should probably be subsumed by https://github.com/observablehq/plot/issues/493, but I was just itching to try something.

**Expose warnings** (https://github.com/observablehq/plot/issues/1192) — In this PR, my approach is to set an array of warnings on the returned figure. Mike has mentioned he might prefer passing in an onWarn callback. I guess the advantage there is that these warnings could be triggered by interaction in the future? Right now, the chart cell almost never runs into warnings, but there are situations where it seems like it should (like ordinal size), so I feel the itch to be able to expose the warnings in its UI.

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -11,14 +11,25 @@ import {frame} from "./frame.js";
 import {line, lineX, lineY} from "./line.js";
 import {rectX, rectY} from "./rect.js";
 import {ruleX, ruleY} from "./rule.js";
+import {warn} from "../warnings.js";
 
 export function autoSpec(data, options) {
   const {x, y, fx, fy, color, size, mark} = autoImpl(data, options);
   return {x, y, fx, fy, color, size, mark};
 }
 
+function isUnderspecifiedReduce({value, reduce}) {
+  // TODO: are there reducers other than count that need no input?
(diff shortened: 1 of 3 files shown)


Fil:
sum does not need an input (it defaults to count)

tophtucker:
Even though sum and distinct work as count, I've seen several people get confused by selecting "distinct" in the chart cell and seeing nothing change, so I'm still sort of inclined to throw (or warn?) on them.

tophtucker:
I broke out the one solid piece (error on underspecified reduce) to https://github.com/observablehq/plot/pull/1833.

https://github.com/observablehq/plot/issues/493 and https://github.com/observablehq/plot/issues/1192 remain as open issues; I don’t feel good enough about the approaches in this PR.

PR #1421, opened 2023-04-03
circular images, clip css basic shapes

Fil:
support for CSS basic shapes in the mark’s **clip** option

the **r** option of the image mark sets a default width = 2*r, and a default clip=circle()

todo:
- [x] variable **r**, associated with the _r_ scale
- [x] complete documentation

closes #1413

PR #1425, opened 2023-04-05
circular images

### LINKED ISSUE #1413: Circular images
E.g. [here](https://twitter.com/tanya_shapiro/status/1641581867781562369), it’d be really nice to have an easy way to apply a circular clip to the image mark

untitled (45)

Maybe rx/ry would work? Or maybe better, you specify an r option and then it works well with the dodge transform.

Fil:
With the **r** option, Plot.image sets the default height, width to twice the effective radius (constant or variable), and clips to a circle(), unless **clip** is explicitly _null_.

Adds circular kitten tests.

Supersedes #1421
closes #1413


src/marks/image.js
```diff
@@ -40,17 +40,19 @@ function maybePathChannel(value) {
...
-    let {x, y, width, height, src, preserveAspectRatio, crossOrigin, frameAnchor, imageRendering} = options;
+    let {x, y, r, width, height, src, preserveAspectRatio, crossOrigin, frameAnchor, imageRendering} = options;
     if (width === undefined && height !== undefined) width = height;
     else if (height === undefined && width !== undefined) height = width;
     const [vs, cs] = maybePathChannel(src);
-    const [vw, cw] = maybeNumberChannel(width, 16);
-    const [vh, ch] = maybeNumberChannel(height, 16);
+    const [vr, cr] = maybeNumberChannel(r);
+    const [vw, cw] = maybeNumberChannel(width, r != null ? cr : 16);
+    const [vh, ch] = maybeNumberChannel(height, r != null ? cr : 16);
```
mbostock, line 49: Slight preference for treating a null `r` as zero instead of undefined, so that we don’t have to accommodate the null type (and instead it is implicitly coerced to a number). Also, I’d prefer if this logic shook out like this, instead of needing new this.sw, this.sh, this.clipR fields:

```suggestion
    const [vw, cw] = maybeNumberChannel(width, r === undefined ? 16 : cr * 2);
    const [vh, ch] = maybeNumberChannel(height, r === undefined ? 16 : cr * 2);
```

src/marks/image.js
```diff
@@ -61,14 +63,18 @@ export class Image extends Mark {
     this.src = cs;
     this.width = cw;
     this.height = ch;
+    this.sw = width == null && r != null ? 1 : 1 / 2;
+    this.sh = height == null && r != null ? 1 : 1 / 2;
+    this.clipR = r != null && options.clip !== null;
```
mbostock, line 68: I think we’ll want to allow **r** to be specified _in addition_ to **width** and **height**, so that, for example, you can have a radius which crops the rectangle (rather than always being a radius of 100%).

src/marks/image.js
```diff
@@ -40,17 +40,19 @@ function maybePathChannel(value) {
...
-    const [vh, ch] = maybeNumberChannel(height, 16);
+    const [vr, cr] = maybeNumberChannel(r);
+    const [vw, cw] = maybeNumberChannel(width, r != null ? cr : 16);
+    const [vh, ch] = maybeNumberChannel(height, r != null ? cr : 16);
     super(
       data,
       {
         x: {value: x, scale: "x", optional: true},
         y: {value: y, scale: "y", optional: true},
+        r: {value: vr, scale: "r", filter: positive, optional: true},
```
mbostock, line 55: Now that the image supports the **r** channel, it should have the same withDefaultSort as the dot mark does.

mbostock (changes requested):
This is closer to what I was expecting. Still some thoughts on the interface and the implementation. I’ll take a crack at this now…

test/plots/kitten.ts
```diff
@@ -0,0 +1,37 @@
+import * as Plot from "@observablehq/plot";
+
+async function kitten({r, clip = undefined, dodge = false}) {
+  let options = {
```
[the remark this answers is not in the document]
mbostock: Like so:

```js
async function kitten({
  x = (d, i) => i % 5,
  y = (d, i) => Math.floor(i / 5),
  src = (d, i) => `https://placekitten.com/${100 + 2 * i}/${100 + 2 * i}`,
  ...options
}: Plot.ImageOptions = {}) {
  return Plot.plot({
    inset: 60,
    width: 520,
    height: 520,
    axis: null,
    r: {range: [10, 60]},
    marks: [Plot.image({length: 25}, {x, y, src, ...options})]
  });
}
```

For the dodge case, you can then invoke it like so:

```js
kitten(Plot.dodgeY({r: (d, i) => i}))
```

This means you don’t need to declare the types of the (d, i), too.

src/marks/image.d.ts
```diff
@@ -31,6 +31,15 @@ export interface ImageOptions extends MarkOptions {
    */
   height?: ChannelValue;
 
+  /**
+   * The image radius, for rounded images. When a number, it is interpreted as a
+   * constant radius in pixels; otherwise it is interpreted as a channel,
+   * typically bound to the *r* scale. Also defaults **height** and **width** to
+   * twice its effective value, and clips the image to a circle, unless **clip**
+   * is set to null.
```
mbostock, line 39: I removed the dependence on **clip**… need to think more though.

mbostock (approved):

mbostock:
We also fixed an unfiled bug where the dodge transform was ignoring the **r** dodge option when the dodge options are passed separately.

PR #1432, opened 2023-04-07
client-side renderer, maybe

Fil:


mbostock:
Probably superseded by 0ff2d22269f1d7f2796b51e9e5847f7bc32429f4 (and later enhancements)? Though I expect we’ll still want a client-only version of :::plot still, too, for the expensive plots (like maps) that we don’t want to serialize as SVG. Perhaps.

PR #1441, opened 2023-04-13
propagate className option to legends

mbostock:
This is needed for deterministic server-side rendering of legends.


src/plot.js
```diff
@@ -141,7 +141,7 @@ export function plot(options = {}) {
   const {fx, fy} = scales;
   const subdimensions = fx || fy ? innerDimensions(scaleDescriptors, dimensions) : dimensions;
   const superdimensions = fx || fy ? actualDimensions(scales, dimensions) : dimensions;
-  const context = createContext(options, subdimensions, scaleDescriptors);
+  const context = createContext(options, subdimensions, className);
```
Fil, line 144: className replaces the spurious argument scaleDescriptors in the call to createContext—this was introduced by #1197, and was ignored in the function's signature.

Fil (approved):

PR #1446, opened 2023-04-14
deterministic class name

### LINKED ISSUE #1433: Plot’s className doesn’t need to be quite so random
For server-side rendering, it’d be nice if Plot used a deterministic name for its classes. The class name should only need to change if its contents (the default styles) change, which means it could probably be fixed depending on which release of Plot you are using.

mbostock:
Adopts a fixed class name as the default, instead of a random one, better determinism by default during server-side rendering. We’ll only need to change the default class name when we change the default styles.

When doing this, note that we now have to be careful that our internal styles don’t conflict. For example, ramp legends, swatches legends, and plots must now have different namespaces for styles. The swatches legend required a few additional changes because it was using a dynamic stylesheet; it now uses inline styles or attributes instead.

Fixes #1433.

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -40,7 +40,7 @@ export function exposeLegends(scales, context, defaults = {}) {
 }
 
 function legendOptions({className, ...context}, {label, ticks, tickFormat} = {}, options) {
-  return inherit(options, {className: `${className}-legend`, ...context}, {label, ticks, tickFormat});
+  return inherit(options, {className, ...context}, {label, ticks, tickFormat});
 }
 
 function legendColor(color, {legend = true, ...options}) {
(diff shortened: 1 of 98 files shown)


src/legends.js
```diff
@@ -40,7 +40,7 @@ export function exposeLegends(scales, context, defaults = {}) {
 }
 
 function legendOptions({className, ...context}, {label, ticks, tickFormat} = {}, options) {
-  return inherit(options, {className: `${className}-legend`, ...context}, {label, ticks, tickFormat});
+  return inherit(options, {className, ...context}, {label, ticks, tickFormat});
```
mbostock, line 43: This suffixing is moved down to the legend implementation because we need a distinct suffix for each legend implementation to avoid conflicting styles.

Fil (approved):

PR #1454, opened 2023-04-16
allow CSS Color Module Level 5

### LINKED ISSUE #1851: allow native color functions with CSS vars as colors
Closely related to https://github.com/observablehq/plot/issues/765.

It would be great if the `isColor` check supported native color functions with CSS vars inside. Some examples:
- `rgba(255, 0, 0, var(--a))`
- `rgb(var(--rgb))`
- `hsl(240deg 50% var(--l))`
- `hsl(var(--hsl) / 50%)`

Thanks!

mbostock:
Ref. https://www.w3.org/TR/css-color-5/

Allows a variety of new ways of specifying colors, including mixing colors:

```
color-mix(in lch, purple 50%, plum 50%)
```

And relative colors:

```
lch(from currentColor calc(1 - l) c h)
```

Though, I can’t really seem to get the latter to work, so probably we should leave this as a draft until it works.

Fixes #1851.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,4 +1,4 @@
-import {color, descending, quantile, range as rangei} from "d3";
+import {descending, quantile, range as rangei} from "d3";
 import {parse as isoParse} from "isoformat";
 import {defined} from "./defined.js";
 import {maybeTimeInterval, maybeUtcInterval} from "./time.js";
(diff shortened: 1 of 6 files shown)


Fil:
I'm fine with merging this earlier than it works in browsers, since we're just "passing colors through". When it works, your second example will be very useful for dark mode (just stating the obvious).

[color-mix](https://caniuse.com/mdn-css_types_color_color-mix) is still waiting for firefox, it seems. (For lch() I couldn't find the additional syntax on caniuse.)

mbostock:
This looks ready — it just needs test. The lack of browser support for level 5 shouldn’t be a blocker because this is also adding support for level 4.

Fil (approved):
p3 colors yay!!

PR #1455, opened 2023-04-16
fix #1453; axis label and href

### LINKED ISSUE #1453: Axis label disappears if href channel is specified for axis mark
When an axis mark is used with both the `label` option and an `href` channel, the label is no longer rendered:
```js
Plot.plot({marks: [
  Plot.tickX("ABCDEF"),
  Plot.axisX({label: "Letter", href: d => `https://en.wikipedia.org/wiki/${d}`})
]})
```
Compare to the same Plot without href:
```js
Plot.plot({marks: [
  Plot.tickX("ABCDEF"),
  Plot.axisX({label: "Letter"})
]})
```

image

mbostock:
Fixes #1453. The problem is that the **href** option was being passed to the axis label mark, too, which presumably caused the mark to be filtered out? I think we might need a more general solution for preventing channels on the singular axis label mark…

### DIFF
--- a/src/marks/axis.js
+++ b/src/marks/axis.js
@@ -2,8 +2,9 @@ import {extent, format, utcFormat} from "d3";
 import {formatDefault} from "../format.js";
 import {marks} from "../mark.js";
 import {radians} from "../math.js";
-import {range, valueof, arrayify, constant, keyword, identity, number} from "../options.js";
-import {isNoneish, isIterable, isTemporal, maybeRangeInterval, orderof} from "../options.js";
+import {arrayify, constant, identity, keyword, number, range, valueof} from "../options.js";
+import {isIterable, isNoneish, isTemporal, orderof} from "../options.js";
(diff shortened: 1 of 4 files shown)


Fil (changes requested):
I haven't given it much thought but I'm not sure how we could avoid having a list of option names we support for the label (whether that list is opt-in or like here, opt-out from all the options supported by the text mark).

After testing all the potentially-channel options (based on the typescript suggestions—thank you typescript!), these are the ones that also disappear the label: ariaLabel, fontSize, fillOpacity, opacity, rotate, title.

The mark crashes if passed a **filter** option.

mbostock:
This is still an improvement over the current behavior, so not sure it’s right to block this PR. I agree we should look for a more comprehensive fix however.

mbostock:
Rewritten to only allow constant options to propagate to the axis label.

Fil (approved):
thanks!

PR #1468, opened 2023-04-19
derive x & y scale domains from geometry

### LINKED ISSUE #1507: aspectRatio fails when there is a geo mark and no projection
```js
values = d3.cross(d3.range(10), d3.range(10)).map(([i, j]) => i * j);
```
```js
Plot.plot({
  aspectRatio: true,
  inset: 40,
  x: { ticks: 10 },
  marks: [
    Plot.dot(values, {
      x: (d, i) => i % 10 + 0.5,
      y: (d, i) => Math.floor(i / 10) + 0.5,
      fill: (d) => d
    }),
    Plot.geo([])
  ]
})
```

Note: in that case aspectRatio is not completely ignored, but defaults to the golden ratio.

mbostock:
While writing the documentation for the geo mark, I was thinking about how we can use the geo mark independent of a projection. It bothered me that you can use the geo mark in conjunction with predefined scales, or scales derived from other marks, but that the geo mark itself could not assist in the automatic definition of the scales like other marks.

So now it does. The _geometry_ channel now produces derived _x_ and _y_ channels that can be used to initialize the _x_ and _y_ scales, just like normal marks. Which means you can now say

```js
Plot.geo({type: "LineString", coordinates: aapl.map((d) => [d.Date, d.Close])}).plot()
```

and it behaves almost identical to

```js
Plot.line(aapl.map((d) => [d.Date, d.Close])).plot()
```

This is still a draft because I think we should skip constructing the derived _x_ and _y_ channels in a couple cases:

- If there’s a **projection** (because then there won’t be _x_ and _y_ scales by definition)
- If the _x_ or _y_ scales have pre-defined domains (because then we don’t care about these derived channels)

This changes the behavior of one of the tests, but I think it makes sense, and you can still get the old behavior by setting the **projection** to _identity_ if you want it.

Fixes #1507.


Fil:
This makes it much easier to just pass in a GeoJSON file. No more quarter-Earth!

Fil:
https://observablehq.com/@observablehq/test-1468

src/plot.js
```diff
@@ -410,21 +410,36 @@ function inferChannelScales(channels) {
...
-        if (scaleChannels !== undefined) scaleChannels.push(channel);
-        else channelsByScale.set(scale, [channel]);
+        // Geo marks affect the default x and y domains if there is no
+        // projection. Skip this (as an optimization) when a projection is
+        // specified, or when the domains for x and y are specified.
+        if (scale === "projection") {
+          if (!hasProjection(options)) {
+            const [x, y] = getGeometryChannels(channel);
+            if (options.x?.domain === undefined) addScaleChannel(channelsByScale, "x", x);
+            if (options.y?.domain === undefined) addScaleChannel(channelsByScale, "y", y);
```
mbostock, line 426: Ideally we skip getGeometryChannels if both options.x?.domain and options.y?.domain are defined.

test/plots/projection-null.ts
```diff
@@ -0,0 +1,12 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+import {feature} from "topojson-client";
+
+export async function projectionNull() {
+  const world = await d3.json<any>("data/countries-110m.json");
+  const land = feature(world, world.objects.land);
+  return Plot.plot({
+    projection: null,
```
mbostock, line 9: The projection is null by default, so this isn’t needed here, right?

mbostock:
This looks good to me. What do you think @Fil?

test/plots/projection-null.ts
Fil: The intent was as a memento that in this case the projection is null(ish), but that's already in the test name!

src/plot.js
Fil: yes we should do this

test/plots/projection-null.ts
mbostock: I’ll leave it as-is since the geoLine test covers the undefined case.

Fil (approved):

Fil:
🎉 We should now be able to throw _any_ valid GeoJSON file at Plot.geo(…).plot(), and have something show up on the screen—independently of the file’s coordinate reference system (CRS). Plus, the x and y axes will show the extent of the coordinates, which will help evaluate if the file is “unprojected” or “projected”, and even maybe  (with a bit of practice) which CRS is more likely.

PR #1479, opened 2023-04-23
precise fractional interval

### LINKED ISSUE #1450: Better precision for fractional intervals
Currently you can say `interval: 0.5` as a scale option, but I think it would be better to also allow `interval: -2` in a similar fashion to d3.tickIncrement for better numeric precision.

mbostock:
Fixes #1450.


test/plots/boxplot.ts
```diff
@@ -1,5 +1,42 @@
...
+  });
+}
+
+export async function boxplotFacetNegativeInterval() {
+  const olympians = await d3.csv<any>("data/athletes.csv", d3.autoType);
+  return Plot.plot({
+    fy: {
+      grid: true,
+      tickFormat: String, // for debugging
+      interval: -10, // 0.1
```
Fil, line 32: shouldn't this be 0.1?

mbostock: No, that’s what the boxplotFacetInterval test is for. This one is explicitly testing a negative interval, hence the name.

Issue #1483, opened 2023-04-25
An easier facet frame label?


mbostock:
It’d be nice when faceting to have an easier way to label the facet. I find myself using the selectFirst transform for this, but I think a singleton mark like the frame mark would be easier to use.

<img width="688" alt="Screenshot 2023-04-24 at 6 05 00 PM" src="https://user-images.githubusercontent.com/230541/234149841-4b44edca-15a3-4353-bdca-9146fc328007.png">

```js
Plot.plot({
  height: 720,
  axis: null,
  marks: [
    Plot.areaY(industries, {x: "date", y: "unemployed", fy: "industry"}),
    Plot.text(industries, Plot.selectFirst({text: "industry", fy: "industry", frameAnchor: "top-left", dx: 6, dy: 6})),
    Plot.frame()
  ]
})
```

Fil:
```js
Plot.plot({
  height: 720,
  axis: null,
  marginRight: 0,
  marks: [
    Plot.areaY(industries, {x: "date", y: "unemployed", fy: "industry"}),
    Plot.axisFy({frameAnchor: "top-left", dy: 10, label: null}),
    Plot.frame()
  ]
})
```

note that this assumes the default axisFy anchor (right)

mbostock:
I found that I needed a longer incantation to get that to work as expected, because even though the axis mark is only rendering the text label, it still introduces margins.

```js
Plot.axisFy({frameAnchor: "top-left", dy: 10, label: null, marginTop: 0, marginRight:0, marginBottom: 0, marginLeft: 0})
```

And that’s longer than what I had previously:

```js
Plot.text(industries, Plot.selectFirst({text: "industry", fy: "industry", frameAnchor: "top-left", dx: 6, dy: 6}))
```

But, it does occur to me there’s a separate bug where the axis marks are ignoring the **margin** option. I should be able to specify it it shorter as:

```js
Plot.axisFy({frameAnchor: "top-left", dy: 10, label: null, margin: 0})
```

And for that matter, maybe the axis mark could even be smarter based on the **frameAnchor** option? Anyway, I still think we want a separate labelFx and labelFy mark or some such for this, though I appreciate you sharing your alternative!

Fil:
related #393

PR #1484, opened 2023-04-25
fix axis margin option

mbostock:
If you specify a **margin** option to an axis mark, it should set the margins.

Ref. https://github.com/observablehq/plot/issues/1483#issuecomment-1521085850


Fil (approved):
that's a bug fix!

PR #1485, opened 2023-04-25
inline facet axis

### LINKED ISSUE #1483: An easier facet frame label?

mbostock:
Fixes #1483. Adds a new *inline* axis **anchor** for facet axes that just draws the text label for the facet. You can then use the **frameAnchor** option to place the text label on the desired side or corner.

For example (a little crowded, but it works):

Screenshot 2023-04-25 at 10 28 44 AM

```js
Plot.plot({
  padding: 0,
  projection: "albers",
  fx: {axis: "inline", interval: d3.utcYear.every(10)},
  marks: [
    Plot.geo(statemesh, {strokeOpacity: 0.2}),
    Plot.geo(nation),
    Plot.geo(walmarts, {fx: (d) => d.properties.date, r: 1.5, fill: "blue"})
  ]
})
```

When using an *inline* facet axis, the tick mark and the axis label are suppressed.


Fil (commented):
I'm adding the types, documentation and a test.

src/marks/axis.d.ts
```diff
@@ -34,7 +37,7 @@ export interface GridOptions extends GridScaleOptions {
    * Plot.gridY({x: (y) => aapl.find((d) => d.Close >= y)?.Date, anchor: "right"})
    * ```
    */
-  anchor?: "top" | "right" | "bottom" | "left";
+  anchor?: "top" | "right" | "bottom" | "left" | "inline";
```
mbostock, line 40: This is a bug (both in the docs and the code). The grid mark doesn’t support the *inline* orientation.

mbostock: I guess we can allow it here in the TypeScript declarations (just like we allow the *right* orientation for the axisX in the TypeScript declarations, even though it isn’t allowed at runtime). I’ll fix it at runtime though.

docs/marks/axis.md
```diff
@@ -337,6 +348,23 @@ Plot.plot({
 ```
 :::
 
+Facet axes also support the *inline* anchor, which draws the facet value inside the frame.
+
+:::plot https://observablehq.com/@observablehq/plot-inline-facet-anchor
```
mbostock, line 353: This needs to be `defer` since the data isn’t loaded until mount.

PR #1489, opened 2023-04-26
remove default white background

### LINKED ISSUE #1434: Plot’s default background: white often isn’t a great default
The background: white is nice for generating opaque screenshots when using Download PNG on Observable, but it makes it harder to use Plot in another web application, say with a dark theme—especially because Plot’s foreground color defaults to _currentColor_ and not _black_. When Plot supports themes #630 (configurable defaults), it might makes sense to have the default background: white only apply on Observable. Perhaps we can have a workaround in the interim, say by looking at location.origin.

### LINKED ISSUE #1463: Markers (and the tree mark) should use a CSS variable instead of white
In dark mode, the *white* stroke on the *circle* marker doesn’t have the intended effect:

Screenshot 2023-04-18 at 10 50 22 AM

If markers used a CSS variable instead of hard-coding *white*, it’d be easier to override the stroke color when in dark mode using an external stylesheet. Related #1434.

mbostock:
Fixes #1434. When upgrading Observable, we can either restore the white background with an external stylesheet, or by changing the download PNG functionality.

Also fixes #1463.


Fil (approved):

mbostock:
I’m not sure I want to merge this. I feel like you can already override the styles if you want to. Maybe this gives better default results, but maybe it just causes churn? 🤔 The shadow DOM approach seems attractive but that seems more disruptive too. Hmm.

Fil:
Can we then close these PRs (or move them to draft status)?

PR #1493, opened 2023-04-26
ignore docs for ci

mbostock:


visnup (approved):

mbostock:
I reverted this because it breaks documentation-only PRs, which now don’t report a status check (because the docs are only built on the main branch when deploying, and the CI doesn’t run since only the docs changed). I think perhaps the right fix is to have a separate workflow that just builds the docs, without deploying them? So that CI tests that the docs build before merging to main.

PR #1497, opened 2023-04-27
quantize transform

mbostock:
For converting quantitative or temporal data into ordinal data.

image

```js
Plot.plot({
  x: {tickFormat: "%Y"},
  marks: [
    Plot.barY(olympians, Plot.groupX({y: "count"}, Plot.quantizeX(d3.utcYear.every(5), {x: "date_of_birth"})))
  ]
})
```

The previous implementation of ordinal intervals #513 #849 uses a scale transform. This is fine if there’s no mark transform involved, but since the mark transforms run _before_ the scale transform (i.e., the scale transform applies to the _output_ of the mark transform rather than the input), it doesn’t help much in the case above, when grouping. I originally considered adding an **interval** option to the group transform, but that starts to get confusing with the existing **interval** mark option (which applies to the opposite dimension in the case of barY!), so I thought it better to have an explicit transform.

Regarding the implementation, I like the reuse of Plot.map, but I don’t like the plumbing required to propagate the knowledge that a channel has an associated interval. I added the plumbing to the group transform for *x* and *y*, but I didn’t add it anywhere else, so there’s probably some cases where this isn’t going to work. This plumbing is needed to ensure that the (ordinal) scale produces a tick for every interval value in the domain, and to suppress the warning that would otherwise occur when using quantitative or temporal data with an ordinal scale.

~~Another confusing aspect is that there’s already the **interval** mark option, but that does something else, which is converting a singular *value* into an extent [*start*, *stop*]. Like, in the case of barY, now I wonder if the **interval** option should instead apply to the *x* channel to quantize values, instead of applying to the *y* channel like it does now. I don’t want to make a backwards-incompatible change, though.~~ Maybe this suggests that we should call this the “quantize” transform instead of the “interval” transform? Update: Yes, we’ll use the name “quantize” instead.

TODO

- [x] Rename to “quantize” transform?
- [x] Add a generic ~~Plot.interval~~ Plot.quantize transform for any channel (akin to Plot.map)?
- [x] Adopt named skip intervals from #1506.
- [ ] Do other channels need the interval hint propagated?
- [ ] Do other transforms need to propagate the interval hint?
- [ ] Create a documentation page for the quantize transform
- [ ] Create a documentation page for intervals (documenting named intervals etc.)


mbostock:
I want to try a different take on this using the existing **interval** scale option. If I can make that available to the group transform, then I can make this work:

```js
Plot.plot({
  x: {
    tickFormat: "%Y",
    interval: "5 years"
  },
  marks: [
    Plot.barY(olympians, Plot.groupX({y: "count"}, {x: "date_of_birth"}))
  ]
})
```

(And in theory we could infer the appropriate **tickFormat**, too.)

mbostock:
Closing in favor of #1511.

PR #1504, opened 2023-04-28
Remove autoImpl and return more from autoSpec

tophtucker:
To support compiling an autoSpec to the equivalent source using explicit marks. This expands the returned surface area of autoSpec a lot; so far we’ve been treating it as an undocumented and unstable function for mostly internal use, so maybe that’s OK.

The giant switch statements could be rewritten to something more like: 

```js
const MARKS = {
  barX,
  barY…
}
```

```
// in auto
markImpl = MARKS[markImpl]
```

but I seem to remember you saying you’d prefer the explicitness of a switch.


test/marks/auto-test.js
```diff
@@ -10,7 +10,20 @@ it("Plot.autoSpec makes a histogram from a quantitative dimension", () => {
...
+    markImpl: "rectY",
+    markOptions: {
+      fx: undefined,
+      fy: undefined,
+      x: {value: Object.assign([1, 1, 38], {label: "value"})},
+      y: undefined,
+      fill: undefined,
+      z: undefined,
+      r: undefined
+    },
```
mbostock, line 23: Remind me why we need to expose `markOptions`? And `transformOptions`?

tophtucker: Yeah it does feel somewhat redundant with the other stuff we pass down. But it's the smallest change that makes this all work. 🤷 

markOptions and transformOptions are used to ultimately invoke the mark and transform implementations (and markOptions is the one place we pass down the materialized columns):

https://github.com/observablehq/plot/blob/e69d9b0b20427c56c58ab8ab0e9a7bd7d6e97e56/src/marks/auto.js#L295

If we wanna fold autoImpl back into autoSpec, we have to either expose markOptions and transformOptions from autoSpec, or down everything that goes into them. The advantage of having them in autoSpec is that it's easier to write a different "compiler" for the last mile, e.g. to construct a string instead of actually invoking the mark and transform: https://observablehq.com/@tophtucker/convert-plot-auto-to-explicit-marks.

It does sorta weird that we pass down y reduce info in both of these ways:

https://github.com/observablehq/plot/blob/e69d9b0b20427c56c58ab8ab0e9a7bd7d6e97e56/src/marks/auto.js#L142-L143

https://github.com/observablehq/plot/blob/e69d9b0b20427c56c58ab8ab0e9a7bd7d6e97e56/src/marks/auto.js#L186

But, the first is in the shape an explicit mark expects (composable transforms held apart from channels); the latter is in the shape the auto mark expects (reducers that imply transforms inside the channel options). If we wanna both 

- return a version of the auto mark options with all the defaults filled in (needed to invoke autoSpec idempotently in a call to auto, or to tell the notebook editor how an auto mark will behave), and
- support either immediately instantiating the mark or representing it as a string without repeating the options construction (needed for ejecting to explicit marks)
 
, then it feels reasonable to return with some redundancy in the shape.

mbostock (approved):
I thought of a way to avoid the verbosity of the switch statements, and to retain the symbolic references internally (and minimize the diff). This removes a little bit of error checking, but I don’t think we really needed that anyway since the names are only generated internally and we already have unit tests for it.

PR #1509, opened 2023-04-29
frame channels

### LINKED ISSUE #1449: Misspelling a frame’s stroke produces a confusing error
The frame mark doesn’t have data, but since it extends Mark, it tries to interpret the **stroke** option as a channel. This leads to a confusing error when you have a typo:

```js
Plot.frame({stroke: "bluez"})
```

The crash occurs when Plot tries to iterate over the data, which is null:

Screenshot 2023-04-14 at 8 35 19 PM

Since the frame mark doesn’t have data, we should never interpret the stroke as a channel. Though, it would be nice if somehow you could specify the stroke as an abstract value (*e.g.*, `stroke: "Biscoe"`) and that could get passed through the _color_ scale…

mbostock:
Fixes #1449. _E.g._ `Plot.frame({fill: "foo"})` sets the **fill** to the abstract value _foo_ by applying the _color_ scale.


mbostock:
Should probably update the documentation with a related example…

Fil (approved):
🧑‍🍳

PR #1511, opened 2023-04-30
interval-aware transforms

mbostock:
In this approach, transforms are passed the plot options as an additional argument; this allows the group transforms (and likely other transforms) to check if the scale has a declared **interval** option, and if so, apply that interval _before_ grouping. This is safe to do because applying an interval is idempotent.

image

```js
Plot.plot({
  x: {tickFormat: "%Y", interval: "5 years"},
  marks: [Plot.barY(olympians, Plot.groupX({y: "count"}, {x: "date_of_birth"}))]
})
```

~~I’m leaving this as a draft because we need to think through what other transforms need this logic. Definitely the stack transform, at least…~~ It’s ready.

This looks more promising than #1497. Previously #513 #849.


test/plots/ibm-trading.ts
```diff
@@ -1,17 +1,12 @@
...
-// not a recommended pattern: you should instead parse strings to dates when
-// loading the data, say by applying d3.autoType or calling array.map.
 export async function ibmTrading() {
-  const ibm = await d3.csv<any>("data/ibm.csv").then((data) => data.slice(-20));
+  const ibm = await d3.csv<any>("data/ibm.csv", d3.autoType).then((data) => data.slice(-20));
   return Plot.plot({
     marginBottom: 65,
     x: {
       interval: "day",
-      transform: (d) => d3.utcDay.floor(d3.isoParse(d)),
```
mbostock, line 14: This technique no longer works here because, again, the scale **transform** is applied _after_ the mark transform, so if we want the mark transform to apply the scale’s **interval**, we need to parse the dates before the mark transform runs. Since this technique was never recommended (as documented), I think it’s acceptable to change the behavior, and I rewrote this example to use the recommended technique instead.

mbostock:
Okay, I added support for interval-aware grouping on **x** and **y** to the bin and stack transforms, and I think this is now ready for review. 👍

It’s possible in the future we might want to add interval-aware grouping for *z* when implicitly defined as **fill** or **stroke** and when the *color* scale has a declared **interval**. But I think that is rare, and it is reasonable for us to punt now to keep things simple.

test/plots/interval-aware.ts
```diff
@@ -0,0 +1,26 @@
...
+  return Plot.plot({
+    x: {interval: 10},
+    marks: [Plot.barY(olympians, Plot.binY({fill: "count"}, {x: "weight", y: "height", inset: 0}))]
+  });
+}
+
+export async function intervalAwareGroup() {
+  const olympians = await d3.csv<any>("data/athletes.csv", d3.autoType);
+  return Plot.plot({
+    x: {tickFormat: "%Y", interval: "5 years"},
```
mbostock, line 15: We can drop the `tickFormat: "%Y"` when #1512 is merged.

test/plots/ibm-trading.ts
Fil: Side note: we could make the previous version of ibm-trading work again if the "day" interval was allowed to read strings (as discussed in #1326).

Fil:
This seems to open the door to many more (custom) behaviors in the transforms. Which could be immensely powerful—I can now imagine a transform that would simplify a geo based on chart width, for example? Are we comfortable with transforms reading options.marks? Or (ab)using this to mutate options (e.g. writing options.width changes the chart's width)?

mbostock:
I definitely wouldn’t support mutating options, but reading options, yes.

test/plots/ibm-trading.ts
mbostock: Yes, but as we discussed previously, then `interval: d3.utcDay` would not behave the same as `interval: "day"`, and I don’t think we should go down further the slippery slope of implicit coercion when it’s better to teach users to set their types earlier.

Fil (approved):

PR #1512, opened 2023-04-30
detect ordinal year intervals

mbostock:
If an ordinal scale has a yearly interval, use the appropriate %Y tick format by default, instead of isoformat which adds a trailing -01-01. Ref. https://github.com/observablehq/plot/issues/768#issuecomment-1528924713.

### DIFF
--- a/src/time.js
+++ b/src/time.js
@@ -70,3 +70,15 @@ export function maybeTimeInterval(interval) {
 export function maybeUtcInterval(interval) {
   return parseInterval(interval, utcIntervals);
 }
+
+export function isUtcYear(i) {
+  if (!i) return false;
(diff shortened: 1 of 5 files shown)


src/time.js
```diff
@@ -70,3 +70,15 @@ export function maybeTimeInterval(interval) {
 export function maybeUtcInterval(interval) {
   return parseInterval(interval, utcIntervals);
 }
+
+export function isUtcYear(i) {
+  if (!i) return false;
+  const date = i.floor(new Date(Date.UTC(2000, 11, 31)));
+  return utcYear(date) >= date;
```
Fil, line 77: do we want a comment here saying that this is in fact checking "equality"?

Fil (approved):

PR #1516, opened 2023-05-01
axis label inset

### LINKED ISSUE #1414: How to offset an axis label
**labelOffset** moves the axis label in the direction orthogonal to the axis (say horizontally for the *y* axis). But how would you move it in the direction of the axis?

Suppose you want to  move the axis label down (because there is a scale.insetTop or a large marginTop).

A hacky solution (implemented in https://observablehq.com/@recifs/a-centered-title) is  an explicit label with lots of \n (spaces if working on the *x* axis) 
>  Plot.axisY({ label: "\n\n\n\n\n↑ flipper_length_mm" }),

Similarly, a text mark with facetAnchor+frameAnchor.

Note: dy doesn't apply to the axis label (this is probably a bug in itself?); if this worked, we could create a specific axis just for the label, and offset it. It would be tedious, but would benefit from the automatic label.

I think we should:

- [ ] fix the bug with *dy*
- [ ] introduce a new option orthogonal to **labelOffset** (name?)

Fil:
closes #1414

Note: I started by supporting offsetTop, offsetBottom, and inset as a shorthand, but supporting only inset with the “natural“ direction appeared like a simpler option.

### DIFF
--- a/docs/marks/axis.md
+++ b/docs/marks/axis.md
@@ -358,6 +358,7 @@ In addition to the [standard mark options](../features/marks.md), the axis mark 
 * **label** - a string to label the axis; defaults to the scale’s label, perhaps with an arrow
 * **labelAnchor** - the label anchor: *top*, *right*, *bottom*, *left*, or *center*
 * **labelOffset** - the label position offset (in pixels; default depends on margins and orientation)
+* **inset** - the label position inset in pixel; defaults to 0
 * **color** - the color of the ticks and labels (defaults to *currentColor*)
 * **textStroke** - the color of the stroke around tick labels (defaults to *none*)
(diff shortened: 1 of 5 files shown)


Fil:
It might be a bit surprising that the name is **inset** since the other direction is **labelOffset**?

Fil:
Merged in #1517 and fixed the _federalFunds_ test (which can't be fixed within #1517).

Fil:
As we discussed it would probably be less confusing to call this labelDx (and y), since it's not a segment that we inset, but just a way of translating the label. (We'd lose the relative positioning though, when switching from left to right labelDx would always go →; I think it's fine.)

PR #1517, opened 2023-05-01
axis dx and dy must be applied to the axis label too

Fil:
for #1414

(note: conflicts with #1516, but the merge is trivial)


Fil:
The [federalFunds](http://localhost:8008/?test=federalFunds) test cannot be repaired in this case, we need #1516 for that.

PR #1525, opened 2023-05-05
tooltip: true

mbostock:
In this branch, I’m trying to figure out how we could provide shorthand syntax for tooltips. So, rather than needing to repeat the channel definitions when declaring a tooltip, we could derive a tooltip from an existing mark (and possibly allow some overrides, say to suppress certain channels from appearing in the tooltip, or altering how the values are formatted).

As a first cut, I recycled the concept of channel aliases from the earlier PR #798. However, I don’t think this will be sufficient for initializers such as hexbin, because initializers can populate channels that are not declared in _mark_.channels, and thus we can’t create static aliases. That suggests that either we’d need to run the initializer on the tooltip mark (which is a bummer because it is duplicate effort and wouldn’t work with a nondeterministic initializer), or the tooltip should instead have a reference to the mark whose channel values we want to show in the tooltip. I think the latter is the way to go, but it’ll require some more plumbing to figure out how that would work… and also we should retain the ability to render an independent tooltip (attached to itself, in effect).

Getting this working is probably a blocker for #1304 because we shouldn’t land the proposed tooltip mark if we’ll have to redesign it substantially to support shorthand. But I think we’re pretty close to figuring this out.

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -5,15 +5,16 @@ import {registry} from "./scales/index.js";
 import {isSymbol, maybeSymbol} from "./symbol.js";
 import {maybeReduce} from "./transforms/group.js";
 
-export function createChannel(data, {scale, type, value, filter, hint}, name) {
-  if (hint === undefined && typeof value?.transform === "function") hint = value.hint;
+export function createChannel(data, channel, name) {
+  if (channel.alias) return channel;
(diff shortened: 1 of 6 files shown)


mbostock:
Superseded by #1527 which, while the syntax is not final, demonstrates how the underlying mechanism will work.

PR #1527, opened 2023-05-06
tip mark + pointer interaction

### LINKED ISSUE #4: Tooltips (hover to read values).

### LINKED ISSUE #443: Crosshairs
Similar to tooltips and an interaction layer, cursors would be 🔥 and IMO, fit pretty naturally into the API by possibly being able to bind arbitrary marks (like `ruleX` and `text`) to your current cursor position. This would probably share a lot of logic with whatever is needed to implement brushing, I would imagine 🤔

mbostock:
This is an alternative to #1304 that decouples the display mark (tip) from interaction (pointer). The tip mark no longer has any built-in interaction and can be used to display static tips. The pointer interaction, meanwhile, handles pointer events and re-renders a downstream mark (often but not always the tip mark) similar to a dynamic select or filter transform. Re-rendering is accomplished using a new render transform mechanism: a **render** option which allows the specified function to wrap _mark_.render.

As with the tooltip mark in the previous PR, the tip mark here uses getBBox to compute text metrics so that the path surrounding the tip text exactly fits the text. This requires the tip mark to render (partially) asynchronously, since plots and marks are initially rendered detached; the tip mark using a microtask (Promise.resolve) or requestAnimationFrame to defer rendering. I would prefer to render the tooltip synchronously, but this would require either approximating text metrics or synchronously inserting the Plot into the document at an arbitrary location, so partial asynchronous rendering feels like the least worst option.

To-do, pointer interaction:

- [x] Find the closest point to the pointer
- [x] Find the closest point in **x** (or **y**), approximately, as for lines and areas
- [x] Find the closest point across all facets
- [x] Limit the search radius
- [x] Click to switch between “sticky” and “free” mode, allowing text selection
- [x] Click to hide the tooltip (clearing the focused point)
- [x] Fix _sticky_ modality when there is faceting; all facets should be sticky, or none
- [x] Fix _sticky_ modality so that if one pointer interaction is sticky, they all are
- [x] Support **x** or **y** not existing; respect **frameAnchor**
- [x] Support **x1** & **y1** and **x2** & **y2** as alternatives to **x** and **y**
- [x] Support **px** & **py** to decouple pointing position from display position
- [x] Render transforms via the **render** mark option
- [x] Do something cleaner than _mark_._render ~~(pass a _next_ option)~~
- [x] Test again on mobile Safari
- [x] Tests
- [ ] Documentation

To-do, tip mark:

- [x] Render the tooltip as SVG ~~or HTML~~, anchored to a point
- [x] Draw a little connection to the anchor point
- [x] Implement four tip orientations
- [x] Option for forcing a specific tip orientation
- [x] Orient the tooltip relative to the anchor automatically based on position
- [x] Minimize changes in tooltip orientation
- [x] Configurable text styles
- [x] Configurable colors (even channels)
- [x] Limit the maximum width of the tooltip, clipping long text
- [x] Show title when tooltip value is truncated
- [x] Add a lil’ drop shadow
- [x] Display unscaled, labeled values for all channels in the tooltip
- [x] Display unscaled facet values, too (index.fx/index.fy)
- [x] Fix the inferred scale label not shown in the axis (_e.g._, “Date”)
- [x] Support **x** or **y** not existing; respect **frameAnchor**
- [x] Support **x1** & **y1** and **x2** & **y2** as alternatives to **x** and **y**
- [x] Display **x1**-**x2** extent for binX (and similar for binY)
- [x] Display **y2**-**y1** length for stackY( and similar for stackX)
- [x] Fix _z_-order of facet axes, which should be underneath facets
- [x] Fix _z_-order of the tip mark, which should be drawn atop all facets
- [x] Fix display of faceted value when the tip mark is not faceted (suppress index.fx/index.fy)
- [x] Fix display values for dodge initializer
- [x] Fix display values for hexbin initializer
- [x] Display inverted hexagon centroids with the hexbin initializer
- [x] Side anchors (not just corners: *top*, *right*, *bottom*, *left*) and *middle*
- [x] Allow the tip drop shadow to be configured #409 
- [x] Allow the tip corner size to be configured
- [x] Don’t have the **fill** and **stroke** channels (etc.) change the appearance of the mark?
- [x] Instead of **fill** and **stroke**, show a color swatch when a channel is bound to the _color_ scale
- [x] Apply the _title_ reducer to **ariaLabel**; use for multi-line tips
- [x] Tests
- [ ] Ability to reorder, hide, and customize the format of the displayed channel values
- [ ] Documentation

To-do, crosshairs:

- [x] Crosshairs composite mark
- [x] Support faceting?
- [x] Support transforms?
- [x] Support initializers?
- [x] Support rule in conjunction with hexbin transform?
- [x] Tests
- [ ] Documentation

To-do, other:

- [x] Shorthand from existing marks #1525

Optional:

- Multi-line text, wrapped text, without a bold label
- For link and arrow, treat {x1, y1} and {x2, y2} as distinct points?
- Option for controlling where a tooltip appears on a rect/area/etc.?
- Respect the **r** channel for hovering large circles?
- Handle multiple dots in the same position (e.g., click to cycle)?
- Automatic dark mode (media query or checking computed background color)?
- Have group, bin, stack transform _etc._ populate a **z** channel for the benefit of tooltips?

Fixes #4.
Fixes #443.


mbostock:
Here’s an example of the facet z-order problem:

Screenshot 2023-05-07 at 9 46 04 AM

Another interesting problem with this approach is that each facet now listens independently for pointer events, so it’s possible to get multiple tooltips from adjacent facets! I’m not sure if this is an urgent problem, but we could certainly add some cross-facet coordination to prevent this from happening. (It can be rare depending on the chosen **maxRadius**, and it might be useful in some cases.)

Screenshot 2023-05-07 at 9 47 18 AM

There’s also a funny problem with the _sticky_ implementation now, in that you can have multiple marks driven by the pointer interaction, and the sticky modality is independent for each mark (based on `g.contains(event.target)`). That means you can click to make one pointer-driven mark sticky while another pointer-driven mark remains free. This also feels like a bit of an edge case, but perhaps we could figure out how to make all the pointer-driven marks for a given plot share the same modality.

mbostock:
Crosshairs!

https://user-images.githubusercontent.com/230541/236692982-1a55285c-0929-438a-a82b-e2ff54c93fc4.mov

mbostock:
I reorganized the to-dos and would like to proceed with this direction instead of #1304. There are a few things still to figure out here, but I think the biggest two are (1) more coordination of the pointer interaction, such that only one point can be focused at a time across facets, and if there are multiple pointer interactions they are required to be defined consistently; and (2) some mechanism that lets the tip mark draw atop other facets.

One possibility for (2) is that we transpose the hierarchy of marks and facets, such that z-order is preserved across facets, rather than only within facets.

mbostock:
Fun experiment using color-mix to apply the **stroke** channel to the drop-shadow filter:

https://user-images.githubusercontent.com/230541/236877133-f054d266-e122-492f-8597-7dbde1a7e277.mov

mbostock:
There are a couple things still to-do above, but I think this is ready for an earnest look.

src/transforms/hexbin.js
```diff
@@ -79,16 +79,21 @@ export function hexbin(outputs = {fill: "count"}, {binWidth, ...options} = {}) {
 
     // Construct the output channels, and populate the radius scale hint.
     const binChannels = {
-      x: {value: BX},
-      y: {value: BY},
+      x: {value: BX, source: null}, // or {value: map(BX, scales.x.invert), scale: "x"}?
+      y: {value: BY, source: null}, // or {value: map(BY, scales.y.invert), scale: "y"}?
```
Fil, line 83: Using scales.x.invert would fail on map projections; instead, I've tried to have both Plot.tip and Plot.crosshair, but crosshair+hexbin doesn't work as I expected.  (Note that crosshair also doesn't work with map projections, independently of hexbin, but maybe that's OK.)

src/marks/tip.js
```diff
@@ -0,0 +1,237 @@
...
+      y1,
+      y2,
+      anchor,
+      monospace,
+      fontFamily = monospace ? "ui-monospace, monospace" : undefined,
+      fontSize,
+      fontStyle,
+      fontVariant,
+      fontWeight,
+      lineHeight = 1,
```
Fil, line 33: lineHeight separates fields, and this default makes the tooltip feel a bit cramped; lineHeight: 1.3 would give more breathing room, or we could separate lineHeight (for text contents) and padding (to separate fields)?

src/interactions/pointer.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+
+      function pointermove(event) {
+        if (sticky || (event.pointerType === "mouse" && event.buttons === 1)) return; // dragging
+        const [xp, yp] = pointof(event, faceted ? g : g.parentNode);
+        let ii = null;
+        let ri = maxRadius * maxRadius;
+        for (const j of index) {
+          const xj = X2 ? (X1[j] + X2[j]) / 2 : X ? X[j] : cx;
+          const yj = Y2 ? (Y1[j] + Y2[j]) / 2 : Y ? Y[j] : cy;
```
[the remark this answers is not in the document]
mbostock: I think **x1** and **x2** should have precedence over **x**, because we generally use that specificity elsewhere (e.g., the stack transform); but I agree that **px** should have precedence over **x1** and **x2** so we’ll need three levels of precedence. 😞

src/channel.js
```diff
@@ -5,8 +5,8 @@ import {registry} from "./scales/index.js";
 import {isSymbol, maybeSymbol} from "./symbol.js";
 import {maybeReduce} from "./transforms/group.js";
 
-// TODO Type coercion?
 export function createChannel(data, {scale, type, value, filter, hint}, name) {
+  if (hint === undefined && typeof value?.transform === "function") hint = value.hint;
```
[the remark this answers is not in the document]
mbostock: To be more restrictive/less ambiguous.

src/marks/tip.js
mbostock: I want to use the same default as the text mark.

src/interactions/pointer.js
```diff
@@ -0,0 +1,159 @@
...
+          g.replaceWith(r);
+        }
+        state.roots[renderIndex] = r;
+        return (g = r);
+      }
+
+      function pointermove(event) {
+        if (state.sticky || (event.pointerType === "mouse" && event.buttons === 1)) return; // dragging
+        let [xp, yp] = pointof(event);
+        if (faceted) (xp -= tx), (yp -= ty);
```
Fil, line 109:
```suggestion
        xp -= tx;
        yp -= ty;
```

Apply tx, ty translation even in the non-faceted case (the tipCell test)

mbostock: Yes, I fixed this and then broke it when staging changes. It’s fixed again.

tophtucker:
easy to add to auto 😅 https://github.com/observablehq/plot/pull/1532

https://github.com/observablehq/plot/assets/841829/e70ae515-9c2b-4f78-8c46-879ff3bec4ed

mbostock:
I would like to get this released soon. I feel the biggest blocking issue is the shorthand syntax, and more specifically, how we might drive the tip mark from an existing mark’s channels rather than duplicating a mark’s definition for the purpose of applying tips. Repeating the channel specifications definitely _works_, as the tests demonstrate, but it can be difficult to use particularly in conjunction with Plot’s many implicit behaviors (_e.g._, implicit stack transform, implicit tuples). My hope is we can somehow describe that the tip mark is derived from another mark, and that, somehow, this derived mark can “inherit” the channels from another mark (even if those channels are created by an initializer).

Related to this, the tip mark is unusual in that the channels are used to control the contents of the tip, but also affect the _appearance_ of the tip. For example, if you’re applying the tip to a mark that uses the **fill** channel, you might want to display the value in the tip; but, you probably _don’t_ want to apply that color as the fill of the tip itself. If there’s a way to tell the tip to display the channels of another mark, such that the channels that control the appearance of the tip are independent of the channels whose values are displayed in the tip, that would be ideal. So, maybe it’s not that the tip mark inherits the channels of some other mark, but instead that the tip mark has a way of _reading_ the channels of another mark (and maybe there’s a restriction that the two marks have the same data and faceting, so they’re parallel). Perhaps the _context_ that is passed to _mark_.render could expose a method for reading the channels of another mark?

I’ll experiment with this tomorrow.

mbostock:
Color swatches! 🟦 🟧 🟥

https://github.com/observablehq/plot/assets/230541/efaacbd9-daa2-4df0-a0de-0a8cd5e890fe

mbostock:
Looks good to me! LGTY, @Fil? 🚢

I suggest we followup with documentation (including the TypeScript interfaces).

mbostock:
Are you saying those are blockers? I suggest we figure those out later as there are many possible solutions and I don’t think we’ll get stuck. (The tipDotFilter test demonstrates one strategy for putting the tips last even when there are multiple tips.)

Fil:
Definitely not blockers. Let's rock :)

tannerlinsley:
Ooo la la 😍 testing asap

net:
Thank you for not binding the tooltip position to the mouse position. I've always hated the aesthetic of tooltips moving around with the mouse. No other UI tooltips do that. ❤️

yurivish:
I am unreasonably excited about these! 

Just tried them out and noticed an interesting consequence of the current rules, mentioning it in case it helps decide whether to keep this behavior or maybe change it:

image

You can cause the same data field to appear twice in the tooltip if more than one channel is driven by it. The above picture is the result of evaluating

```
Plot.dot([{ x: 1 }], {
  x: "x",
  y: "x",
  tip: true
}).plot({ width: 100, height: 100 })
```

using the tip-option branch.

[Notebook link](https://observablehq.com/d/0c340e783256f7e8)

mbostock:
Yes, @yurivish. The tip can’t currently tell which channels are redundant. The tip has access to the materialized channel’s array of values, but not the channel’s definition. Even if it had access to the definition, if in your example you wrote the definitions as `x: (d) => d.x` and `y: (d) => d.x` it would still “look” like two different definitions.

Instead of trying to detect such a case, the current plan is to give the user control over which channels appear in the tip; by default we’ll try to show (almost) everything. If the problem of redundant channels occurs more often than expected then we can try to detect the common forms, but I’d rather not add the complexity if possible.

yurivish:
Ah, yeah, that's a very good point. Makes sense – thanks for explaining!

tannerlinsley:
Any way we could get an `@next` npm tag release? This needs to get into Nozzle asap.

PR #1537, opened 2023-05-11
tip option

mbostock:
This adds a **tip** option for generating a tip mark + pointer interaction from a given mark. It can be set to _x_, _y_, or _xy_ for pointerX, pointerY, or pointer respectively, or to true in which case the default orientation is used (such as _x_ in conjunction with implicit stackY). When plotting, these marks are created implemented and drawn atop all other marks.

In the future, perhaps the **tip** option could be specified as an object allowing additional tip options such as **anchor**.

### DIFF
--- a/src/mark.d.ts
+++ b/src/mark.d.ts
@@ -271,6 +271,9 @@ export interface MarkOptions {
    */
   title?: ChannelValue;
 
+  /** Whether to generate a tooltip for this mark. */
+  tip?: boolean | "x" | "y" | "xy";
+
   /**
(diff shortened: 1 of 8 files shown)


Fil:
Love it! One thing to consider, is whether dodgeY should default to *x*—using *xy* in that case feels more natural to me, and also more usable on mobile. ymmv

mbostock:
You’re right; I overindexed on the test case. I’ve removed the one-dimensional defaults for dodge. 🙏

Fil (approved):

PR #1540, opened 2023-05-12
click-click

Fil:
When playing with http://localhost:8008/?test=tipGeoCentroid on the trackpad, I found that, after a first click that unsticks the tip, I want a second click to show the tip for the new location. Currently a pointermove is needed, and clicking did nothing.


mbostock:
I have a mild preference to not do this, because the current behavior is the same as native tooltips, and if people double-click to dismiss the tick, it’ll still work rather than bringing it back or toggling repeatedly as you click. Is it unreasonable to move the pointer if you want a new tip?

Fil:
OK. Maybe I could try with a more elaborate PR, but then the risk is that it gets too complicated.

PR #1546, opened 2023-05-13
default tip for auto

mbostock:
Supersedes #1532.


mbostock:
I don’t have a good fix for the challenges of doing server-side rendering with Vue while adding interaction during client-side hydration. The problem is the pointer interaction captures a reference to the DOM element during the initial render, but this is Vue’s _virtual_ DOM node, which isn’t what we want, and I don’t know if it’s feasible for us to maintain a mapping from the virtual DOM node to the corresponding real one after Vue renders.

But anyway, we can just defer rendering to client-side for interactive charts for now. If we can figure out a way to opt-out of Vue’s client-side hydration, then we could restore server-side rendering but replace it with the real DOM on the client.

tophtucker:
wow `tip: true` easy!!

PR #1553, opened 2023-05-15
index-aware reducer functions

mbostock:
I want to write an interactive index chart where the user specifies the date at which to normalize series. This means I can’t use the built-in _first_ normalize basis; I need to find the appropriate starting value for each series based on the given date. While I could do this outside of Plot, I’d prefer to implement this within the existing normalize transform, which means I want access to the index (`I`) within the normalize basis method. This PR allows that, maintaining backwards compatibility by checking the length of the reduce function. This applies to both the normalize and window transforms.


Fil (approved):
It's always delicate when the arguments change with the function's length. For example you you passed d3.min as a reducer you'd now have to pass `(I, v) => d3.min(I, i => v[i])` instead. But I don't see a working alternative, so I guess it's the way to go.

mbostock:
Yes, [relevant xkcd](https://xkcd.com/1172/). I don’t think we can avoid backwards incompatibility entirely but this feels reasonable to me. I would expect people to specify the _min_ (named) reducer instead of passing in d3.min directly, for example.

mbostock:
I should have added this to the map transform, too. This should use a taker instead:

https://github.com/observablehq/plot/blob/42f5d9fdac734c11bf2de5f4e2f72e5b5f0d1bc0/src/transforms/map.js#L78

PR #1556, opened 2023-05-15
shorter isoformat

mbostock:
Like isoformat, but drops -01-01 or -01 whenever implied. Related #768.


mbostock:
Closing per https://github.com/observablehq/plot/issues/768#issuecomment-1548234063.

PR #1557, opened 2023-05-15
0.6.7

mbostock:
To-do:

- [x] Separate CHANGELOG by year (because it’s so big!)
- [x] Update CHANGELOG
- [x] Document interactions feature
- [x] Document tip mark
- [x] Document **tip** mark option
- [x] Document pointer transform
- [x] Document crosshair mark
- [x] Implement TypeScript declarations
- [x] Fix crosshair in dark mode


docs/marks/tip.md
```diff
@@ -79,29 +85,66 @@ Plot.dot(olympians, {
...
+    ].map((anchor) => [
+      Plot.dot({length: 1}, {frameAnchor: anchor, fill: "blue"}),
+      Plot.tip([anchor], {frameAnchor: anchor, anchor})
+    ])
+  ]
+})
+```
+:::
+
+If you don’t specify an **anchor**, the tip mark will choose one automatically. It will prefer *top-left* if the tip fits; otherwise the it will switch sides to try to contain the tip within the plot’s frame. When dynamically rendering the tip mark, say with the [pointer interaction](../interactions/pointer.md), the tip will also attempt to use the anchor it chose previously, making the tip more stable as you move the pointer and improving readability. In some cases, it may not be possible to fit the tip within the plot’s frame; consider setting the plot’s **style** to `overflow: visible;` to prevent the tip from being truncated.
```
Fil, line 129:
```suggestion
If you don’t specify an **anchor**, the tip mark will choose one automatically. It will prefer *top-left* if the tip fits; otherwise it will switch sides to try to contain the tip within the plot’s frame. When dynamically rendering the tip mark, say with the [pointer interaction](../interactions/pointer.md), the tip will also attempt to use the anchor it chose previously, making the tip more stable as you move the pointer and improving readability. In some cases, it may not be possible to fit the tip within the plot’s frame; consider setting the plot’s **style** to `overflow: visible;` to prevent the tip from being truncated.
```

CHANGELOG.md
```diff
@@ -1,5 +1,145 @@
...
+
+Fix a bug where the position defaults for the rectX, rectY, barX, and barY marks were only applied if the *options* object was strictly undefined; now the position defaults apply if the relevant position options are undefined.
+
+Fix a bug when filtering facets and computing facet anchors with non-primitive facet domains, such as dates. Fix *z*-order across facets: each mark now draws atop all earlier marks across facets. Fix *z*-order of facet axes, which should be drawn below rather than above other marks, as other axes do.
+
+Fix the auto mark to chose the rect mark instead of rectX or rectY when appropriate, and likewise choose correctly between line, lineX, and lineY, and areaX and areaY. Also, the autoSpec method now returns the name of the explicit mark and transform implementations as **markImpl** and **transformImpl** respectively, along with the **markOptions** and **transformOptions** needed to instantiate them.
```
Fil, line 135:
```suggestion
Fix the auto mark to choose the rect mark instead of rectX or rectY when appropriate, and likewise choose correctly between line, lineX, and lineY, and areaX and areaY. Also, the autoSpec method now returns the name of the explicit mark and transform implementations as **markImpl** and **transformImpl** respectively, along with the **markOptions** and **transformOptions** needed to instantiate them.
```

docs/features/interactions.md
```diff
@@ -0,0 +1,75 @@
...
+
+onMounted(() => {
+  d3.csv("../data/athletes.csv", d3.autoType).then((data) => (olympians.value = data));
+});
+
+</script>
+
+# Interactions
+
+Interaction can allow readers to read values out of a plot (details on demand), or to fluidly change a view of data without having to edit code (zoom and filter). There are a variety of ways to achieve interaction with Plot, including built-in interaction features and development techniques with frameworks such as Observable and React.
```
Fil, line 20:
```suggestion
Interaction can allow viewers to read values out of a plot (details on demand), or to fluidly change a view of data without having to edit code (zoom and filter). There are a variety of ways to achieve interaction with Plot, including built-in interaction features and development techniques with frameworks such as Observable and React.
```

mbostock: Tweaked in 9362245a.

PR #1558, opened 2023-05-15
prefer lineX when Y is monotonic, lineY when X is monotonic, for better tooltips

### LINKED ISSUE #1547: The auto mark could be smarter and choose lineY
Currently the auto mark tends to prefer line over lineX and lineY. There is no visible difference, but there is a difference in interaction behavior if we enable the default **tip** option as in #1546: with line, we get the two-dimensional pointer, but with lineY (the common choice for _x_ = time and _y_ = value) we would get the one-dimensional pointerX. Using the isMonotonic heuristic we should be able to prefer lineY or lineX in some cases ~~and perhaps the same for areaY and areaX~~ (we already prefer areaY).

Fil:
This applies apply the same logic as for area to decide between lineX and lineY, but, when not sure, defaults to line (whereas area defaults to areaY).

closes #1547

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -91,10 +91,10 @@ export function autoSpec(data, options) {
       break;
     case "line":
       markImpl =
-        X && Y // same logic as area (see below), but default to line
-          ? yZero || isMonotonic(X)
+        (X && Y) || xReduce || yReduce // same logic as area (see below), but default to line
+          ? yZero || yReduce || (X && isMonotonic(X))
             ? lineY
-            : xZero || isMonotonic(Y)
+            : xZero || xReduce || (Y && isMonotonic(Y))
             ? lineX
             : line
           : X // 1d line by index
@@ -104,7 +104,7 @@ export function autoSpec(data, options) {
       if (isHighCardinality(C)) Z = null; // TODO only if z not set by user
       break;
     case "area":
-      markImpl = yZero ? areaY : xZero || (Y && isMonotonic(Y)) ? areaX : areaY; // favor areaY if unsure
+      markImpl = !(yZero || yReduce) && (xZero || xReduce || (Y && isMonotonic(Y))) ? areaX : areaY; // favor areaY if unsure
       colorMode = "fill";
       if (isHighCardinality(C)) Z = null; // TODO only if z not set by user
       break;
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -92,12 +92,10 @@ export function autoSpec(data, options) {
     case "line":
       markImpl =
         X && Y // same logic as area (see below), but default to line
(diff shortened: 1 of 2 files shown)


Fil:
Sorry this is not ready yet.

src/marks/auto.js
```diff
@@ -90,7 +90,18 @@ export function autoSpec(data, options) {
...
+          ? yZero
+            ? lineY
+            : xZero || isMonotonic(Y)
+            ? lineX
+            : isMonotonic(X)
+            ? lineY
+            : line
+          : X // 1d line by index
+          ? lineX
+          : lineY;
```
mbostock, line 104: If both **x** and **y** are monotonic, and neither **xZero** nor **yZero** are specified, I think we should default to lineY rather than lineX. If this works, I think I’ll push this:

```suggestion
      markImpl =
        X && Y // same logic as area (see below), but default to line
          ? yZero || isMonotonic(X)
            ? lineY
            : xZero || isMonotonic(Y)
            ? lineX
            : line
          : X // 1d line by index
          ? lineX
          : lineY;
```

mbostock (approved):
Looks good to me with the latest commit I pushed, but I can re-review if there’s something else I missed! 🙏

Fil:
I'm testing it with #1546, and it looks good for lines with two channels.

However I don't like that autoLineHistogram is using lineX (not a new thing); it's specified in the 1d line by index part, and it seems we must reverse lineX and lineY here.

src/marks/auto.js
```diff
@@ -90,7 +90,16 @@ export function autoSpec(data, options) {
...
+      markImpl =
+        X && Y // same logic as area (see below), but default to line
+          ? yZero || isMonotonic(X)
+            ? lineY
+            : xZero || isMonotonic(Y)
+            ? lineX
+            : line
+          : X // 1d line by index
+          ? lineX
+          : lineY;
```
Fil, line 102:
```suggestion
          ? lineY
          : lineX;
```

i'd also want to add this test

```
export async function autoLineHistogramY() {
  const aapl = await d3.csv<any>("data/aapl.csv", d3.autoType);
  return Plot.auto(aapl, {y: "Volume", mark: "line"}).plot({marginLeft: 80});
}
```

mbostock: I see the problem, but I don’t think that’s the right fix. I think the additional context in this example is that there’s a reducer for _y_, and so even those only _x_ is provided as input, _x_ is binned, effectively meaning that _x_ is monotonic (and _y_ is derived). So, we need to consider the reducers when determining the mark implementation, too.

mbostock: Okay, I think the latest commit fixes the problem.

mbostock (approved):

PR #1559, opened 2023-05-15
index-aware map method

mbostock:
Ref. https://github.com/observablehq/plot/pull/1553#issuecomment-1548266009


mbostock:
I added some tests for custom map methods.

Fil (approved):

PR #1561, opened 2023-05-15
quick fix for tooltips in dark mode, which should not be white on white

Fil:
see https://observablehq.com/plot/marks/auto


mbostock:
Good catch! I went with a different fix in c3798316 to match the same pattern we used for markers. I added an external stylesheet rule:

```css
:root.dark [aria-label=tip] {
  fill: var(--vp-c-bg-alt);
}
```

PR #1562, opened 2023-05-16
tip should ignore p[xy]

mbostock:
When refactoring I accidentally made the tip mark aware of the **px** and **py** channels; it should ignore those channels.

Ref. https://observablehq.com/@observablehq/geo-shape-aware-pointer


Fil (approved):
yay! it works

PR #1563, opened 2023-05-16
auto chose rect

mbostock:
The auto was always choosing rectX or rectY, but sometimes we want rect because it has more appropriate pointing behavior. Ref. https://twitter.com/angiehjort/status/1658474295788724229


mbostock:
Going to merge optimistically but @Fil please let me know if you want me to change anything! 🙏

Fil:
👍

PR #1566, opened 2023-05-16
labelArrow option

### LINKED ISSUE #510: Axis label helper?

mbostock:
Fixes #510. Supersedes #702.

Plot now applies a directional arrow such as → or ↑ automatically, even if the scale **label** is specified explicitly, as long as the given label does not already include a directional arrow. The arrow can be suppressed by setting the **labelArrow** option to *none*, null, or false. In addition, the arrow can be forced in cases where it would not appear automatically, such as for ordinal scales, by setting the **labelArrow** option to true or one of the four orientations: *up*, *right*, *bottom*, or *left*.

Previously in our repo there were 175 instances of Unicode arrows in specified labels; with this change, there are now only 14 instances! (−92%) The remaining instances are solely for double-ended arrows such as `← Women · Men →`.

### DIFF
--- a/docs/features/legends.md
+++ b/docs/features/legends.md
@@ -62,8 +62,8 @@ When an ordinal *color* scale is used redundantly with a *symbol* scale, the *sy
 ```js
 Plot.plot({
   grid: true,
-  x: {label: "Body mass (g) →"},
-  y: {label: "↑ Flipper length (mm)"},
+  x: {label: "Body mass (g)"},
+  y: {label: "Flipper length (mm)"},
   symbol: {legend: true},
   marks: [
     Plot.dot(penguins, {x: "body_mass_g", y: "flipper_length_mm", stroke: "species", symbol: "species"})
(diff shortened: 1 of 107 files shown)


mbostock:
As context, there are several benefits of removing the arrows:

1. Arrows are hard to type. (You need to copy-paste an existing one, or bring up the special characters palette.)
2. Arrows are context-dependent: they are appropriate for axes, but not for tips.
3. Arrows are orientation-dependent, making it harder to update the plot (_e.g._, to reverse a scale or swap _x_ and _y_).

Fil (approved):
This is great, thank you!
I'm adding a suggestion for docs in #1568, I also figured we might appreciate an example with an explicit labelArrow where true would not do the right thing? Like a depth transect of some sort, where you'd want depth to ↓ to the negatives.

PR #1567, opened 2023-05-17
implicit categorical color scale

### LINKED ISSUE #1523: Assume a categorical scale when scheme is categorical
Currently this errors because we assume that the *color* scale is *linear* because the input is numbers:

```
Error: unknown quantitative scheme: tableau10
```
```js
Plot.cellX(d3.range(10)).plot({color: {scheme: "Tableau10"}})
```

But if you explicitly specify a categorical color scheme, we could instead switch the inferred *color* scale to be *categorical* (*ordinal*):

untitled (74)

```js
Plot.cellX(d3.range(10)).plot({color: {type: "categorical", scheme: "Tableau10"}})
```

We already do something similar for *diverging* schemes:

untitled (75)

```js
Plot.cellX(d3.range(10)).plot({color: {scheme: "BuRd"}})
```

mbostock:
Fixes #1523.

### DIFF
--- a/src/scales.js
+++ b/src/scales.js
@@ -413,12 +413,6 @@ function inferScaleType(key, channels, {type, domain, range, scheme, pivot, proj
   // type explicitly.
   if ((domain || range || []).length > 2) return asOrdinalType(kind);
 
-  // For color scales, take a hint from the color scheme and pivot option.
-  if (kind === color) {
-    if (pivot != null || isDivergingScheme(scheme)) return "diverging";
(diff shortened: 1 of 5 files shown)


Fil:
There was a regression on the walmart raster tests, but I don't think it makes sense to force a diverging scheme onto a temporal scale (with a default pivot at unix epoch?). I'm suggesting a different priority order in #1569.

Fil (approved):

PR #1568, opened 2023-05-17
labelArrow docs

Fil:


docs/features/scales.md
```diff
@@ -949,6 +949,7 @@ Plot automatically generates [axis](../marks/axis.md) and optionally [grid](../m
 * **label** - a string to label the axis
 * **labelAnchor** - the label anchor: *top*, *right*, *bottom*, *left*, or *center*
 * **labelOffset** - the label position offset (in pixels; default depends on margins and orientation)
+* **labelArrow** - the label arrow: *left*, *right*, *top* or *bottom*; *none*, or null, or false to suppress; true to indicate the scale’s ascending order; defaults to true for quantitative and temporal scales, false for ordinal scales
```
mbostock, line 952: It’s up, right, down, left instead of top, right, bottom, left (because it’s a direction rather than a side).

mbostock (commented):
I try to make the description short enough for these options to fit on one line. If that’s not enough, I use a separate paragraph below the list to provide more details. The goal is to facilitate faster scanning of the available options.

Screenshot 2023-05-17 at 7 56 13 AM

I’ll take a pass at editing this.

docs/features/scales.md
mbostock: Also, this is missing *auto*, which is the default value.

Fil:
thanks for the edits!

PR #1569, opened 2023-05-17
implicit categorical color scale

Fil:
Complement to #1567: we do the scheme test _after_ the domain (or values) have been tested. Fixes the regressions with diverging schemes and temporal scales.

### DIFF
--- a/src/scales.js
+++ b/src/scales.js
@@ -411,13 +411,7 @@ function inferScaleType(key, channels, {type, domain, range, scheme, pivot, proj
   // If the domain or range has more than two values, assume it’s ordinal. You
   // can still use a “piecewise” (or “polylinear”) scale, but you must set the
   // type explicitly.
-  if ((domain || range || []).length > 2) return asOrdinalType(kind);
-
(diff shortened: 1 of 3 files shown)


src/scales.js
```diff
@@ -411,39 +411,37 @@ function inferScaleType(key, channels, {type, domain, range, scheme, pivot, proj
...
 
 // Positional scales default to a point scale instead of an ordinal scale.
-function asOrdinalType(kind) {
+function asOrdinalType(kind, scheme) {
   switch (kind) {
     case position:
       return "point";
     case color:
-      return ordinalImplicit;
+      return isCategoricalScheme(scheme) ? "categorical" : ordinalImplicit;
```
mbostock, line 444: There’s not an effective difference between ordinalImplicit and categorical here. The only thing ordinalImplicit gets you over categorical is that if the domain is booleans, it produces a same-ordered two-element range from the scheme.

mbostock (commented):
Oh, thanks for catching this. I overlooked this behavior. I’ll take another look shortly.

mbostock (approved):

PR #1576, opened 2023-05-17
composable render transforms

mbostock:
Rather than expecting the render transform to call mark.render, this passes a *next* function to the render transform. This would allow us to compose render transforms, potentially. (That wouldn’t be possible if the render transform is expected to call this.render internally.)


Fil (approved):
I realize that my mental model of the operations is reversed, as I see "next" as being "already there before" the new transform. But it's "called next", for sure :-)

mbostock:
Here’s a similar example using `next`: https://expressjs.com/en/guide/using-middleware.html#middleware.router

PR #1579, opened 2023-05-18
tip: respect empty lines

### LINKED ISSUE #1578: Tip text positioning doesn't always pad top correctly
Using. "@observablehq/plot-0-6-7" tip:  In chrome browser (Version 113.0.5672.93 ) on win 10

Notebook that it happens on. https://observablehq.com/@hellonearthis/looking-at-using-ai-classify-an-age-specific-english-word-di#clu

The top padding seems to not happen sometimes.

image

Fil:
closes #1578

### DIFF
--- a/src/marks/tip.js
+++ b/src/marks/tip.js
@@ -188,8 +188,8 @@ export class Tip extends Mark {
       if (j >= 0) {
         // name is truncated
         name = name.slice(0, j).trimEnd() + ellipsis;
-        value = "";
         title = value.trim();
+        value = "";
       } else {
         if (name) value = " " + value;
         const [k] = cut(value, w - widthof(name), widthof, ee);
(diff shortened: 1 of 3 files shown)


src/marks/tip.js
```diff
@@ -188,8 +188,8 @@ export class Tip extends Mark {
       if (j >= 0) {
         // name is truncated
         name = name.slice(0, j).trimEnd() + ellipsis;
-        value = "";
         title = value.trim();
+        value = "";
```
Fil, line 192: this change is for an unrelated bug, where trimming the label would not result in the title being empty instead of showing the value

src/marks/tip.js
```diff
@@ -200,8 +200,9 @@ export class Tip extends Mark {
         }
       }
       const line = selection.append("tspan").attr("x", 0).attr("dy", `${lineHeight}em`);
-      line.append("tspan").attr("font-weight", "bold").text(name);
+      if (name) line.append("tspan").attr("font-weight", "bold").text(name);
       if (value) line.append(() => document.createTextNode(value));
+      if (!name && !value) line.append(() => document.createTextNode(" "));
```
mbostock, line 205:
```suggestion
      if (value || !name) line.append(() => document.createTextNode(value || "\u200b"));
```

mbostock (changes requested):

src/marks/tip.js
Fil: zwsp on its own doesn't create a line that is "meaty" enough to inform the height in all cases—the "after" test fails (both on Chrome and Safari) if we don't use a space.

mbostock:
Argh, I was able to reproduce this, but then it suddenly stopped happening and I can’t debug. ☹️

mbostock (approved):
Screenshot 2023-05-18 at 10 26 11 AM

PR #1582, opened 2023-05-18
use dy to position multi-line text

mbostock:
Instead of setting the `y` attribute to a multiple of the line height, only set the `y` attribute on the first tspan, and then use the `dy` attribute for subsequent lines. The result should be identical, but avoids floating point error such as `y="6.851999999999999em"`.

Note that since empty lines are omitted, we have to count skipped lines (`m`). I’m not sure this ends up being any simpler, but I thought of it since we use this technique for the tip mark.


Fil (approved):
nice!

PR #1585, opened 2023-05-18
[WIP] types: data generics for text * sort order fn

tannerlinsley:
This PR is a tiny preview of the much-needed typescript generic upgrades that I would like to see come to the library. The existing types are really thorough when it comes to option discoverability and static value/option analysis, but as soon as you get into anything that relies on user types, it's the wild-wild west again.

I would *really* love to help make the generics of Plot better and have *a lot* of practice and experience in this area from my other OSS projects.

If this is interesting and something you'd be open to handing off, I would happily take it on. I am relatively certain that I would be able to do this across the board with probably few to no breaking changes to the types. So to summarize my pitch:

- Add generics everywhere possible with non-breaking defaults and constraints
- Do not touch any runtime code
- Profit

### DIFF
--- a/src/mark.d.ts
+++ b/src/mark.d.ts
@@ -63,7 +63,7 @@ export type RenderFunction = (
 export type Markish = RenderableMark | RenderFunction | Markish[] | null | undefined;
 
 /** Shared options for all marks. */
-export interface MarkOptions {
+export interface MarkOptions<T = unknown> {
   /**
    * Applies a transform to filter the mark’s index according to the given
    * channel values; only truthy values are retained. For example, to show only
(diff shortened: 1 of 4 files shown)


mbostock (commented):
Do we want T = any as the default instead of T = unknown? We’re already using T = any in a few places (e.g., Interval, ReducerFunction, MapFunction).

mbostock:
Converting to draft since marked as WIP.

tannerlinsley:
I think that depends on the default experience. I personally prefer unknown to surface the lack of types earlier and force stated intention as early as possible. What do you think?

mbostock:
I didn’t want to force people to declare generic types; I would prefer it to be opt-in for those that want stricter typing. But I can see it either way (and the TypeScript declarations themselves are opt-in, so maybe it doesn’t make sense to have it both ways).

tannerlinsley:
I’d love an example from your perspective to make sure we’re on the same page. I’ll do the same when I’m back in front of my keyboard.

PR #1588, opened 2023-05-19
show tips in generated snapshots

Fil:
dispatch fake pointermove events to have tips show up in the generated snapshots in the correct position

notes:
* it doesn't work with faceted marks, because pointer.js assumes that requestAnimationFrame is present to dedupe events; we could check for that if this unit testing seems useful.
* this happens before postrender, so still no callout box around the tips


mbostock (commented):
I’m not sure how helpful this is given that getBBox doesn’t work in JSDOM, so we can’t draw the box around the tips. It looks broken—though, I guess it at least lets us see something and check the formatting within the tip.

And I don’t like having to call pointermove explicitly within each test. That makes the tests less functional as examples and harder to copy-paste between these tests and Observable or elsewhere.

Fil:
Yes the idea was to have the interactive tip show in the snapshot unit tests, mostly to check that their position is correct. But I agree it's ugly.

PR #1591, opened 2023-05-19
descending shorthand

### LINKED ISSUE #1458: Shorthand descending sort
It’d be nice to have a shorthand for descending sort such as:

```js
Plot.barY(alphabet, {x: "letter", y: "frequency", sort: {x: "-y"}}).plot()
```

This would be equivalent to:

```js
Plot.barY(alphabet, {x: "letter", y: "frequency", sort: {x: "y", reverse: true}}).plot()
```

mbostock:
Fixes #1458. This adds *-name* shorthand to the mark **sort** option, applying to both imputed ordinal scale domains and the basic sort transform.

There are still two more places where we could support *-name* shorthand: the stack transform’s **order** option, and the bin and group transform’s **sort** option. I guess I would like to support these, too, but it felt a bit burdensome to make more changes, so I’m hoping this can stand on its own.


Fil:
It's very nice to use. I hope we can make the extra mile that will make it work everywhere there is sorting/ordering (tentatively #1606).

Fil (approved):

mbostock:
This should wait for #1607 (at the very least for better testing, but also because that PR clarifies the desired behavior).

mbostock:
Okay, I have merged with #1607, and also made the behavior more conservative: now the *-channel* shorthand only changes the default **order** option; it doesn’t attempt to invert the meaning of the given order if one is specified explicitly. I think this is easier to understand and simplifies the implementation slightly since we don’t need the reverseOrder wrapper.

It should be possible to extend this to the stack, group, and bin transforms. For the group and bin transforms, we’ll need some plumbing to change the ascendingDefined here to a dynamic comparator (descendingDefined or something custom):

https://github.com/observablehq/plot/blob/5c5acb7b6f92572c812ba1387441b179f774cbc1/src/transforms/group.js#L296-L305

Similarly for the stack transform we’ll need plumbing to override the ascendingDefined here:

https://github.com/observablehq/plot/blob/5c5acb7b6f92572c812ba1387441b179f774cbc1/src/transforms/stack.js#L334-L338

I feel this work is distracting from the release of 0.6.7 and tooltips, so I would ask that we do _not_ work on extending this PR to stack, group, and bin transform today. I would like to get back to writing the documentation for 0.6.7 and determining if there is any additional blocking functionality (such as custom formats for crosshairs #1597).

docs/features/scales.md
```diff
@@ -969,21 +969,35 @@ Plot.barY(alphabet, {x: "letter", y: "frequency", sort: {x: "y"}})
...
+Plot.barY(alphabet, {x: "letter", y: "frequency", sort: {x: "y", order: "descending"}})
+```
+
+Alternatively, the *-channel* shorthand option, which changes the default **order** to *descending*:
+
+```js
+Plot.barY(alphabet, {x: "letter", y: "frequency", sort: {x: "-y"}})
+```
+
+Setting **order** to null will disable sorting, preserving the order of the data. (When an aggregating transform is used, such as [group](../transforms/group.md) or [bin](../transforms/bin.md), note that the data may already have been sorted and thus may the order may differ from the input data.)
```
Fil, line 992:
```suggestion
Setting **order** to null will disable sorting, preserving the order of the data. (When an aggregating transform is used, such as [group](../transforms/group.md) or [bin](../transforms/bin.md), note that the data may already have been sorted and thus the order may differ from the input data.)
```

Fil (approved):
I like how small the code change is. It's mostly docs and examples!

PR #1592, opened 2023-05-19
dark-mode logo

mbostock:
Before:
Screenshot 2023-05-19 at 10 52 17 AM

After:
Screenshot 2023-05-19 at 10 52 00 AM

No changes to light mode. 😌


mbostock:
It doesn’t work in Safari. ☹️ I guess we’ll leave this as-is for now.

PR #1597, opened 2023-05-20
crosshair format option

### LINKED ISSUE #1596: Custom format for the crosshair mark?
The crosshair mark doesn’t currently provide an option to control how the values are formatted. The **text** channel in constructed using an initializer. Maybe it needs a format option, but we’d need separate options for *x* and *y*?

mbostock:
Fixes #1596. E.g., 

```js
Plot.crosshair(olympians, {x: "weight", y: "height", format: {x: ".1f", y: ".3f"}})
```

Maybe it should pick a smarter default format, similar to scale ticks (but like, assuming ~100 ticks instead of ~10, and always trimming trailing zeroes)? Maybe you should be able to specify `format: ".1f"` shorthand for both *x* and *y*?

### DIFF
--- a/src/marks/crosshair.js
+++ b/src/marks/crosshair.js
@@ -1,6 +1,8 @@
+import {maybeAutoTickFormat} from "../axes.js";
 import {getSource} from "../channel.js";
 import {pointer, pointerX, pointerY} from "../interactions/pointer.js";
 import {marks} from "../mark.js";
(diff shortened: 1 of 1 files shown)


Fil:
Good idea(s). The shorthand "single format" is not high priority. We'll probably want a similar behavior for the tip mark?

PR #1599, opened 2023-05-20
pointer[XY] for crosshair[XY]

### LINKED ISSUE #1584: The crosshair mark, but with pointerX or pointerY
Currently there’s no way to tell the crosshair mark to use pointerX or pointerY; if you use crosshairX or crosshairY, it suppresses pointing in y. This feels wrong: I think crosshairX should imply pointerX, and crosshairY should imply pointerY, but whether we show **x** or **y** should depend on whether those channels are available, not whether you call crosshairX or crosshairY.

mbostock:
Fixes #1584. See the crosshairLine test.


Fil (approved):
cool!

PR #1600, opened 2023-05-20
rule & tick markers; arrow-reverse marker

### LINKED ISSUE #1412: The rule and tick marks could support marker options.
Dunno how useful it would be, but I don’t think it would be hard to add, either.

mbostock:
Fixes #1412.


src/marker.js
```diff
@@ -31,7 +31,7 @@ function markerArrow(color, context) {
     .attr("viewBox", "-5 -5 10 10")
     .attr("markerWidth", 6.67)
     .attr("markerHeight", 6.67)
-    .attr("orient", "auto")
+    .attr("orient", "auto-start-reverse")
```
mbostock, line 34: This is especially noticeable with tick and rule (as shown in the tests); I think this is a better default for the *arrow* marker.

Fil: Agree it's a better default for rules and ticks, that indicate an extent; but for a line that indicates a sequence, it feels a bit wrong if the first arrow is pointing outwards.

mbostock: It feels less wrong than having the arrows point the same way for a rule and a line. I don’t see any practical demand for having an arrow at the start of the line; you’d typically use markerMid instead, and possible markerEnd.

The alternative is we have to introduce two types of marker, one for line (*arrow* if we retain backwards compatibility) which will almost never be used for that purpose, and another for rule and tick (*arrow-reverse*, say) which won’t be as discoverable.

mbostock: I’ve reverted the changes to the *arrow* marker and added an *arrow-reverse* marker in the latest commit.

Fil (approved):
If you wanted to have the best defaults, we could imagine two new options _arrow-auto_ and _arrow-reverse_, with _arrow_ defaulting to _arrow-reverse_ for rules and ticks, and to _arrow-auto_ for lines?

Just a suggestion, I'm OK with both approaches.

mbostock:
Interesting; I hadn’t consider mark-specific defaults for markers but that would be possible. I think this is fine for now. Perhaps it’s better to be more explicit when you want bidirectional arrows. And I don’t know if there will be much demand for markers on rules and ticks in any case.

PR #1604, opened 2023-05-21
simpler defaults for arrow markerStart orient

Fil:
hard-code these defaults, instead of exposing arrow-reverse (?)


mbostock (commented):
I think we should leave things as-is. In the parent PR I wrote:

> Perhaps it’s better to be more explicit when you want bidirectional arrows.

I also don’t think we should spend much more time thinking about this because I’m not sure there’s really demand for this feature besides. 😅

Issue #1460, opened 2023-04-17
The SortOrder interface could support a reverse option?


mbostock:
I was struggling to remember I needed to specify `order: "descending"` instead of `reverse: true`. Perhaps we could support both?

mbostock:
The use of `reverse: boolean` in ChannelDomainOptions is confusing because it doesn’t mean (exactly) to reverse the order: it means to use descending natural order instead of ascending natural order. The difference is that descending natural order still puts null/unorderable values last, whereas reversing ascending natural order would put null/unorderable values first. Therefore it would probably make more sense to add a **order** option to the ChannelDomainOptions interface, and change the meaning of the **reverse** option so that it puts nulls first (assuming that you leave **order** as the default *ascending*). We could also allow the **order** option in ChannelDomainOptions to be specified as a comparator function.

PR #1606, opened 2023-05-21
more descending shorthand

Fil:
I've struggled a bit with the types, but ~~I think the feature is complete~~ (EDIT: well, no).

### DIFF
--- a/src/transforms/basic.js
+++ b/src/transforms/basic.js
@@ -125,6 +125,10 @@ function sortValue(value) {
     channel = channel.slice(1);
     order = reverseOrder(order);
   }
+  if (typeof value === "string" && value?.startsWith("-")) {
+    value = value.slice(1);
+    order = reverseOrder(order);
+  }
   return (data, facets, channels) => {
     let V;
     if (channel === undefined) {
--- a/src/transforms/basic.js
+++ b/src/transforms/basic.js
@@ -121,6 +121,10 @@ function sortValue(value) {
         throw new Error(`invalid order: ${order}`);
     }
   }
+  if (channel?.startsWith("-")) {
+    channel = channel.slice(1);
+    order = reverseOrder(order);
+  }
   return (data, facets, channels) => {
     let V;
     if (channel === undefined) {
@@ -134,4 +138,12 @@ function sortValue(value) {
     const compareValue = (i, j) => order(V[i], V[j]);
     return {data, facets: facets.map((I) => I.slice().sort(compareValue))};
   };
+}
+
+function reverseOrder(order) {
+  return order === ascendingDefined
+    ? descendingDefined
+    : order === descendingDefined
+    ? ascendingDefined
+    : (i, j) => order(j, i);
 }
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -98,6 +98,10 @@ function binn(
   // Compute the outputs.
   outputs = maybeBinOutputs(outputs, inputs);
   reduceData = maybeBinReduce(reduceData, identity);
+  if (typeof sort === "string" && sort.startsWith("-")) {
+    sort = sort.slice(1);
+    reverse = !reverse;
+  }
   sort = sort == null ? undefined : maybeBinOutput("sort", sort, inputs);
   filter = filter == null ? undefined : maybeBinEvaluator("filter", filter, inputs);
 
--- a/src/transforms/group.d.ts
+++ b/src/transforms/group.d.ts
@@ -1,5 +1,5 @@
 import type {ChannelReducers, ChannelValue} from "../channel.js";
-import type {Reducer} from "../reducer.js";
+import type {ReducerSort, Reducer} from "../reducer.js";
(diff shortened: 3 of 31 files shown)


src/transforms/basic.js
```diff
@@ -125,6 +125,10 @@ function sortValue(value) {
     channel = channel.slice(1);
     order = reverseOrder(order);
   }
+  if (typeof value === "string" && value?.startsWith("-")) {
+    value = value.slice(1);
+    order = reverseOrder(order);
+  }
```
mbostock, line 131: If *value* is a string here, it is a field name rather than a reducer or a channel name. Because field names come from the user’s data, they are much more diverse than channel names and reducer names (which are generally fixed by Plot, though we do allow “extra” channels for the tip mark). This would break field names that start with a hyphen, which is admittedly unlikely and uncommon, but still possible. And since we aren’t going to allow `x: "-foo"` for field names in general, this is somewhat inconsistent.

I don’t think we should introduce `sort: "-foo"` shorthand for field names. There are already three ways to do this, and that seems reasonable enough:

1. `sort: "foo", reverse: true`
2. `sort: {value: "foo", order: "descending"}`
3. `sort: (d) => -d.foo` (assuming foo is numeric)

We could introduce a value helper, perhaps: `sort: Plot.negate("foo")`. This would take the same arguments that Plot.valueof takes for *value*, but would return the negative value. It would only work for numbers, though, of course…

Or perhaps we could also introduce a sort helper: `sort: Plot.descending("foo")` which just returns `{value: "foo", order: "descending"}`.

src/transforms/bin.js
```diff
@@ -98,6 +98,10 @@ function binn(
   // Compute the outputs.
   outputs = maybeBinOutputs(outputs, inputs);
   reduceData = maybeBinReduce(reduceData, identity);
+  if (typeof sort === "string" && sort.startsWith("-")) {
+    sort = sort.slice(1);
+    reverse = !reverse;
+  }
```
mbostock, line 104: A subtle gotcha here: `sort: "-x"` doesn’t mean “sort by *x* in natural ascending order and then reverse the resulting order”; it means “sort by *x* in natural descending order”. The difference is that we want nulls/unordered values to be last when using *-x*, whereas reversing after sorting would put them first.

The ChannelDomainOptions interface is confusing in this regard because it calls the option **reverse**, but it really means whether to use ascending or descending order (per the documentation):

https://github.com/observablehq/plot/blob/0490ba6f1dba3403ed050e453e9c0839ecf635a5/src/channel.d.ts#L195-L196

I commented on this indirectly in #1460, and I’m going to add some context to that issue now to clarify the confusion.

In any case, here we ideally don’t want to flip the **reverse** option; we want to use descending rather than ascending natural order. That’s why I added the reverseOrder helper for the basic sort transform:

https://github.com/observablehq/plot/blob/bef0ec98085669f0eb1a979af8bdaf534087aaad/src/transforms/basic.js#L143-L149

I think this might mean that the maybeSort helper should take an order argument, and that the bin and group transform should support `order: "descending"` and `order: "ascending"` as complements to the sort option, instead of only supporting a reverse option?

src/transforms/group.d.ts
```diff
@@ -29,7 +29,7 @@ export interface GroupOutputOptions<T = Reducer> {
    * Plot.groupX({y: "count", sort: "count"}, {fill: "sex", x: "sport"})
    * ```
    */
-  sort?: T | null;
+  sort?: ReducerSort | null;
```
mbostock, line 32: This is incorrectly narrowing the type of `T` (_e.g._, in the case that `T` is `BinReducer` instead of `Reducer`). We’ll need to find another way to type this.

src/transforms/group.js
```diff
@@ -78,6 +78,10 @@ function groupn(
   // Compute the outputs.
   outputs = maybeOutputs(outputs, inputs);
   reduceData = maybeReduce(reduceData, identity);
+  if (typeof sort === "string" && sort.startsWith("-")) {
+    sort = sort.slice(1);
+    reverse = !reverse;
+  }
```
mbostock, line 84: Same here re. reverse.

src/transforms/stack.js
```diff
@@ -81,7 +81,11 @@ function stack(x, y = one, kx, ky, {offset, order, reverse}, options) {
   const [Y2, setY2] = column(y);
   Y1.hint = Y2.hint = lengthy;
   offset = maybeOffset(offset);
-  order = maybeOrder(order, offset, ky); // TODO shorthand -order with reverse?
+  if (typeof order === "string" && order.startsWith("-")) {
+    order = order.slice(1);
+    reverse = !reverse;
+  }
```
mbostock, line 87: Same here re. reverse. And also as commented above, *order* can be a field name, and I don’t think we should support *-field* syntax since it is more likely to conflict with a user-supplied column name.

mbostock (changes requested):
Thanks for the contribution. There are a variety of subtle considerations here.

PR #1607, opened 2023-05-22
descending order

### LINKED ISSUE #1460: The SortOrder interface could support a reverse option?

mbostock:
Instead of interpreting `reverse: true` as “sort by descending natural order”, the ChannelDomainOptions interface now has an **order** option which can be *ascending*, *descending*, or a comparator function (which compares two [*name*, *value*] tuples). This better matches the SortOrder interface, and enables two new features:

1. If you set `order: null` you get input order (order of first occurrence in the data).
2. If you set `reverse: true` you get true reverse order, putting null/unorderable values first when sorting, or reverse input order when not sorting, in the same manner as the basic sort transform.

~~I’m leaving this as draft because we should probably add tests for both of these new features.~~ Added tests!

Fixes #1460.  Ref. https://github.com/observablehq/plot/pull/1606#discussion_r1199844959


test/plots/index.ts
```diff
@@ -46,6 +45,7 @@ export * from "./cars-hexbin.js";
 export * from "./cars-jitter.js";
 export * from "./cars-mpg.js";
 export * from "./cars-parcoords.js";
+export * from "./channel-domain.js";
```
mbostock, line 48: I’m following the principle here that test names should indicate which features are being tested, rather than tests being feature-agnostic examples. We can have the latter, too, but here the tests are intended to test imputed channel domains so it’s better to call them that.

src/channel.d.ts
```diff
@@ -192,7 +193,10 @@ export interface ChannelDomainOptions {
    */
   reduce?: Reducer | boolean | null;
 
-  /** If true, use descending instead of ascending order. */
+  /** How to order reduces values. */
```
Fil, line 196:
```suggestion
  /** How to order reduced values. */
```

Fil (approved):
As an aside, we can now do this if we need to use the raw grouped values for sorting (not that I'd recommend it, but…)

```
  channels: {index: (d, i) => i},
  sort: {
    fy: "index",
    reduce: (index) => index,
    order: ([group, index]) => console.warn(Array.from(index, (i) => movies[i])) // replace with a comparator on ([groupA, indexA], [groupB, indexB])
  }
```

PR #1610, opened 2023-05-22
balanced tree

### LINKED ISSUE #1589: Easier tree label positioning?
In https://observablehq.com/@observablehq/plot-tree-tidy if you want to vary the orientation of node labels (so that leaves point to the right and parent nodes to the left, as in https://observablehq.com/@d3/tree), you can do:

```
Plot.plot({
  axis: null,
  margin: 10,
  marginLeft: 40,
  marginRight: 160,
  width: 928,
  height: 1800,
  marks: [
    Plot.tree(flare, {
      path: "name",
      delimiter: ".",
      dx: -6,
      textAnchor: "end",
      fontSize: { node: (d) => (d.children ? 10 : NaN) }
    }),
    Plot.text(flare, Plot.treeNode({
      path: "name",
      delimiter: ".",
      dx: 6,
      text: "node:name",
      fontSize: { node: (d) => (d.children ? NaN : 10) }
    }))
  ]
})
```

This could be easier if dx and textAnchor were channels (#1528, #909), or maybe we need a specific option to Plot.tree.

Fil:
todo:
- [x] see how we can make it configurable (either by the user or by the layout function), see https://github.com/observablehq/plot/pull/1611#discussion_r1200753188

closes #1589

- [x] ~~what do we want to expose as *data*? I decided on an object with the "node:*" values, but that's tbd.~~ (unchanged)
- [x] ~~we'll probably need to make this configurable (opt-in or out, or smarter anyway, maybe it could be set _by the layout_), otherwise flare-indent is going sideways.~~
Capture d’écran 2023-05-22 à 17 51 40


src/transforms/tree.js
```diff
@@ -43,7 +43,14 @@ export function treeNode({
...
           treeFacet.push(++treeIndex);
-          treeData[treeIndex] = node.data;
+          treeData[treeIndex] = {
+            data: node.data,
+            name: nodeName(node),
+            path: nodePath(node),
+            internal: nodeInternal(node),
+            depth: node.depth, // nodeDepth(node)
+            height: node.height // nodeHeight(node)
+          };
```
mbostock, line 53: This breaks backwards compatibility, and I think we want to preserve the original data (which is why I originally designed it this way).

Can we not use the existing *node:internal* method because that only works for channels that are visible to the tree transform? I would think that we could pass `filter: "node:internal"` as an option to the treeNode transform? (And we could have a *node:external* or *node:leaf* for the opposite meaning.) But maybe we have to pull out the filter option and do something special; I haven’t tested yet.

mbostock (changes requested):

src/transforms/tree.js
Fil: I couldn't find how to filter inside the transform; the naive approach removes those nodes too early, even before the determination of the tree—basically crashing with a "missing root". There's certainly a better way.

mbostock: I guess the confusion is whether the filter applies before or after the tree is constructed? I guess the filter operation should run before the tree layout… 🤔 Maybe we could have a **treeFilter** option that runs after the tree layout, and supports the same *node:internal* method?

mbostock (commented):
I think I prefer `textLayout: "balanced"` instead of `textBalanced: true`, because what if we think of a different text layout strategy in the future? I think I would also prefer *mirrored* to *balanced*. What do you think?

src/marks/tree.js
```diff
@@ -27,10 +28,34 @@ export function tree(
...
     dy,
+    textAnchor,
+    textLayout,
     ...options
   } = {}
 ) {
   if (dx === undefined) dx = maybeTreeAnchor(options.treeAnchor).dx;
+  if (textAnchor !== undefined) throw new Error("textAnchor is not a configurable tree option");
+  textLayout = keyword(
+    textLayout === undefined ? (options.treeLayout === undefined ? "mirrored" : "normal") : textLayout,
```
mbostock, line 39: I think we should check whether treeLayout is Tree or Cluster here. I’ll make this change and then merge. (I was just thinking this would be nice for the D3 documentation!)

mbostock (approved):

PR #1623, opened 2023-05-23
Voronoi initializer (for pointer interactions)

### LINKED ISSUE #1622: the voronoi mark should consider the pointer as indicating which cell to show
The voronoi diagram is computed at render for the subset (grouped by Z) of the index, but with Plot.pointer what we mean is "show this cell from the (unpointed) voronoi", not "create a voronoi from this one point".

### LINKED ISSUE #1858: The voronoi transform requires facets to be exclusive
Followup to #1648; we fixed the stack transform, but not the voronoi transform.

Fil:
_new description!_

This PR chiefly addresses #1622, making the voronoi mark work with the tip option. A notable performance issue in the modified usStateCapitalsVoronoi test is that the clip path is created anew each time we rerender the mark (this is addressed separately, in #1624).

A secondary issue (#1858, making voronoi work with exclusive facets) was first addressed in an ad hoc fashion (in https://github.com/observablehq/plot/pull/1623/commits/71c49a2178bc8eb40db547f83665ac4087d3fe80). After [this review comment](https://github.com/observablehq/plot/pull/1623#pullrequestreview-1448325313) suggesting that this should be centralized, I reverted the change in https://github.com/observablehq/plot/pull/1623/commits/70f23521a4deb48ab34726398e833b540d4b9580, then used the new **exclusiveFacets** transform (introduced in #1648) in https://github.com/observablehq/plot/pull/1623/commits/972558cdee4fcfb1f04e20411664f92006a93940 et sq.

Closes #1622
Closes #1858.


src/marks/delaunay.js
```diff
@@ -221,44 +221,60 @@ class Voronoi extends Mark {
...
-      options,
+      {
+        ...options,
+        // TODO compose with existing initializer
+        initializer: function (data, facets) {
+          const D = new Array(data.length);
+          for (const f of facets) {
+            for (const i of f) D[i] = f;
+          }
+          return {data, facets, channels: {diagrams: {value: D}}};
```
mbostock, line 232: I think we want to compute the Delaunay/Voronoi here?

mbostock:
I couldn’t resist trying the Voronoi initializer idea… I dunno if we need to make similar changes to the Delaunay marks? I think maybe it doesn’t make sense there, though, because there’s a not a 1:1 correspondence between the data and the geometry.

https://github.com/observablehq/plot/assets/230541/067c54b5-801f-49cb-bea7-fabf355a0c3c

src/marks/delaunay.js
```diff
@@ -221,45 +221,45 @@ class Voronoi extends Mark {
...
+        let {x: X, y: Y, z: Z} = channels;
+        ({x: X, y: Y} = applyPosition(channels, scales, context));
+        Z = Z?.value;
+        const C = new Array(data.length);
+        const [cx, cy] = applyFrameAnchor(this, dimensions);
+        const xi = X ? (i) => X[i] : constant(cx);
+        const yi = Y ? (i) => Y[i] : constant(cy);
+        for (let facet of facets) {
+          if (X) facet = facet.filter((i) => isFinite(X[i]));
+          if (Y) facet = facet.filter((i) => isFinite(Y[i]));
```
mbostock, line 234: Should we use defined here instead of isFinite to match normal channel filtering? And also because isFinite(null) is true which is undesirable.

Fil: yes

Fil:
the voronoi cell now depends on i and fi (with a test showing facet: exclude)

mbostock (commented):
This seems _mostly_ fine… but there’s something that bothers me. I guess it’s that we need to create nested/faceted channels because we can’t assume that the facets are exclusive. #1648 And our workaround abuses how channels are intended to be represented (parallel to the data). And it’s especially funny that the Voronoi mark handles this but the stack transform doesn’t—the stack transform is far more frequently used!

I would rather fix the non-exclusive facet problem generally rather than apply the workaround in this PR. I think we probably need a redux of the approach in #1069. Ideally, it’s like a transform wrapper that says “this transform/initializer needs exclusive facets”, and it expands the faceted index and channels as needed to meet the constraint that the facets are exclusive, and the downstream code therefore doesn’t need any changing.

Fil:
It works! Applying the exclusiveFacets to the voronoi mark only for now (I saw that it didn't work with the voronoiMesh).

Fil:
In the documentation there is a voronoi mark with the centroid initializer, and the reindexation logic fails in that case.

Fil:
I'm not sure how to apply exclusiveFacets properly. At least we now have test cases.

Fil:
I've updated the description of this PR. I'd like to find a cleaner way to apply the exclusiveFacets transform, but it seems complete feature-wise.

src/marks/delaunay.js
```diff
@@ -279,9 +279,13 @@ function voronoiof(delaunay, dimensions) {
...
 }
 
-function delaunayMark(DelaunayMark, data, {x, y, ...options} = {}) {
+function delaunayMark(DelaunayMark, data, {x, y, transform, initializer, ...options} = {}) {
   [x, y] = maybeTuple(x, y);
-  return new DelaunayMark(data, {...options, x, y});
+  // Apply the exclusiveFacets transform before any other transform or initializer.
+  options = basic(options, exclusiveFacets);
+  if (transform != null) options = basic(options, transform);
+  options = {...basic(options, exclusiveFacets), initializer, x, y};
```
mbostock, line 287: Is this applying the exclusiveFacets transform twice? It should be harmless but I wonder if and why it is necessary.

Fil: reworked that part — the solution was indeed much simpler (and, trying to break it, I found another potential crash!)

src/style.js
```diff
@@ -297,6 +297,16 @@ export function* groupIndex(I, position, mark, channels) {
...
 }
 
+// TODO Accept other types of clips (paths, urls, x, y, other marks…)?
+// https://github.com/observablehq/plot/issues/181
+export function maybeClip(clip) {
+  if (clip === true) clip = "frame";
+  else if (clip === false) clip = null;
+  else if (clip != null) clip = keyword(clip, "clip", ["frame", "sphere"]);
+  return clip;
+}
```
mbostock, line 307: This doesn’t look used. Is this needed as part of this PR?

src/marks/delaunay.js
```diff
@@ -279,9 +279,9 @@ function voronoiof(delaunay, dimensions) {
   return delaunay.voronoi([marginLeft, marginTop, width - marginRight, height - marginBottom]);
 }
 
-function delaunayMark(DelaunayMark, data, {x, y, ...options} = {}) {
+function delaunayMark(DelaunayMark, data, {x, y, initializer, ...options} = {}) {
   [x, y] = maybeTuple(x, y);
-  return new DelaunayMark(data, {...options, x, y});
+  return new DelaunayMark(data, {...basic({...options, x, y}, exclusiveFacets), initializer});
```
mbostock, line 284: Is `exclusiveFacets` only needed in the `Voronoi` case, not the general case of all Delaunay marks?

src/style.js
Fil: Good catch. This was the result of a bad merge operation (this part belongs to #1624).

src/marks/delaunay.js
Fil: Yes I think so. The hand-wavy "proof" is that the voronoi cells are "more" associated to each data point, and can be separated by facet after the transform, whereas the delaunay triangles (and mesh) are tied to the whole group of points, and can't be separated post-transform (any triangle that is not joining 3 points in the same facet would be inconsistent).

mbostock (approved):

PR #1624, opened 2023-05-24
stable clip paths

Fil:
_new description!_

By using (and reusing) the same clip-path when two renders are clipped with the same shape, we get:
- smaller files
- better performance during pointer interactions

Tested manually against #1623. The only reason why I based this PR on #1623 originally is that this forms a good test case: [demo](https://observablehq.com/@observablehq/stable-clippaths-1624)); for the same reason, merging this will change the snapshots for the new tests added in #1623, but they are independent concerns.


Fil:
Marking as a draft since it needs to be rebased on #1623.

Fil:
This seems ready now.

Fil:
rebased

src/style.js
```diff
@@ -297,43 +297,62 @@ export function* groupIndex(I, position, mark, channels) {
...
+      }
+      selection = create("svg:g", context).each(function () {
+        this.appendChild(selection.node());
+        selection.node = () => this; // Note: mutation!
+      });
+      clipUrl = `url(#${clips.get("frame")})`;
       break;
     }
     case "sphere": {
+      const clips = context.clips ?? (context.clips = new Map());
```
mbostock, line 343: These probably shouldn’t be exposed publicly on the context. Maybe use a WeakMap keyed by the context instead of assigning to a new property on the context? (Also note that the `Context` interface is publicly documented in `context.d.ts`.)

mbostock: Are there only at most two keys in the clip cache (`frame` and` projection`)?

Fil: Currently yes, these are the only keys; related: https://github.com/observablehq/plot/issues/1109

src/style.js
```diff
@@ -302,43 +302,62 @@ export function* groupIndex(I, position, mark, channels) {
...
+  if (clip === true) clip = "frame";
+  else if (clip === false) clip = null;
+  else if (clip != null) clip = keyword(clip, "clip", ["frame", "sphere"]);
+  return clip;
+}
+
+function clipDefs({ownerSVGElement}) {
+  const svg = select(ownerSVGElement);
+  const defs = svg.select("defs.clip");
+  return defs.size() ? defs : svg.insert("defs", ":first-child").attr("class", "clip");
```
mbostock, line 317: Is there any benefit to a `defs` element? A `clipPath` element won’t render anything anyway and could just live at the top level. Also feels unnecessary to use the `class="clip"`.

Fil:
This was clearly over engineered. I don't think we even need a WeakMap, as we could just do a quick check through the DOM with querySelector; we just need to pass the type of clip (frame, sphere) in a class or in the id, then select against it _e.g._ with `[id^=plot-clip-sphere-]`.

mbostock:
> I don't think we even need a WeakMap, as we could just do a quick check through the DOM with querySelector

Why do you think that’s simpler or faster than a WeakMap?

mbostock (approved):

Fil:
Using the DOM feels simpler, in the sense that we wouldn't need to track the association of type of clip and clipPath in parallel, as we could just look it up. (It's hard to remember what I had in mind a year ago, though — I think I was probably trying to change the output as little as possible.) But I'm happy with this version too.

mbostock:
> Using the DOM feels simpler [information-wise], in the sense that we wouldn't need to track the association of type of clip and clipPath in parallel, as we could just look it up.

I don’t follow the difference. In the DOM approach, you have to use some (serializable) identifier in the DOM such as an id prefix (`[id^=plot-clip-sphere-]`), and then use `querySelector` to find it. In the WeakMap approach, you use a symbolic identifier (the reference to the WeakMap) and you use `map.get` to find it. The main difference I see is that the WeakMap approach is completely symbolic and hidden, whereas the DOM approach exposes its implementation, and requires pattern matching or other messiness via `querySelector` since it’s not a symbolic reference.

mbostock:
The way I think about the WeakMap is that there’s a `context.frameClip` and `context.projectionClip` property that are lazily set in `getFrameClip` and `getProjectionClip`. The only thing the WeakMap provides is that these properties aren’t publicly exposed.

Fil:
Makes sense; you're right that there is as much tracking in both cases, just differently. To be clear my comment about over engineering was about the original introduction of the parallel tracking with Map; I could have used a className instead, but indeed it would have been visible in the output.

PR #1627, opened 2023-05-24
Self-host Dr. Kirsten Gorman’s picture of sea ice

Fil:
I've reduced the image from 1.5MB to 106K

Which URL do we want:
* a) https://observablehq.com/plot/sea-ice.jpg
* b) /plot/sea-ice.jpg
* c) ../sea-ice.jpg

(a works as is in the forkable notebook, b and c work locally offline… so probably a?)


mbostock:
In the notebook we should use a file attachment so that if people fork it, it won’t break if we decide to stop using the image.

(We could use the file attachment URL from the notebook here too, if we want to avoid keeping a copy in the repo).

Fil:
it's a looooong URL :( https://static.observableusercontent.com/files/82bea9fc4b884b1dafeefa29476944ada8c68e4d54fb1dcdbf95c2378a61cf5effd497638b3b067b26eade8a5982cec71032cdacada4305982fbb0af0174ee14

mbostock (approved):

tophtucker:
I appreciate the diligence here; I didn’t love my use of imgur! Thank you! 🙏 

Our strong institutional commitment to the verisimilitude of depicted penguin habitats continues…

PR #1633, opened 2023-05-24
fix misaligned range

### LINKED ISSUE #1632: Continuous color ranges with more than two elements are handled inconsistently
Below the **domain** implicitly has two elements [0, 39] while the specified **range** has three [*red*, *green*, *blue*]. The intended behavior is that the third element *green* in the range is ignored. However, the *color* legend incorrectly shows all three colors.

Screenshot 2023-05-24 at 12 42 39 PM

```js
plot = Plot.plot({
  color: {
    legend: true,
    type: "linear",
    range: ["red", "blue", "green"]
  },
  marks: [
    Plot.cellX(d3.range(40), {fill: Plot.identity})
  ]
})
```

In addition, the exposed *color* scale incorrectly shows a three-element range:

```js
plot.scale("color").range // ["red", "blue", "green"]
```

If the domain of the same length is specified, the piecewise color scale works fine. It’s only a problem when the two don’t match.

Screenshot 2023-05-24 at 12 44 24 PM

```js
Plot.plot({
  color: {
    legend: true,
    type: "linear",
    domain: [0, 39 / 2, 39],
    range: ["red", "blue", "green"]
  },
  marks: [
    Plot.cellX(d3.range(40), {fill: Plot.identity})
  ]
})
```

mbostock:
Fixes #1632.

### DIFF
--- a/src/scales/diverging.js
+++ b/src/scales/diverging.js
@@ -8,10 +8,12 @@ import {
   scaleDivergingPow,
   scaleDivergingSymlog
 } from "d3";
-import {positive, negative} from "../defined.js";
+import {negative, positive} from "../defined.js";
+import {arrayify} from "../options.js";
+import {warn} from "../warnings.js";
+import {color, registry} from "./index.js";
+import {flip, inferDomain, interpolatePiecewise, maybeInterpolator} from "./quantitative.js";
 import {quantitativeScheme} from "./schemes.js";
-import {registry, color} from "./index.js";
-import {inferDomain, maybeInterpolator, flip, interpolatePiecewise} from "./quantitative.js";
 
 function createScaleD(
   key,
@@ -37,7 +39,10 @@ function createScaleD(
   }
 ) {
   pivot = +pivot;
+  domain = arrayify(domain);
   let [min, max] = domain;
+  if (domain.length > 2) warn(`Warning: the diverging ${key} scale domain contains extra elements.`);
+
   if (descending(min, max) < 0) ([min, max] = [max, min]), (reverse = !reverse);
   min = Math.min(min, pivot);
   max = Math.max(max, pivot);
--- a/src/scales/quantitative.js
+++ b/src/scales/quantitative.js
@@ -7,25 +7,26 @@ import {
   interpolateNumber,
   interpolateRgb,
   interpolateRound,
-  min,
   max,
   median,
+  min,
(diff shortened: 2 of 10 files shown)


Fil (commented):
I'm not sure I understand the rationale. I see three possibilities:
* We consider it as a user mistake; in that case, we should add a warning, instead of/in addition to truncating.
* It is expected that users might pass mismatching domain+range in some legitimate cases (which ones?). Then again, why truncate it: it's only the legend that would need to be fixed.
* It's a missed opportunity to define a color ramp for a quantitative color scale with a simple [min, max] domain.

mbostock:
To elaborate:

* I don’t want to change the current scale behavior (in the plot), which I consider to be correct; this PR only changes the legend. Changing the appearance of the legend to match the plot is more conservative than the reverse. I am willing to change the behavior of a plot if we think it’s a bug, but as I said, I think the scale is functioning correctly here so I don’t want to change the behavior. We favor backwards compatibility and stability.

* The exposed scale included extra (unused) elements in the domain and range. These did not affect the behavior of the scale because the scale ignores these extra elements, but I think it’s clearer to show that they are ignored by truncating the domain or range to match the scale’s behavior.

* The challenge with shorthand color ramps is: what happens if the domain has more than two elements, too? The current requirement is that the domain and range are parallel. Say we allow the range to contain an arbitrary number of elements if the domain contains two elements. What if the domain contains three elements [d1, d2, d3] and the range contains four elements [r1, r2, r3, r4]? Do we ignore the extra range element r4 in _that_ case? Or do we ignore the domain element d3 and partition [d1, d3] uniformly into three intervals [r1, r2], [r2, r3], and [r3, r4]? Do we partition [d1, d2] using [r1, r2, r3], and partition [d2, d3] using [r3, r4]? I think it’s better to either be explicit and require parallel arrays or for the user to provide an interpolator, say using d3.piecewise. We could provide shorthand syntax for d3.piecewise, but I don’t think we should do it just by allowing the range to contain extra elements; there should be a more explicit signal.

* I added a warning when we detect extra elements in the domain or range.

Fil:
Thanks for the detailed answer! I was just reading the [documentation for scales](https://observablehq.com/plot/features/scales) to clarify my comment—and I realize I can't find any mention of _polylinear_ or _piecewise_ scales.

PR #1639, opened 2023-05-25
one-dimensional arrow, line and link

### LINKED ISSUE #1541: One-dimension link and arrow
We should allow link and arrow to be one-dimensional

Fil:
TODO:
- [ ] vertical tests ?
- [ ] restore errors when both x and y are missing?

closes #1541


Fil:
It might be too disruptive for the case of lines—adding a feature at the expense of giving a good error message when people forget to specify x or y. Should we punt on that part?

PR #1641, opened 2023-05-26
fix hexbin default stroke

mbostock:
Fixes the default **stroke** option declared by the hexbin transform. Instead of declaring a default **stroke** (which was wrong in the case of the tip mark), it now declares an initially-empty **fill** channel. This allows the mark constructor to know that a **fill** channel will be populated by the initializer, and hence chose the correct default **stroke**.


Fil (approved):
love the attention to details!

PR #1644, opened 2023-05-27
more example links

Fil:
(and more to come)


mbostock (approved):

docs/features/interactions.md
```diff
@@ -70,6 +70,8 @@ With the exception of render transforms (see the [pointer transform](https://git
...
+On Observable, you can use [viewof](https://observablehq.com/@observablehq/views) in conjunction with [Observable Inputs](https://observablehq.com/@observablehq/inputs) for interactivity. If your cell references another cell, it will automatically re-run whenever the upstream cell’s value changes. For example, try dragging the slider in this [hexbin example](https://observablehq.com/@observablehq/plot-hexbin-binwidth?intent=fork). The pointer transform (and by extension any automatic tip) will set the value property of the plot to the datum of the closest point, then fire an [*input* event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event). This allows to use a plot as an input.
```
mbostock, line 73: The purpose of this paragraph is to describe how to re-render plots in response to external interaction; it does not specify how the external interaction is implemented (other than referencing inputs, which is intended as an example). I think it would be better to explain the pointer transform’s support for viewof on the pointer transform documentation rather than here.

Fil:
Note: the fork link on the first chart of https://observablehq.com/plot/interactions/pointer does not appear, because it is added to the "hidden" part.

PR #1645, opened 2023-05-27
fix: avoid double application of a scale transform 

Fil:
when using a derived mark (such as a tip), a scale transform was applied twice

### DIFF
--- a/src/channel.d.ts
+++ b/src/channel.d.ts
@@ -109,6 +109,9 @@ export interface Channel {
    */
   filter?: ((value: any) => boolean) | null;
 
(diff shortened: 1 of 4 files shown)


mbostock (approved):
Nice! This is almost exactly the solution I imagined too.

I renamed channel.transformed to channel.transform and inverted the meaning: it now means whether to apply the scale’s transform and defaults to true; if false, then the scale’s transform is bypassed. I think this is clearer if an initializer wants to declare that a channel should bypass the scale’s transform, and is similar to what we do for channel.filter. I also added it to the Channel interface.

Issue #1634, opened 2023-05-25
pointing to a raster


Fil:
For a data-based raster, the pointer will target the underlying points; it could be useful to show the computed value instead. For example:
- iris + random walk: show the color of the interpolated point under the mouse
- Mandelbrot : like crosshair, but see the value/color of the point

Fil:
Several ways of writing the contour and raster channels don't support pointer (and tip: true). I haven't checked all of the cases, but here are two categories that fail to show tips, or sometimes only show the tip in the middle, or the tip with index 0 (I'm not sure). 
1.  when x1, x2, y1, y2 are set to constant values, e.g. rasterVaporPeters
2. when the x and y channels are implicit, e.g. contourVapor and rasterVapor

Fil:
The issue is because we (ab)use the x1, x2, y1, y2 channels to pass the options of the same name—which are constants—as fake channels of length 1. We'll have to either rename these (to a1, b1, a2, b2…), or find a way to let pointer know that it must ignore them.

Fil:
The smallest patch consists of adding a flag in raster.js to say that x1,x2… are not truly channels, and then reading that flag from pointer.js. But I don't think this is a correct approach—pointer should not have to know about raster's “trick”.

```patch
diff --git a/src/marks/raster.js b/src/marks/raster.js
index fd601e3c..3ee8eac0 100644
--- a/src/marks/raster.js
+++ b/src/marks/raster.js
@@ -69,6 +69,7 @@ export class AbstractRaster extends Mark {
         y1: {value: y1 == null ? null : [y1], scale: "y", optional: true, filter: null},
         x2: {value: x2 == null ? null : [x2], scale: "x", optional: true, filter: null},
         y2: {value: y2 == null ? null : [y2], scale: "y", optional: true, filter: null},
+        fake: {value: [], filter: null},
         ...channels
       },
       options,

diff --git a/src/interactions/pointer.js b/src/interactions/pointer.js
index 72f1b2d7..f0a2d725 100644
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -23,6 +23,7 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
     render: composeRender(function (index, scales, values, dimensions, context, next) {
       const svg = context.ownerSVGElement;
       const {data} = context.getMarkState(this);
+      const {fake} = values.channels; // ignore fake channels x1,y1,x2,y2

[...]

PR #1646, opened 2023-05-27
test tip for primitive data

mbostock:
This initially surprised me, but now I think it might be desirable/acceptable: if the mark has primitive data, we consider it to be “textual”, and hence default the **title** channel to *identity*. And hence the tip displays the data rather than the channels. If the mark has primitive data, then maybe it’s unlikely that the channels would show anything else interesting? So this might be a reasonable default. But I can’t think of a way to opt-out of this behavior and show all channels, either, since the **title** option isn’t visible to the derived tip—it can only see the data and the channels…


Fil:
Yes, similar to what happens with a raster mark with data = volcano.values (or rather, when #1634 is fixed).

Fil (approved):

Issue #1629, opened 2023-05-24
Continuous crosshair?


mbostock:
The crosshair mark uses the pointer transform to snap to the data, but it’d be nice to also support a continuous mode that inverted the x and y scales instead.

Fil:
Here's a proposal
https://observablehq.com/@observablehq/plot-continuous-crosshair

Fil:
Another proposal
https://observablehq.com/@observablehq/cross-facet-crosshair

PR #1647, opened 2023-05-28
tip for dense raster

mbostock:
Ref. #1634. Crashes on raster mandelbrot, but works on raster volcano.

Screenshot 2023-05-27 at 5 03 15 PM

### DIFF
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -179,10 +179,18 @@ export function pointerY(options) {
   return pointerK(0.01, 1, options);
 }
 
-export function anchorX({x1: X1, x2: X2, x: X = X1}, cx) {
-  return X1 && X2 ? (i) => (X1[i] + X2[i]) / 2 : X ? (i) => X[i] : () => cx;
+function anchorK(k, values, c) {
+  let {[`${k}1`]: V1, [`${k}2`]: V2, [k]: V, channels} = values;
+  if (V1 && channels[`${k}1`].hint?.singleton) V1 = null;
+  if (V2 && channels[`${k}2`].hint?.singleton) V2 = null;
+  if (V === undefined) V = V1;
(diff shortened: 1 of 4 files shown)


Fil:
I've played around this a bit. Nothing conclusive at this stage but here are some notes about various issues that this raised:

1. We'd want a color swatch next to the value, when applicable (_e.g.,_ in the screenshot above, tipRasterDense). Using title seems like too much of a shortcut.

2. When experimenting with [custom spatial interpolators](https://observablehq.com/@observablehq/ais-transponders-events), it appears that —if possible— I’d prefer the color scale to be determined by the rasterized values rather than by the original data values. In that sense a raster interpolator feels like a transform.

With this in mind, we might want to compute the raster (in data space) as an initializer, like we do for voronoi in #1623. However, it might be tricky because the pointer will still want the untransformed index, X, and Y, _if they exist_.

When using a function sampler (mandelbrot), or a rectangular dataset (volcano), otoh, it would be quite slow if the transform were to create a huge index, X and Y channels, and value, then look for the closest point pixel by pixel. In that case we should directly look up the value in the derived raster.

mbostock:
The (1) issue you raise is related to what I raised yesterday in #1646: the reason we’re seeing the title tip here is because the data is an array of numbers (primitives). I also feel it would be better to see **x** and **y**, and **fill**  with a color swatch. Perhaps we can somehow opt-out of the isTextual test when deriving a tip for the raster mark… (and ideally there is a way the user can explicitly opt-out of that behavior with tip: true, too, but fixing the automatic behavior is more important.)

For (2) I don’t think it will be practical to use the pointer transform as currently implemented for dense grids; I think it could add a lot of complexity. I would prefer if we could build a specialized “rasterPointer” instead that understands dense grids and does the lookup directly, and keep these problems decoupled. I also want to build continuous crosshairs.

Fil:
related #1629

Issue #1648, opened 2023-05-28
Some transforms (and marks) require facets to be exclusive


mbostock:
The design of channels assumes that either facets are exclusive (i.e., each datum appears in at most one facet) or that if facets are not exclusive, that the derived channel value is the same across all facets. Yet this is not true in all cases, such as the stack transform in conjunction with the *exclude* **facet** mode. This problem will become more common if we extend faceting to support animation #995 with filtering to show earlier/latter values, and was also encountered when switching the Voronoi mark to compute the Voronoi diagram in an initializer #1623.

Related #1041 #1057 #1068 #1069 #1070 (but we never filed an issue for the bug).

mbostock:
From #1623:

> I think we probably need a redux of the approach in #1069. Ideally, it’s like a transform wrapper that says “this transform/initializer needs exclusive facets”, and it expands the faceted index and channels as needed to meet the constraint that the facets are exclusive, and the downstream code therefore doesn’t need any changing.

PR #1649, opened 2023-05-28
exclusiveFacets

mbostock:
Sketch of minimal solution to #1648.

This needs to pad the data with duplicates, too.

I think we might also read using non-exclusive faceted index, while writing using the exclusive faceted index. Otherwise channels supplied as arrays of values would not work. I’m not sure whether we would do anything about channels that are supplied as arrays of values that are not read by the transform; that feels like a generic issue where transforms can reindex and break the association with data, and maybe if we consider that an issue, Plot should provide a generic mechanism to reindex channel values after transforming data.

### DIFF
--- a/src/transforms/exclusiveFacets.js
+++ b/src/transforms/exclusiveFacets.js
@@ -1,0 +1,39 @@
+import {slice} from "../options.js";
+
+// TODO How to reindex channels supplied as arrays? I don’t want to inspect
+// arbitrary values on the options; maybe we could use this.channels?
+export function exclusiveFacets(data, facets) {
+  if (facets.length === 1) return {data, facets}; // only one facet; trivially exclusive
+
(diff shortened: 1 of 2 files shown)


Fil:
#1662 tries to solve the array channel reindexing issue

we also need to list any other transform or initializer that needs reindexing, and create a test for it

mbostock:
Looks good to me. 👍

Fil (approved):
Yes! And it works (almost) when merged into #1623,  by adding this:

```diff
 function delaunayMark(DelaunayMark, data, {x, y, ...options} = {}) {
   [x, y] = maybeTuple(x, y);
+  options = basic(options, (data, facets) => exclusiveFacets(data, facets));
   return new DelaunayMark(data, {...options, x, y});
 }
```

PR #1653, opened 2023-05-29
brush interaction

### LINKED ISSUE #5: Brushing.

Fil:
closes #5


src/plot.js
```diff
@@ -172,7 +172,7 @@ export function plot(options = {}) {
 
   // Allows e.g. the pointer transform to support viewof.
   context.dispatchValue = (value) => {
-    if (figure.value === value) return;
+    if (figure.value === value || selectionEquals(figure.value, value)) return;
```
mbostock, line 175: This feels overfit to the brush interaction; I would rather push this logic down into the brush interaction.

src/interactions/brush.js
```diff
@@ -0,0 +1,93 @@
...
+      const {x, y} = scales;
+      const {x: X, x1: X1, x2: X2, y: Y, y1: Y1, y2: Y2} = values;
+
+      const Xl = X1 && X2 ? X1.map((d, i) => Math.min(d, X2[i])) : X;
+      const Xm = X1 && X2 ? X1.map((d, i) => Math.max(d, X2[i])) : X;
+      const Yl = Y1 && Y2 ? Y1.map((d, i) => Math.min(d, Y2[i])) : Y;
+      const Ym = Y1 && Y2 ? Y1.map((d, i) => Math.max(d, Y2[i])) : Y;
+
+      // Use a RAF so we have access to the (facet) transform
+      window.requestAnimationFrame?.(() => {
```
mbostock, line 25: Avoid `window`. Use the same pattern the tip mark uses.

```js
if (typeof requestAnimationFrame !== "undefined") requestAnimationFrame(…);
```

src/interactions/brush.js
```diff
@@ -0,0 +1,93 @@
...
+              }
+            }
+            const filtered = brushes.map(({index}) => index).filter((d) => d);
+            context.dispatchValue(
+              filtered.length === 0 ? data : Array.from(union(...filtered), (i) => data[i]) // 🌶 todo typed array if data is numeric
+            );
+          }
+        );
+
+        this.brushes.push({brush, selection: g.append("g").call(brush.extent(extent))});
```
mbostock, line 75: This is deferring the initialization of state to the request animation frame. This may lead to unpredictable behavior. We should do as much as we can synchronously and only defer what is absolutely necessary. (I haven’t understood this code enough yet to to know why the “facet transform” isn’t available synchronously…)

src/interactions/brush.js
```diff
@@ -0,0 +1,93 @@
...
+      const Xl = X1 && X2 ? X1.map((d, i) => Math.min(d, X2[i])) : X;
+      const Xm = X1 && X2 ? X1.map((d, i) => Math.max(d, X2[i])) : X;
+      const Yl = Y1 && Y2 ? Y1.map((d, i) => Math.min(d, Y2[i])) : Y;
+      const Ym = Y1 && Y2 ? Y1.map((d, i) => Math.max(d, Y2[i])) : Y;
+
+      // Use a RAF so we have access to the (facet) transform
+      window.requestAnimationFrame?.(() => {
+        const transform = viz.getAttribute("transform"); // Facet transform
+        // Register all brushes on the mark so they can communicate
+        if (!this.brushes) this.brushes = [];
```
mbostock, line 28: This needs to use the per-plot states WeakMap approach that the pointer interaction does. If the same mark/interaction is rendered across multiple plots there should be no shared state (and no memory leak).

src/interactions/brush.js
Fil: In fact the RAF wasn't necessary, since the transform property will be added to the wrapper element which we now return. I had over complicated it.

src/interactions/brush.js
```diff
@@ -1,31 +1,40 @@
...
     render: composeRender(function (index, scales, values, dimensions, context, next) {
       const svg = context.ownerSVGElement;
       const {data} = context.getMarkState(this);
 
       // Isolate state per-brush, per-plot; if the brush is reused by multiple
       // marks, they will share the same state.
       let state = states.get(svg);
-      if (!state) states.set(svg, (state = {brushes: [], selection: null}));
-      const {brushes} = state;
+      if (!state) states.set(svg, (state = {brushes: [], bounds: new WeakMap(), selection: null}));
```
mbostock: This doesn’t need to be a WeakMap because the state itself is already weakly held.

```suggestion
      if (!state) states.set(svg, (state = {brushes: [], bounds: new Map(), selection: null}));
```

src/interactions/brush.js
```diff
@@ -0,0 +1,138 @@
...
+        wrapper.append(() => display);
+      }
+
+      // Translate the brush target to match the facet’s frame position, and
+      // initialize the plot’s value. We skip a beat so all the marks have been
+      // added before the brush (to top z-index).
+      Promise.resolve().then(() => {
+        node.setAttribute("transform", wrapper.attr("transform"));
+        svg.appendChild(node);
+        context.dispatchValue(selectionMode === "data" ? data : null);
```
mbostock, line 113: It’s not ideal to dispatch an _input_ event on mount since the user didn’t really interact with the plot. If it were synchronous, it’d be okay since the element by definition wouldn’t be attached to the DOM and thus the event wouldn’t go anywhere. But here it’s asynchronous, so it could trigger redundant computation. The chart’s value should ideally be set synchronously during render so that it’s resolved when render returns. (I think the pointer transform has this problem too, but only when the chart is faceted… there’s probably a way we could fix that.)

I also don’t like making this asynchronous as a way of drawing the brush on top. That feels like it violates our normal way of controlling z-order through the order of the **marks** option; what if there are other things that want to be “on top” in the future, too. Like the tip mark? Or some other interaction we haven’t designed yet? Perhaps we can avoid this for now by only applying the brush to the last mark, or perhaps we need to provide a separate explicit way of declaring marks that go on top.

src/interactions/brush.js
```diff
@@ -0,0 +1,138 @@
...
+              }
+            }
+            // "extent"
+            else {
+              if (selection === null) {
+                context.dispatchValue(null);
+              } else {
+                const value = {};
+                if (X) value.x = x.invert ? X.map(x.invert).sort(ascending) : X;
+                if (Y) value.y = y.invert ? Y.map(y.invert).sort(ascending) : Y;
```
mbostock, line 79: I want to think through the design here a bit more, and in particular, how we’d handle brushing over ordinal (point and band) scales. In the continuous case, I thought we might expose the selection using _x1_ and _x2_ properties, though saying _x_ = [_x1_, _x2_] also makes sense. In the ordinal case, I suppose we could use the same representation, but it would be more natural to use _x_ = [… the subset of values from the _x_ scale’s domain that are selected…].

src/interactions/brush.d.ts
```diff
@@ -0,0 +1,41 @@
...
+   * The brush’s selection mode determines the contents of the plot’s value
+   * property when the user manipulates the brush:
+   * * **data** - default; the selected data
+   * * **extent** - the selection extent, in data space
+   *
+   * The extent is an object with properties *x*: [x1, x2] for brushX, *y*: [y1,
+   * y2] for brushY, and both *x* and *y* for brush. Additionally, when
+   * faceting, it contains the facet’s *fx* and *fy* properties.
+   */
+  selectionMode?: "data" | "extent";
```
mbostock, line 15: I don’t think we’ll want this option; we should always provide both. To do this, the brush’s value could be the selection array decorated with the extent. I.e., _selection_ = […the subset of _data_ that is selected by the brush…], and _selection_.x = [_x1_, _x2_], _selection_.y = [_y1_, _y2_]. Or perhaps better, _selection_.x1, _selection_.x2, _selection_.y1, _selection_.y2, more like we do to represent the extent of bins with the bin transform. This way you always get both representations.

mbostock (commented):
This approach — brushing as a filtering render transform similar to the pointer interaction — is interesting! I hadn’t yet considered it, and I believe we never discussed it; our prior approaches to brushing were as a mark (#71 #721).

I like that this approach gives some ability to control how the selected data are displayed. For the pointer interaction, the primary motivation was to display a tip selectively for the point closest to the mouse. Here I believe the primary motivation is to alter the display so as to highlight what data are selected _vs._ not selected. For example in a scatterplot, the dots might start out as black, and then while brushing we might draw the selected points red and the unselected points light gray.

This approach doesn’t yet give us quite what we need, though: it lets us conditionally draw _selected_ data, but it doesn’t let us conditionally draw _unselected_ data, and so the selected data are typically drawn atop a “non-interactive” mark which is always visible.

This approach also requires us to specify how to display selected data, even if we don’t want anything above the intrinsic display of the brush. We would have to declare an invisible mark, say by setting the **stroke** to _none_:

```js
Plot.plot({
  marks: [
    Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm"}),
    Plot.dot(penguins, Plot.brush({x: "culmen_length_mm", y: "culmen_depth_mm", stroke: "none"}))
  ]
})
```

Whereas a brush mark would encourage this minimal display by default:

```js
Plot.plot({
  marks: [
    Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm"}),
    Plot.brush(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm"})
  ]
})
```

We don’t have to solve this problem with the brush transform, though; ideally, we have a shorthand **brush** option like we did for the tip mark and pointer interaction. In that case, we could say:

```js
Plot.plot({
  marks: [
    Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm", brush: true})
  ]
})
```

So the real question is: what is the default display when applying the brush interaction to an existing mark? And will we support altering the display of what is _not_ selected while brushing?

My ideal would be to start with an existing (“non-interactive”) mark definition, and then define “interactive overrides” (additional options) for selected and unselected data while brushing. I like the idea of doing this as a render transform — I just think we need three possible states (non-interactive, interactive and selected, interactive and unselected), and they should probably all be defined on the same mark rather than on separate marks.

If this proves too difficult, we could only offer the intrinsic visual representation of the brush, and have a brush mark instead of a brush transform. But I think we should continue exploring this direction! It seems promising!

mbostock:
I exploring the use of render transforms to override the appearance of a mark during interaction in the [mbostock/pointer-highlight branch](https://github.com/observablehq/plot/compare/mbostock/pointer-highlight).

Fil:
This branch now adopts the creator() API. I still have to figure how to type the selected and unselected options, though.

seans84:
Is this PR dead? I'd really like to be able to use this.

mishatsvelik:
@Fil Would be really nice to be able to have this working alongside tip/pointer in such a way that does not prevent them from emitting data. I.e. if a data point is clicked on via pointer and the plot chart is declared as a `viewof`, it should emit the data for that point rather than all the data (or at least configure brush to be able to be overridden in this way).

Fil:
@mishatsvelik ref. your notebook https://observablehq.com/d/6f3b6e6206c3451b.

PR #1658, opened 2023-05-30
no implicit title on derived tip

### LINKED ISSUE #1657: Derived tip when grouping primitive values shows the wrong value
Ref. https://observablehq.com/d/42b72061ec8200e4

https://github.com/observablehq/plot/assets/230541/00a7266c-fc31-44bf-973b-7cbe0f50318f

It appears to have something to do with the default *identity* **title**.

```js
Plot.plot({
  height: 80,
  x: {type: "band"},
  marks: [
    Plot.barY(chars, Plot.groupX({y: "count"}, {x: (d) => d, tip: true}))
  ]
})
```

If you change the definition to avoid this behavior, the tip shows the correct value:

```js
Plot.plot({
  height: 80,
  x: {type: "band"},
  marks: [
    Plot.barY(chars.map((d) => ({char: d})), Plot.groupX({y: "count"}, {x: "char", tip: true}))
  ]
})
```

This feels related to the discussion we were having in #1649 on channel values supplied as arrays of values, and transforms breaking that association. I’m guessing that the isTextual heuristic is wrong here (related #1646): the heuristic is being applied to the mark’s raw data, but that data is subsequently transformed (by the group transform), and so maybe the heuristic shouldn’t apply?

mbostock:
Fixes #1657.

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -492,7 +492,9 @@ function inferTips(marks) {
     const t = mark.tip;
     if (t) {
       const p = t === "x" ? pointerX : t === "y" ? pointerY : pointer;
-      tips.push(tip(mark.data, p(derive(mark)))); // TODO tip options?
+      const options = p(derive(mark)); // TODO tip options?
+      options.title = null; // prevent implicit title for primitive data
+      tips.push(tip(mark.data, options));
     }
   }
   return tips;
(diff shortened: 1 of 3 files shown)


Fil (approved):
Thanks for the quick bug fix. I'd like to explore an alternative (described in #1659), but that might be for later, though.

PR #1662, opened 2023-06-01
reindex

Fil:
This seems to be working? (ref. #1649)

demo contrasting the three branches:
https://observablehq.com/@observablehq/plot-reindex-1662


src/options.js
```diff
@@ -17,6 +17,8 @@ export function valueof(data, value, type) {
     ? map(data, constant(value), type)
     : typeof value?.transform === "function"
     ? maybeTypedArrayify(value.transform(data), type)
+    : value && data?.reindex
+    ? maybeTypedMap(data.reindex, (i) => value[i], type)
```
mbostock, line 21: This assumes that `value` is an array, but it’s technically possible to be an arbitrary iterator. (Though, I discovered there’s some confusion in isOptions.)

mbostock (approved):
Nice! 👏 

I switched to using a symbol since that seems cleaner/safer than assigning to _array_.reindex. (Who knows, maybe that’ll be a method in JavaScript in the future.) I also tweaked the logic slightly to handle channel values specified as iterables (even though there’s a latent confusion in isOptions or isObject that I’m not fixing here).

PR #1663, opened 2023-06-01
follow-up: derive x & y scale domains from geometry

Fil:
addresses TODO items in #1468 and adds/tweaks tests


src/plot.js
```diff
@@ -398,17 +398,19 @@ function inferChannelScales(channels) {
...
       const channel = channels[name];
       const {scale} = channel;
       if (scale != null && filter(scale)) {
+        // Geo marks participate in the default x and y domains.
         if (scale === "projection") {
-          // TODO only do this if there’s no projection
-          const [x, y] = getGeometryChannels(channel);
-          addScaleChannel(channelsByScale, "x", x);
-          addScaleChannel(channelsByScale, "y", y);
+          if (projection == null && x?.domain === undefined && y?.domain === undefined) {
```
mbostock, line 409: I think we’ll need a fancier test than `projection == null`, per this comment on projectionAspectRatio:

https://github.com/observablehq/plot/blob/df3a3add051049986c029fdad727bd64c929e622/src/projection.js#L225-L241

A quick take, but should have more thought:

```js
function hasProjection({projection} = {}) {
  if (projection == null) return false;
  if (typeof projection.stream === "function") return true;
  if (isObject(projection)) projection = projection.type;
  return projection != null;
}
```

But the bad part is that the projection can be specified as a function (a “projection initializer”) which can return null indicating that there isn’t actually a projection. But maybe we don’t care about that nuance here, since this is really just a performance optimization and shouldn’t change the behavior?

Fil: I don't think we even need to cover the `projection: null` and `projection: {type: null}` cases, tbh. Having a smart domain when `projection===undefined` would be enough to solve the casual use of Plot.geo with some geojson. If the user starts to specify a projection, then they know what they're doing.

Fil: Hmm it was the other way round: we can only skip that geometry look-up if the projection is guaranteed to be non-nullish. Which is what you were saying :)

mbostock (approved):

PR #1665, opened 2023-06-01
Axes and grids now support the filter option

### LINKED ISSUE #1457: The axis mark doesn’t support the filter option
It’d be nice if the **filter** option worked after the axis initializer so that you can easily filter ticks. For example here I want some ticks to use the *left* anchor, and others to use the *right* anchor:

Screenshot 2023-04-16 at 3 30 39 PM

```js
Plot.plot({
  label: null,
  x: {
    axis: "top",
    grid: true,
    label: "← decrease · Change in population, 2010–2019 (%) · increase →",
    labelAnchor: "center",
    tickFormat: "+",
    percent: true
  },
  color: {
    scheme: "PiYg",
    type: "ordinal"
  },
  marks: [
    Plot.barX(statepop, {y: "State", x: (d) => (d[2019] - d[2010]) / d[2010], fill: (d) => Math.sign(d[2019] - d[2010]), sort: {y: "x"}}),
    Plot.axisY(statepop.filter((d) => d[2019] >= d[2010]).map((d) => d.State), {x: 0}),
    Plot.axisY(statepop.filter((d) => d[2019] < d[2010]).map((d) => d.State), {x: 0, anchor: "right"}),
    Plot.gridX({stroke: "var(--vp-c-bg)", strokeOpacity: 0.5}),
    Plot.ruleX([0])
  ]
})
```

I can do it by passing the tick values to the axis mark constructor, but I think using the **filter** option would be more intuitive. The reason that it crashes right now is that the default tick values aren’t known until the axis mark’s initializer runs (after all transforms). So, the axis mark would probably need to pull out the **filter** option and run it after the built-in initializer.

### LINKED ISSUE #1655: The reverse option crashes explicit axis
Just saw that as I was making a facet axis explicit; of course I should have left the **reverse** option in the _fy_ specification, but copied it over and got a `TypeError: Cannot read properties of undefined (reading 'map')`.

minimal repro:
>  Plot.axisX({ reverse: true }).plot()

Fil:
The data for the axis and grid marks (the associated scale’s ticks) is computed (in the general case) after the scales have been initialized. As a consequence,

1. These marks did not support the **filter** option.
2. The **sort** and **reverse** option crashed as the mark’s data is undefined.

In this PR, by moving the filter logic to their initializer function, we allow the user to filter the axis ticks; for example in usStatePopulationDiverging, to draw ticks on the right only for states with a population decrease, and on the left only for states with a population increase. We also throw a proper error message when the **sort** and **reverse** options are specified.

closes #1457
closes #1655


Fil:
Got side tracked by a typescript issue: I want to remove the first and last ticks of an x axis (as in [this D3 horizon chart](https://observablehq.com/@d3/horizon-chart)), and for this I use the third argument *ticks* that the filter function is getting:

`Plot.axisX({anchor: "top", filter: (tick, index, ticks) => index > 0 && index < ticks.length - 1})`

It works; however it's not documented: the function is typed as `((d: any, i: number) => any) // function of data`

Fil:
the secondary remark is discussed in https://github.com/observablehq/plot/pull/1670 and is not blocking for this PR.

Fil:
rebased

src/marks/axis.js
```diff
@@ -505,56 +505,61 @@ function labelOptions(
...
+      })
+    );
+    if (initializeFacets) facets = context.filterFacets(data, initializedChannels);
+    return {data, facets, channels: initializedChannels};
+  });
+
+  // Apply the filter option after the initializer has determined the mark’s data.
+  if (f != null) i = filter(f, i);
+  if (s != null) throw new Error("Unsupported axis option: sort");
+  if (r != null) throw new Error("Unsupported axis option: reverse");
```
mbostock, line 560: Match the style of our other errors:

```suggestion
  if (s != null) throw new Error("unsupported axis option: sort");
  if (r != null) throw new Error("unsupported axis option: reverse");
```

mbostock (commented):
Thanks for the context. It also helped to hide the whitespace in the diff.

Could we also support **sort** and **reverse** by applying them after the axis initializer, like we do for filter? Couldn’t we compose them using the same strategy that the `initializer` does, just after the axis initializer instead of before? Something like this…

```
i.initializer = composeInitializer(i.initializer, initializer({filter, sort, reverse}).initializer);
```

src/marks/axis.js
Fil: The version with composeInitializer works (including sort+reverse!); I'm still trying to make the second version work.

Fil: (Couldn't make the second version work.)

mbostock: Yep, I got the order reversed in my suggestion. I’ll take one more look but I think the composeInitializer approach is fine too! Thanks.

test/plots/axis-labels.ts
```diff
@@ -65,3 +65,20 @@ export async function axisLabelHref() {
     marks: [Plot.axisX({label: "Letter", href: (d) => `https://en.wikipedia.org/wiki/${d}`})]
   });
 }
+
+export async function axesFilter() {
```
mbostock, line 69: This test doesn’t really belong here. I’ll move to a more appropriate location.

mbostock (approved):

src/marks/axis.js
mbostock: Got it working. One call to construct the “basic” initializer (which could even be a custom **initializer**!) and a second to compose that with the “axis” initializer.

Fil: much better! this consistency will hopefully unlock more things in the future (such as zooming)

PR #1668, opened 2023-06-02
Refine “high cardinality” warning

### LINKED ISSUE #1667: spurious "high cardinality" warning
We're sometimes seeing "high cardinality" warnings when in fact the cardinality is not so high.
Here's a slightly contrived example (simplified down from a real-life use case):

```
Plot.plot({
  height: 200,
  marks: [
    Plot.frame(),
    Plot.lineY(aapl.slice(0, 9), {
      x: "Date",
      y: "Close",
      stroke: (d) => d.Date.getUTCDay(),
      fy: (d) => d.Date.getUTCDay(),
      marker: true
    })
  ]
})
```

Here we have 5 facets, four of which have 2 data points, and the last one has 1—which triggers the warning.

we should probably check that G.size is greater than, say, 10, before warning:
https://github.com/observablehq/plot/blob/df3a3add051049986c029fdad727bd64c929e622/src/style.js#L255-257

“high cardinality” is also tested in Plot.auto, I think.

Fil:
The intent of this “high cardinality” warning (as specified [here](https://github.com/observablehq/plot/pull/761#issuecomment-1043121845)) is the following: when a mark only¹ makes sense with grouping—currently area, line, linearRegression and density—, we want to warn when the number of groups (G.size) is higher than the smallest possible number of _meaningful_ groups, which is I.length >> 1 (obtained when the groups are pairs).

However when there is an odd number of elements, having 1 element that falls out is to be expected and should not invalidate the chart.

Closes #1667

¹ It's not a hard rule: the areaY mark with a stroke, the lineY mark with a marker, and the density mark tolerate single points. But these are not the default usage.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -252,7 +252,7 @@ function groupAesthetics(
 
 export function groupZ(I, Z, z) {
   const G = group(I, (i) => Z[i]);
-  if (z === undefined && G.size > I.length >> 1) {
+  if (z === undefined && G.size > (1 + I.length) >> 1) {
     warn(
       `Warning: the implicit z channel has high cardinality. This may occur when the fill or stroke channel is associated with quantitative data rather than ordinal or categorical data. You can suppress this warning by setting the z option explicitly; if this data represents a single series, set z to null.`
     );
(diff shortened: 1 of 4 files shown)


mbostock (approved):
Nice!

PR #1669, opened 2023-06-02
normalize tickFormat's arguments

### LINKED ISSUE #1640: `tickFormat` type error
` tickFormat?: string | ((t: any, i: number) => any) | null; `

should it be ` tickFormat?: string | ((t: any, i: number, D:any) => any) | null; `

Fil:
closes #1640

### DIFF
--- a/src/axes.js
+++ b/src/axes.js
@@ -17,6 +17,17 @@ export function maybeAutoTickFormat(tickFormat, domain) {
       ? formatIsoDate
       : string
     : typeof tickFormat === "function"
-    ? tickFormat
+    ? tickFormatWrap(tickFormat)
     : (typeof tickFormat === "string" ? (isTemporal(domain) ? utcFormat : format) : constant)(tickFormat);
+}
+
+// D3 axes typically call the tickFormat function with the tick value, index,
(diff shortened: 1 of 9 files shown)


src/axes.js
```diff
@@ -17,6 +17,17 @@ export function maybeAutoTickFormat(tickFormat, domain) {
...
 }
+
+// D3 axes typically call the tickFormat function with the tick value, index,
+// and array of nodes; for Plot, the third argument is the array of tick values.
+export function tickFormatWrap(tickFormat) {
+  return (t, i, nodes) =>
+    tickFormat(
+      t,
+      i,
+      nodes.map((d) => d.__data__)
```
mbostock, line 31: This is quite inefficient because it has to reconstruct the array of data each time the tick format function is called.

Do we really need this? Perhaps we can just close the issue and say the last argument is not documented and may vary.

Fil: I don't know. Pro: the examples show that the feature is (a bit) useful, and the tighter code is both more predictable and avoids a leaky abstraction. Con: the double loop (on a small number of values, though), and the additional thing that will need to be maintained.

Fil:
closing due to lack of interest

Issue #1621, opened 2023-05-23
Pointing at coincident points


mbostock:
Since the pointer transform only focuses one point at a time, if points are coincident (or nearly so), some points may not be focusable. Should the pointer transform allow focusing multiple points simultaneously, and then the tip mark could handle cycling between points, or showing multiple tips with collision detection? Or should the pointer transform support an interactive method of cycling through nearby points, say by clicking?

mbostock:
Yep, that’s what I meant by “allow focusing multiple points simultaneously”. 👍

net:
I think the mark should detect collisions and try to gracefully render colliding tips without further interaction, that way the solution is also applicable to non-interactive plots with static tips (such as the US states example in the docs).

mbostock:
Yes, @net, perhaps we should have another feature request for that. Though even so that will only scale to at best ~4 tips, so we’ll still need some way to disambiguate pointing at coincident/dense points.

net:
@mbostock if resolving the collisions through positioning fails, the mark could fall back on merging the tips into one, positioning the point of the tip at the average of the collided tips, and rendering the merged tip contents in one tip. Of course, that only scales so far too, but it would fix the US states example 🙂

```
+-------------+
| weight 54   |
| height 1.63 |
+-------------+
| weight 55   |
| height 1.63 |
+-   ---------+
  |/

korvin89:
Hi @mbostock! First of all, thank you for your great work! :)
We are looking forward to this feature. We want to replace the current library for drawing graphs with Observable Plot, but we lack the functionality that is discussed in this issue.

What do you think about adding access to all points data with the same values of the corresponding axis via pointer? It will help to render one tip for all data(points) represented by the same value.

<details>
<summary>Here is our user case</summary>

![image](https://github.com/observablehq/plot/assets/26501073/900b15bf-805d-44d7-bc97-138aa0737090)
</details>

Fil:
See https://github.com/observablehq/plot/pull/1671

korvin89:
> See #1671

Great! Is it possible to stack tips into one?
[Fork of your example](https://observablehq.com/d/8f73b3deba37213b)

![image](https://github.com/observablehq/plot/assets/26501073/daf014a9-a212-4a9f-88ba-4b184866bcc8)

Fil:
[...]

PR #1671, opened 2023-06-06
z-aware pointer

### LINKED ISSUE #1574: Pointer, but that filters everything with a matching z value
For hovering a multi-series line chart, we don’t want to just highlight the current point; we want the highlight the entire line. I expect it’d be pretty easy to extend the current pointer interaction, and instead of just selecting a single index *i*, it filters the index to everything that has the same **z** value. Then you could apply this “pointerZ” interaction (name TBD) to a line mark, such that the line closest to the pointer is drawn in a different color.

(We probably also want an “exclude” mode for the pointer interaction to draw things differently when not pointed.)

Fil:
Do we always follow an explicit *z*; do lines etc set z with stroke or fill; is this something we want to opt-in more explicitely (e.g. zPointer: true)?

https://github.com/observablehq/plot/assets/7001/6dafd782-6c06-4304-9eb2-2178af847531

closes #1574 

maybe also part of a solution for #1621?


Fil:
Note: in order to minimize the diff, I did not wrap the call to initializer(options, ƒ). To do before merging.

Fil:
rebased

mootari:
It would be great to have this in Plot! 🙏🏻 What remaining issues need to be resolved?

Fil:
Here's a simpler approach:
https://observablehq.observablehq.cloud/pangea/plot/multiple-line-chart-hover

PR #1688, opened 2023-06-12
add docs-test

Fil:
from https://github.com/d3/d3/pull/3673

fix a few malformed links; the automatic renumbering of duplicate links with -1 seems fragile, so we make them explicit rather than making the link checker smarter.


mbostock:
I think it might be an leftover from when you crawled the generated HTML, because there is an A tag here for downloading the pre-built UMD bundles:

https://github.com/observablehq/plot/blob/2d4a2f63283b817c6f1aa75b2715d006a68f68e9/docs/getting-started.md?plain=1#L85-L86

But the new code only looks for Markdown links, so I think we don’t need this exemption anymore. (And I think it’s okay that you can opt-out of link checking by using an A element instead of a Markdown link.)

mbostock (approved):
This reminds me that we adopted a more explicit system for naming anchors in D3 (e.g., the anchor for Plot.scale would be `#scale` instead of the default `#scale-options` or disambiguated `#scale-options-1`). But I think we could adopt that naming convention in a subsequent PR, and when we do, it’ll be great to have this test in place to make sure we don’t miss anything!

mbostock:
Oh, *I* see now. It’s just crawling the files and breaking on the symlink to dist/plot.umd.js because it hasn’t been built. Yeah. There’s no need to stat .js files. I’ll revert the last commit.

PR #1692, opened 2023-06-14
fixes an alignment issue for radio buttons

Fil:
the issue was affecting only safari, e.g. on /plot/transforms/window

thanks to @mootari for the fix

before

after


mbostock (approved):
Thanks. Should copy this over to D3, too.

PR #1693, opened 2023-06-14
Observable footer

Fil:
ported from https://github.com/d3/d3/commit/1582f4afb151f9705ebd43b7839456c841af3319

changes:
- copyright: Observable, Inc.
- releases URL
- license URL

footer


mbostock (approved):
Looks good. I wonder if we should link to D3 in the footer, too?

Fil:
I think it's fine with the link in the header, but we can revisit

PR #1697, opened 2023-06-16
pointer-events: none when not sticky

mbostock:
The pointer transform now applies pointer-events: none to the downstream mark when the pointer is not sticky.


src/interactions/pointer.js
```diff
@@ -97,11 +98,12 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
...
 
       function render(ii) {
-        if (i === ii) return; // the tooltip hasn’t moved
+        if (i === ii && s === state.sticky) return; // the tooltip hasn’t moved
         i = ii;
+        s = state.sticky;
         const I = i == null ? [] : [i];
         if (faceted) (I.fx = index.fx), (I.fy = index.fy), (I.fi = index.fi);
-        const r = next(I, scales, values, dimensions, context);
+        const r = next(I, scales, values, dimensions, {...context, pointerSticky: s});
```
mbostock, line 106: I want to change this to not create a new context object on each render, but to create it once above, and then update the value of pointerSticky as it changes.

PR #1698, opened 2023-06-17
documentation tree

Fil:
from the d3 website: https://github.com/d3/d3/commit/fbf24bb271759bb8297722377fb767d5aa12d78a and https://github.com/d3/d3/commit/9f88f025e8fc5903b70ec80a105606508056d0bb

Capture d’écran 2023-06-17 à 12 11 05

### DIFF
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -31,6 +31,25 @@
   z-index: 1;
 }
 
+
+:root [aria-label="tip"][fill="white"],
+:root [aria-label="tip"] [fill="white"] {
+  fill: var(--vp-c-bg-alt);
+}
+
+:root.dark marker[stroke="white"],
+:root.dark [aria-label^="crosshair"][stroke="white"] {
+  stroke: var(--vp-c-bg);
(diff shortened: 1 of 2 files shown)


docs/.vitepress/theme/custom.css
```diff
@@ -31,6 +31,25 @@
...
+}
+
+:root.dark marker[stroke="white"],
+:root.dark [aria-label^="crosshair"][stroke="white"] {
+  stroke: var(--vp-c-bg);
+}
+
+:root.dark marker[fill="white"] {
+  fill: var(--vp-c-bg);
+}
```
mbostock, line 47: These styles already exist here.

```suggestion
```

mbostock (commented):
I’m not sure if this is worth the attention. I thought it was necessary in the D3 case because the sidebar is more deeply hierarchical, but here are there are only four sections, so you can already see fairly easy what’s available?

Fil:
It's such a nice overview! When I saw it in D3, I felt it was missing in Plot.

mbostock (approved):

PR #1699, opened 2023-06-17
null filter on the href channel

### LINKED ISSUE #1696: The href channel shouldn’t filter nullish values by default?
It seems overly strict that the **href** channel causes data to be filtered if the provided values are null or undefined; instead, we should just not link those elements.

You can override the filter using the **channels** option as a workaround:

```js
channels: {href: {value: "link", filter: null}}
```

Fil:
closes #1696

I wonder if we should not extend this to any channel that “enhances” the mark but is not a visual channel. That is, title and **ariaLabel**? (The “use case” of using them to filter marks out seems far-fetched.)


mbostock (approved):
I looked at doing title before, and I wouldn’t change title’s filter; the olympians scatterplot with info tip is a nice example of where you _do_ want to filter by the title channel.

I think we can always change the ariaLabel filter in the future if we can think of a use case. Let’s start here.

Fil:
it seems to work in https://github.com/observablehq/plot/pull/1704!

mbostock (approved):

PR #1701, opened 2023-06-18
fix barycentric extrapolation

### LINKED ISSUE #1700: barycentric extrapolation should not override interpolation
I have made a mistake when designing the “extrapolation” part of the barycentric interpolator. Adding points to the frame can create huge distortions when the frame doesn't include all data points—for example when we zoom in on a region.

If we think that points that live inside the convex hull should be governed by the normal triangulation (as I do now), we have to rethink how extrapolation works.

Fil:
closes #1700

Some of the tests look worse, but it's an expected artifact of the delaunay triangulation.


Fil:
Wait… It's still incorrect, since it doesn't check if the point is on the exterior side of the segment.

Fil:
Double-checked and optimized: running the whole hull for each pixel was quite slow.

Fil:
I've now fixed the performance issue in #1705. Leaving both PRs open, but I think I prefer the exact algorithm.

src/marks/raster.js
```diff
@@ -282,30 +282,14 @@ export function interpolateNone(index, width, height, X, Y, V) {
...
-
     // Interpolate the interior of all triangles with barycentric coordinates
-    const {points, triangles} = new Delaunay(XY);
+    const {points, triangles, hull} = Delaunay.from(
+      index,
+      (i) => X[i],
+      (i) => Y[i]
+    );
     const W = new V.constructor(width * height);
+    const I = new Uint8Array(width * height);
```
mbostock, line 292: Having both `index` and `I` here is hard to follow, since both are indexes. Maybe `I` should be renamed to `delaunayIndex` or something? I’m not entirely sure that the right index is being used in the right place… I wonder if we are assuming that `index` is [0, 1, 2, 3, …]. Worth double-checking if you haven’t already.

mbostock:
The variant in #1705 fixes a few artifacts on the edge, so that looks like a good addition to me. I haven’t grokked the algorithm yet, but I enjoyed the visual explanation of the “distance to segment” and “ray out of a convex hull” notebooks!

src/marks/raster.js
Fil: I've renamed I into S and added a comment.

The values are extracted here from the indexed data—I've added a test to triple-check.
```
  X = Float64Array.from(hull, (i) => X[index[i]]);
  Y = Float64Array.from(hull, (i) => Y[index[i]]);
  V = Array.from(hull, (i) => V[index[i]]);
```

src/marks/raster.js
```diff
@@ -337,14 +322,91 @@ export function interpolatorBarycentric({random = randomLcg(42)} = {}) {
...
         }
       }
     }
+    extrapolateBarycentric(W, S, X, Y, V, width, height, hull, index, mix);
     return W;
   };
 }
 
+// Extrapolate by finding the closest point on the hull.
+function extrapolateBarycentric(W, I, X, Y, V, width, height, hull, index, mix) {
```
mbostock, line 337: You’re still using `I` here instead of `S` (the earlier confusion I mentioned).

```suggestion
function extrapolateBarycentric(W, S, X, Y, V, width, height, hull, index, mix) {
```

mbostock (commented):
Looks good other than the I/S comment.

mbostock (approved):

PR #1704, opened 2023-06-19
the title channel is not filtering the mark, unless it's a tip mark

Fil:
(any mark that depends on this channel must declare it)

ref. https://github.com/observablehq/plot/pull/1699#issuecomment-1596884532


src/marks/tip.js
```diff
@@ -56,7 +57,8 @@ export class Tip extends Mark {
         x1: {value: x1, scale: "x", optional: x2 == null},
         y1: {value: y1, scale: "y", optional: y2 == null},
         x2: {value: x2, scale: "x", optional: x1 == null},
-        y2: {value: y2, scale: "y", optional: y1 == null}
+        y2: {value: y2, scale: "y", optional: y1 == null},
+        title: {value: title, optional: true} // filter: defined
```
mbostock, line 61: So, just for posterity: this works because the channels that are defined here override the default channels that are declared by styles. And that’s specifically because styles is listed first in the spread here:

https://github.com/observablehq/plot/blob/5e313d9a3c4bb8158bf5fa4d1286386282c1faa3/src/mark.js#L44

That’s fine. It’s a little subtle, but it makes sense that the mark-specific channel definitions would override the defaults.

mbostock (approved):

PR #1705, opened 2023-06-19
fix barycentric extrapolation variant

Fil:


src/marks/raster.js
```diff
@@ -383,6 +373,42 @@ function segmentProject(x1, y1, x2, y2, x, y) {
...
+  const hbd = Math.hypot(dxbd, dybd);
+  return (x, y) => {
+    const dxa = x - xa;
+    const dya = y - ya;
+    const dxb = x - xb;
+    const dyb = y - yb;
+    return (
+      cross(dxa, dya, dxb, dyb) >= 0 &&
+      cross(dxa, dya, dxab, dyab) * hca >= cross(dxa, dya, dxca, dyca) * hab &&
+      cross(dxb, dyb, dxab, dyab) * hbd >= cross(dxb, dyb, dxbd, dybd) * hab
```
Fil, line 407: maybe compare `> -0.01` to avoid precision issues that would make a point not part of the polygon on its left and not part of the polygon on its right?

src/marks/raster.js
```diff
@@ -340,34 +340,24 @@ function extrapolateBarycentric(W, I, X, Y, V, width, height, hull, index, mix)
...
-    for (let d = 0; d < 1; d += s) {
-      points.push(xa + d * dx, ya + d * dy);
-      refs.push(j);
-    }
-  }
-  const delaunay = new Delaunay(points);
-  let iy, ix;
+  const n = X.length;
+  const rays = Array.from({length: n}, (_, j) => ray(j, X, Y));
+  let k = 0;
```
Fil, line 345: Maybe use the I array to note the value of `k` and retrieve it sparsely when starting a new line or emerging through the hull.

mbostock (approved):

src/marks/raster.js
Fil: I've tried this and it's not making a measurable difference in speed , so dropped it.

PR #1713, opened 2023-06-22
fix non-faceted pointer

### LINKED ISSUE #1677: The pointer transform doesn’t work if non-faceted in a faceted plot
E.g., here I would expect the pointer to highlight the same rule _across_ facets. But instead the pointer’d rule is not displayed at all.

untitled (16)

```js
Plot.plot({
  marks: [
    Plot.lineY(aapl, {x: "Date", y: "Close", fy: d => (d.Close % 2) === 0}),
    Plot.ruleX(aapl, {x: "Date", strokeOpacity: 0.1}),
    Plot.ruleX(aapl, Plot.pointerX({x: "Date", _fy: d => (d.Close % 2) === 0, stroke: "red"})),
  ]
})
```

https://observablehq.com/d/82e93a91379bd230

mbostock:
Fixes #1677.

The code was confusing the plot being faceted vs. the current mark being faceted. If the plot is faceted, but the current mark is _not_ faceted, we still need to populate the _index_.fx, _index_.fy, and _index_.fi fields so that the mark knows which facet is being rendered. Since the _index_ is reused across facets, we create a defensive copy so that these fields aren’t mutated after the mark renders (_e.g._, in the case of an asynchronous render transform). We use _index_.subarray if the index is a typed array, which should be faster than create a copy using _index_.slice.


Fil (approved):
nice!

PR #1716, opened 2023-06-23
Remove reference to penguins in group transform docs

nachocab:


Fil (approved):
thank you!

mbostock (approved):
Thank you!

Issue #1101, opened 2022-10-25
Axis label position: Allow customizing vertical position of axis label for Y axis


nobarany:
The distance between the axis label and the chart is customizable in one direction. It'd be great to be able to expose a setting for the other direction (or for styling it in general) to allow more flexibility.
 ```
.attr("dy", labelAnchor === "center"
                ? axis === "right"
                    ? "-0.32em"
                    : "0.75em"
                : labelAnchor === "bottom"
                    ? "1.4em"
                    : "-1em")
```

The last condition ("-1em") seems too tight for my chart:
<img width="84" alt="Screen Shot 2022-10-25 at 10 53 19 AM" src="https://user-images.githubusercontent.com/926082/197846665-881b3be9-5611-4d9f-9e8a-27faeb171402.png">

**Comment by Fil** [comment]
Can you share your chart?

**Comment by nobarany** [comment]
Sure:
```
Plot.plot({
    width,
    y: {
      grid: true,
      zero: true,
      label: 'Cumulative spending',
      tickFormat: (d: number) => `${formatCurrencySimple(d)}`,
      nice: true,
    },
    x: { tickRotate: -45, type: 'utc' },
    marks: [
      Plot.lineY(
        budget,
        Plot.map(
          { y: 'cumsum' },
          {
            x: xProp,
            y: yProp,
            sort: xProp,
            curve: 'step',
            stroke: 'gray',
          }
        )
      ),
    ],
    marginTop: 40,
    marginBottom: 60,
    marginLeft: 80,
  });
```

llimllib:
I'd also like this; here's a chart I generated where the X axis label is unpleasantely close to the right margin:

![plot(5)](https://user-images.githubusercontent.com/7150/198727617-617b9fae-f263-473c-ab40-ec236b145119.png)

(Probably axis titles should respect the graph margin? But I could also fix it manually if I could tweak its horizontal offset)

mbostock:
Related #375.

PR #1722, opened 2023-06-25
auto margins

### LINKED ISSUE #1451: A larger default marginLeft for y axes

Fil:
Alternative to #1714.

TODO:
- [ ] facet axes?
- [ ] find a “good” heuristic (consider the scales’ domain, ticks, tickFormat…)
- [ ] consider labelAnchor ("center" needs more space)
- [ ] use approx. text metrics instead of string.length

closes #1451


mbostock:
Can you regenerate the snapshots so we can more easily review the impact of this change? Thank you. 🙏

Fil:
Here's a playground to develop tests: https://observablehq.com/@observablehq/auto-margins-1722

test/output/beckerBarley.svg
mbostock, line 1: Is this a regression? I think it’s a little nicer if the y-axis label feels more attached to the y axis. Though, when the left margin is small, it’s nice to have the y-axis label all the way in the corner. In other words, the approach in this PR is more consistent, but I’m not sure it’s better. I’d be curious to see two alternatives: (1) aligning the y-axis label with the start or end of the range rather than bleeding into the margins, and (2) limiting the amount the y-axis label can bleed into the margins to, say, 30 or 40px or whatever feels right.

test/output/usPopulationStateAgeGrouped.svg
mbostock, line 1: This one is interesting; presumably we are ignoring the use of the `s` tick format, and hence think that we need a much bigger left margin than we actually do?

test/output/usPresidentialElection2020.svg
mbostock, line 1: Same here — a log y scale probably doesn’t need as much padding since the ticks using the s format.

mbostock (commented):
I think we’ll need to account for the **tickFormat** option when computing the automatic margins (even if it’s a custom function). It should be possible to compute the formatted tick strings, and then use their text metrics to more accurately infer the needed margins? Rather than assuming that numbers/dates/etc. are formatted a specific way?

test/output/beckerBarley.svg
Fil: Good catch, not sure what happens in this case; maybe related to #375 #1101 #1414?

Fil:
Computing all the ticks is probably more adequate, but it is also quite a bit more complicated, since the ticks are computed from the instantiated scales. I've tried to approximate it by doing half-way measures (like, apply tickFormat if given… but on what ticks?) and it seemed worse. My conclusion was that either we have a light system (this one, to which we could add a few common cases such as type: "log"), or we go full-in and measure the actual text that will be rendered, by running a (fake) axisY mark with all the options?

mbostock:
Yes, it’s complicated. 🤔

src/dimensions.js
```diff
@@ -11,7 +38,12 @@ export function createDimensions(scales, marks, options = {}) {
     marginBottomDefault = 0.5 + offset,
     marginLeftDefault = 0.5 - offset;
 
-  for (const {marginTop, marginRight, marginBottom, marginLeft} of marks) {
+  // The left and right margins default to a value inferred from the y (and fy)
+  // scales, if present.
+  const yflip = options.y?.axis === "right" || options.fy?.axis === "left";
+  for (let {marginTop, marginRight, marginBottom, marginLeft} of marks) {
+    if (marginLeft === "auto") marginLeft = autoMarginH(yflip ? [scales.fy, options.fy] : [scales.y, options.y]);
+    if (marginRight === "auto") marginRight = autoMarginH(yflip ? [scales.y, options.y] : [scales.fy, options.fy]);
```
mbostock, line 46:
```suggestion
    if (marginRight === "auto") marginRight = autoMarginK(yflip ? "y" : "fy", scales, options);
```

mbostock (commented):
If we support _auto_ for **marginLeft** and **marginRight**, we should also support it for **marginTop** and **marginBottom** (even if those margins are comparatively “dumb” and just have the current behavior).

test/output/usPopulationStateAgeGrouped.svg
Fil: this example is now working

test/output/usStatePopulationChangeRelative.svg
Fil, line 1: new problem! no need for extra margins when the axes are in the middle of the chart!

Fil:
I've implemented the new strategy. Still needs a bit of work:
* clean up the code
* fix the two test cases mentioned above
* keyword check on "auto"
* consider auto for top and bottom margins

test/output/penguinMassSexSpecies.svg
Fil, line 1: same problem here with the x axis label

test/output/penguinFacetDodgeSymbol.html
Fil, line 1: not sure why this one has changed, I'll check

test/output/penguinDodgeHexbin.svg
Fil, line 1: feels like a lot of marginLeft!

test/output/beckerBarley.svg
Fil: Fixed by controlling y and fy differently

test/output/usStatePopulationChangeRelative.svg
Fil: fixed

Fil:
I think this is good to go! Does it need documentation?

src/dimensions.js
```diff
@@ -1,17 +1,94 @@
...
-export function createDimensions(scales, marks, options = {}) {
+const marginMedium = 60;
+const marginLarge = 90;
+
+// When axes have "auto" margins, we might need to adjust the margins, after
+// seeing the actual tick labels. In that case we’ll compute the dimensions and
+// scales a second time.
+function autoMarginK(margin, {k: scale, labelAnchor, label}, options, mark, stateByMark, scales, dimensions, context) {
+  const {data, facets} = stateByMark.get(mark);
+  const {channels} = mark.initializer(data, facets, {}, scales, dimensions, context);
```
mbostock, line 16: Don’t you want to call mark.initialize here, not mark.initializer? The latter only applies to marks with an initializer, but here we want to initialize the mark (without assuming that it has an initializer).

src/marks/axis.js
```diff
@@ -81,9 +81,9 @@ function axisKy(
     x,
     margin,
     marginTop = margin === undefined ? 20 : margin,
-    marginRight = margin === undefined ? (anchor === "right" ? 40 : 0) : margin,
+    marginRight = margin === undefined ? (anchor === "right" ? (x == null ? k : 40) : 0) : margin,
```
mbostock, line 84: Temporarily setting this to `k` feels a bit dirty… it’s effectively a hidden API if you set the marginRight option to "x" or "y", right?

mbostock (commented):
This looks great overall. I think I want to fiddle with the implementation a bit, and in particular I want to use mark.initialize instead of mark.initializer, but the results seem good and the implementation manageable. Do you feel good about it?

src/dimensions.js
Fil: I think we need the initializer because the actual ticks depend on the instantiated scales. We know that the mark has an initializer because it can only be a (tick label) axis mark.

src/marks/axis.js
Fil: Agree, but I didn't find a better way.

src/plot.js
```diff
@@ -171,6 +158,17 @@ export function plot(options = {}) {
...
 
+  // Initalize the dimensions and scales.
+  const channels = addScaleChannels(channelsByScale, stateByMark, options);
+  const {scaleDescriptors, scales, dimensions, subdimensions, superdimensions} = createDimensionsScales(
+    channels,
+    marks,
+    stateByMark,
+    options,
+    context
+  );
```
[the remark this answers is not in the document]
Fil: The thing is we need to run it twice, so inlining it means we have to repeat these lines. Also it felt better to have the part about autoMargins live in dimensions.js (although I'd prefer to have the scales part live in plot.js… and they're intertwined!

Fil:
Yes I feel good about the result: the tests are all looking great and the only changes are improvements. I think the algorithm is correct. The implementation can certainly be better.

src/dimensions.js
mbostock: My point is that we should follow the same code path that Plot.plot follows:

- call _mark_.initialize first
- if _mark_.initializer exists, call that second

We shouldn’t make such a strong assumption about how the mark is implemented. Also, I wasn’t expecting that the mark would already be initialized (using stateByMark); I was thinking we would initialize it ourselves here. So that further complicates matters.

src/marks/axis.js
mbostock: You’ll likely have to change how the defaults are initialized rather than using default initializers.

Fil:
One thing that bugged me is that monospaceWidth and defaultWidth are not aligned; maybe there's still time to change this by making monospace chars 67 instead of 100 — it would be a slightly breaking change, but it would make the two lengths comparable, which is something that was needed here.

src/dimensions.js
Fil: When we call autoMarginK all the marks have been initialized already, which is why we can avoid the cost of re-initializing them. I can see how reaching for stateByMark can feel messy, though. 🥺

mbostock: If the mark is already initialized here, why do we need to invoke its initializer again?

Fil:
There is still a bug as this breaks the aspectRatio option.

AidMen:
bump

PR #1726, opened 2023-06-26
closes #1723

### LINKED ISSUE #1723: For the tree compound mark, the tip option should only apply to the tree node
Currently it applies to both the node and link, which can lead to dual tips.

Fil:
I pondered a bit what happens when dotDot is falsy: we could then add the tip option on the text mark, but this would in turn create more complexity. Decided to keep it simple.


mbostock:
For future, a self-contained description is preferred (in case we end up searching commits or issues), e.g. “fix tip option for tree mark”.

PR #1718, opened 2023-06-23
default multi-line time format

mbostock:
Fixes #1285.

TODO

- [x] utc time
- [x] local time
- [x] bottom orientation
- [x] top orientation
- [x] left orientation
- [x] right orientation
- [x] reversed domains

It would also be nice to have a better format for explicit tick time intervals #932, but I’ve left that for future work.

PR #1728, opened 2023-06-27
document multi-line tick format; improve format for custom ticks

Fil:
(ref: #1718, #1725)


docs/marks/axis.md
```diff
@@ -143,34 +143,22 @@ Plot.plot({
 ```
 :::
 
-You can emulate [Datawrapper’s time axes](https://blog.datawrapper.de/new-axis-ticks/) using `\n` (the line feed character) for multi-line tick labels, plus a bit of date math to detect the first month of each year.
+Time axes default to a consistent multi-line tick format, [à la Datawrapper](https://blog.datawrapper.de/new-axis-ticks/), for example showing the first month of each quarter, and the year:
 
 :::plot https://observablehq.com/@observablehq/plot-datawrapper-style-date-axis
 ```js
 Plot.plot({
+  x: {tickSpacing: 60},
```
mbostock, line 151: This example is a little less compelling having to customize the tick settings. (Presumably, `ticks: 30` or `ticks: d3.utcMonth.every(3)` would produce equivalent output?) Maybe that’s fine, but maybe there’s a dataset with a smaller domain that would produce multi-line ticks by default, resulting in a simpler example?

docs/marks/axis.md
```diff
@@ -143,34 +143,22 @@ Plot.plot({
...
 })
 ```
 :::
 
-:::tip
-In the future, Plot may generate multi-line time axis labels by default. If you’re interested in this feature, please upvote [#1285](https://github.com/observablehq/plot/issues/1285).
-:::
-
-Alternatively, you can add multiple axes with options for hierarchical time intervals, here showing weeks, months, and years.
+The format is inferred from the tick interval, and consists of two fields (*e.g.*, month and year, day and month, minutes and hours…); when a tick has the same second field value as the previous tick (*e.g.*, “19 Jan” after “17 Jan”), only the first field (“19”) is shown for brevity. Alternatively, you can specify multiple explicit axes with options for hierarchical time intervals, here showing weeks, months, and years.
```
mbostock, line 161: Don’t need the ellipsis because *e.g.* implies that you are just giving examples, not an exhaustive list.

```suggestion
The format is inferred from the tick interval, and consists of two fields (*e.g.*, month and year, day and month, minutes and hours); when a tick has the same second field value as the previous tick (*e.g.*, “19 Jan” after “17 Jan”), only the first field (“19”) is shown for brevity. Alternatively, you can specify multiple explicit axes with options for hierarchical time intervals, here showing weeks, months, and years.
```

docs/marks/axis.md
Fil: we could do
`Plot.line(sftemp, Plot.windowY(7, {x: "date", y: "high"})).plot({y: {nice: true}})`

mbostock (approved):

PR #1740, opened 2023-07-08
arrow sweep

### LINKED ISSUE #1739: Arrow sweep strategy
To make an arc diagram such as https://observablehq.com/@d3/arc-diagram, I want to use the arrow mark with a bend angle of 90°, to draw the half-circles connecting a source to a target in the graph.

However currently this always draws the arc in a “clockwise”[1] manner from the first point (x1, y1) to the second point (x2, y2), which means that in most cases the arrow will go either to the left or to the right, depending on whether y1 > y2.

[1] clockwise, becomes counter-clockwise when bend is < 0.

Visually, what I get is the image on the left; what I want is the image on the right:
Capture d’écran 2023-07-08 à 16 30 37

To reach this goal, I can:
1. check how the source and target are ordered in the y domain, and reverse them as needed. This implies that the channels now depend on the scale (not good); 
2. draw both source-target and target-source arrows, and clip them to the frame—provided the main axis is the left edge of the frame (also not very good);
3. write a render transform that reverses the sweep-flag on the generated path (it works, but… 😵‍💫).

PR #1740 introduces a new option **sweep** that allows to control this at render time.

Fil:
Introduces the new **sweep** option for the arrow mark, that allows to ensure that all arrows bulge in the same direction.

Also ensures that no arrowhead is drawn if headLength = 0 (as indicated in the comment in src/marks/arrow).

TODO:
- [x] wording? sweep could be something else, order-x could be x, order-y could be y, and order could be xy… or something else.
- [x] test the horizontal case
- [x] fix bug with the arrowhead when the sweep flag is flipped
- [x] document

closes #1739


src/marks/arrow.js
```diff
@@ -91,6 +87,23 @@ export class Arrow extends Mark {
             // wings, but that’s okay since vectors are usually small.)
             const headLength = Math.min(wingScale * sw(i), lineLength / 3);
 
+            // Maybe flip the bending, if a sweep order is applied.
+            const flip =
+              this.sweep == null
+                ? 1
+                : this.sweep === "order-x"
+                ? descending(x1, x2)
```
mbostock, line 95: This will return 0 when x1 == x2, which will result in a bend angle of 0 below, which I don’t think is correct?

Fil: The use case I had in mind for *order-x* is when *y* is constant; in that case this test would be for links of length 0, and bending can be 0. 

However I hadn't really given much thought to the fact that *y* can be variable. In that case, I don't really know what the user intent might be by specifying *order-x* (instead of *order*, which always “bends” unless the link has length 0). The current code, which makes a straight line when the order is a tie on x, doesn't look wrong and might even be useful (?):
```
Plot.arrow([-2, -1, 0, 1, 2], {
  x1: Plot.identity,
  x2: (d) => d * 1.1,
  y1: 0,
  y2: 1,
  sweep: "order-x",
  bend: -20
}).plot()
```

docs/marks/arrow.md
```diff
@@ -111,9 +111,12 @@ The arrow mark supports the [standard mark options](../features/marks.md#mark-op
...
+* **sweep** - the sweep order
 
 The **bend** option sets the angle between the straight line connecting the two points and the outgoing direction of the arrow from the start point. It must be within ±90°. A positive angle will produce a clockwise curve; a negative angle will produce a counterclockwise curve; zero will produce a straight line. The **headAngle** determines how pointy the arrowhead is; it is typically between 0° and 180°. The **headLength** determines the scale of the arrowhead relative to the stroke width. Assuming the default of stroke width 1.5px, the **headLength** is the length of the arrowhead’s side in pixels.
 
+The **sweep** option can be used to make arrows bend in the same direction, independently of the relative positions of the starting and ending points. If set to *order-x*, the bend angle is flipped when the ending point is to the left of the starting point—ensuring all arrows bulge up (down if bend is negative). If set to *order-y*, the bend angle is flipped when the ending point is above the starting point—ensuring all arrows bulge right (left if bend is negative). If set to *order*, applies an *order-x* sweep, and breaks ties with *order-y*.
```
mbostock, line 118:
```suggestion
The **sweep** option can be used to make arrows bend in the same direction, independently of the relative positions of the starting and ending points. If set to *order-x*, the bend angle is flipped when the ending point is to the left of the starting point — ensuring all arrows bulge up (down if bend is negative). If set to *order-y*, the bend angle is flipped when the ending point is above the starting point — ensuring all arrows bulge right (left if bend is negative). If set to *order*, applies an *order-x* sweep, and breaks ties with *order-y*.
```

src/marks/arrow.d.ts
```diff
@@ -84,6 +84,16 @@ export interface ArrowOptions extends MarkOptions {
...
   insetEnd?: number;
+
+  /**
+   * The sweep order; defaults to null. If set to order-x, the bend angle is
+   * flipped when the ending point is to the left of the starting point—ensuring
+   * all arrows bulge up (down if bend is negative). If set to order-y, the bend
+   * angle is flipped when the ending point is above the starting point—ensuring
+   * all arrows bulge right (left if bend is negative). If set to order, applies
+   * an order-x sweep, and breaks ties with order-y.
+   */
```
mbostock, line 95:
```suggestion
  /**
   * The sweep order; defaults to null. If set to order-x, the bend angle is
   * flipped when the ending point is to the left of the starting point —
   * ensuring all arrows bulge up (down if bend is negative). If set to order-y,
   * the bend angle is flipped when the ending point is above the starting point
   * — ensuring all arrows bulge right (left if bend is negative). If set to
   * order, applies an order-x sweep, and breaks ties with order-y.
   */
```

mbostock (changes requested):
Per 5e313d9a3c4bb8158bf5fa4d1286386282c1faa3.

src/marks/arrow.js
mbostock: I think it’s okay if you document it.

src/marks/arrow.js
```diff
@@ -91,6 +87,23 @@ export class Arrow extends Mark {
...
+                : this.sweep === "order-y"
+                ? descending(y1, y2)
+                : descending(x1, x2) || descending(y1, y2); // "order"
```
mbostock, line 98: I see the use case for *order-x* and *order-y*, but what’s the use case for *order*? Can we remove it? I prefer to add as little functionality as possible unless there is an articulated use case.

mbostock:
If you merge #1741, this looks good to me.

(I also thought about renaming the tests to group the arc diagrams together, instead of making the dataset the primary aspect of the test name… but I’m not sure it matters. The test names should articulate what they are trying to achieve, and I think “arc diagram” is sufficient here. Though “arcs” would also suffice.)

src/marks/arrow.js
Fil: OK; it was more for “completeness” but I have not found a use case.

mbostock (approved):

PR #1741, opened 2023-07-09
functional sweep

mbostock:
I prefer to encapsulate the logic in a function. This also makes it easier for a user to extend with custom logic — even if we don’t chose to expose that extensibility yet.

### DIFF
--- a/src/marks/arrow.d.ts
+++ b/src/marks/arrow.d.ts
@@ -86,14 +86,15 @@ export interface ArrowOptions extends MarkOptions {
   insetEnd?: number;
 
   /**
-   * The sweep order; defaults to null. If set to order-x, the bend angle is
-   * flipped when the ending point is to the left of the starting point—ensuring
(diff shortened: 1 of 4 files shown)


mbostock (commented):
Other changes that I want to make:

- Allow the **sweep** option to be specified as a constant number (typically 1, 0, -1) with the same semantics as a function that returns this number. -1 would effectively negate the bend angle.
- Change the default **sweep** option to 1 instead of null.
- If the **sweep** option is null, it should either be treated as a number and coerced to 0, or throw an error. (We don’t want to treat null as redundant with undefined for specifying the default value; it means “no” sweep.)
- Remove the *order* **sweep** option. (Because I can’t think of a valid use case… but if there is one, please let me know.)
- Allow the **sweep** option to be specified as a function. (Because why not.)

mbostock:
Another thought I had was that we could have the **sweep** `+x` for `ascending(x1, x2)` and `-x` for `descending(x1, x2)`. That would be more flexible than `order-x`. And we could do the same for y.

PR #1744, opened 2023-07-10
geo + tip: true

### LINKED ISSUE #1742: Plot.geo with tip: true collapses the chart
As initially reported by @yurivish ([notebook](https://observablehq.com/d/b8718b11bae1455a)), using tip: true on a non-projected geo makes the map almost vanish; if you look more closely you can see that the map exists but the x and y scales have a range [0, 1] which breaks everything (the geo path, and the axes).

### LINKED ISSUE #1743: Plot.geo with tip: true should use Plot.centroid
If there is no *x* and *y* channels (or maybe *px* and *py* ??), but we have a geometry channel, the tip: true option should derive these coordinates with Plot.centroid.

Fil:
This changes the derivation of a tip mark when there is only a geometry channel (typically, the geo mark) by:
- fixing a collapsing bug (#1742)
- computing a centroid (unless there already is a channel for x or y) (#1743)

Demo: https://observablehq.com/@fil/geo-tip-true-1744

fixes #1742
closes #1743 

cc: @yurivish


Fil:
closing, I'm going to separate fixing the bug and the proposed new feature

PR #1746, opened 2023-07-10
tip on a geo mark without a projection should not collapse the chart

### LINKED ISSUE #1742: Plot.geo with tip: true collapses the chart

Fil:
fixes #1742

when adding tip: true on a geo mark, we create an initializer that generates new channels for the derived tip mark, typically geometry, and x and y if the geoCentroid transform or the centroid initializer were applied.

However:
1. _geometry_ is associated to the "projection" scale, resulting in the "collapsing" of x and y when considered as a new scale
2. _x_ and _y_, as produced by the centroid initializer, were not scaled 

Demo: https://observablehq.com/@fil/geo-tip-true-1746


mbostock (approved):
Nice!

PR #1761, opened 2023-07-20
title, subtitle

### LINKED ISSUE #92: Plot title (and signature, sources).

### LINKED ISSUE #382: Title, subtitle, source, copyright
According to the usual rules of legibility, a chart should include an informative title, coming first (above the chart), possibly a subtitle, obligatorily a source, at the bottom of the chart, more optionally a copyright (which can also be arranged vertically, e.g. bottom right).

image
 
ggplot2 and vega-lite allow you to specify these essential elements, it would be very nice if Plot could as well.

The SVG export of the Plot should allow you to get the whole graphic, with titles and sources.

As vega-lite proposes, specifying a title as an array allows it to be spread over several lines.

The height specified for the graphic should not take these elements into account, but only the central body of the graphic.

PS: Plot is a great library, very intellectually stimulating, and the 21 beautifully written articles in the https://observablehq.com/collection/@observablehq/plot collection should be read by all data scientists!

Fil:
Add top-level **title** and **subtitle** options.

Like caption, these options make a figure element. By default the title is considered text, and set in a H2 element. The subtitle in a H3 element. If an HTML node is passed, it is used as is (this differs from caption, where the contents are always wrapped in a figcaption element).

I'd like to add default styles, but better discuss them here before, because changing the styles will modify all the snapshot tests. And, in that case, maybe add `${className}-figure` on the figure element?

Default styles could be something like:
```js
.${className}-h2 {
  margin-top: 0.3em;
  margin-bottom: 0.3em;
}
.${className}-h3 {
  margin-top: 0.3em;
  margin-bottom: 0.3em;
  font-weight: normal;
  font-style: italic;
}
```

Demo: https://observablehq.com/@observablehq/plot-title-1761

closes #92
supersedes #423

### DIFF
--- a/docs/features/plots.md
+++ b/docs/features/plots.md
@@ -250,7 +250,7 @@ When using facets, set the *fx* and *fy* scales’ **round** option to false if 
 
 By default, [plot](#plot) returns an SVG element; however, if the plot includes a title, subtitle, [legend](./legends.md), or caption, plot wraps the SVG element with an HTML figure element. You can also force Plot to generate a figure element by setting the **figure** option <VersionBadge pr="1792" /> to true.
 
-The **title** <VersionBadge pr="1792" />, **subtitle** <VersionBadge pr="1792" />, and **caption** options accept either a string or an HTML element. If given an HTML element, say using the [`html` tagged template literal](http://github.com/observablehq/htl), the title and subtitle are used as-is while the caption is wrapped in a figcaption element; otherwise, the specified text will be escaped and wrapped in an H2, H3, or figcaption, respectively.
+The **title** <VersionBadge pr="1792" />, **subtitle** <VersionBadge pr="1792" />, and **caption** options accept either a string or an HTML element. If given an HTML element, say using the [`html` tagged template literal](http://github.com/observablehq/htl), the title and subtitle are used as-is while the caption is wrapped in a figcaption element; otherwise, the specified text will be escaped and wrapped in an h2, h3, or figcaption, respectively.
 
 :::plot https://observablehq.com/@observablehq/plot-caption
 ```js
--- a/docs/features/plots.md
+++ b/docs/features/plots.md
@@ -255,12 +255,12 @@ The **title** <VersionBadge pr="1792" />, **subtitle** <VersionBadge pr="1792" /
 :::plot https://observablehq.com/@observablehq/plot-caption
 ```js
 Plot.plot({
-  title: "For charts, a title assists interpretation",
-  subtitle: "Tell the reader what’s interesting instead of forcing them to discover it.",
+  title: "For charts, an informative title",
+  subtitle: "Subtitle to follow with additional context",
   caption: "Figure 1. A chart with a title, subtitle, and caption.",
   marks: [
     Plot.frame(),
-    Plot.text(["Hello, world!"], {frameAnchor: "middle"})
+    Plot.text(["Titles, subtitles, captions, and annotations assist inter­pretation by telling the reader what’s interesting. Don’t make the reader work to find what you already know."], {lineWidth: 30, frameAnchor: "middle"})
   ]
 })
 ```
--- a/docs/features/plots.md
+++ b/docs/features/plots.md
@@ -248,9 +248,9 @@ When using facets, set the *fx* and *fy* scales’ **round** option to false if 
 
 ## Other options
 
-By default, [plot](#plot) returns an SVG element; however, if the plot includes a title, subtitle, [legend](./legends.md), or caption, plot wraps the SVG element with an HTML figure element. You can also force Plot to generate a figure element by setting the **figure** option <VersionBadge pr="1792" /> to true.
+By default, [plot](#plot) returns an SVG element; however, if the plot includes a title, subtitle, [legend](./legends.md), or caption, plot wraps the SVG element with an HTML figure element. You can also force Plot to generate a figure element by setting the **figure** option <VersionBadge pr="1761" /> to true.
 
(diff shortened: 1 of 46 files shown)


Fil:
(I don't know why the test failed; it passes locally.)

src/plot.js
```diff
@@ -320,17 +320,20 @@ export function plot(options = {}) {
...
-      figure.appendChild(figcaption);
+    for (const e of [[title, "h2"], [subtitle, "h3"], ...legends, svg, [caption, "figcaption"]]) {
+      const [contents, tag] = Array.isArray(e) ? e : [e];
+      if (contents == null) continue;
+      if (tag && (!contents.ownerDocument || tag === "figcaption")) {
+        const c = document.createElement(tag);
+        if (tag === "h2" || tag === "h3") c.className = `${className}-${tag}`;
+        c.appendChild(contents.ownerDocument ? contents : document.createTextNode(contents));
+        figure.appendChild(c);
+      } else figure.appendChild(contents);
```
mbostock, line 336: This feels like it’s written a bit too generically; probably better to just break out the code separately for title, subtitle, legends, svg, and caption.

Note that we could also do `figure.append(...legends, svg)` for that case.

src/plot.d.ts
```diff
@@ -107,6 +107,38 @@ export interface PlotOptions extends ScaleDefaults {
    */
   className?: string;
 
+  /**
+   * The figure title. If present, Plot wraps the generated SVG element in an
+   * HTML figure element with the title in a h2 element, returning the figure. To
+   * specify an HTML title, consider using the [`html` tagged template
+   * literal](http://github.com/observablehq/htl); otherwise, the specified
```
mbostock, line 114: Use a [1] link here. (See examples elsewhere if needed.)

src/plot.js
Fil: I've updated the logic, hopefully it feels more readable/less messy?

src/plot.js
```diff
@@ -320,18 +320,15 @@ export function plot(options = {}) {
...
-    figure.appendChild(svg);
-    if (caption != null) {
-      const figcaption = document.createElement("figcaption");
-      figcaption.appendChild(caption?.ownerDocument ? caption : document.createTextNode(caption));
-      figure.appendChild(figcaption);
-    }
+    if (title != null) figure.appendChild(createTitleElement(title, "h2", className));
+    if (subtitle != null) figure.appendChild(createTitleElement(subtitle, "h3", className));
+    figure.append(...legends, svg);
+    if (caption != null) figure.appendChild(createFigcaption(caption));
```
mbostock, line 331:
```suggestion
    if (title != null) figure.append(createTitleElement(title, "h2", className));
    if (subtitle != null) figure.append(createTitleElement(subtitle, "h3", className));
    figure.append(...legends, svg);
    if (caption != null) figure.append(createFigcaption(caption));
```

src/plot.js
```diff
@@ -354,6 +351,20 @@ export function plot(options = {}) {
   return figure;
 }
 
+function createTitleElement(contents, tag, className) {
+  if (contents.ownerDocument) return contents;
+  const e = document.createElement(tag);
```
mbostock, line 356: This needs to use context.document instead of the global document.

src/plot.js
```diff
@@ -354,6 +351,20 @@ export function plot(options = {}) {
...
+  e.className = `${className}-${tag}`;
+  e.appendChild(document.createTextNode(contents));
```
mbostock, line 358:
```suggestion
  e.append(document.createTextNode(contents));
```

src/plot.js
```diff
@@ -354,6 +351,20 @@ export function plot(options = {}) {
...
+  return e;
+}
+
+function createFigcaption(caption) {
+  const e = document.createElement("figcaption");
+  e.appendChild(caption.ownerDocument ? caption : document.createTextNode(caption));
```
mbostock, line 364:
```suggestion
  e.append(caption.ownerDocument ? caption : document.createTextNode(caption));
```

mbostock (changes requested):
Better! But we need to respect the context.document instead of using the global document.

src/plot.js
Fil: oops! good catch, thanks!

Fil:
I've now added documentation, with a custom stylesheet to bypass vitepress margins on h2 and h3.

docs/features/plots.md
```diff
@@ -248,15 +248,19 @@ When using facets, set the *fx* and *fy* scales’ **round** option to false if
 
 ## Other options
 
-If a **caption** is specified, Plot.plot wraps the generated SVG element in an HTML figure element with a figcaption, returning the figure. To specify an HTML caption, the caption can be specified as an HTML element, say using the [`html` tagged template literal](http://github.com/observablehq/htl); otherwise, the specified string represents text that will be escaped as needed.
+By default, [plot](#plot) returns an SVG element; however, if the plot includes a title, subtitle, [legend](./legends.md), or caption, plot wraps the SVG element with an HTML figure element. You can also force Plot to generate a figure element by setting the **figure** option <VersionBadge pr="1792" /> to true.
+
+The **title** <VersionBadge pr="1792" />, **subtitle** <VersionBadge pr="1792" />, and **caption** options accept either a string or an HTML element. If given an HTML element, say using the [`html` tagged template literal](http://github.com/observablehq/htl), the title and subtitle are used as-is while the caption is wrapped in a figcaption element; otherwise, the specified text will be escaped and wrapped in an h2, h3, or figcaption, respectively.
```
mbostock, line 253: *cough* wrong PR linked here 😞

mbostock (approved):

docs/features/plots.md
Fil: oops! thank you

Issue #1763, opened 2023-07-21
Auto mark isMonotonic should distinguish sorted within series


tophtucker:
If data is monotonic in X or Y, the auto mark may draw a line. But we have a longstanding TODO in isMonotonic:

https://github.com/observablehq/plot/blob/ee5eff7844a48453a4c6239d9f11363edc0879c6/src/marks/auto.js#L246-L263

For example, this data:

```
const data = [
    {date: new Date(2020, 0, 1), value: 5, series: 1},
    {date: new Date(2020, 0, 1), value: 50, series: 2},
    {date: new Date(2020, 0, 1), value: 500, series: 3},
    {date: new Date(2020, 0, 2), value: 6, series: 1},
    {date: new Date(2020, 0, 2), value: 60, series: 2},
    {date: new Date(2020, 0, 2), value: 600, series: 3},
    {date: new Date(2020, 0, 3), value: 7, series: 1},
    {date: new Date(2020, 0, 3), value: 70, series: 2},
    {date: new Date(2020, 0, 3), value: 700, series: 3}
  ]
  return Plot.auto(data, {x: "date", y: "value"}).plot();
```

Produces a weird line: 

<img width="452" alt="image" src="https://github.com/observablehq/plot/assets/841829/eb4c4d18-c605-49ad-ba95-c4aa8562ff85">

If two points have the same value (so order === 0), should isMonotonic be false? There are fancier things we could do, like trying to infer a z channel, but…

(Idk if this is a bug or enhancement!)

mbostock:
This seems like the expected/desired behavior to me. The only way to detect “sorted/monotonic within series” would be to tell the auto mark the z channel, to group by z, to apply the monotonicity test to each series, requiring them all to be monotonic.

mbostock:
The case I meant by the TODO is:

```js
data = [
  {date: new Date(2020, 0, 1), value: 5, series: 1},
  {date: new Date(2020, 0, 2), value: 7, series: 1},
  {date: new Date(2020, 0, 3), value: 6, series: 1},
  {date: new Date(2020, 0, 1), value: 50, series: 2},
  {date: new Date(2020, 0, 2), value: 70, series: 2},
  {date: new Date(2020, 0, 3), value: 60, series: 2},
  {date: new Date(2020, 0, 1), value: 500, series: 3},
  {date: new Date(2020, 0, 2), value: 700, series: 3},
  {date: new Date(2020, 0, 3), value: 600, series: 3}
]
```

[...]

PR #1764, opened 2023-07-21
Auto mark: isMonotonic return false if points collide

tophtucker:
Fixes https://github.com/observablehq/plot/issues/1763. Like making it a `> || <` check instead of `>= || <=`. Not sure about this approach but wanted to put something up.


mbostock:
Don’t think this is right. There could be two coincident observations without violating monotonicity. I think the only way to detect series is explicitly with z (or fill or stroke).

PR #1765, opened 2023-07-22
api index

### LINKED ISSUE #1752: API index
We should have an API index, like https://d3js.org/api. It could just be a list of all options, grouped by mark/transform/method?

mbostock:
Fixes #1752.

Enumerates every method with a brief description pulled from the documentation:

Screenshot 2023-07-26 at 4 09 09 PM

Also enumerates every option, showing which interface the option is defined:

Screenshot 2023-07-26 at 4 09 58 PM

It would perhaps be better to provide more specific links (e.g., to the `#bend` anchor fragment instead of the general arrow mark documentation), but I think that’s non-blocking and we could add that in the future as desired.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -130,7 +130,8 @@ export default defineConfig({
           {text: "Crosshair", link: "/interactions/crosshair"},
           {text: "Pointer", link: "/interactions/pointer"}
         ]
-      }
+      },
+      {text: "API index", link: "/api"}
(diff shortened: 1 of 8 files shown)


mbostock:
Trying this for now. Let me know if you want changes. I think I’d like to use shorter, explicit anchor fragments across the site, for example, such as `#pointerX` instead of `#pointerx-options`.

Fil:
Love it!

PR #1771, opened 2023-07-29
better anchors

mbostock:


docs/api.md
```diff
@@ -10,15 +10,15 @@ import {data} from "./data/api.data";
...
+    <span><a :href="`${href}#${name}`">{{ name }}</a> - {{ comment }}</span>
   </li>
 </ul>
 
 ## Options
 
 <ul>
   <li v-for="[name, contexts] in data.options">
-    <b>{{ name }}</b> - <span v-for="({name, href}, index) in contexts"><a :href="href">{{ name }}</a><span v-if="index < contexts.length - 1">, </span></span>
+    <b>{{ name }}</b> - <span v-for="({name: context, href}, index) in contexts"><a :href="href">{{ context }}</a><span v-if="index < contexts.length - 1">, </span></span>
```
mbostock, line 21: We could use `${href}#${name}` here too, if we were to add corresponding anchors on the destination page for each option.

PR #1772, opened 2023-07-29
bollinger mark & transform

### LINKED ISSUE #548: Built-in Bollinger band transform
Like so

```js
function bollinger(N, K) {
  return values => {
    let i = 0;
    let sum = 0;
    let sum2 = 0;
    const Y = new Float64Array(values.length).fill(NaN);
    for (let n = Math.min(N - 1, values.length); i < n; ++i) {
      const value = values[i];
      sum += value, sum2 += value ** 2;
    }
    for (let n = values.length; i < n; ++i) {
      const value = values[i];
      sum += value, sum2 += value ** 2;
      const mean = sum / N;
      const deviation = Math.sqrt((sum2 - sum ** 2 / N) / (N - 1));
      Y[i] = mean + deviation * K;
      const value0 = values[i - N + 1];
      sum -= value0, sum2 -= value0 ** 2;
    }
    return Y;
  };
}
```

mbostock:
Fixes #548.

### DIFF
--- a/src/marks/bollinger.js
+++ b/src/marks/bollinger.js
@@ -1,0 +1,57 @@
+import {deviation, mean} from "d3";
+import {marks} from "../mark.js";
+import {map} from "../transforms/map.js";
+import {window} from "../transforms/window.js";
+import {areaX, areaY} from "./area.js";
+import {lineX, lineY} from "./line.js";
+import {identity} from "../options.js";
+
+export function bollingerX(
+  data,
+  {
+    x = identity,
+    y,
+    n = 20,
+    k = 2,
+    color = "currentColor",
+    opacity = 0.2,
+    fill = color,
+    fillOpacity = opacity,
+    stroke = color,
+    strokeOpacity,
+    strokeWidth,
+    ...options
+  } = {}
+) {
+  return marks(
+    areaX(data, map({x1: bollinger(n, -k), x2: bollinger(n, k)}, {x1: x, x2: x, y, fill, fillOpacity, ...options})),
+    lineX(data, map({x: bollinger(n, 0)}, {x, y, stroke, strokeOpacity, strokeWidth, ...options}))
+  );
+}
+
+export function bollingerY(
+  data,
+  {
+    x,
+    y = identity,
+    n = 20,
+    k = 2,
+    color = "currentColor",
+    opacity = 0.2,
+    fill = color,
+    fillOpacity = opacity,
+    stroke = color,
+    strokeOpacity,
+    strokeWidth,
+    ...options
+  } = {}
+) {
+  return marks(
+    areaY(data, map({y1: bollinger(n, -k), y2: bollinger(n, k)}, {x, y1: y, y2: y, fill, fillOpacity, ...options})),
+    lineY(data, map({y: bollinger(n, 0)}, {x, y, stroke, strokeOpacity, strokeWidth, ...options}))
+  );
+}
+
+export function bollinger(n, k) {
+  return window({k: n, reduce: (Y) => mean(Y) + k * (deviation(Y) || 0), strict: true, anchor: "end"});
+}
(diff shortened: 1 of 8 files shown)


src/marks/bollinger.js
```diff
@@ -0,0 +1,75 @@
...
+    ...options
+  } = {}
+) {
+  return marks(
+    areaY(data, bollingerY(n, k, {x, y1: y, y2: y, fill, fillOpacity, ...options})),
+    lineY(data, bollingerY(n, 0, {x, y, stroke, strokeOpacity, strokeWidth, ...options}))
+  );
+}
+
+export function bollingerX(n, k, options) {
```
mbostock, line 53: Not sure *n* and *k* should be required positional arguments here. Probably would be more consistent to have a *bollingerOptions* argument, and allow a single combined *options*, too. And maybe a single number could be used to specify *n*; *n* defaults to 20 and *k* defaults to 2.

src/marks/bollinger.js
```diff
@@ -0,0 +1,75 @@
...
+  let {x, x1, x2} = options;
+  if (x === undefined && x1 === undefined && x2 === undefined) options = {...options, x: (x = identity)};
+  const outputs = {};
+  if (x != null) outputs.x = bollinger(n, k);
+  if (x1 != null) outputs.x1 = bollinger(n, -k);
```
mbostock, line 58: By negating *k* here, bollingerX behaves like mapX, and can be used for either area or line!

mbostock:
Nice, we get a build failure now when we don’t document something. 😁 I’ll add that to the PR.

Fil:
Nice! It’s a useful and sound method for general time series analysis.

I'm just wondering if we should keep that name and defaults of the transform. “Bollinger” seems to be confined to finance (according to [wikipedia](https://en.wikipedia.org/wiki/Bollinger_Bands#Bollinger_bands_outside_of_finance)), and has the default multiplier = 2 (not a bad choice, but mathematically surprising).

From https://www.bollingerbands.com/bollinger-bands, that "bands" is plural—they represent the three lines y = moving average MA, y2 = MA+2 Dev, and y1 = MA-2 Dev. Whereas how I see it, is a singular band, the area between y1 and y2.

Also, there is a trademark on the name.

Even in finance, people seem to be using “volatility” often (shown as a 1-sigma band). The 1-sigma band excludes 31.8% of the distribution on a normal model, so it makes sense to use a 2-sigma band as the default (4.6% outliers), but it's also legit to do 3-sigma (0.3%), etc. Other data science domains seem to be using words like “rolling window variance” or  “rolling window deviation”, and plotting them with 1-sigma, 2-sigma, 3-sigma lines or bands, or adapting the sigmas to match 10%, 5%, 1%….

Fil (commented):
The only blocking remark is that there is a TODO in the documentation.

_Non-blocking remarks & questions:_

1. To apply in log-space rather than in linear-space (e.g., to avoid negative values if the multiplier &times; the deviation is high), one can do:

```
Plot.bollingerY(aapl, { x: "Date", y: (d) => t(d.Close), k }).plot({
  y: { transform: (d) => t.invert(d), domain: [0, 300], nice: true }
})
```

with `t = d3.scaleLog()` 👍. (Note that the difference on aapl with the default parameters is totally imperceptible.)


2. Should we add the **anchor** option? For time series we obviously don't want to use “values from the future“, and if the data is coming in reverse chronological order (e.g. `data = d3.reverse(aapl)`, it seems important to either anchor: "start" or use an explicit **sort: "Date"**. Similarly, it could be useful to anchor “middle” if _x_ is not time, and we want to gauge local variance.

3. In that vein I'd also want to expose the underlying **strict** option, which would be particularly useful to set to false when the data has gaps (_e.g._, interval: "day" on the aapl dataset). (If we support that option and default it to _true_, we should document it as it is the converse of windowY's default.)

Fil (approved):

PR #1777, opened 2023-08-01
fix pointerX & faceting

### LINKED ISSUE #1776: pointerX can fail with faceting
as reported by @yurivish in https://observablehq.com/d/3264dbb558b44080, using pointerX (e.g. to browse faceted timelines) fails when the facets are aligned vertically can in some cases feel “wrong”.

minimal repro:

```
Plot.lineY(industries, {
  x: "date",
  y: "unemployed",
  fx: (d) => d.date.getDay(),
  fy: (d) => d.date.getFullYear(),
  stroke: "industry",
  tip: true
}).plot()
```

https://github.com/observablehq/plot/assets/7001/e9ae447f-2724-41d6-9322-b592ca7ca3ba

Fil:
Return the usual euclidian distance to determine the winner across facets, not the squashed distance.

fixes #1776 cc @yurivish 

Demo here https://observablehq.com/@observablehq/facet-and-pointerx-1777

I'm not adding that test in the repo, because it's not an interesting chart / use case.

### DIFF
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -127,6 +127,10 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
         return r;
       }
 
+      // Select the closest point to the mouse in the current facet; for
+      // pointerX or pointerY, the orthogonal component of the distance is
+      // squashed, selecting primarily on the dominant dimension. Return the
+      // usual euclidian distance to determine the winner across facets.
       function pointermove(event) {
         if (state.sticky || (event.pointerType === "mouse" && event.buttons === 1)) return; // dragging
         let [xp, yp] = pointof(event);
@@ -139,7 +143,7 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
           const rj = dx * dx + dy * dy;
           if (rj <= ri) (ii = j), (ri = rj);
         }
-        update(ii, ri);
+        update(ii, Math.hypot(px(ii) - xp, py(ii) - yp));
       }
 
       function pointerdown(event) {


Fil:
I've added a unit test for @tophtucker ; see the demo notebook.

mbostock:
I’m afraid I don’t see any effective difference in behavior in the liborProjectionsFacet. I do see this improves the behavior in the case that @yurivish provided. Am I missing something? Can we make the test more obvious and still “interesting” or “valid”?

https://github.com/observablehq/plot/assets/230541/d31414d5-5fea-4b0e-879f-348ef72d5f11

Also, even in the fixed version, I think it’s a little surprising that the tip appears on the facets far below (because there’s nothing interesting on the current facet). I’d prefer that across facets we respect the **maxRadius** option, too.

src/interactions/pointer.js
```diff
@@ -139,7 +143,7 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
           const rj = dx * dx + dy * dy;
           if (rj <= ri) (ii = j), (ri = rj);
         }
-        update(ii, ri);
+        update(ii, Math.hypot(px(ii) - xp, py(ii) - yp));
```
mbostock, line 146: Also, `ii` can be null here, and in that case we shouldn’t attempt to compute the distance.

mbostock (changes requested):

Fil:
I agree that it's unnerving when the selection goes in the "wrong" facet, but I'm not sure how to determine that we're in the wrong facet (especially since a value can be outside the facet's frame). Here's a notebook that demonstrates a difficult case:
https://observablehq.com/@observablehq/closest-point-in-facets-1777
untitled (4)

On the left part of the frame, where points are alternatively in the blue and orange facets, the current branch is much better than main (main jumps between top and bottom).

However, the right part of the frame is still a bit unsettling. What should happen if the pointer is at the center of one of the large circles? The top circle belongs to the top facet, so we're tempted to reject the point that has the tip (based on euclidian distance); however, the orange circle belongs to the same facet as that point, so we should accept it (based on the "squished" horizontal distance).

Maybe the correct criterion is “if the pointer is in the frame *fi*, filter out solutions that belong to other facets, based on euclidian distance”. But we need to keep the solutions that are in the same facet, even if they're too far in euclidian distance. I'll try and implement that.

---

Regarding the unit test, I reckon it's subtle, and I could substitute a more obvious chart (maybe with a subset of libor, or with the test above). But I do see the difference when I mouse in the following region:
Capture d’écran 2023-08-07 à 12 26 15

Fil:
The updated strategy is available by selecting `#1777/6cb2d` in [the demo](https://observablehq.com/@observablehq/closest-point-in-facets-1777).

mbostock:
Thanks, that’s a very helpful demo! I want to look at this a little more but I think this is good. 👍

mbostock:
I tweaked this logic to be a little simpler (in my opinion): in addition to using the unscaled distance to compare across facets, if the pointer is outside the frame, then the scaling factors _kx_ and _ky_ are ignored.

mbostock (approved):

PR #1779, opened 2023-08-02
fix: pointer interaction was not always dispatching the new value

### LINKED ISSUE #1778: Tip & faceting
I happened to notice an oddity with selections and faceting: https://observablehq.com/d/a03f70ae5c5962c4 

It might be related to https://github.com/observablehq/plot/issues/1776 but I can still reproduce the behavior I'm seeing with `PlotF` from the [notebook](https://observablehq.com/@observablehq/facet-and-pointerx-1777) linked in https://github.com/observablehq/plot/pull/1777 so it does not look like the fix for that one also fixes this one.

image

In this plot, the selection can sometimes be `null` even though the user is hovering over an item. Giving the plot a wide `width` fixes it, so it might have to do with the transition between facets and the sequencing of the events that triggers. Interestingly, moving from left to right seems to work, but not right to left.

Fil:
When simultaneously exiting a facet and entering a new one, prioritize dispatching the value from the entering facet.

Fixes #1778 

Cc: @yurivish

### DIFF
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -123,7 +123,10 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
           g.replaceWith(r);
         }
         state.roots[renderIndex] = g = r;
-        context.dispatchValue(i == null ? null : data[i]);
+
+        // Dispatch the value. When simultaneously exiting this facet and
+        // entering a new one, prioritize the entering facet.
+        if (!(i === null && facetState?.size > 1)) context.dispatchValue(i == null ? null : data[i]);
         return r;
       }
 


src/interactions/pointer.js
```diff
@@ -123,7 +123,10 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
           g.replaceWith(r);
         }
         state.roots[renderIndex] = g = r;
-        context.dispatchValue(i == null ? null : data[i]);
+
+        // Dispatch the value. When simultaneously exiting this facet and
+        // entering a new one, prioritize the entering facet.
+        if (!(i === null && facetState?.size > 1)) context.dispatchValue(i == null ? null : data[i]);
```
mbostock, line 129:
```suggestion
        if (!(i == null && facetState?.size > 1)) context.dispatchValue(i == null ? null : data[i]);
```

Elsewhere in this function we’re using a loose nullish test, so I assume that’s what we want here too?

PR #1794, opened 2023-08-07
normalize links in JSDoc comments

Fil:
Normalize to using numbered links everywhere. Change a few http to https. 

Note that this changes the aspect of one element (_locale_), which goes from "paragraph" to "inline" mode:
 
Capture d’écran 2023-08-07 à 16 28 16

(Also, when the URL is long, we get bad reflows when using option-Q in VS Code.)

### DIFF
--- a/src/format.d.ts
+++ b/src/format.d.ts
@@ -2,11 +2,12 @@
  * Returns a function that formats a given month number (from 0 = January to 11
  * = December) according to the specified *locale* and *format*.
  *
- * @param locale a [BCP 47 language tag](https://tools.ietf.org/html/bcp47);
- * defaults to U.S. English.
- * @param format a [month
- * format](https://tc39.es/ecma402/#datetimeformat-objects): either *2-digit*,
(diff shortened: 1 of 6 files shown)


src/format.d.ts
```diff
@@ -17,11 +18,12 @@ export function formatMonth(
...
  * = Saturday) according to the specified *locale* and *format*.
  *
- * @param locale a [BCP 47 language tag](https://tools.ietf.org/html/bcp47);
- * defaults to U.S. English.
- * @param format a [weekday
- * format](https://tc39.es/ecma402/#datetimeformat-objects): either *narrow*,
- * *short*, or *long*; defaults to *short*.
+ * @param locale a [BCP 47 language tag][1]; defaults to U.S. English.
+ * @param format a [weekday format][2]: either *narrow*, *short*, or *long*;
+ * defaults to *short*.
```
mbostock, line 23: Aside: I think we could drop the `either *narrow*, *short*, or *long*` here because that’s already clear from the type.

mbostock (approved):

PR #1796, opened 2023-08-09
avoid repetition of format options

Fil:
(as suggested by https://github.com/observablehq/plot/pull/1794#discussion_r1287786741; note that the second line is still in the ugly "paragraph" mode in VSCode, even though it's shorter that the first line 🤷

Capture d’écran 2023-08-09 à 10 01 10

### DIFF
--- a/src/format.d.ts
+++ b/src/format.d.ts
@@ -2,11 +2,11 @@
  * Returns a function that formats a given month number (from 0 = January to 11
  * = December) according to the specified *locale* and *format*.
  *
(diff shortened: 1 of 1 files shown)


mbostock:
Confusing, but we have to put the links _above_ the param declarations. Otherwise VS Code thinks that there’s a paragraph of text for the last param, and hence uses the paragraph format.

Screenshot 2023-08-09 at 8 41 22 AM

mbostock (approved):

PR #1801, opened 2023-08-11
refine auto bar mark determination

### LINKED ISSUE #1800: Auto: can it be made to make a timeseries bar chart?
Currently, you can only make a bar chart with a temporal/quantitative independent axis if the dependent axis has a reducer specified (implicitly or explicitly). But you cannot make a temporal/quantitative bar chart with a simple value as the dependent encoding.

- `Plot.auto(olympians, {x: "date_of_birth"}).plot()` → implicit count reducer on y, hence temporal bar chart
- `Plot.auto(aapl, {x: "Date", y: "Volume", mark: "bar"}).plot()` → surprising sparse matrix of cells

I know the bar-interval thing is a perennial headache lol, and others have raised it, but I’m feeling the pain right now and feeling motivated to come up with _something_!

E.g. for `Plot.auto(aapl, {x: "Date", y: "Volume"}).plot()`, we get a line, but it is [inappropriate](https://observablehq.com/@tophtucker/inferring-chart-type-from-autocorrelation-and-other-evils) to interpolate values between successive days, so (hairline) bars would be better.

~~I may have made this worse in https://github.com/observablehq/plot/pull/1674, which changed it to choose cell over rect if there's no reducer, which might've been too aggressive and prevents us from taking advantage of the rect's implicit y1 = 0. (But I'm sleepy and not sure it ever would've worked; can check later.)~~ (Nope, see below.)

Might be improved by https://github.com/observablehq/plot/pull/1790. 

[Example notebook](https://observablehq.com/d/60896fc6389d1341):

image

mbostock:
Fixes #1800. The logic is quite complicated — but I think it helped to start by looking at the reducers, since that determines whether there is a bin or group transform.

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -1,6 +1,6 @@
 import {ascending, InternSet} from "d3";
 import {marks} from "../mark.js";
-import {isColor, isObject, isOptions, isOrdinal, labelof, valueof} from "../options.js";
+import {isColor, isNumeric, isObject, isOptions, isOrdinal, labelof, valueof} from "../options.js";
 import {bin, binX, binY} from "../transforms/bin.js";
 import {group, groupX, groupY} from "../transforms/group.js";
 import {areaX, areaY} from "./area.js";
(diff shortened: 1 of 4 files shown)


mbostock:
Ah, foo. Looks like I’ve broken the autoRectColorReducer test… I’ll need to investigate.

mbostock:
Fix was easy! I just needed to make colorReduce take priority over xReduce and yReduce.

tophtucker (approved):
Nice!! So I guess isOrdinalReduce can be effectively inlined because you made the reducer the top-level switch?

Nice that it no longer depends on xZero / yZero, given the intuition that "bar" implies zero-ness. (Maybe explicit zero-ness let us break ties between barX and barY? But I can't think of a case.)

Played around and didn‘t see anything unexpected.

Test seems good. None of our sample data seems to naturally hit this case. I feel like it’s more about small-N stuff. Like one revenue number for each of three business segments for each of the last five years.

PR #1804, opened 2023-08-14
fix axis label being clipped with the clip plot option

### LINKED ISSUE #1803: clip: true clips the axes’ labels
This should show "↑ Close" as the y axis label, but the label gets clipped.

```js
Plot.plot({
  clip: true,
  x: { domain: [new Date(2015, 0, 1), new Date(2015, 3, 1)] },
  marks: [
    Plot.areaY(aapl, { x: "Date", y: "Close", fill: "#eee" }),
    Plot.lineY(aapl, { x: "Date", y: "Close" })
  ]
})
```

Ref: #1792

Fil:
closes #1803


src/marks/axis.js
```diff
@@ -501,6 +501,7 @@ function labelOptions(
     monospace,
     pointerEvents,
     shapeRendering,
+    clip: clip === undefined ? false : clip,
```
mbostock, line 504: We should use `clip = false` when destructuring the arg.

mbostock (approved):

PR #1807, opened 2023-08-15
re-encode image data

mbostock:


Fil (approved):
Excellent! I feel that the thresholds are very generous — how much error did you notice in practice?

mbostock:
In the opacityLegendColor example, which has 256 pixels (1,024 RGBA values), I found:

- 47 values with any error
- 14 values with an error of at least 2
- 6 values with an error of at least 3
- 2 values with an error of at least 4
- no values with an error of more than 4
- a mean error of 0.0673828125
- 95% of pixels had no error
- 96% of pixels had 1 or less error
- 99% of pixels had 2 or less error

We could probably reduce the max error threshold slightly, but that is the noisiest since it can be affected by a single value. I think the other ones are already tight.

tophtucker:
😻 

2023-08-15 - plot tests passing

PR #1809, opened 2023-08-16
update vitepress, fix a build crash, add a sitemap

Fil:
- fix a crash with docs:build and figure.append / el.append
- upgrade vitepress to 1.0.0-rc.4 (https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md)
- generate a sitemap at ./docs/.vitepress/dist/sitemap.xml (available with the vitepress update, see https://github.com/vuejs/vitepress/pull/2691; documentation: https://vitepress.dev/guide/sitemap-generation#sitemap-generation)

Re: the crash, I don't understand why it doesn't seem to happen in production.  When I call `yarn docs:build` with the current main branch, I get dozens of errors:

```
✓ building client + server bundles...
⠹ rendering pages...TypeError: Function.prototype.apply was called on undefined, which is a undefined and not a function
    at Module.plot (file:///Users/fil/Sites/plot/docs/.vitepress/.temp/plot.c0318f47.js:5839:12)
    at Proxy.render (file:///Users/fil/Sites/plot/docs/.vitepress/.temp/PlotRender.53d19871.js:788:35)
    at renderComponentRoot (/Users/fil/Sites/plot/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:399:16)
    at renderComponentSubTree (/Users/fil/Sites/plot/node_modules/@vue/server-renderer/dist/server-renderer.cjs.prod.js:440:28)
    at renderComponentVNode (/Users/fil/Sites/plot/node_modules/@vue/server-renderer/dist/server-renderer.cjs.prod.js:371:12)
    at ssrRenderComponent (/Users/fil/Sites/plot/node_modules/@vue/server-renderer/dist/server-renderer.cjs.prod.js:86:10)
```

plot.c0318f47.js:5839 is our `figure.append(...legends, svg)`.

### DIFF
--- a/docs/components/PlotRender.js
+++ b/docs/components/PlotRender.js
@@ -69,6 +69,9 @@ class Element {
   dispatchEvent() {
     // ignored; interaction needs real DOM
   }
+  append(...children) {
+    for (const child of children) this.appendChild(child);
+  }
   appendChild(child) {
     this.children.push(child);
     child.parentNode = this;
--- a/docs/components/PlotRender.js
+++ b/docs/components/PlotRender.js
@@ -70,7 +70,9 @@ class Element {
     // ignored; interaction needs real DOM
   }
   append(...children) {
-    for (const child of children) this.appendChild(child);
+    for (const child of children) {
+      this.appendChild(child?.ownerDocument ? child : this.ownerDocument.createTextNode(child));
+    }
   }
   appendChild(child) {
     this.children.push(child);
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "ts-morph": "^19.0.0",
     "typescript": "^5.0.2",
     "vite": "^4.0.0",
-    "vitepress": "^1.0.0-beta.2"
+    "vitepress": "^1.0.0-rc.4"
   },
   "dependencies": {
     "d3": "^7.8.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1026,34 +1026,34 @@
   resolved "https://registry.yarnpkg.com/@vue/shared/-/shared-3.3.4.tgz#06e83c5027f464eef861c329be81454bc8b70780"
   integrity sha512-7OjdcV8vQ74eiz1TZLzZP4JwqM5fA94K6yntPS5Z25r9HDuGNzaGdgvwKYq6S+MxwF0TFRwe50fIR/MYnakdkQ==
 
-"@vueuse/core@10.2.1", "@vueuse/core@^10.2.1":
-  version "10.2.1"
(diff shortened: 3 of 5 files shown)


docs/components/PlotRender.js
```diff
@@ -69,6 +69,9 @@ class Element {
   dispatchEvent() {
     // ignored; interaction needs real DOM
   }
+  append(...children) {
+    for (const child of children) this.appendChild(child);
```
mbostock, line 73: element.append is a little fancier than this since it also automatically promotes anything that’s not a Node to a TextNode by coercing it to a string and then wrapping it. We don’t need that functionality yet, I believe, but we should either implement it preemptively or put a comment here as a reminder.

package.json
```diff
@@ -72,7 +72,7 @@
     "ts-morph": "^19.0.0",
     "typescript": "^5.0.2",
     "vite": "^4.0.0",
-    "vitepress": "^1.0.0-beta.2"
+    "vitepress": "^1.0.0-rc.4"
```
mbostock, line 75: This isn’t strictly necessary (we could have upgraded vitepress with just `yarn upgrade` and changing the yarn.lock file. But I guess since we’re using the sitemap feature added in beta.7, this is okay. The point being, `1.0.0-rc.4` is already consistent with `^1.0.0-beta.2`.

yarn.lock
```diff
@@ -3176,23 +3192,34 @@ vite@^4.0.0, vite@^4.4.6:
   optionalDependencies:
     fsevents "~2.3.2"
 
-vitepress@^1.0.0-beta.2:
-  version "1.0.0-beta.6"
-  resolved "https://registry.yarnpkg.com/vitepress/-/vitepress-1.0.0-beta.6.tgz#f5439bcbb53ab9a51f6f2b2a57d1bfad19cff476"
-  integrity sha512-xK/ulKgQpKZVbvlL4+/vW49VG7ySi5nmSoKUNH1G4kM+Cj9JwYM+PDJO7jSJROv8zW99G0ise+maDYnaLlbGBQ==
+vite@^4.4.9:
```
mbostock, line 3195: Looks like we’ve picked up multiple copies of rollup and vite in this upgrade. This is why I generally prefer to upgrade everything at once via `yarn upgrade` rather than doing it piecemeal. I’ll do that now and push to this branch.

mbostock:
I had to downgrade Vite & Vitepress in 2511ef3f43917d2d1c9886481c74179dd8a96448 due to client-side navigation being broken. (The left nav would often scroll up to the top and reset which sections were open.)

Fil:
issue reported upstream https://github.com/vuejs/vitepress/issues/2796

PR #1810, opened 2023-08-16
expose instantiated scales descriptors in the render API

Fil:
Address the most important part of #1263. In this simplest approach, we add the instantiated scales descriptors to the otherwise unmodified scales object.

the render method is thus called as 
> render(index, {x: xScale, y: yScale, color: colorScale, scales: {x, y, color}}, values, dimensions, context[, next])

where xScale, yScale, colorScale (etc) are D3 scales, and x, y, color are [instantiated scale objects](https://observablehq.com/plot/features/scales#scale).

xScale, yScale… are documented as functions that map from the data domain to the visual domain (the apply function), but still leak the D3 abstraction. The long-term plan is to replace them with the apply function only; this means chasing all the internal occurrences — for example, of xScale.bandwidth() or xScale.ticks() — and replacing them with a proper look-up in the scale descriptor (scales.x.bandwidth). (It will be a different PR.)


src/plot.js
```diff
@@ -221,9 +221,10 @@ export function plot(options = {}) {
     addScaleChannels(newChannelsByScale, stateByMark, options, (key) => newByScale.has(key));
     addScaleChannels(channelsByScale, stateByMark, options, (key) => newByScale.has(key));
     const newScaleDescriptors = inheritScaleLabels(createScales(newChannelsByScale, options), scaleDescriptors);
-    const newScales = createScaleFunctions(newScaleDescriptors);
+    const {scales: newIntantiatedScales, ...newScales} = createScaleFunctions(newScaleDescriptors);
```
mbostock, line 224: Typo “intantiated”, but also this is a new word for what we call an “exposed” scale elsewhere.

src/scales.d.ts
```diff
@@ -161,9 +161,11 @@ export type ScaleName = "x" | "y" | "fx" | "fy" | "r" | "color" | "opacity" | "s
 
 /**
  * The instantiated scales’ apply functions; passed to marks and initializers
- * for rendering.
+ * for rendering. The scales property exposes all the scale definitions.
  */
-export type ScaleFunctions = {[key in ScaleName]?: (value: any) => any};
+export type ScaleFunctions = {
+  [key in ScaleName]?: (value: any) => any;
+} & {scales: {[key in ScaleName]?: Scale}};
```
mbostock, line 168:
```suggestion
export type ScaleFunctions = {[key in ScaleName]?: (value: any) => any} & {scales: {[key in ScaleName]?: Scale}};
```

src/scales.js
```diff
@@ -97,17 +97,21 @@ export function createScales(
...
-        if (label != null) scale.label = label; // for axis
-        return [name, scale];
-      })
-  );
+export function createScaleFunctions(descriptors) {
+  const scales = {};
+  const scaleFunctions = {scales};
+  for (const [key, desc] of Object.entries(descriptors)) {
+    const {scale, type, interval, label} = desc;
+    scales[key] = exposeScale(desc);
```
mbostock, line 105: This involves a bunch of copying and these exposed scales are rarely used. I think we should investigate whether we can make this lazy using a getter (but probably a caching getter so that if you access the same scale multiple times it returns the same instance).

src/scales.js
```diff
@@ -513,21 +517,22 @@ export function scale(options = {}) {
...
 // tickFormat) are not included here as they do not affect the scale’s behavior.
 function exposeScale({scale, type, domain, range, interpolate, interval, transform, percent, pivot}) {
-  if (type === "identity") return {type: "identity", apply: (d) => d, invert: (d) => d};
+  if (type === "identity")
+    return {type: "identity", apply: (d) => d, invert: (d) => d, ...(range !== undefined && {range})};
   const unknown = scale.unknown ? scale.unknown() : undefined;
   return {
     type,
-    domain: slice(domain), // defensive copy
+    domain: [...new Set(domain)], // defensive copy, ensure uniqueness
```
mbostock, line 535: I don’t think this is appropriate here; this only applies to ordinal scales, and we should do it earlier. I’ve noticed that we render duplicate elements in color legends when the domain repeats values, and that’s definitely a bug we should fix.

test/plots/render.ts
```diff
@@ -0,0 +1,37 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+import * as htl from "htl";
+
+export async function renderInstantiatedScales() {
```
mbostock, line 5: The file name “render.ts” isn’t very descriptive.

And I think this would be better as a unit test instead of a snapshot test since most of it is about checking the exposed scales, not the visual appearance of the chart.

mbostock:
I like the progressive roll-out plan, starting here and then following up with tightening down the scale function representation to just the apply function. 👍

mbostock (commented):
This also changes _plot_.scale(…) to return the existing instantiated scale instance, rather than creating a new instance each time you call it. I think that’s okay, but we should note that in the release notes and the documentation since it’s a slight change in semantics. Possibly we should still make a defensive copy to protect against external mutation, but… I dunno, maybe it’s better to minimize the number of copies instead.

src/scales.js
mbostock: Fil’s comment: We’ll need these exposed scales in the near future anyway when we adopt them internally for axes etc., and therefore it’s premature optimization (or even slower) to make these lazy.

src/scales.js
```diff
@@ -513,21 +515,24 @@ export function scale(options = {}) {
...
 
 // Note: axis- and legend-related properties (such as label, ticks and
 // tickFormat) are not included here as they do not affect the scale’s behavior.
-function exposeScale({scale, type, domain, range, interpolate, interval, transform, percent, pivot}) {
-  if (type === "identity") return {type: "identity", apply: (d) => d, invert: (d) => d};
+function exposeScale({scale, type, range, domain, interpolate, interval, transform, percent, pivot}) {
+  // The domain and range may be missing for non-position identity scales (e.g.,
+  // color), and for position identity scales, only the range is computed
+  // internally (by autoScaleRange) and then promoted to the domain here.
+  if (type === "identity") domain = range;
```
mbostock, line 531: This change is, in a sense, optional to this PR. We’ll need it when we adopt the exposed scale representation internally to render axes for position scales, but we don’t need it yet. And actually… we don’t technically need it because the axis mark could detect when the scale is an identity scale, and then use the dimensions of the chart to re-infer the domain.

I think we should add this when we need it, so we aren’t preemptively adding something we later realize we don’t need.

mbostock (approved):

Issue #1711, opened 2023-06-22
Tracking issue for 1.0


mbostock:
- [x] Pointer transform
- [ ] #5 
- [ ] #1590
- [ ] Crossfiltering (at the very least, a template/snippet/example if not a built-in component)
- [ ] #1285 
- [ ] #630 
- [ ] #1648
- [ ] Finalize render API for extensibility

mbostock:
Sketch of new “Extending Plot” documentation page that gives an overview of all the ways Plot can be extended with custom logic (JavaScript): https://observablehq.com/d/4e732d61f48c4925

jaanli:
I’m curious what the trade-offs are of Plot’s planned interactivity versus how Mosaic accomplishes it (by downsampling views and treating marks as SQL queries to the duckdb WASM, run client-side):

https://uwdata.github.io/mosaic/

Fil:
@jaanli Happy to hear more thoughts about this, but would you open a [discussion](https://github.com/observablehq/plot/discussions) instead?

PR #1811, opened 2023-08-16
document the custom render option (#1711)

Fil:
If we do an “Extending Plot” page, maybe it should cover more than the render function?


docs/features/marks.md
```diff
@@ -553,6 +553,8 @@ All marks support the following [transform](./transforms.md) options:
 
 The **sort** option, when not specified as a channel value (such as a field name or an accessor function), can also be used to [impute ordinal scale domains](./scales.md#sort-mark-option).
 
+The **render** option allows to override or extend the default mark’s rendering method. See [extengin Plot](./extensions.md#render) for details.
```
Fil, line 556:
```suggestion
The **render** option allows to override or extend the default mark’s rendering method. See [extending Plot](./extensions.md#render) for details.
```

mbostock:
Should we also document the other public Mark methods, and add them to the Mark class TypeScript declaration? Specifically:

- the Mark constructor
- _mark_.initialize
- _mark_.filter
- _mark_project
- _mark_.scale
- _mark_.render

(I also thought about all the public fields on Mark and subclasses, such as mark.data, mark.sort, mark.initializer, _etc._ But I don’t really want those to be part of the public API. 🤔)

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
 ```
 
 A convenience method for composing a mark from a series of other marks. Returns an array of marks that implements the *mark*.plot function. See the [box mark](../marks/box.md) implementation for an example.
+
+## Custom marks
+
+Each [mark](./marks.md) has a render method that is called once for each facet (unless its data for that facet is empty). This method is responsible for drawing the mark. You can extend (or replace) this method by specifying the **render** mark option as a custom function.
```
mbostock, line 572:
```suggestion
A mark’s render method is called once for each facet, unless its data for that facet is empty. The render method is responsible for drawing the mark by producing an SVG element. You can extend or replace this method by specifying the **render** mark option as a custom function.
```

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
...
+
+:::warning
+Note that, in general, we do not recommend using this low-level interface when a more high-level option exists. It is meant for use by extension developers more than by users.
```
mbostock, line 575:
```suggestion
Note that, in general, we do not recommend using this low-level interface when a more high-level option exists such as a [data transform](https://observablehq.com/plot/features/transforms). It is meant for use by extension developers more than by users.
```

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
 ```
 
 A convenience method for composing a mark from a series of other marks. Returns an array of marks that implements the *mark*.plot function. See the [box mark](../marks/box.md) implementation for an example.
+
+## Custom marks
```
mbostock, line 570:
```suggestion
## Rendering
```

If this section is about custom marks, it should also cover extending the Mark class and implementing other Mark methods such as *mark*.initialize. But that adds scope to this PR, so let’s keep it focused on rendering.

Also, this section describes both **render functions** (functional marks, shorthand for a custom mark) and **render transforms**. I think we should make clear that these are two different concepts even though they are similar.

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
...
+:::
+
+The render function is called with the following six arguments:
+
+* *index*: the index of the facet
+* *scales*: the scale functions and descriptors
+* *values*: the scaled and raw channels
+* *dimensions*: the dimensions of the facet
+* *context*: the context
+* *next*: the next render method in the chain
```
mbostock, line 585: The *next* argument only applies to render transforms, not render functions.

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
...
+
+The function is expected to return a single SVG node, or null or undefined if the facet should be skipped. Typically, it returns a [&lt;g>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g) container, with a child node (say, a [&lt;circle>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle)) for each valid data point.
```
mbostock, line 587:
```suggestion
The function is expected to return a single SVG node, or null or undefined if no output is desired for the current facet. Typically, it returns a [G element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g), with a child node (say, a [circle element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle)) for each valid data point.
```

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
...
+
+For example, the following code will log the color associated with the Torgersen category ("#e15759") and the [instantiated color scale object](./plots.md#plot_scale), and will not render anything to the chart.
+
+```js
+Plot.dot(penguins, {
+  x: "culmen_length_mm",
+  y: "culmen_depth_mm",
+  fill: "island",
+  render(index, scales, values, dimensions, context, next) {
+    console.log(scales.color("Torgersen"));
```
mbostock, line 599:
```suggestion
    console.log(scales.color("Torgersen")); // "#e15759"
```

docs/features/marks.md
```diff
@@ -564,3 +566,137 @@ Plot.marks(
...
+152 '#f28e2c' 'Dream'
+220 '#4e79a7' 'Biscoe'
+```
+
+The *dimensions* object contains the marginTop, marginRight, marginLeft,marginBottom, and width and height of the chart. For example, to draw an ellipse that extends to the edges:
+
+```js
+Plot.plot({
+  marks: [
+    function (index, scales, values, dimensions, context, next) {
```
mbostock, line 633:
```suggestion
    function (index, scales, values, dimensions, context) {
```

PR #1822, opened 2023-08-22
Add ignoreChannels option for tip mark

mkfreeman:
Adds an `ignoreChannels` option for the tip mark that allows users to ignore channels that appear in the tooltip.


mbostock:
Thanks for the suggestion. I’m working on this in #1823, and I’m going to take a different approach rather than exposing the **ignoreChannels** option directly.

Issue #1767, opened 2023-07-23
feature request: to allow the title channel to contain rich text or markup


aborruso:
Hi,
I would like for example to have in bold the first line of tooltip (it’s a title) and to add a hyperlink at the end of the tooltip text.

I think it would be very helpful and informative.

Thank you

![image](https://github.com/observablehq/plot/assets/30607/e2b67e26-6917-484a-ad84-a88a412d2974)

yurivish:
It would be very cool if one could use a feature like the one suggested here to put a plot inside of a tooltip – eg. imagine a scatterplot where each dot represents a group of datapoints, and hovering on a datapoint can show you more details about the points inside that particular group (eg. the distributions of some other variables).

CobusT:
Saw this interesting use of `tippy` to generate html tooltips for Plot.

https://observablehq.com/d/baeb8cbead1ed3d2

hausdorff:
Just to add a bit to this (I was encouraged to post here by @Fil): it would be very nice to be able to have "aggregated" tips, where `pointerX` somehow allows us to report all values on the chart at the X position the mouse cursor exists at. Like this:

![Screenshot 2023-08-07 at 2 54 23 AM](https://github.com/observablehq/plot/assets/1409156/65caa618-c241-4eb0-91c9-f47bcc499608)

Happy to try to work on this together if people are interested in this.

Fil:
Here’s an example of grouped tips: not as advanced as what you're suggesting, but I'm using Plot to generate it so it's easy to expand with marks and transforms.
https://observablehq.com/@recifs/plot-grouped-tips

joe-bruce:
Upvoting as I also have a use case for rich text, etc tool tips. In particular, I want to be able to embed images or custom marks in the tooltip. Thanks!

mbostock:
@joe-bruce Can you please also upvote by adding a 👍 reaction to the top post? That way we can track the number of upvotes by sorting issues. Thank you.

<img width="183" alt="Screenshot 2023-08-16 at 3 30 31 PM" src="https://github.com/observablehq/plot/assets/230541/5e31dc00-e17b-41f5-a72e-32f5f0b91fa6">

mbostock:
Something relatively “easy” we could do: a custom **format** function for a channel value could return an SVG TSPAN or A element, allowing the text to be styled and hyperlinked. Essentially we’d change this line here to append the returned _value_ directly rather always wrapping it in a text node:

[...]

PR #1823, opened 2023-08-22
custom tip format

### LINKED ISSUE #1612: Custom format for the tip mark?
Like #1596, but for the tip mark.

### LINKED ISSUE #1733: Channel label option
In [this discussion](https://talk.observablehq.com/t/custom-tooltip-text-and-keeping-stroke-colour/8068/6) we realize that to specify a channel label (for the interactive tips #1612), we need either a {transform, label} _value_,  or an arcane combination of an accessor function with a label property:
~~~
channels: {year: Object.assign((d) => d.year, {label: "My custom year label"})} 
~~~

This is because labelof only reads the *value*.

It would be nice to be allowed to write instead:
~~~
channels: {year: {value: "year", label: "My custom year label"}}
~~~

Note that this is not strictly necessary for the given use case, since we can also name the channel as we want (although not with a string which has precedence over the channel’s key name):
~~~
channels: {["My custom year label"]: d => d.year}
~~~

(maybe related to #1364)

mbostock:
This implements a new **format** option for the tip mark that specifies how to format values. The option is specified as an object whose property names are channel names and whose values are formats; each format may either be a shorthand string for a number or time format, or a function that takes a channel value and zero-based index and returns the corresponding formatted string.

For example, to customize the number format of the *x* channel:

```js
format: {x: ".2f"}
```

Equivalently:

```js
format: {x: (d) => d.toFixed(2)}
```

To suppress a channel (similar to #1822), set the corresponding format to null:

```js
format: {x: null}
```

As before, if you want the tip to display a paragraph of text, you can use the **title** channel; if you want to add additional name-value pairs to the tip, you can declare extra channels using the **channels** option; if you want to change the label, you can either change the corresponding channel or scale label. This PR improves the **channels** option slightly, allowing the **label** option to be declared alongside a channel #1734, and defaulting the channel label to its name when expressed as a shorthand string (e.g., `channels: {Sport: "sport"}` now uses the label “Sport” instead of “sport”.)

Lastly, the **tip** mark option now supports passing arbitrary options through to the derived tip mark, including the **format** option. The **tip** mark option also now takes a **pointer** option to control which pointer mode is used (_x_, _y_, or _xy_, for pointerX, pointerY, or pointer respectively).

Fixes #1612.
Fixes #1733.
Fixes #1734.

### DIFF
--- a/src/marks/tip.js
+++ b/src/marks/tip.js
@@ -350,14 +350,10 @@ function getSourceChannels({channels}, scales) {
       const value = sources[key]?.value ?? scales[key]?.domain() ?? [];
       this.format[key] = (isTemporal(value) ? utcFormat : numberFormat)(format);
     } else if (format === undefined || format === true) {
-      // Borrow the scale’s tick format for facet channels; this is
-      // generally better than the default (and safe for ordinal scales).
-      if (key === "fx" || key === "fy") {
-        const scale = scales[key];
-        this.format[key] = inferTickFormat(scale, scale.domain());
-      } else {
-        this.format[key] = formatDefault;
-      }
+      // For ordinal scales, the inferred tick format can be more concise, such
+      // as only showing the year for yearly data.
+      const scale = scales[key];
+      this.format[key] = scale?.bandwidth ? inferTickFormat(scale, scale.domain()) : formatDefault;
     }
   }
 
--- a/src/marks/tip.js
+++ b/src/marks/tip.js
@@ -124,26 +124,6 @@ export class Tip extends Mark {
     } else {
       sources = getSourceChannels.call(this, index, values, scales);
       format = formatChannels;
-
-      // Promote shorthand string formats to functions, and materialize default
-      // formats. Note: this mutates this.format, but that should be safe since
-      // we made a defensive copy.
-      for (const key in sources) {
-        const format = this.format[key];
-        if (typeof format === "string") {
-          const value = sources[key]?.value ?? scales[key]?.domain() ?? [];
-          this.format[key] = (isTemporal(value) ? utcFormat : numberFormat)(format);
-        } else if (format === undefined || format === true) {
-          // Borrow the scale’s tick format for facet channels; this is
-          // generally better than the default (and safe for ordinal scales).
-          if (key === "fx" || key === "fy") {
-            const scale = scales[key];
-            this.format[key] = inferTickFormat(scale, scale.domain());
-          } else {
-            this.format[key] = formatDefault;
-          }
-        }
-      }
     }
 
     // We don’t call applyChannelStyles because we only use the channels to
@@ -327,29 +307,68 @@ function getPath(anchor, m, r, width, height) {
   }
 }
 
+// Note: mutates this.format!
 function getSourceChannels(index, {channels}, scales) {
   const {facet, format} = this;
   const sources = {};
+
   // Prioritize channels with explicit formats, in the given order.
   for (const key in format) {
-    if (format[key] === null || format[key] === false) continue;
-    if (key === "fx" || key === "fy") sources[key] = true;
-    else {
-      const source = getSource(channels, key);
+    const value = format[key];
+    if (value === null || value === false) {
+      // Promote x and y null shorthand for paired channels (in order).
+      if (key === "x" || key === "y") {
+        const key1 = `${key}1`;
+        const key2 = `${key}2`;
+        if (!(key1 in format)) format[key1] = null;
+        if (!(key2 in format)) format[key2] = null;
+      }
+    } else if (key === "fx" || key === "fy") {
+      sources[key] = true;
+    } else {
+      let source = getSource(channels, key);
       if (source) sources[key] = source;
+      // Promote x and y non-null shorthand for paired channels (in order).
+      if (key === "x" || key === "y") {
+        const key1 = `${key}1`;
+        const key2 = `${key}2`;
+        if (!(key1 in format) && (source = getSource(channels, key1))) (format[key1] = value), (sources[key1] = source);
+        if (!(key2 in format) && (source = getSource(channels, key2))) (format[key2] = value), (sources[key2] = source);
+      }
     }
   }
+
   // Then fallback to all other (non-ignored) channels.
   for (const key in channels) {
     if (key in sources || key in format || ignoreChannels.has(key)) continue;
     const source = getSource(channels, key);
     if (source) sources[key] = source;
   }
+
   // And lastly facet channels, but only if this mark is faceted.
   if (facet) {
     if (scales.fx && !("fx" in format)) sources.fx = true;
     if (scales.fy && !("fy" in format)) sources.fy = true;
   }
+
+  // Promote shorthand string formats, and materialize default formats.
+  for (const key in sources) {
+    const format = this.format[key];
+    if (typeof format === "string") {
+      const value = sources[key]?.value ?? scales[key]?.domain() ?? [];
+      this.format[key] = (isTemporal(value) ? utcFormat : numberFormat)(format);
+    } else if (format === undefined || format === true) {
+      // Borrow the scale’s tick format for facet channels; this is
+      // generally better than the default (and safe for ordinal scales).
+      if (key === "fx" || key === "fy") {
+        const scale = scales[key];
+        this.format[key] = inferTickFormat(scale, scale.domain());
+      } else {
+        this.format[key] = formatDefault;
+      }
+    }
+  }
+
   return sources;
 }
 
@@ -372,12 +391,12 @@ function* formatChannels(i, index, channels, scales, values) {
     if (key === "x2" && "x1" in channels) {
       yield {
         label: formatPairLabel(scales, channels, "x"),
-        value: formatPair(this.format.x, channels.x1, channel, i)
+        value: formatPair(this.format.x2, channels.x1, channel, i)
       };
     } else if (key === "y2" && "y1" in channels) {
       yield {
         label: formatPairLabel(scales, channels, "y"),
-        value: formatPair(this.format.y, channels.y1, channel, i)
+        value: formatPair(this.format.y2, channels.y1, channel, i)
       };
     } else {
       const value = channel.value[i];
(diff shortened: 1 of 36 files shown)


Fil:
Love the API design.

color, opacity… can we have symbol too, maybe? :-)

mkfreeman:
Very exciting! Would it be possible to pass up the `format` option from other marks in the `inferTip` [function](https://github.com/observablehq/plot/blob/main/src/plot.js#L523)? The first thing I tried to do was use a tip formatter with another mark where `tip: true`. 

I think it would be a big win for new-ish users to be able to format their tips without ejecting from `tip: true` to adding another tip mark (might also make it easier to add the UI in the chart cell?).

mbostock:
Yes @mkfreeman; see the top description.

mkfreeman:
Does it make sense to put the swatch at the beginning of each line (to match legends)? Just wanted to compare:

| Start  | Middle | end |
| ------------- | ------------- | ------------- |
| beginning |middle |  end|

mbostock:
@mkfreeman I’m not changing the display functionality of tips in this PR; I’m just giving more control how the contents of the tip are specified. See #1767 for a more general issue, but feel free to open another issue if you want to suggest changes to the default design of tips.

mkfreeman:
@mbostock thanks, will take a look at that issue!

mbostock:
I’m wondering if I over-engineered this by adding the ability to derive anonymous channels as part of the format. I could do something that is more focused on just how the channels are formatted. For example, instead of this:

```js
format: [{channel: "x", format: ".2f"}]
```

maybe you’d say:

```js
format: {x: ".2f"}
```

and it wouldn’t change which channels were displayed, but it would change how the _x_ channel is formatted (or perhaps how any channel bound to the _x_ scale is formatted? not sure). And maybe you could suppress a channel by saying:

```js
format: {x: ".2f", y: null}
```

I think that might be enough. 🤔

mbostock:
After much exploration, I’ve decided to pare this PR down to its essentials, focusing solely on formatting channel values and suppressing channels. The other functionality (declaring additional channels, and overriding labels) is already covered by the existing API, so I didn’t want to introduce a competing mechanism.

src/marks/tip.js
```diff
@@ -332,18 +334,73 @@ function getSources({channels}) {
...
+    } else if (key === "y2" && "y1" in channels) {
+      const format = this.format?.y; // TODO y1, y2?
+      if (format === null) continue;
+      yield {
+        label: formatPairLabel(scales, channels, "y"),
+        value: formatPair(format ?? formatDefault, channels.y1, channel, i)
+      };
+    } else {
+      const format = this.format?.[key];
+      if (format === null) continue;
```
mbostock, line 363: Possible future improvement: we could prune the result of getSources based on this.format; then we wouldn’t need to skip here. Also, we could materialize the formatDefault as needed.

src/plot.js
```diff
@@ -239,6 +239,7 @@ export function plot(options = {}) {
   // Compute value objects, applying scales and projection as needed.
   for (const [mark, state] of stateByMark) {
     state.values = mark.scale(state.channels, scales, context);
+    state.values.data = state.data; // expose transformed data for advanced usage
```
Fil, line 242: It will be immensely useful in the render API to always have access to mark’s (arrayified, possibly transformed) data. I just wonder why it appears here, as it seems unrelated to this PR. Needs to be added to the documentation #1811.

Fil:
Great!! Besides the missing documentation, two minor (non-blocking) remarks:

1. I was tempted to say `format: {x: true} `to just use the standard format (for instance, when I wanted to revert `{x: null}`).


2. Looks like this enables to set the order of the fields by setting the order in the channels options, and possibly suppressing the corresponding standard channels. For example:

```
Plot.dot(penguins, {
  x: "body_mass_g",
  y: "culmen_length_mm",
  fill: "species",
  channels: {
    culmen_length_mm: "culmen_length_mm",
    species: { value: "species", label: "species", scale: "color" },
    body_mass_g: "body_mass_g"
  },
  tip: { format: { fill: null, culmen_length_mm: ".1f", x: null } }
}).plot()
```

results in the order (culmen_length_mm, species, body_mass_g)

Capture d’écran 2023-08-25 à 11 07 14

I feel that this is a bit arbitrary and hacky (and also bad performance-wise). Since it's something that I think users will want to do anyways, it would be nicer to design this such that it followed the order given in the _format_ instead. This would be an additional reason to support `{x: true}`, btw.

Fil:
To expand my remark about symbols: it would be great to have a symbol (where there is a swatch) for channels that use the symbol scale. And the combo color+symbol should apply too. Again not a blocker, I can try to do this separately.

src/plot.js
mbostock: This was needed in an earlier variation of this PR where you could specify the format as a function of data. I don’t think it’s needed anymore, so I can remove it and we can add it as a separate PR when we are ready.

mbostock:
Let’s do the symbol swatch separately. 

The order of channels is hard and not addressed in this PR. The formats are sparse and I think using “true” as a format is weird — that doesn’t work for a tickFormat, for example. I think channel order might need to be a separate PR or option too, but I’ll think about it today. (I’ve already spent a lot of time on this and want to move forward to other things for 1.0.)

mbostock:
I could change it to do whatever is listed in the **format** option first, followed by any remaining channels (in arbitrary order). That seems like a reasonable improvement although it might complicate the code some. We will see! I can probably address the improvement I mentioned above at the same time. But then I want to move on!

mbostock:
The **format** option now determines the order in which channels are shown, and you can set the format to true (or equivalently undefined) if you just want to use it to specify the order. I also added the optimization I described above to materialize the defaults, fixed a bug with facet settings (**facet** and **facetAnchor**) for the **tip** mark option, and removed the _data_ pseudochannel.

Still need documentation and tests, I guess. Maybe that could be a followup PR? 😬

src/marks/tip.js
```diff
@@ -322,28 +327,86 @@ function getPath(anchor, m, r, width, height) {
...
+      };
+      continue;
+    }
+    if (key === "x1" && "x2" in channels) continue;
+    if (key === "y1" && "y2" in channels) continue;
+    const channel = channels[key];
+    if (key === "x2" && "x1" in channels) {
+      yield {
+        label: formatPairLabel(scales, channels, "x"),
+        value: formatPair(this.format.x, channels.x1, channel, i)
```
mbostock, line 375: I think this might approach not work now, at least if you use a shorthand string for the _x_ or _y_ format, because the name of the format (*x*) does not correspond to the name of the channel (*x1* and *x2*). I could switch to using the *x2* here, but I think supporting *x* is more natural since the same format is used for both channels. 🤔

Fil: Ah yes this breaks the tipAreaBand, tipAreaStack tests (and maybe others), because format.y is undefined. (Too bad we can't catch those automagically.)

Fil: I need to run, but here's a fix:

```patch
diff --git a/src/marks/tip.js b/src/marks/tip.js
index 0a271e1d..97fa52bb 100644
--- a/src/marks/tip.js
+++ b/src/marks/tip.js
@@ -372,12 +372,12 @@ function* formatChannels(i, index, channels, scales, values) {
     if (key === "x2" && "x1" in channels) {
       yield {
         label: formatPairLabel(scales, channels, "x"),
-        value: formatPair(this.format.x, channels.x1, channel, i)
+        value: formatPair(this.format.x1 ?? this.format.x, this.format.x2 ?? this.format.x, channels.x1, channel, i)
       };
     } else if (key === "y2" && "y1" in channels) {
       yield {
         label: formatPairLabel(scales, channels, "y"),
-        value: formatPair(this.format.y, channels.y1, channel, i)
+        value: formatPair(this.format.y1 ?? this.format.y, this.format.y2 ?? this.format.y, channels.y1, channel, i)
       };
     } else {
       const value = channel.value[i];
@@ -393,10 +393,10 @@ function* formatChannels(i, index, channels, scales, values) {
   }
 }

-function formatPair(formatValue, c1, c2, i) {
+function formatPair(formatValue1, formatValue2, c1, c2, i) {
   return c2.hint?.length // e.g., stackY’s y1 and y2
-    ? `${formatValue(c2.value[i] - c1.value[i], i)}`
-    : `${formatValue(c1.value[i], i)}–${formatValue(c2.value[i], i)}`;
+    ? `${formatValue1(c2.value[i] - c1.value[i], i)}`
+    : `${formatValue1(c1.value[i], i)}–${formatValue2(c2.value[i], i)}`;
 }

 function formatPairLabel(scales, channels, key) {

```

It seems to work on tipAreaStack and tipBin (I can format separately x1 and x2). I haven't investigated what it gives on tips that show a length/difference instead of an extent (?).

mbostock: I pushed what I think is a more general fix, allowing you to specify a format for the _x_ channel as shorthand for both _x1_ and _x2_ (and same for _y_). This means you can control the order in which paired channels appear, and even suppress one of the two paired channels to just show the other. For example, `format: {y1: false}` will just show the high:

Screenshot 2023-08-25 at 12 52 59 PM

And `format: {y: true}` will give _y1_ and _y2_ higher priority:

Screenshot 2023-08-25 at 12 53 37 PM

Fil (approved):

src/marks/tip.js
```diff
@@ -322,28 +307,125 @@ function getPath(anchor, m, r, width, height) {
...
+      sources[key] = true;
+    } else {
+      let source = getSource(channels, key);
+      if (source) sources[key] = source;
+      // Promote x and y non-null shorthand for paired channels (in order).
+      if (key === "x" || key === "y") {
+        const key1 = `${key}1`;
+        const key2 = `${key}2`;
+        if (!(key1 in format) && (source = getSource(channels, key1))) (format[key1] = value), (sources[key1] = source);
+        if (!(key2 in format) && (source = getSource(channels, key2))) (format[key2] = value), (sources[key2] = source);
```
mbostock, line 336: Ugh, this only works the first time the mark is rendered. Because we mutate `format` here, thus adding the paired channels to the end, we’re changing the order the second time around. 🤦 I need to do this in a way that puts the paired channels (_x1_ and _x2_) at the same priority as the shorthand (_x_).

dperetti:
Is it supposed to work for default channels? I'm not able to suppress / reorder default those, only works for extra channels.

mbostock:
@dperetti Yes, it does. Please open a discussion here or in the Observable Forum if you’d like help, and please include a live code example to demonstrate. Thank you.

dperetti:
Thanks, finally got it! Turns out the channels were `x` and `y`, not the keys in the data object. This was too much convolution for my aging brain.

PR #1825, opened 2023-08-23
tests for coverage

Fil:
Now that we have test:coverage, it becomes easier to add more tests.

From the starting point of (#1818):
```
----------------------|---------|----------|---------|---------|
File                  | % Stmts | % Branch | % Funcs | % Lines |
----------------------|---------|----------|---------|---------|
All files             |   95.54 |    89.37 |   92.78 |   95.54 |
```

Current status:

```
All files             |   95.66 |    89.65 |    92.8 |   95.66 |
```

(It doesn’t look like much, but each addition is worth it!)


mbostock (commented):
I made some changes:

1. I folded the tests into the parent class or function. For example we already had tests for the line mark’s **curve** option, so I moved these new tests to be adjacent.
2. I changed the test descriptions to match the current pattern (which is mostly `method(...arguments) does something`).
3. I adopted TypeScript for the tests whenever possible. (We should file an issue for going back and moving all the tests to TypeScript, but we’ll need to use a lot of `as any` to test the undocumented implementation.)
4. I adopted [jsdom.js](https://github.com/observablehq/plot/blob/main/test/jsdom.js); the eliminates the need for instantiating JSOM and manually passing the **document** option to Plot.

mbostock (approved):

PR #1826, opened 2023-08-23
fix tip flicker

### LINKED ISSUE #1805: Flickering tip
I recently built a page that has some charts using `tip`s as chart annotations controlled by a range slider, and noticed that the tooltips sometimes "flicker" before being rendered at their final position. I was following along with tip development a while back and if I remember, there's some layout computation that happens post-initial render, so I wonder if it might have to do with that.

I created a mini test case [here](https://observablehq.com/d/6ea54bc9b4b83de8). It's hard to get a screenshot, but in Chrome & Safari on macOS I see the tip text appear at the bottom right of the annotated datum for just a moment while I slide the slider, before re-appearing at the top left as requested by the `anchor` option, along with the rest of the tip.

Fil:
cc: @yurivish 

closes #1805


src/marks/tip.js
```diff
@@ -155,6 +155,7 @@ export class Tip extends Mark {
       .call(applyIndirectStyles, this, dimensions, context)
       .call(applyIndirectTextStyles, this)
       .call(applyTransform, this, {x: X && x, y: Y && y})
+      .style("visibility", "hidden") // avoid flickering
```
mbostock, line 158: We should prefer attr here instead of style. (We only use style when attr is not supported.)

```suggestion
      .attr("visibility", "hidden") // hide until postrender
```

src/marks/tip.js
```diff
@@ -241,6 +242,7 @@ export class Tip extends Mark {
         text.setAttribute("y", `${+getLineOffset(a, text.childNodes.length, lineHeight).toFixed(6)}em`);
         text.setAttribute("transform", `translate(${getTextTranslate(a, m, r, w, h)})`);
       });
+      g.style("visibility", null);
```
mbostock, line 245:
```suggestion
      g.attr("visibility", null);
```

mbostock (changes requested):
We should use attr instead of style.

Also, looking at the diff makes me realize that we know at render time whether the tip is empty, and hence we could avoid calling postrender entirely in that case. I’m going to take a quick look at that now.

src/marks/tip.js
Fil: My thinking was that this would be orthogonal if we wanted to support the visibility attribute. But yeah, we can see this when we (or… if we ever) support it.

PR #1827, opened 2023-08-23
pass fontVariant to the axis label

### LINKED ISSUE #1821: fontVariant should be applied to axis labels
In https://observablehq.com/@recifs/plot-large-axis-label--support, try to add fontVariant: "small-caps" to the second axis with the huge label.

related: #1414

Fil:
closes #1821


mbostock:
This omission was somewhat intentional: the most common use of **fontVariant** is for tabular-nums with numeric or temporal ticks so that the numerals align across ticks. That need does not apply to the axis label, since there is only one label per axis and hence no need for the numerals to have the same width — and indeed we prefer _not_ to have tabular-nums because numerals look better with the default variable width.

Though, I guess it might be confusing if some of the font-related options apply to both the ticks and the label, but some don’t. We could have separate options (**tickFontVariant**, **labelFontVariant**, and **tickFontSize**, **labelFontSize**, etc.). Though maybe it’d be better to just have separate marks… though there we already have a tick mark that is something else too? 😭

test/plots/axis-labels.ts
```diff
@@ -52,6 +52,12 @@ export async function axisLabelBothReverse() {
   });
 }
 
+export async function axisLabelFontVariant() {
+  return Plot.plot({
+    x: {domain: "abcde"},
+    marks: [Plot.axisX({label: "Letter", fontVariant: "small-caps"})]
+  });
+}
```
mbostock, line 60:
```suggestion
}

```

mbostock (approved):
We could always introduce more fine-grained options in the future if we need it.

PR #1828, opened 2023-08-23
the denseInterval option for k must also reduce k1 and k2 if present

### LINKED ISSUE #1784: The areaY mark’s interval option doesn’t work with y1 and y2
The maybeDenseIntervalX transform used by areaY’s **interval** option uses the binX transform, but only outputs the **y** channel:

https://github.com/observablehq/plot/blob/194e1f46fa432d8666c4f05099f1a7c534dd911d/src/transforms/bin.js#L70

This means it doesn’t work if the areaY mark is using the **y1** and **y2** options:

untitled (49)

```js
Plot.areaY(aapl.slice(0, 81), {
  x: "Date",
  y1: "Low",
  y2: "High",
  interval: "day",
  curve: "step"
}).plot()
```

It should instead be equivalent to this:

untitled (50)

```js
Plot.areaY(
  aapl.slice(0, 81),
  Plot.binX(
    { y1: "first", y2: "first", filter: null },
    { x: "Date", y1: "Low", y2: "High", interval: "day", curve: "step" }
  )
).plot()
```

So somehow maybeDenseIntervalX will need to know to output **y1** and **y2** instead.

Ref. https://observablehq.com/d/39a059b4b6322d37

Fil:
closes #1784


src/transforms/bin.js
```diff
@@ -65,9 +65,12 @@ export function bin(outputs = {fill: "count"}, options = {}) {
 }
 
 function maybeDenseInterval(bin, k, options = {}) {
-  return options?.interval == null
-    ? options
-    : bin({[k]: options?.reduce === undefined ? reduceFirst : options.reduce, filter: null}, options);
+  if (options?.interval == null) return options;
+  const reduce = options?.reduce === undefined ? reduceFirst : options.reduce;
```
mbostock, line 69: Since `options?.interval` is non-nullish here, `options` is guaranteed to be non-nullish, and we can remove the optional chaining. Hence, this can be written as:

```suggestion
  const {reduce = reduceFirst} = options;
```

src/transforms/bin.js
```diff
@@ -65,9 +65,12 @@ export function bin(outputs = {fill: "count"}, options = {}) {
...
+  const outputs = {[k]: reduce, filter: null};
```
mbostock, line 70: Do we want to check if `options[k] != null`, too?

mbostock (approved):

PR #1830, opened 2023-08-24
fix symbol set hint when fill is a constant currentColor

### LINKED ISSUE #1462: The correct default symbol set is only inferred when fill is a channel, not a constant
For example, this works fine:

Screenshot 2023-04-17 at 6 40 12 PM

```js
Plot.dotX([
  "circle",
  "cross",
  "diamond",
  "square",
  "star",
  "triangle",
  "wye"
], {fill: () => "currentColor", symbol: (d, i) => i}).plot()
```

But this does not (because it uses the stroked symbols, even though the dots are filled):

Screenshot 2023-04-17 at 6 40 50 PM

```js
Plot.dotX([
  "circle",
  "cross",
  "diamond",
  "square",
  "star",
  "triangle",
  "wye"
], {fill: "currentColor", symbol: (d, i) => i}).plot()
```

The dots can be made visible by applying a stroke:

Screenshot 2023-04-17 at 6 41 34 PM

```js
Plot.dotX([
  "circle",
  "cross",
  "diamond",
  "square",
  "star",
  "triangle",
  "wye"
], {fill: "currentColor", stroke: "red", symbol: (d, i) => i}).plot()
```

Fil:
when we specify fill: "currentColor", this.fill is undefined, but we still want to pass "currentColor" as a hint to the symbol scale

closes #1462


mbostock (commented):
Shouldn’t our handling of _fill_ and _stroke_ be the same?

mbostock (approved):

PR #1831, opened 2023-08-24
The geometry channels accepts a channelSpec 

### LINKED ISSUE #1673: The geo mark ignores scale overrides for the geometry channel
The geo mark should check whether the geometry channel should be scaled before applying the projection.

Fil:
The geometry channels accepts a specialized channelSpec `{value: …, scale: null}` that allows a geo mark to opt out of the projection.

TODO:
- [ ] we need to do the same for the centroid transform.

closes #1673


src/marks/geo.d.ts
```diff
@@ -8,7 +8,7 @@ export interface GeoOptions extends MarkOptions {
    * A required channel for the geometry to render; defaults to identity,
    * assuming *data* is a GeoJSON object or an iterable of GeoJSON objects.
    */
-  geometry?: ChannelValue;
+  geometry?: ChannelValueSpec | {value: ChannelValue; scale?: null | "projection"};
```
mbostock, line 11: We should also allow boolean and _auto_.

src/marks/geo.d.ts
```diff
@@ -8,7 +8,7 @@ export interface GeoOptions extends MarkOptions {
    * A required channel for the geometry to render; defaults to identity,
    * assuming *data* is a GeoJSON object or an iterable of GeoJSON objects.
    */
-  geometry?: ChannelValue;
+  geometry?: ChannelValueSpec | {value: ChannelValue; scale?: null | boolean | "auto" | "projection"};
```
mbostock, line 11: I think we should add _projection_ to ScaleName so we can use ChannelValueSpec here. Otherwise you’ll have to also add the label property that was just added. And it’s still the case that this produces a Channel, and hence we need to allow _projection_ in Channel["scale"].

src/marks/geo.js
```diff
@@ -21,18 +21,17 @@ export class Geo extends Mark {
     const [vr, cr] = maybeNumberChannel(options.r, 3);
     super(
       data,
-      {
-        geometry: {value: options.geometry, scale: "projection"},
-        r: {value: vr, scale: "r", filter: positive, optional: true}
-      },
+      {geometry: maybeGeometry(options.geometry), r: {value: vr, scale: "r", filter: positive, optional: true}},
```
mbostock, line 24: Style nit: would still prefer each channel to be on a separate line.

src/marks/geo.js
```diff
@@ -95,3 +94,19 @@ export function sphere({strokeWidth = 1.5, ...options} = {}) {
 export function graticule({strokeOpacity = 0.1, ...options} = {}) {
   return geo(geoGraticule10(), {strokeOpacity, ...options});
 }
+
+export function maybeGeometry(geometry) {
+  const {value, scale} = geometry?.value ? geometry : {value: geometry};
```
mbostock, line 99: I expect we want to use maybeValue here for consistency.

mbostock: More importantly, it should be necessary to apply the scale override here. Scale overrides are applied here:

https://github.com/observablehq/plot/blob/26b5d14ca94697efb1229c38b2ebbc3fb4318291/src/mark.js#L48-L52

We probably need to add some logic to inferChannelScale to handle the _projection_ scale being applied to the _geometry_ channel.

mbostock (changes requested):

PR #1833, opened 2023-08-24
Auto: error on underspecified reduce

tophtucker:
Watching people use the chart cell, they often set a reducer without understanding what it’s reducing. Min _of what_, max _of what_, mean _of what_? Setting most reducers without a field is incoherent; even if it produces something, it’ll be misleading. But people often don’t realize they have to do something else, because they see some output, and scratch their heads trying to interpret it.

Since it’s easy to statically tell if a configuration is invalid like this, this PR throws an error in those cases. Demo: https://observablehq.com/d/932dd87a4c129e16

image

I tried to think of a more plain-English way to say this. We could do something like `Error: ${channel} is "${reducer}" of what? reducer requires ${channel} field`. People don’t know the word “reducer”, but they know the prepositional relationship expressed in “mean” _of_ “height”. It’s gotta be _of_ something.

Broken out from https://github.com/observablehq/plot/pull/1424.

### DIFF
--- a/src/marks/auto.js
+++ b/src/marks/auto.js
@@ -15,6 +15,11 @@ import {ruleX, ruleY} from "./rule.js";
 export function autoSpec(data, options) {
   options = normalizeOptions(options);
 
+  for (const channel of ["x", "y", "color", "size"]) {
+    if (isUnderspecifiedReduce(options[channel]))
+      throw new Error(`setting ${channel} reducer to "${options[channel].reduce}" requires setting ${channel} field`);
+  }
+
   // Greedily materialize columns for type inference; we’ll need them anyway to
   // plot! Note that we don’t apply any type inference to the fx and fy
   // channels, if present; these are always ordinal (at least for now).
@@ -338,6 +343,10 @@ function isReducer(reduce) {
       return true;
   }
   return false;
+}
+
+function isUnderspecifiedReduce({value, reduce}) {
+  return value === undefined && reduce !== undefined && reduce !== "count";
 }
 
 function isHighCardinality(value) {


src/marks/auto.js
```diff
@@ -340,6 +345,10 @@ function isReducer(reduce) {
   return false;
 }
 
+function isUnderspecifiedReduce({value, reduce}) {
+  return value === undefined && reduce !== undefined && reduce !== "count";
+}
```
mbostock: Suggestion:

```suggestion
function isUnderspecifiedReduce({value, reduce}) {
  return value == null && reduce != null && /^count$/i.test(reduce);
}
```

This:

- Treats null and undefined as the same for _value_ and _reduce_
- Coerces _reduce_ to a string
- Does case-insensitive match equivalent to the keyword helper

mbostock (commented):
The behavior seems desirable.

But I wonder if we could solve this at a lower level: would it be possible for the bin and group transform to detect this instead, when you’ve specified a reducer that requires an input channel but not provided said channel? I think it would be a matter of denoting whether a reducer requires an input channel when it is declared — or perhaps conversely, when the input is optional, since _count_ is really the exception rather than the default, and then enforcing that the input exists when the evaluator is initialized…

Oh, wait. 🤔💡 The reason this isn’t an error is because if the input channel doesn’t exist, then the data itself is fed to the reducer in lieu of the channel:

https://github.com/observablehq/plot/blob/1ff36a3ff859354cb16c4284d0811ed2e804c7dc/src/transforms/group.js#L212-L217

So, if you have an array of numbers as data, you should be able to feed that directly to the _median_ reducer without specifying an input channel. And hence this enforcement is limiting desirable shorthand functionality. (I don’t think the chart cell allows arrays of primitives to be data sources?)

I think either we have to solve this more generally #493 where we detect that the materialized input is entirely nullish or NaN, or we do a more specific heuristic here, like bypass the check if _data_ is an array of primitives (numbers, dates, strings, booleans, etc.).?

src/marks/auto.js
tophtucker: nice!! ive always wondered why you use regexes in this sort of helper, this is a good explanation

tophtucker:
>Oh, wait. 🤔💡 The reason this isn’t an error is because if the input channel doesn’t exist, then the data itself is fed to the reducer in lieu of the channel:

ohhh ya. makes sense. plot.auto doesn't support shorthand / primitive arrays, though i've always wanted it to:

image

so this error wouldn't limit plot.auto any more than it's already limited. though it'd add some inertia to the limit.

>Another potential direction is issue a warning when the input is more than 50% nullish or NaN

is the warning approach only bc of the shorthand case? at one point i think you described to me that things should be errors when they're statically incoherent in the options, or warnings if they're about the dynamic data. for plot.auto as it stands today it feels more like an error.

>(https://github.com/observablehq/plot/issues/1192, although that issue presupposes a specific solution and we may want to consider other approaches).

yeah good aside, i've rewritten that issue description to be more agnostic 👍

mbostock:
> plot.auto doesn't support shorthand / primitive arrays, though i've always wanted it to

Plot.auto doesn’t support _shorthand_ (you have to specify some options), but it _does_ support primitive arrays:

untitled (66)

```js
Plot.auto(d3.range(1000).map(d3.randomNormal()), {x: Plot.identity}).plot()
```

tophtucker:
took a stab at it throwing only if not primitive (using a simpler primitivity check than stdlib table stuff). doesn't feel great to iterate over the data once more each time 🤷

src/marks/auto.js
```diff
@@ -346,7 +348,11 @@ function isReducer(reduce) {
 }
 
 function isUnderspecifiedReduce({value, reduce}) {
-  return value == null && reduce != null && /^count$/i.test(reduce);
+  return value == null && reduce != null && !/^count$/i.test(reduce);
+}
+
+function isPrimitive(values) {
+  return isEvery(values, (d) => ["number", "boolean", "string"].includes(typeof d) || d instanceof Date);
```
mbostock, line 355: Couple nits here.

First, isEvery requires values to be an iterable, but mark data can also be an “arrayish” value that is supported by Array.from such as {length: 10}. So, we’d need to handle that case, too. I recommend protecting the call as `isIterable(data) && !isPrimitive(data)` (since any data is not iterable by definition can only contain undefined, as in a sparse array, which we would not consider to be primitive).

Second, we should also treat bigints as primitive. (There’s also the *symbol* type to consider. I guess you could use the _mode_ reducer on non-primitive values? But that’s obscure enough that I wouldn’t worry about it. If you’re doing these things you probably can debug the error.)

Last, I would avoid array.includes(type) in the inner loop. Unless the JavaScript engine is smart, it’ll have to allocate that array afresh for each value, and the call to array.includes is also probably slower than an inline logical expression. It’s better to just write this by hand; see isOrdinal for an example.

test/marks/auto-test.js
```diff
@@ -243,3 +243,42 @@ it("Plot.autoSpec makes a faceted heatmap", () => {
     colorMode: "fill"
   });
 });
+
+it.only("autoSpec rejects an underspecified reducer", () => {
```
mbostock, line 247: ☝️ it.only

tophtucker:
@mbostock I think I’ve addressed all comments:

- Check isIterable before isPrimitive
- Treat bigint as primitive
- Write out "or" by hand instead of array.includes
- Fix it.only

Fil:
Codewise this lgtm. But can we rephrase the error message a little bit? `setting y reducer to "mean" requires setting y field` uses "reducer" and "field", but the properties are named **reduce** and **value**. Maybe something like “a value is required when setting the reduce to "mean"”, “provide a value to reduce y to "mean"”, “missing y value to reduce with "mean"”, or just “missing value for y”.

PR #1836, opened 2023-08-25
document plot extensibility

Fil:
There are a few todos… it's a lot of prose, and a bit of code.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -70,7 +70,8 @@ export default defineConfig({
           {text: "Formats", link: "/features/formats"},
           {text: "Markers", link: "/features/markers"},
           {text: "Shorthand", link: "/features/shorthand"},
-          {text: "Accessibility", link: "/features/accessibility"}
+          {text: "Accessibility", link: "/features/accessibility"},
+          {text: "Extensibility", link: "/features/extensibility"}
         ]
       },
       {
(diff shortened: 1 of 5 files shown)


mbostock:
An initial thought…

Actual difficulty is unique to each reader because it depends on the reader’s own experience and way of thinking. Regardless of what we estimate to be easy or hard, each reader will feel differently, and if our estimations don’t match, it can be discouraging or alienating.

So instead of classifying by difficulty, how about we classify by commonness or expected relevance? How likely is it that the reader will want to extend it? If we rate things by how rare or common they are, we can put the common stuff at the top to prioritize what is more likely to be relevant. It’s okay, even expected, if readers are interested in uncommon things. For example, a scale transform, bin reducer, channel transform, and custom interval implementation might all be the same assumed level of difficulty 🌶️, but I expect that the scale transform and bin reducer will be more common. 

Another approach would be to group extensibility by topic area: data transformations, visual encoding, mark rendering, interaction, etc. Then within these groups, we can try to order by commonness. The current flat list feels a bit daunting!

Fil:
Yeah I don't think there should be so many 🌶, it was more a manner of mapping these in my head. 

Maybe we could keep just one emoji when something can be deemed more difficult, like Knuth does in the [TeX book](https://visualmatheditor.equatheque.net/doc/texbook.pdf) with curvy road signs:
TEX Book

The topic/commonness structure seems more natural and allows to find what you're looking for more quickly — then you look for the simplest solution that might help in that category.

PR #1837, opened 2023-08-26
Empty bins fall back on the first element of their group

### LINKED ISSUE #1775: Area + interval creates a spurious series
as reported by @yurivish in https://observablehq.com/d/3264dbb558b44080, the following chart creates a spurious series, with a color (_stroke_) that is undefined. I've tweaked Yuri’s example to make the issue more visually salient, and to clarify that it is not the mix of *z* and *fill* that creates the problem. Inspecting the channels with mark.initialize reveals that the _fill_ channel is ["A", "A"…, "B", "B", ...], but the _stroke_ channel is full of [undefined, undefined…, "A", …, "B", …].

Capture d’écran 2023-08-01 à 11 23 28

```
Plot.plot({
  marks: [
    Plot.areaY(
      [
        {
          x: 0,
          y: 5,
          type: "B"
        },
        {
          x: 5,
          y: 5,
          type: "B"
        },
        {
          x: 5,
          y: 5,
          type: "A"
        },
        {
          x: 10,
          y: 5,
          type: "A"
        }
      ],
      {
        interval: 1,
        x: "x",
        y: "y",
        fill: "type",
        stroke: "type",
        strokeWidth: 2,
        strokeOpacity: 0.7
      }
    )
  ],
  color: { legend: true }
})
```

Fil:
For z, stroke and fill. Previously b[0] was undefined, so a spurious undefined value was potentially added to GZ, GF or GS.

cc @yurivish 

closes #1775

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -184,9 +184,10 @@ function binn(
               groupFacet.push(i++);
               groupData.push(reduceData.reduceIndex(b, data, extent));
               if (K) GK.push(k);
-              if (Z) GZ.push(G === Z ? f : Z[b[0]]);
-              if (F) GF.push(G === F ? f : F[b[0]]);
-              if (S) GS.push(G === S ? f : S[b[0]]);
+              const u = b.length === 0 ? g[0] : b[0];
+              if (Z) GZ.push(G === Z ? f : Z[u]);
+              if (F) GF.push(G === F ? f : F[u]);
+              if (S) GS.push(G === S ? f : S[u]);
               if (BX1) BX1.push(extent.x1), BX2.push(extent.x2);
               if (BY1) BY1.push(extent.y1), BY2.push(extent.y2);
               for (const o of outputs) o.reduce(b, extent);
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -184,10 +184,9 @@ function binn(
               groupFacet.push(i++);
               groupData.push(reduceData.reduceIndex(b, data, extent));
               if (K) GK.push(k);
-              const u = b.length === 0 ? g[0] : b[0];
-              if (Z) GZ.push(G === Z ? f : Z[u]);
-              if (F) GF.push(G === F ? f : F[u]);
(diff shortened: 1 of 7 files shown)


src/transforms/bin.js
```diff
@@ -184,9 +184,10 @@ function binn(
               groupFacet.push(i++);
               groupData.push(reduceData.reduceIndex(b, data, extent));
               if (K) GK.push(k);
-              if (Z) GZ.push(G === Z ? f : Z[b[0]]);
-              if (F) GF.push(G === F ? f : F[b[0]]);
-              if (S) GS.push(G === S ? f : S[b[0]]);
+              const u = b.length === 0 ? g[0] : b[0];
```
mbostock, line 187: If we always use `g[0]` here instead of `b[0]`, the tests still pass. So, should we do that? Or can we add a test for this case. I’m struggling to think through the implications here.

mbostock: I added a test to understand the implication here. 👍

mbostock (approved):

Fil:
Thanks! Here's another test:

Capture d’écran 2023-08-26 à 18 56 09

```js
export async function test2() {
  const fills = {red: d3.interpolateReds, green: d3.interpolateGreens, blue: d3.interpolateBlues};
  const data = []
    .concat(
      d3
        .range(100)
        .map((d) => ({x: d, y: 0.4 + 0.6 * Math.cos(d / 10), z: "red"}))
        .filter((d) => d.y > 0)
    )
    .concat(
      d3
        .range(100)
        .map((d) => ({x: d, y: 0.4 + 0.6 * Math.cos(d / 3), z: "green"}))
        .filter((d) => d.y > 0)
    )
    .concat(
      d3
        .range(100)
        .map((d) => ({x: d, y: 0.4 + 0.6 * Math.cos(d / 7), z: "blue"}))
        .filter((d) => d.y > 0)
    );
  return Plot.areaY(data, {
    x: "x",
    y: "y",
    z: "z",
    fill: (d) => fills[d.z](d.y),
    stroke: "z",
    interval: 1
  }).plot();
}

```

PR #1839, opened 2023-08-27
box plot tips

### LINKED ISSUE #1835: Tips are chaotic on compound marks
When you pass `tip: true` to boxX, I think it passes it on to each of its constituent marks, which results in chaos ([demo](https://observablehq.com/d/47d3cc61a9cc4d96)):

https://github.com/observablehq/plot/assets/841829/87ff4529-6a31-4850-b099-82f837494209

I noticed this when playing with adding box to Plot.auto, which automatically sets tip: true. But what would I expect? It's a little hard to say. I guess I'd want one tip that lands on a point for each quartile, and maybe for the outliers. Maybe this just means compound marks should ideally sometimes intercept the tip option and implement their own tip instead of passing the option through to each of their marks.

Fil:
For each box plot we have 5 elements to report (+ the outliers).

When there are less than 5 values, a choice must be made. The "report" array is in order of priority (highest priority last, since we use Array.pop).

This seems to work well with 1, 2, 3, 4, and 5 values. I wonder if q1 and q3 make sense when there are < 5 values, by the way, and we should probably censor them in the visible mark too?

closes #1835

(also adds a test for boxY)

### DIFF
--- a/src/marks/box.js
+++ b/src/marks/box.js
@@ -109,5 +109,5 @@ function boxStats(values) {
     }
   }
   const report = [q3, q1, mi, hiqr2, loqr1];
-  return values.map((d) => (d < loqr1 || d > hiqr2 ? d : report.pop() ?? NaN));
+  return values.map((d) => (d !== null && (d < loqr1 || d > hiqr2) ? d : report.pop() ?? NaN));
 }
--- a/src/marks/box.js
+++ b/src/marks/box.js
@@ -33,7 +33,7 @@ export function boxX(
     barX(data, group({x1: "p25", x2: "p75"}, {x, y, fill, fillOpacity, ...options})),
     tickX(data, group({x: "p50"}, {x, y, stroke, strokeOpacity, strokeWidth, sort, ...options})),
     dot(data, map({x: oqr}, {x, y, z: y, stroke, strokeOpacity, ...options})),
-    tip && tipmark(data, pointerX(map({x: boxStats}, {x, y, z: y, ...options})))
+    tip && tipmark(data, pointerY(map({x: boxStats}, {x, y, z: y, ...options})))
   );
 }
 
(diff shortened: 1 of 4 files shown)


tophtucker (approved):
Tip behavior works exactly like I’d expect. 👍 

boxStats recomputes some things that the other marks were already computing; I guess it’d be more efficient to precompute them all and then pass them into both the graphical marks and the tip. But, as ever, I’m not sure how much it matters. 😄

mbostock (commented):
I would like to take a look but haven’t had a chance yet.

mbostock (commented):
I tried this on a non-faceted plot:

```js
Plot.plot({
  y: {
    interval: 0.1
  },
  marks: [
    Plot.boxX(
      olympians.filter((d) => d.height),
      {x: "weight", y: "height", tip: true}
    )
  ]
})
```

I don’t think it’s functioning as intended in this case? I was able to hover points within the box:

https://github.com/observablehq/plot/assets/230541/70b813c0-53c8-4f5b-b58f-b53429166d7f

It also feels to me like the pointer modes are switched. For boxX, I think we want pointerY so that you can brush horizontally and it sticks to the current box; likewise we want pointerX for boxY. Unfortunately when using faceting instead of position, it looks like pointerX and pointerY doesn’t really make a difference (since #1777); I wonder if we want to visit that so _y_ and _fy_ behave the same way? 🤔 

Design-wise, I think I would prefer one combined tip per box with the summary statistics labeled (min, q1, median, q3, max). And maybe not bother giving a tip to the outliers? You could always add an invisible dot mark overlay if you wanted to label all the points.

Fil:
> the pointer modes are switched

Very true.



For the test case with interval on the y channel, I think the problem is with the map transform, which groups by y and not by intervaled y. If you replace `y: "height"` by `y: (d) => ((d.height * 10) | 0) / 10` it works. But that means the box mark is broken, too — we didn't see it because it just returns more outliers than necessary.

Here's a test with the correct outliers:
```
export async function boxplotX() {
  const olympians = await d3.csv<any>("data/athletes.csv", d3.autoType);
  return Plot.plot({
    y: {interval: 0.1},
    marks: [
      Plot.boxX(
        olympians.filter((d) => d.height),
        {x: "weight", y: "height", z: (d) => Math.floor(d.height * 10)}
      )
    ]
  });
}
```

correct

now, remove the *z* specified above, and you get this incorrect chart, where you have "outliers" inside the q1-q3 band:

incorrect

mbostock:
Sounds like the map transform needs to use maybeApplyInterval, like the group transform does? Can you file a bug? 🙏

Fil:
An alternative with a summary table: #1843

Fil:
closing in favor of #1843

PR #1841, opened 2023-08-29
Add possibility to include a data attribute to text element.

lotusjohn:
I want to be able to scale text with css to offset text scaling together with the svg.
Ex, svg width=1000 and fontSize=16, if this svg then has a width: 100% with css on it, the text will be super small if the svg container scales down on for ex mobile devices.

With this we can now choose to add a data attribute to the text node which can be selected in css. This enables us to target different text nodes in the svg to apply different styling on different nodes. 

A really nice example is using container query units to offset the scaling text with css only. Then we can just do something like this with the cqi unit.
```
.plotcontainer text[data-name="my-regular-text"] {
    font-size: calc(2rem - 1.85cqi);
}

.plotcontainer text[data-name="my-large-text"] {
    font-size: calc(8rem - 2cqi);
}
```

### DIFF
--- a/src/marks/text.d.ts
+++ b/src/marks/text.d.ts
@@ -151,6 +151,13 @@ export interface TextOptions extends MarkOptions, TextStyles {
    * interpreted as a channel.
    */
   rotate?: ChannelValue;
+
+  /**
+   * Add a data attribute to the text node, which could be used for targeting with CSS.
+   * If a string is provided the data attribute key will be "data-text",
+   * for more control provide an array with the attribute name and value, e.g. ["data-name", "my-text"].
+   */
+  dataAttr?: string | [`data-${string}`, string];
 }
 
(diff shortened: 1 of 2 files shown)


mbostock:
I would prefer to solve this at the base Mark level rather than being specific to the Text mark.

Could you use a class name #1002 or the existing **ariaLabel** or **ariaDescription** options to target the elements instead? Seems like that would already work.

Regarding data attributes, would you want these to be channels (varying on each instance of a mark), or just constant? It sounds like a constant would be sufficient here, and simpler. If a channel, we would have to declare it and initialize it in the constructor.

lotusjohn:
You're right, className is probably a better way to go for this.

Regarding channels or not, im not super clear on this at the moment, so far i've only had use cases where it would not have been necessary.

PR #1843, opened 2023-08-31
box tip aggregate

Fil:
An alternative to #1839 for boxplot tips; this one shows the # of items in each quartile (and # of outliers), using numbers and math. symbols (as opposed to English language). We don't show empty quartiles and redundant information (i.e. when the # of values is small).

I don't know what users of box plots expect from a tip. The visual summary is one thing, but making it into “sentences” is quite challenging, given the complex definition of the elements.

For example, we can't really show percentages, as they are misleading when the data is “discrete” (few values). For instance, if you have 5 values, the proportion that are lower to the median is 2 out of 5 (40%), and the proportion that are lower or equal to the median is 3 out of 5 (60%). There is never a 50%. So, recount everything wrt the quartiles, and don't display %. But of course now it makes things much more difficult to grasp, because these statistics are coming from percentages, and it's hard to understand why the tip would show “made up” values of 4.25, 7, 10.75 as interesting edges to summarize an array of integers.

Capture d’écran 2023-08-31 à 13 26 39

This is my gripe with the whole concept of a box plot, maybe, more than with the tip :)

### DIFF
--- a/src/marks/box.js
+++ b/src/marks/box.js
@@ -109,5 +109,5 @@ function boxStats(values) {
     }
   }
   const report = [q3, q1, mi, hiqr2, loqr1];
-  return values.map((d) => (d < loqr1 || d > hiqr2 ? d : report.pop() ?? NaN));
+  return values.map((d) => (d !== null && (d < loqr1 || d > hiqr2) ? d : report.pop() ?? NaN));
 }
--- a/src/marks/box.js
+++ b/src/marks/box.js
@@ -1,4 +1,4 @@
-import {max, min, quantile, quantileSorted} from "d3";
+import {count, max, min, quantile, quantileSorted} from "d3";
 import {marks} from "../mark.js";
 import {identity} from "../options.js";
 import {groupX, groupY, groupZ} from "../transforms/group.js";
@@ -9,6 +9,7 @@ import {ruleX, ruleY} from "./rule.js";
 import {tickX, tickY} from "./tick.js";
 import {pointerX, pointerY} from "../interactions/pointer.js";
 import {tip as tipmark} from "./tip.js";
+import {formatDefault} from "../format.js";
 
 // Returns a composite mark for producing a horizontal box plot, applying the
 // necessary statistical transforms. The boxes are grouped by y, if present.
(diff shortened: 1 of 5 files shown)


mbostock:
I think I would find it more readable if the statistics were labeled with their names, min, max, median, q1, q3 (maybe count or n also?). I guess there is the question of whether we use the simple definitions or the more nuanced ones the box plot uses. I would look for precedence at other tools that provide one dimensional statistical summaries such as DuckDB and R.

Fil:
Maybe we don't need to reinforce the difference between counts (123#) and values? Also, we could use the p25 notation, for consistency.

If we could somehow format as a table (or correctly aligned columns), I'd see something like this (made up numbers):

```
count
    n: 31,256 (+ 10 n/a)
quartiles
  p25:  7,530 < 23.1
  p50: 16,230 < 45
  p75: 23,530 < 56.1
outliers
  low:    156 < 12.3
 high:    347 > 89.1
```

(12 numbers, that's quite a lot.) 

For small _n_ we wouldn't always display all the stuff, but rather:

```
     n: 3
median: 42
```

Fil:
A simpler version, doesn't show counts nor tries to explain what the 99.3% range is:

Capture d’écran 2023-09-01 à 06 32 03

tophtucker:
I like this general direction! I think “Min: 1.7” and “Max: 2.08” would be clearer than the [1.7 – 2.08] interval notation. 

It seems OK to me if the tooltip ignores the concept of outliers… even though “how is outlier defined?” is always my first question when I see a box plot with them!!

Fil:
The point is, they're not min and max — instead, they're “low” and “high” values that capture 99.7% of the (estimated) underlying normal distribution, thus defining “outliers”. If you know how a box plot is defined, the bracket notation is fine; if you don't, you’ll need a key, and I don't think that a proper definition can be conveyed in a concise tip. (Words such as “low” and “high” are not self-sufficient.)

mbostock:
I think low and high is better (more readable) than the bracket notation, and agree it avoids the confusion that min and max would introduce due to outliers. In either case you’d have to look up the details of what the number means, so better to give it a name at least so people know what to look for and how to refer to the presented values.

mbostock:
I would also prefer if the names were bolded like we do with channels, i.e., “**p25** 42” instead of “p25: 42”.

Fil:
I can't find how to add low, high, etc as independent channels on the tip mark. 

I think there's a bug somewhere since, when I do 
```
group({low: loqr1, x: "p50"}, {x, y, z: y, channels: {low: x}, ...options}))
```

the low channel is generated with the correct number of values (as many as there are groups), but all its values are `undefined`.

PR #1846, opened 2023-09-01
A transform that makes z interval-aware

### LINKED ISSUE #1842: The map transform should maybe apply maybeApplyInterval?
The map transform groups by y and not by intervaled y. If you replace `y: (d) => ((d.height * 10) | 0) / 10` below by `y: "height"`, the box plot shows too many outliers (the outliers are using a map transform).

Here's a test with the correct outliers:
```
export async function boxplotXInterval() {
  const olympians = await d3.csv<any>("data/athletes.csv", d3.autoType);
  return Plot.plot({
    y: {interval: 0.1, reverse: true},
    marks: [
      Plot.boxX(
        olympians.filter((d) => d.height),
        {x: "weight", y: "height", z: (d) => Math.floor(d.height * 10)}
      )
    ]
  });
}
```

correct

now, remove the *z* specified above, and you get this incorrect chart, where you have "outliers" inside the q1-q3 band:

incorrect

(excerpted from #1839; note that I've updated the code slightly to add y: {reverse: true}, for when we create the actual unit test.)

Fil:
TODO: This should probably go in the map.js section, together with a specific option that we'd pass to the map transform?

closes #1842

### DIFF
--- a/src/marks/box.js
+++ b/src/marks/box.js
@@ -1,7 +1,8 @@
 import {max, min, quantile} from "d3";
 import {marks} from "../mark.js";
-import {identity} from "../options.js";
+import {column, identity, maybeApplyInterval, valueof} from "../options.js";
 import {groupX, groupY, groupZ} from "../transforms/group.js";
+import {basic} from "../transforms/basic.js";
 import {map} from "../transforms/map.js";
 import {barX, barY} from "./bar.js";
 import {dot} from "./dot.js";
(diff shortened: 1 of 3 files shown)


mbostock:
I think this should be solved by the map transform rather than by the box mark.

Fil:
To pass the information that "z must be intervaled with the y scale", we can test whether options.z===options.y, which is simple ([2c61dda](https://github.com/observablehq/plot/pull/1846/commits/2c61ddaceabda74d5eb9dfbb1380069ae3e2759e)).

Before that I tried to pass something like "z.scale = 'y'" to make it more specific, but my solution used ChannelValueSpec, which was wrong typewise; here z is a ChannelValue, and this type doesn't support scale hints. If we had a way to specify a scale on a ChannelValue, or if z could be a ChannelValueSpec, this behavior would be better defined.

Fil:
I realize this issue was already mentioned in https://github.com/observablehq/plot/issues/1330#issue-1622482129!

PR #1871, opened 2023-09-22
Filter render transform

### LINKED ISSUE #1857: API: Add data to the values?
I'd like to reverse this reverse commit:
https://github.com/observablehq/plot/pull/1823/commits/09f9182bbb7d166fb7c6f661c848e2c275e08597

mbostock:
Here’s a sketch on a more imperative, lower-level interaction primitive: a filter render transform that you can dynamically update whenever you like. It starts by declaring some filters (and their initial filter state):

```js
const pointerInactiveFilter = renderFilter(true);
const pointerContextFilter = renderFilter(false);
const pointerFocusFilter = renderFilter(false);
```

Then you apply the filters to the desired marks like any other transform:

```js
const plot = Plot.plot({
  y: {grid: true},
  color: {scheme: "BuRd", symmetric: false},
  marks: [
    Plot.ruleY([0]),
    Plot.lineY(bls, pointerInactiveFilter({x: "date", y: "unemployment", z: "division", tip: true})),
    Plot.lineY(bls, pointerContextFilter({x: "date", y: "unemployment", z: "division", stroke: "#ccc"})),
    Plot.lineY(bls, pointerFocusFilter({x: "date", y: "unemployment", z: "division", stroke: "red"}))
  ]
});
```

Lastly here I use an event listener (here driven by the tip mark of the plot itself!) to update the filters:

```js
plot.addEventListener("input", () => {
  if (plot.value === null) {
    pointerInactiveFilter.update(true);
    pointerContextFilter.update(false);
    pointerFocusFilter.update(false);
  } else {
    const division = plot.value.division;
    pointerInactiveFilter.update(false);
    pointerContextFilter.update((d) => d.division !== division);
    pointerFocusFilter.update((d) => d.division === division);
  }
});
```

https://github.com/observablehq/plot/assets/230541/bfbb8274-05e9-4945-b22d-87ea0ba7b302

Fixes #1857.
Related #1574.


Fil:
Exciting!

To support faceting we only need to re-apply the translate attribute (we can copy it from the old `g`). https://observablehq.com/@observablehq/filtering-several-charts-1871

I'd like to also support sharing the behavior between charts, but it's a bit more delicate (it needs an invalidation strategy).

PR #1872, opened 2023-09-24
tip preferredAnchor

### LINKED ISSUE #1876: tip measurement too optimistic
Seen in the tiplineX example of #1872 , for some values we select an anchor that doesn't fit. (Most probably not related to that PR, since it doesn't change the measurements.) 
Capture d’écran 2023-09-25 à 08 43 58

mbostock:
Introduces a **preferredAnchor** option for the tip mark, setting the desired anchor when the tip fits. (As opposed to the **anchor** option, which always uses the specified anchor regardless of whether it fits.) This causes the tip mark to prefer the _bottom_ orientation, except when using the **tip** option with the _y_ pointer mode, in which case it prefers the _left_ orientation.

This example scatterplot now prefers *bottom* instead of *top-left*:
Screenshot 2023-09-24 at 9 20 53 AM

This example horizontal bar chart now prefers *left* instead of *top-left*:
Screenshot 2023-09-24 at 9 20 35 AM

I think this is what most people will prefer, and avoids the default “comic book” appearance that Gregor mentioned.

Fixes #1876.


Fil:
In the tiplineX example, for some values we select an anchor that doesn't fit. (Most probably not related to this PR, since it doesn't change the measurements, so I'll open a new issue) 
Capture d’écran 2023-09-25 à 08 43 58

Fil (approved):

PR #1883, opened 2023-09-29
fix tip width & add missing types for tip options

### LINKED ISSUE #1876: tip measurement too optimistic

Fil:
(based on #1872 because I couldn't reproduce the issue otherwise)

closes #1876


mbostock:
Folded into #1872.

PR #1895, opened 2023-10-17
default observable10 scheme

mbostock:


Fil:
Do we want to update the defaults green and blue for the differenceY mark to these colors?

Issue #159, opened 2021-02-24
Separate positive and negative fill colors for areas (for difference charts), or an area clipping channel or mark (and then use two marks).


mbostock:

Fil:
My plugin mentioned above is infinitely more complex than filling the space with vertical rules, as Zan suggested.

mbostock:
It looks like since June 29 the [difference chart example](https://observablehq.com/@observablehq/plot-difference-chart@327) no longer uses clipping? But I believe that approach isn’t viable when the curve is affected by multiple points (even in the case of _linear_, but most acutely with _basis_)?

Here is a take using a render transform and a compound mark:

https://observablehq.com/d/1e37e3c09876e460

Before:
![untitled (60)](https://github.com/observablehq/plot/assets/230541/0cbf0c11-46ae-4c85-9626-8b58c004b3ef)

After:
![untitled (61)](https://github.com/observablehq/plot/assets/230541/9ec70806-14d4-40d4-9d7e-129323df1753)

Fil:
Yes I simplified it since in that case the clipping approach was overkill, and I wanted to revisit the implementation when we did #1109  (Also, you may have noticed that I had inadvertently switched the colors!)

Issue #1866, opened 2023-09-20
Add a line option to the area mark


Fil:
this creates a derived line on x2/y2 which consumes the stroke; this is the "top" line for areaY, and the "right-most" line on areaX. It might be a bit unexpected or useless on area, needs experimenting to see if we must ignore it.

Typical use case https://observablehq.com/@recifs/spectrum-plotter-basic

PR #1896, opened 2023-10-17
difference mark

mbostock:
First cut at a difference mark for #159.


src/marks/difference.js
```diff
@@ -0,0 +1,109 @@
...
+      options,
+      defaults
+    );
+    this.curve = maybeCurve(curve, tension);
+    this.positiveColor = maybeColor(positiveColor);
+    this.negativeColor = maybeColor(negativeColor);
+  }
+  filter(index) {
+    return index;
+  }
```
Fil, line 45: I don't understand why we're not using the default filter; if any point's y1 or y2 is invalid, the whole area disappears.

src/marks/difference.js
```diff
@@ -0,0 +1,109 @@
...
+      )
+      .call((g) => g.append("path").attr("d", renderLine(X1, Y1, this, index)))
+      .node();
+  }
+}
+
+function renderArea(X, Y, y0, {curve}, index) {
+  return shapeArea()
+    .curve(curve)
+    .defined((i) => i >= 0)
```
Fil, line 93: I don't understand this either :) And it's not tested.

src/marks/difference.js
```diff
@@ -0,0 +1,109 @@
...
+    .x((i) => X[i])
+    .y1((i) => Y[i])
+    .y0(y0)(index);
+}
+
+function renderLine(X, Y, {curve}, index) {
+  return shapeLine()
+    .curve(curve)
+    .defined((i) => i >= 0)
```
Fil, line 102: ditto

test/plots/difference.ts
```diff
@@ -0,0 +1,13 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export async function differenceY() {
+  const aapl = await d3.csv<any>("data/aapl.csv", d3.autoType);
+  const goog = await d3.csv<any>("data/goog.csv", d3.autoType);
+  const x = aapl.map((d) => d.Date);
+  const y1 = aapl.map((d, i, data) => d.Close / data[0].Close);
+  const y2 = goog.map((d, i, data) => d.Close / data[0].Close);
```
Fil, line 9: It makes sense to normalize, but the chart is difficult to read: which part is "goog > aapl" and vice-versa? Maybe we can solve it by adding annotations, rather than a color legend?

src/marks/difference.js
```diff
@@ -0,0 +1,109 @@
...
+
+function maybeColor(value) {
+  if (value == null) return "none";
+  if (!isColor(value)) throw new Error(`invalid color: ${value}`);
+  return value;
+}
+
+class Difference extends Mark {
+  constructor(data, options = {}) {
+    const {x1, y1, x2, y2, curve, tension, positiveColor = "green", negativeColor = "blue"} = options;
```
Fil, line 27: these default colors are not very pretty; maybe use blue/orange instead?

Fil (commented):
Nice! A few comments for now.

src/marks/difference.js
mbostock: I copied this from area/line. I think the idea is that it uses the _line_.defined and _area_.defined check instead to interrupt the path, rather pretending the point doesn’t exist and connecting across the gap. I’ll add some tests to make sure this works.

src/marks/difference.js
mbostock: I’ll pick some prettier colors.

test/plots/difference.ts
mbostock: I also want to try comparing week-over-week or month-over-month change. This might involve shifting _x1_ or _x2_ instead of separating _y1_ and _y2_. We should test this use case.

src/marks/difference.js
mbostock: The filtering is handled by groupIndex, which I’ve now restored.

mbostock:
Now improved! I have an open question as to whether this should support the **z** channel, and what to do if **fill**, **fillOpacity**, **stroke**, **strokeOpacity**, or **strokeWidth** are specified as channels. I’m inclined to say that these are not supported (at least for now — I’m trying to keep things simple). But I also don’t protect against them, and I’m not entirely sure how we would protect against them. What do you think? Perhaps this elucidates why I think requests like #175 would be so difficult to support!

Fil:
Toyed with this in #1897 in a composite mark; I also worked on the default tip contents, to show y1, y2, and the actual difference = y1-y2 (which is the point of this mark).

I'd like to suggest two changes to the options, for a bit more consistency:
- use **positive** and **negative**, or **positiveFill** and **negativeFill** (instead of positiveColor and negativeColor)
- make **positiveOpacity** and **negativeOpacity** default to fillOpacity (instead of opacity)

this way **opacity** would be the usual overall mark opacity (line + areas), **strokeOpacity** would be used on the line.

Fil:
For the record, this (correctly) ignores the computed difference value from the default tip which I suggested in #1897. It was in fact a wrong good idea, because the difference mark is showing the _visual_ difference between scaled values, which is not necessarily encoding y2 &minus; y1. (Adding the difference to the default tip would have been correct only if we could assume a linear scale for *y*.)

Users can add it explicitly with the channels option, e.g.:
> `channels: {difference: y1.map((d, i) => d - y2[i])},`

or `d/y2[i]` if using a log scale and comparing ratios.

src/marks/difference.js
```diff
@@ -15,11 +15,11 @@ export function differenceY(
     y,
     y1,
     y2,
-    positiveColor = "#01ab63",
-    negativeColor = "#4269d0",
-    opacity = 1,
-    positiveOpacity = opacity,
-    negativeOpacity = opacity,
+    positive = "#01ab63",
+    negative = "#4269d0",
```
mbostock, line 19: These sound like booleans to me. I definitely prefer either the *Color* or *Fill* suffix.

Fil: OK, I'd favor positiveFill/negativeFill in that case (pushing this for now).

mbostock: I think we should use positiveFillOpacity & negativeFillOpacity, too (for symmetry, and to make clear that these affect the fillOpacity and not the opacity of the constituent areas).

Fil:
Two more suggestions:

a) flip the meaning of y1 and y2, to make it closer to areaY. Instead of having y1 be the primary metric, we'd have y1 as the base metric ("bottom" when positive), and y2 as the reference metric ("top" when positive). Switching from areaY to differenceY would not mean exchanging y1 and y2.

b) refer to #1866 when it comes to drawing the line on the area.

With these two changes, differenceY would behave like areaY, except the area part would be split into its positive and negative sections. (We could even decide to make this a simple option of areaY and avoid introducing a new symbol.)

mbostock:
The difference mark is hard to use when you have a tidy dataset like this, of temperature observations from two stations (San Francisco and San Jose):

```
date,station,tmax,tmin
2020-12-31,SJ,59,43
2020-12-31,SF,60,47
2020-12-30,SJ,58,33
2020-12-30,SF,57,40
2020-12-29,SJ,62,38
2020-12-29,SF,61,41
2020-12-28,SJ,57,43
2020-12-28,SF,56,47
```

You can say this, but it’s pretty magical and assumes that the two stations exactly alternate rows!

```js
Plot.differenceY(temperature, Plot.windowY(14, {
  filter: (d) => d.station === "SF",
  x: "date",
  y: "tmin",
  y1: (d, i) => temperature[i + 1]?.tmin
}))
```

Perhaps we need a [pivot wider](https://tidyr.tidyverse.org/reference/pivot_wider.html) transform for cases like this. This could produce a more natural shape for this case (in tidyspeak, names_from = station, values_from = c(tmin, tmax), and names_sep = ':'… and I guess because the date column isn’t mentioned, it becomes id_cols?):

```
date,SJ:tmax,SJ:tmin,SF:tmax,SF:tmin
2020-12-31,59,43,60,47
2020-12-30,58,33,57,40
2020-12-29,62,38,61,41
2020-12-28,57,43,56,47
```

Fil:
Addendum to my previous remark: "positive" and "negative" get flipped when we use the `y: {reverse: true}` option. This felt a bit unexpected—in the sense that when flipping the y axis I expected the chart to flip, but instead it changed the meaning of the colors.

| usual y | flipped y |
| --- | -- |
| difference | difference-flipped |
| negative | negative-flipped |

Not sure if this can be fixed. At least it's a reminder that this mark is not about the (sign of the) difference, but rather about the orientation of the difference between the scaled values (i.e., up or down).

mbostock:
We could use the same `inferScaleOrder` technique that the axis label uses to infer the arrow orientation to address this. I think we should fix it — we want the color to represent the sign of the difference in the abstract value, not the encoded value.

mbostock:
I broke out #1915 and #1916 from this PR.

mbostock:
In the latest commits, I introduced an explicit shift transform rather than limiting it a **shift** option on the difference mark. I’ve also changed to setting a domain hint on the channel rather than filtering the data to produce the same effect; I think this is cleaner and more precise.

If you can review this, @Fil, and the related PRs that I extracted, I can finish up the documentation and we should be able to do a release this week. I’d like to move on from further polishing of this feature (and start using it in dashboards)!

Fil:
The **z** branch (#1906) now supports clip: "frame".

mbostock:
LGTY @Fil?

Fil (approved):
yes! Thanks for fixing the documentation on **z**

PR #1904, opened 2023-10-26
case independent scale type matching

### LINKED ISSUE #1894: support scale types written in CAPS
such as `x: {type:"UTC"}`

Fil:
closes #1894


src/scales.js
```diff
@@ -285,11 +285,11 @@ function createScale(key, channels = [], options = {}) {
       );
   }
 
-  options.type = type; // Mutates input!
+  options.type = typeof type === "string" ? `${type}`.toLowerCase() : type; // Mutates input!
```
mbostock, line 288: This is happening a little too late. For example, we do an isOrdinalScale test on L267 prior to normalization.

We should move this logic into `inferScaleType` and try to do it as early as possible. For example, there could be a similar error due to lack of normalization when we check for inconsistent channel type declarations:

https://github.com/observablehq/plot/blob/7452deb48d9f88269b7b0676bdcb8cdf0c2990c1/src/scales.js#L397

I suggest that we add a helper,

```js
function maybeType(type) {
  return typeof type === "string" ? `${type}`.toLowerCase() : type;
}
```

and then we do that at the top of `inferScaleType` for the passed-in `type`, as well as for the `t` for each channel in `channels`. This should guarantee that `inferScaleType` returns a normalized scale type, so we shouldn’t have to do any subsequent normalization in `createScale`.

mbostock (changes requested):

src/scales.js
mbostock: Added in cab2d38.

mbostock (approved):

Fil:
thanks!

PR #1906, opened 2023-10-30
Support z in the difference mark

Fil:
I'm putting this in a separate PR because it's a bit of code and I'd like a review. (An alternative would be to say we don't support **z**, but in that case we'd have to be explicitly removing it; the current situation in which we pass z to the area, but merge the clip-paths &c, is halfway and broken.)

In the example chart, we follow three time series (age cohorts) by country, and by sex (the difference makes the chart switch to grey on the only case where young men are less likely than young women of the same cohort to be living with their parents, top-left (Sweden, Y16-19).

For comparison, I'm adding the faceted version (fy: "age").

| buggy | fixed | faceted |
| --- | --- | --- |
| young-adults-no-z | young-adults | young-adults-facet |

TODO:
- [x] use Plot.find when #1914 is merged


mbostock:
Can we make this look like a more traditional difference chart? I was confused why the line was being drawn in the middle of the filled area, but it’s because the stroke is disabled on the difference mark.

Fil:
good call… it's not even very useful to have the line for T, since it's most probably the midpoint (the numbers of F and M in each cohort being sensibly the same).

mbostock (commented):
Rather than repeatedly calling `next` and grouping by z, I think we just want to call `next` twice like we do in the parent branch, but collate the clipping paths with the paths. So for example if you have two series, each call to `next` returns a G element with two path elements; the first path element from the first G becomes the first clipPath element for the first path element in the second G, and so on.

PR #1909, opened 2023-10-31
rect support for band scales

### LINKED ISSUE #1899: The rect mark could work better with band scales
Currently the rect mark is a bit pedantic regarding band scales — it generally requires that you specify **x1**, **x2**, **y1**, **y2** explicitly. So in a situation like this, you see nothing:

untitled (99)

```js
Plot.plot({
  marks: [
    Plot.rectX(data, {x1: "start", x2: "end", y: "name"})
  ]
})
```

You have to switch to barX to see what you likely intend:

untitled (100)

```js
Plot.plot({
  marks: [
    Plot.barX(data, {x1: "start", x2: "end", y: "name"})
  ]
})
```

A related question is whether we could automatically infer the _band_ scale for _y_ (rather than the default _point_). That would also be necessary for the rect to appear.

Another possibility, to at least show _something_ that maybe guides someone to a better outcome, would be to apply default negative insets in the case whether the rect would otherwise have zero width or height.

untitled - 2023-10-20T111404 886

```js
Plot.plot({
  marks: [
    Plot.rectX(data, {x1: "start", x2: "end", y: "name", inset: -0.5})
  ]
})
```

Notebook: https://observablehq.com/d/87107ffe90ae73be

mbostock:
Fixes #1899.

### DIFF
--- a/docs/marks/rect.md
+++ b/docs/marks/rect.md
@@ -26,10 +26,6 @@ onMounted(() => {
 
 # Rect mark
 
-:::tip
-The rect mark is one of several marks in Plot for drawing rectangles; it should be used when both dimensions are quantitative. See also [bar](./bar.md) and [cell](./cell.md).
-:::
-
 The **rect mark** draws axis-aligned rectangles defined by **x1**, **y1**, **x2**, and **y2**. For example, here we display geographic bounding boxes of U.S. counties represented as [*x1*, *y1*, *x2*, *y2*] tuples, where *x1* & *x2* are degrees longitude and *y1* & *y2* are degrees latitude.
 
 :::plot defer https://observablehq.com/@observablehq/plot-county-boxes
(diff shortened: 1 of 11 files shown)


Fil:
This is great! For the documentation, I think we need to amend the [paragraph](https://observablehq.com/plot/marks/rect#rect-options) that says “Typically either x1 and x2 are specified, or y1 and y2, or both.”?

docs/marks/rect.md
```diff
@@ -199,7 +195,7 @@ The following channels are optional:
 * **x2** - the ending horizontal position; bound to the *x* scale
 * **y2** - the ending vertical position; bound to the *y* scale
 
-Typically either **x1** and **x2** are specified, or **y1** and **y2**, or both.
+If **x1** is specified but **x2** is not specified, then *x* must be a *band* scale; if **y1** is specified but **y2** is not specified, theN *y* must be a *band* scale.
```
Fil, line 198:
```suggestion
If **x1** is specified but **x2** is not specified, then *x* must be a *band* scale; if **y1** is specified but **y2** is not specified, then *y* must be a *band* scale.
```

Fil (approved):

PR #1915, opened 2023-11-03
fix overload precedence

mbostock:
This fixes a TypeScript type inference bug in a case like this:

```js
Plot.plot({
  marks: [
    Plot.differenceY(
      stocks,
      Plot.normalizeY(
        Plot.groupX(
          {y1: Plot.find((d) => d.Symbol === "GOOG"), y2: Plot.find((d) => d.Symbol === "AAPL")},
          {x: "Date", y: "Close", curve: "cardinal", tension: 0.1}
        )
      )
    )
  ]
})
```

Previously, TypeScript would assume that the normalizeY transform was being passed a single argument of _basis_; now it correctly assumes that it is being passed a single argument of _options_.

### DIFF
--- a/src/transforms/dodge.d.ts
+++ b/src/transforms/dodge.d.ts
@@ -59,8 +59,8 @@ export interface DodgeYOptions extends DodgeOptions {
  *
  * If *dodgeOptions* is a string, it is shorthand for the dodge **anchor**.
  */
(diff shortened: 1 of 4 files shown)


Fil:
Confirmed (both error and fix) with areaY.

Fil (approved):

PR #1916, opened 2023-11-03
x and y reducers for group and hexbin

### LINKED ISSUE #1530: The group and hexbin transforms could support x and y reducers
The bin transform supports it, but the group and hexbin transforms don’t. We could extend support to the group and hexbin transforms.

mbostock:
Fixes #1530.


src/transforms/group.d.ts
```diff
@@ -38,8 +38,42 @@ export interface GroupOutputOptions<T = Reducer> {
...
+export interface GroupReducerImplementation<S = any, T = S> {
+  /**
+   * Given an *index* representing the contents of the current group, the input
+   * channel’s array of *values*, and the current group’s *extent*, returns the
+   * corresponding reduced output value. If no input channel is supplied (e.g.,
+   * as with the *count* reducer) then *values* may be undefined.
+   */
+  reduceIndex(index: number[], values: S[], extent: {x: any; y: any}): T;
+  // TODO scope
+  // TODO label
```
Fil, line 72: This made me think of how we would type the reduceIndex method for scope = "data" or "facet". Maybe something like the following, but it seems a bit messy:

```ts
  reduceIndex(
    index: number[],
    values: undefined | S[],
    valueOrExtent: any | {data: any[]},
    maybeExtent: undefined | {data: any[]}
  ): T;
```

Fil (approved):

PR #1922, opened 2023-11-12
differenceX and shiftY

### LINKED ISSUE #1920: differenceX and shiftY
as counterparts of #1896

also fix the examples (in particular https://observablehq.com/@observablehq/plot-difference-chart)

Fil:
closes #1920

(I don't think these need too much documentation, since they are principally motivated by symmetry.)


src/marks/difference.js
```diff
@@ -6,15 +6,23 @@ import {getClipId} from "../style.js";
 import {area} from "./area.js";
 import {line} from "./line.js";
 
-export function differenceY(
+export function differenceX(data, options) {
+  return differenceK("x", data, options);
+}
+export function differenceY(data, options) {
```
mbostock, line 12:
```suggestion

export function differenceY(data, options) {
```

src/marks/difference.d.ts
```diff
@@ -69,6 +69,52 @@ export interface DifferenceOptions extends MarkOptions, CurveOptions {
   z?: ChannelValue;
 }
 
+export interface DifferenceXOptions extends DifferenceYOptions {
```
mbostock, line 72: We should make a shared `DifferenceOptions` based instead of having `DifferenceXOptions` extend `DifferenceYOptions`.

mbostock (changes requested):
Small type declaration suggestion. I’ll push a fix.

src/marks/difference.d.ts
mbostock: Actually it seems to me that the types are identical, and only the comments are difference? So I’ll unify the interfaces.

mbostock (approved):

PR #1930, opened 2023-11-17
fix!: reduce specificity of generated stylesheets to 0

mstade:
### What is this, and why is it necessary?

By reducing the specificity of selectors in the generated stylesheets to 0, consumers of Plot can much more easily override these styles using their own stylesheets while still enjoying the default styling of Plot.

Without this, consumers of Plot must resort to increasing the specificity of selectors or individual styles, which can be both awkward and hard to do.

### A breaking change, technically

While this change is meant to be safe, it is technically a breaking change. Any consumer of Plot that is trying to override css, but failing due to lower specificity, should start seeing those styles applied. While that may actually fix a few bugs out there, it could also cause some others, where those consumers have simply missed that those styles aren't being applied on top of the defaults.

### Why so many snapshot updates?

Because the change in selector specificity affects every chart, all snapshots will need regenerating. The rendered output should yield no differences however. If it does, that's a bug.

Alas – this many snapshot updates seem to make GitHubs UI practically worthless for reviewing the code however. My apologies, I don't have a good answer for how to fix that. I did however split the changes into two commits for this reason:

- CSS changes: ~[b2ff68e](https://github.com/observablehq/plot/pull/1930/commits/b2ff68eacf41fb7618509859a9510d8fd6fb4653)~ [755954e](https://github.com/observablehq/plot/pull/1930/commits/755954ee08b3a5f317d615b2d4015887f5c5515c) (Edit: updated SHA post rebase)
- Snapshot updates: ~[300c400](https://github.com/observablehq/plot/pull/1930/commits/300c40002c5ac8eb3ec74f092cff0945378c546d)~ [fbcfa1f](https://github.com/observablehq/plot/pull/1930/commits/fbcfa1f311f801fff025006c349544933437e9a1) (Edit: updated SHA post rebase)

I hope this helps.

### Related resources

- MDN Docs on `:where`: https://developer.mozilla.org/en-US/docs/Web/CSS/:where

### DIFF
--- a/src/legends/ramp.js
+++ b/src/legends/ramp.js
@@ -38,14 +38,14 @@ export function legendRamp(color, options) {
     .call((svg) =>
       // Warning: if you edit this, change defaultClassName.
       svg.append("style").text(
-        `.${className}-ramp {
+        `:where(.${className}-ramp) {
   display: block;
   height: auto;
   height: intrinsic;
   max-width: 100%;
   overflow: visible;
 }
-.${className}-ramp text {
+:where(.${className}-ramp text) {
   white-space: pre;
 }`
       )
--- a/src/legends/swatches.js
+++ b/src/legends/swatches.js
@@ -96,16 +96,16 @@ function legendItems(scale, options = {}, swatch) {
   let extraStyle;
 
   if (columns != null) {
-    extraStyle = `.${className}-swatches-columns .${className}-swatch {
+    extraStyle = `:where(.${className}-swatches-columns .${className}-swatch) {
   display: flex;
   align-items: center;
   break-inside: avoid;
   padding-bottom: 1px;
 }
-.${className}-swatches-columns .${className}-swatch::before {
+:where(.${className}-swatches-columns .${className}-swatch::before) {
   flex-shrink: 0;
 }
-.${className}-swatches-columns .${className}-swatch-label {
+:where(.${className}-swatches-columns .${className}-swatch-label) {
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
@@ -123,13 +123,13 @@ function legendItems(scale, options = {}, swatch) {
         item.append("div").attr("class", `${className}-swatch-label`).attr("title", tickFormat).text(tickFormat)
       );
   } else {
-    extraStyle = `.${className}-swatches-wrap {
+    extraStyle = `:where(.${className}-swatches-wrap) {
   display: flex;
   align-items: center;
   min-height: 33px;
   flex-wrap: wrap;
 }
-.${className}-swatches-wrap .${className}-swatch {
+:where(.${className}-swatches-wrap .${className}-swatch) {
   display: inline-flex;
   align-items: center;
   margin-right: 1em;
@@ -150,12 +150,12 @@ function legendItems(scale, options = {}, swatch) {
   return swatches
     .call((div) =>
       div.insert("style", "*").text(
-        `.${className}-swatches {
+        `:where(.${className}-swatches) {
   font-family: system-ui, sans-serif;
   font-size: 10px;
   margin-bottom: 0.5em;
 }
-.${className}-swatch > svg {
+:where(.${className}-swatch > svg) {
   margin-right: 0.5em;
   overflow: visible;
 }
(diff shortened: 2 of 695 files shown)


mbostock (commented):
Thanks, I didn’t know about `:where`. I can see this being useful. My consideration now is compatibility, including backwards compatibility. It looks like `:where` is fairly widely supported, but it’s only since 2021 that it’s been implemented in browsers. Can you speak a little more to the use case that this solved for you? Is there an example where you tried to override Plot’s default styles, and were surprised or stymied by the specificity of those styles?

mstade:
These are great questions @mbostock – thank you so much for reviewing!

The compatibility concern is of course very relevant, and I must admit I have not done due diligence here – my apologies! I have the luxury of not really having to target legacy environments very often, and have been using this selector for some time with great success. It would be possible of course to use something like `@supports selector(:where)`, but I fear that feature may be about as well supported as `:where` to be honest, so might not be all the rage after all.

In any event, to your questions.

We've been working with Plot the past week to put together some dead simple time series charts, and when we were happy with the functionality and configuration I got to work trying to style these. In our application we make heavy use of [CSS modules](https://github.com/css-modules/css-modules) for locally scoped class selectors. The specificity of these are `0,1,0` – the same as any other pure class selector.

When applying this to Plot, I noticed that my styles for the legend didn't seem to take, at all. It turns out, while setting `className` on the plot will just set it outright, for the legend this isn't true. Instead it sets names like `${className}-swatches ${className}-wrap`, and similar. This is a little awkward, but possible to do in CSS modules using the `:global` scope – e.g. something like `.legend:global(-swatches)` will generate css looking something like `.Xyz123-swatches`, where `Xyz123` is a unique hash generated at build time. 

So I did this, and _some_ of the styles took, but not all. For example, I tried setting `font-size` but the style generated by Plot took precedence because the style is injected later in the document than my CSS module styles. And with the same specificity (`0,1,0`), Plot's generated style won the day. Changing Plot's selectors to `:where(...)` however resets the specificity to `0,0,0`, meaning it doesn't matter _where_ Plot injects these styles, a regular class selector will take precedence.

Of course, it's possible to override these styles by increasing the specificity. Using something like `!important` or even just adding another type of selector would do, like an id or attribute selector, just bumping that selector ever so slightly higher. But this is not without issues either, and can break in subtle and hard-to-catch ways – for example if Plot changes its selectors to be more specific.

That's why many reusable libraries that ship styles these days make heavy use of the `:where` selector, to reset the specificity all the way down to 0. Since any other selector will increase the specificity this basically guarantees the styles will be overridable. `@layer` is also gaining in popularity, and may be an even better solution, but that's even more bleeding edge... 😅

It may be best to actually implement a new test case for this, and my apologies for not doing this – time permitting I will look into it over the weekend. It really shouldn't be particularly difficult. Maybe that will be a better answer to your questions?

mstade:
@mbostock I implemented a very basic test case in [7c2c24e](https://github.com/observablehq/plot/pull/1930/commits/7c2c24e6b9fab217ee1a79df6c4a029eabf433f6). These changes are wider than just categorical legends of course, but before I spend more effort in developing further test cases I'd like to just check that this:

1. Is in line with how you test things in this project and
2. You believe these changes to be relevant for the project

Verifying that these changes actually do what they say on the tin is very simply, just revert [b2ff68e](https://github.com/observablehq/plot/pull/1930/commits/b2ff68eacf41fb7618509859a9510d8fd6fb4653). I've recorded a brief session of me doing just this, to show you how reverting the changes makes the second legend ignore the styles we're trying to set by specifying `className`. The recording first shows the result of the test case with the fix implemented, and then I revert [b2ff68e](https://github.com/observablehq/plot/pull/1930/commits/b2ff68eacf41fb7618509859a9510d8fd6fb4653) to show how the styles revert to the default, due to the specificity conflict. Finally I reset `head` to remove the revert commit and it shows the fixes being applied again:

Screen Recording 2023-11-18 at 16 03 00

I hope this is a sufficiently small and good example of the use case I'm trying to solve with these changes – many thanks for your review!

mstade:
@mbostock / @Fil – appreciating you have other things to do and this PR may not be a priority – I'm wondering if you think the proposed changes are at all relevant to this project? If so if I should implement more tests or have any thoughts on other changes before this can be brought to `main`? Thank you for taking the time to review!

mbostock:
Sorry for the delay on this. We’ll get this merged. I should have included it in 0.6.12 but I’ll take it from here. Thank you!

mstade:
Thank you @mbostock – much obliged! If you'd like me to look into expanding the test suite or such related to this, please do feel free to let me know. Happy to contribute!

PR #1957, opened 2023-12-21
occlusionY initializer

Fil:
https://github.com/observablehq/plot/assets/7001/76810a97-44c4-4caf-afb9-2519e5ed7ada

quoting @visnup: _I always need some occlusionX or Y_ 😅

PS: not too happy with the name… it seems to say the opposite of what it does?


mbostock (commented):
This sounds a lot like the dodge transform and this statement isn’t quite enough to help me understand the difference:

> The occlusion transform differs from the [dodge transform](./dodge.md) in that it only adjusts the nodes’ existing positions.

We should draw the distinction in a tip at the top of the page, and expand it. And when you say “rearranges the **x** (horizontal) position” I interpret “rearranges” as “reorders” the _x_-coordinates, but that’s not what you mean at all.

I guess the difference is that dodgeY doesn’t take a **y** channel as input, it just computes _y_ from the anchor (baseline), whereas occlusionY does take a **y** channel as input and adjusts those values to enforce a minimum distance, while minimizing movement and preserving order? I wonder if we could instead incorporate this into the dodge transform rather than making a new transform.

We should continue to name transforms with a verb. This isn’t “occlusion-ing” and it’s not “occluding” either because it’s trying to _prevent_ occlusion. I think “dodge” is the best name for this type of thing. “Repel”, “offset”, “collide”, and “avoid” are other verbs that come to mind…

Your earlier [occlusion helper](https://observablehq.com/@d3/occlusion) might better be named an “occlude” transform as that one hides things! 😅

Fil:
> I wonder if we could instead incorporate this into the dodge transform rather than making a new transform.

Yes I had the same feeling, but didn't want to complicate things at first. It might mean renaming *minDistance* (to *r*, with a default of 5.5 instead of 11, but then it's not the same default as dodge…), and applying it on the opposite direction too? I can investigate whether that's possible. Now that we have this baseline, maybe it's worth doing this extra step.

I like "repel" a lot.

Fil:
Other verbs: Fan out, Spread, Stretch, Radiate, Unfold...

visnup:
I had to check but disocclude is an actual word. https://en.wiktionary.org/wiki/disocclude

Fil:
I would love to unify this with the dodge transform, unfortunately I'm not seeing a lot of common ground. Maybe this needs more work.

What's different:

| | dodgeY | occlusionY |
|--| -- | -- |
|  **y** channel | not supported | required |
|  **x** channel | required | optional |
| spread |  **x** + **y** |  **y** only |
| default | _r_ 3, channel | _minDist_ 11, constant |
| anchor | supported | not supported |
 
A way to unify them would be to let dodgeY switch to the occlusionY algo when a **y** channel is specified, but this would mean that almost everything changes meaning:
- what to do when *x* is a channel?
- *minDist* could become *r* (at the price of not having the best default for text labels), but then what to do when *r* is a channel?

mbostock:
“Repel” 👍

PR #1959, opened 2023-12-23
preserve group input order

### LINKED ISSUE #1678: The group transform always sorts groups naturally; should this be configurable?
maybeGroup always sorts by the group key:

https://github.com/observablehq/plot/blob/98cb5538104b7d5493867757fc1ad475abf6df34/src/transforms/group.js#L229-L236

Maybe there should be a way to turn this off if you want to render the groups in input order? It’s confusing because if you set the **sort** option to null, you don’t get input order, you get ascending natural order, which is different from most other cases. (It makes sense with the bin transform, though.)

https://github.com/observablehq/plot/blob/98cb5538104b7d5493867757fc1ad475abf6df34/src/transforms/group.d.ts#L21-L32

Ref. https://talk.observablehq.com/t/facet-custom-sort/8002/6

mbostock:
Fixes #1678. If we simply don’t sort the groups naturally by key, then the groups will be in input order and hence the **sort** option will generally work as expected.

~~I’ll need to fix a number of the tests that relied on the implicit sort.~~ Done.


test/output/athletesHeightWeightSex.svg
mbostock, line 1: This should probably use multiply blending so the appearance is not affected by order, but that makes it unacceptably slow. 😞

docs/transforms/group.md
```diff
@@ -329,7 +329,7 @@ Although barX applies an implicit stackX transform, [textX](../marks/text.md) do
 
 ## Group options
 
-Given input *data* = [*d₀*, *d₁*, *d₂*, …], by default the resulting grouped data is an array of arrays where each inner array is a subset of the input data such as [[*d₁*, *d₂*, …], [*d₀*, …], …]. Each inner array is in input order. The outer array is in natural ascending order according to the associated dimension (*x* then *y*).
+Given input *data* = [*d₀*, *d₁*, *d₂*, …], by default the resulting grouped data is an array of arrays where each inner array is a subset of the input data such as [[*d₁*, *d₂*, …], [*d₀*, …], …]. Each inner array is in input order. The outer array is input order according to the first element of each group.
```
Fil, line 332:
```suggestion
Given input *data* = [*d₀*, *d₁*, *d₂*, …], by default the resulting grouped data is an array of arrays where each inner array is a subset of the input data such as [[*d₁*, *d₂*, …], [*d₀*, …], …]. Each inner array is in input order. The outer array is in input order according to the first element of each group.
```

test/output/athletesHeightWeightSex.svg
Fil: we could switch to computing the end color directly? challenge unrelated to this PR though :)

Fil (approved):

PR #1967, opened 2024-01-02
stack handles negative zero

mbostock:
Ref. https://github.com/observablehq/plot/discussions/1963#discussioncomment-7993585

### DIFF
--- a/src/transforms/stack.js
+++ b/src/transforms/stack.js
@@ -104,9 +104,9 @@ function stack(x, y = one, kx, ky, {offset, order, reverse}, options) {
           if (reverse) stack.reverse();
           for (const i of stack) {
             const y = Y[i];
-            if (y < 0) yn = Y2[i] = (Y1[i] = yn) + y;
-            else if (y > 0) yp = Y2[i] = (Y1[i] = yp) + y;
-            else Y2[i] = Y1[i] = yp; // NaN or zero
+            if (y < 0 || 1 / y === -Infinity) yn = Y2[i] = (Y1[i] = yn) + y; // negative maybe zero
+            else if (y >= 0) yp = Y2[i] = (Y1[i] = yp) + y; // positive maybe zero
+            else Y2[i] = Y1[i] = yp; // NaN
           }
         }
         facetstacks.push(stacks);


Fil:
🤔 We don't support -0 anywhere else, so this introduces an incongruity. For example, if the value we need to stack is the output of an aggregation with a "sum" reducer, we wouldn't be able to generate -0 (although the whole series might be sums of negative numbers).

As an alternative, I'd suggest adding any zeroes to yn if yp is zero. The patch can't be simpler:

```diff
-            else Y2[i] = Y1[i] = yp; // NaN or zero
+            else Y2[i] = Y1[i] = yp || yn; // NaN or zero
           }
```

Fil:
(Also re-reading this code I wonder why compare is in its own loop, and reverse is not. For readability, could we move compare just above reverse, or vice-versa?)

Fil:
My (very unimaginative) unit test:

```
export async function stackZeroes() {
  const data = Array.from({length: 100}, d3.randomNormal.source(d3.randomLcg(42))()).map((value, i) => ({
    value: i % 7 === 0 ? 0 : value,
    x: Math.floor(i / 3),
    series: i % 3
  }));
  return Plot.plot({
    marks: [
      Plot.areaY(data, {x: "x", fill: "series", y: (d) => Math.min(d.value, 0)}),
      Plot.areaY(data, {x: "x", fill: "series", y: (d) => Math.max(d.value, 0)})
    ]
  });
}
```

mbostock:
The alternative you suggest is not desirable as it will cause zero values to swap to the bottom when all the other values are positive, rather than maintaining the input order.

The purpose of this change is simply to allow -0 to indicate a zero value that stacks with the other negative values. We don’t need to do anything else.

Fil:
I've spent a moment trying to understand the situation you're describing, but I don't get it. Seems to me that zeroes in a stack of positives are properly stacked in input order, as are zeroes in a stack of negatives. This solves the original issue with two separate series min(V,0) and max(V, 0).

The "known unknown" case is when a zero arrives in a stack where we've already seen non-negatives and non-positives. In that case it goes arbitrarily to the top of the positive stack (yp); this might not be desirable and could (additionally) be explicitly routed with -0.

mbostock:
@Hvass-Labs says it doesn’t work with 0 as I linked in the OP https://github.com/observablehq/plot/discussions/1963#discussioncomment-7993585. But seems like we need the actual data to confirm that this is the problem and not something else?

mbostock:
You can see the problem in the linked images:

image

The negative zero values jump up to the _y_ = 0 baseline because they are erroneously considered positive, hence the need to use -1e-35 as the workaround.

Fil:
I've edited my comment above to clarify that it relates to my suggested patch.

mbostock:
Conceptually, -0 is a negative number and hence should be stacked with the other negative values. This feels more explicit than having the treatment of negative zero depend on whether there’s any preceding positive value in the current stack.

Fil:
Agree that if the user passes -0 it is logical to stack it on yn. But should we _depend_ on this peculiar value to answer the OP? It feels like an easter egg. (I understand it's a Number, but it feels like a special symbol that needs specific documentation and is not so easily generated).

I've pushed my suggestion to #1968 — with a funky test case.

But note that we could have both: stack a 0 value on the "active" side (yp || yn), and stack an explicit -0 on the negative side yn:

```diff
-            else Y2[i] = Y1[i] = yp; // NaN or zero
+            else Y2[i] = Y1[i] = 1 / y === -Infinity ? yn : yp || yn; // NaN or zero
```

I don't want to block this further, I just hope I've made my case :) Whatever you prefer, it will be good to include several test plots.

Issue #811, opened 2022-03-16
document warnings in README


Fil:
they are only documented in https://github.com/observablehq/plot/blob/main/CHANGELOG.md#041 for now

ref. https://talk.observablehq.com/t/bar-chart-axis-is-inverted/6359/3

LeonxLJX:
I'd like to take this one (`Plot’s warning feature ⚠️ needs documentation`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

PR #1969, opened 2024-01-04
#811 Custom document waring clearing

Rahmaaaan:


mbostock:
Thanks, but this isn’t what we meant — #811 refers to Plot’s warnings feature, not Prettier.

PR #1974, opened 2024-01-15
fix facetAnchor interactive doc crashing when it receives "null"

### LINKED ISSUE #1972: Documentation page about axis mark turns empty if an interactive selection is made
image

https://observablehq.com/plot/marks/axis
If I go to the middle and change for the interactive chart facetAnchor to be null, the whole page turns empty.
(Browser Google Chrome Version 120.0.6099.217 (Offizieller Build) (64-Bit) on Windows PC)

Fil:
closes #1972


mbostock (approved):
Thanks!

Issue #1832, opened 2023-08-24
A pointer interaction that handles click differently?


Fil:
When building an interactive app, we might want:

1. a pointer that only sets the value when we click
2. a pointer that doesn't stick when we click

mbostock:
Maybe we can call it the “clicker” interaction in homage to TLOU. 😂

yurivish:
I've been using Plot to build early product prototypes recently and something along these lines would have been extremely useful (though this use case is admittedly stretching the boundaries of "exploratory data visualization" a bit!)

A bit of context in case it's useful: one of the prototypes explored ways to browse the results of a set of benchmarks/performance experiments. The prototype looks like this:

<img width="885" alt="image" src="https://github.com/observablehq/plot/assets/114388/d4f83650-fd50-4e0f-b78b-d5a38590a0bb">

The top "chart" shows a grid of cells, where each cell represents a specific set of experimental parameters. The charts below correspond to detailed information about the selected cell in the top chart.

The interaction I wanted was that the user would click on a cell in the top chart to explore the associated experiments, with the secondary charts appearing only once a cell had been clicked.

I couldn't figure out a way to distinguish hover from click, though, so I ended up doing some lower-level event handling to maintain a variable that held the "last-pointed-at" element as a way to model a sticky hover selection – otherwise, moving the cursor away from the cells would result in a very visible reflow as the associated charts disappeared.

I would have liked to be able to show a tip in the top chart on hover, but not show the associated charts until the user clicks on a cell, ie. to distinguish a hover selection from a sticky selection. The tip represents a first level of detail, and the associated charts represent a second, much greater level of detail (akin to a page navigation). 

A related question that came up was how to persist the pointer state across re-renders of the chart. The prototype had a few range sliders that all of the visualizations depended on. Since these included the top chart (which held the selection state), the selection would be reset whenever the user interacted with any of the sliders. Because I was storing the "last-pointed-at" element separately, I was able to keep the associated charts on screen as the user slid the sliders, but the selection state in the top chart (ie. the tooltip indicating the selected experiment) would disappear and I couldn't figure out how to bring it back without introducing dependency cycles.

(One thing I tried setting the `value` of the top plot and triggering an `input` event, but that did not seem to cause the tips to show up -- maybe because I was setting the `value` of the `figure` element and not the `svg` itself, though?)

Fil:
Another example https://observablehq.com/d/7e5f1eec1354f1a2

ezralee:
[...]

PR #1979, opened 2024-01-30
Experiment: add support for distinguishing hover & click events

yurivish:
This is a tiny experimental patch that addresses https://github.com/observablehq/plot/issues/1832 by sending Boolean "sticky" information via the event's `detail` option. 

Plot's behavior is otherwise unchanged. Since the sticky data is part of the event, a user can now create a generator cell that responds to the "substream" of click-relevant events. [Example](https://observablehq.com/@iopsystems/hover-and-click):

```js
click = Generators.observe((notify) => {
  let lastValue = null;
  const listener = (e) => {
    // If e.detail is true, then this was a click (sticky) event.
    // If e.detail is false, then (this feels untidy) deselect if the value is null.
    if (e.detail || (viewof hover.value === null && lastValue !== null)) {
      notify(viewof hover.value);
    }
  };
  viewof hover.addEventListener("input", listener);
  notify(lastValue);
  return () => viewof hover.removeEventListener("input", listener);
})
```

There might be unfavorable trade-offs to the idea of using the `detail` option for this, and the `null` handling is somewhat untidy (though maybe there's a better way?) so please feel free to close if it isn't a good idea. But I was surprised and pleased with the fact that this was a three-line change.

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -172,10 +172,11 @@ export function plot(options = {}) {
   };
 
   // Allows e.g. the pointer transform to support viewof.
-  context.dispatchValue = (value) => {
-    if (figure.value === value) return;
+  context.dispatchValue = (value, detail) => {
+    if (figure.value === value && figure._detail === detail) return;
     figure.value = value;
-    figure.dispatchEvent(new Event("input", {bubbles: true}));
+    figure._detail = detail;
+    figure.dispatchEvent(new CustomEvent("input", {bubbles: true, detail}));
   };
 
   // Reinitialize; for deriving channels dependent on other channels.
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -126,7 +126,7 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
 
         // Dispatch the value. When simultaneously exiting this facet and
         // entering a new one, prioritize the entering facet.
-        if (!(i == null && facetState?.size > 1)) context.dispatchValue(i == null ? null : data[i]);
+        if (!(i == null && facetState?.size > 1)) context.dispatchValue(i == null ? null : data[i], s);
         return r;
       }
 


mbostock (commented):
I’m cool with this conceptually, but I’m wondering if we can

- create a custom subclass of Event instead of using CustomEvent
- expose the sticky state as event.sticky instead of event.detail

If a subclass isn’t possible, then I would prefer detail to be an object {sticky} instead of the boolean for better future compat.

yurivish:
Hmm. I changed it to use an Event subclass, but jsdom is still unhappy. There are lots of errors like this:

```
 1) plot autoArea:
     TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.
      at exports.convert (node_modules/jsdom/lib/jsdom/living/generated/Event.js:22:9)
      at SVGSVGElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:236:24)
      at context.dispatchValue (file:///home/runner/work/plot/plot/src/plot.js:179:12)
      at render (file:///home/runner/work/plot/plot/src/interactions/pointer.js:129:59)
      at Tip.<anonymous> (file:///home/runner/work/plot/plot/src/interactions/pointer.js:182:14)
      at Tip.render (file:///home/runner/work/plot/plot/src/mark.js:144:15)
      at plot (file:///home/runner/work/plot/plot/src/plot.js:288:25)
      at Array.plotThis [as plot] (file:///home/runner/work/plot/plot/src/plot.js:372:10)
      at autoArea (test/plots/autoplot.ts:58:65)
      at async doesNotWarnAsync (file:///home/runner/work/plot/plot/test/assert.js:53:14)
      at async file:///home/runner/work/plot/plot/test/plot.js:12:18
      at async Context.<anonymous> (file:///home/runner/work/plot/plot/test/jsdom.js:29:14)
```

The error is thrown from a generated file that is not in source control, but here are the [relevant lines](https://gist.github.com/yurivish/2447bd2d9cca6f0b1b296b1a3df4fce1#file-jsdom-generated-event-js-L12-L23) from my local copy.

I am not familiar with [jsdom](https://github.com/jsdom/jsdom), but some searching found a [similar issue](https://github.com/jsdom/jsdom/issues/3331) where the issue was mitigated with a "fix" that looks similar to what Plot already does:

> I can "fix" it by setting `global.Event = window.Event`, but realize from this thread that this is discouraged.

https://github.com/observablehq/plot/blob/main/test/jsdom.js#L23

I don't quite understand what is happening here, but can have another look tomorrow...

yurivish:
Moving the event definition inside the `plot` function resolves the jsdom issue, though I can't say I understand why...

Fil:
> I wonder why that wasn't the case outside the function.

As I understand it it's because it happens earlier and thus the class extends whatever is the `global.Event` at that point.

The addition of the custom `_sticky` property to the figure element exposes this value; what we want is to cache it so we can compare it on the next event. So I wonder if we could store this state in a variable named `stickyState` and created parallel to `context`?, or add it to the `context` instead?

Please add a representative use case in test/plots/ so we can test these changes more easily. (See https://github.com/observablehq/plot/blob/main/test/plots/pointer.ts#L28 for an example.)

yurivish:
I'm not sure if it's due to [this issue](https://github.com/observablehq/plot/issues/1980) or something else but the test command fails to regenerate output files if I delete them as per the [instructions](https://github.com/observablehq/plot/blob/main/CONTRIBUTING.md) for regenerating the snapshot tests. This is on an M2 mac on Sonoma 14.2.1 with node v21.5.0.

I tried generating an initial snapshot manually, but this gets rejected due to differences between the true output and the manual version.  Comparing to the HTML I copied from a notebook and the HTML in the pointerViewof output test, there seem to be lots of small differences so it doesn't seem to trivial to make a correct snapshot manually. Maybe someone else can generate the initial snapshot? Let me know if the general idea of the test is reasonable – it dynamically shows the click selection in a textarea similar to the way the viewof test shows the value.

src/plot.js
```diff
@@ -171,11 +171,21 @@ export function plot(options = {}) {
...
 
+  class StickyEvent extends Event {
+    constructor(type, options) {
+      super(type, options);
+      this.sticky = options.sticky;
+    }
+  }
+
+  context.stickyState = undefined;
+
```
Fil, line 182:
```suggestion
```

initial value is undefined

Fil (commented):
I'll take a pass at generating the test

Fil:
To generate the test output I did:

> nvm use 20
> yarn
> rm test/output/pointerSticky.html
> yarn test

I also had to run
> yarn prettier -w test/plots/pointer.ts

And finally
> git add test/output/pointerSticky.html test/plots/pointer.ts

I don't think I can push to your branch, though.

yurivish:
Thank you! I did not know about `nvm`. Tests passed locally on node 20.

Fil:
In the brush PR #1653 I'm following a slightly different pattern (with the same goal of telling the listener whether we're "browsing" or "selecting"). This PR [attaches a .done property to the value](https://github.com/observablehq/plot/pull/1653/files#diff-ddb6135aeab503983fe7ffd8bd7b8cb129f59ac11e0343961daf8e4bd44efd3cR106) on brush `end`.

Pros and cons:
- `value.done` is simpler to read; 👍
- `value.done` is consistent with what Observable does when streaming sql; 👍
- `done` as a concept is more generic than `sticky`; 👍
- `value.done` messes with the value, and cannot be applied generically on any value (it needs to be an array, and we should not mutate the original array of data when the selection is "full"). 👎

Whatever route we take we should try to be as generic as possible (and as consistent as what makes sense).

yurivish:
Another aspect of generality is whether you anticipate ever wanting to support simultaneous hover / click selections. With the current event stream we assume that once a new non-sticky event comes in, the selected datum has been deselected. I've found the ability to support hovers and clicks at the same time to be very useful in more app-like contexts, where you can use them to support showing comparisons between the selected element and other (hovered) elements.

yurivish:
I think that the information about the _kind_ of selection should be communicated out-of-band, ie. not on the datum itself. I think this means that it would either be a property of the event object, or a wrapper object that contains the datum along with any metadata, eg. `plot.value = { done: ..., value: ... }`. There might be other options too.

To me word `done` seems to carry an implication that each hover selection is just an unfinished click selection, but I'll leave it to you and Mike to pick the best fit for the idea; sticky is also a bit odd to me, though both make sense :)

src/plot.js
```diff
@@ -171,11 +171,19 @@ export function plot(options = {}) {
     return {...state, channels: {...state.channels, ...facetState?.channels}};
   };
 
+  class StickyEvent extends Event {
+    constructor(type, options) {
+      super(type, options);
+      this.sticky = options.sticky;
+    }
+  }
```
mbostock, line 179: This should be moved outside the plot function so there’s a single (global) class.

yurivish:
I noticed that on my iPhone, the [example notebook](https://observablehq.com/@iopsystems/hover-and-click) only ever sets the `hover` variable, and not `click`. It's not immediately obvious to me why that might be – any ideas?

yurivish:
Right – there was a reason we moved the event inside the `plot` function: https://github.com/observablehq/plot/pull/1979#issuecomment-1918352258

PR #1995, opened 2024-02-14
scales: fix description

vwkd:
Found this reference in the description, which seems to not exist (anymore?).

Feel free to edit, if it should be changed instead of being deleted.


mbostock:
It’s here: https://observablehq.com/plot/features/plots#plot_legend

PR #2020, opened 2024-03-12
d3 7.9.0

mbostock:


Fil:
Should we not bump this dependency to 7.9.0? To guarantee d3.schemeObservable10?

  "dependencies": {
    "d3": "^7.8.0",

Fil:
The 5 tests that change are expected (and welcome… better geo circles).

Fil:
I'm still unsure if this works!?

in a notebook if I try to import 031cecf:

```js
Plot = import("https://esm.sh/gh/observablehq/plot@031cecfb4de3863403dd7ec67db4c48219de415b")
```

it errors like so:

```
Plot = SyntaxError: The requested module '/v135/d3@7.9.0/es2022/d3.mjs' does not provide an export named 'schemeObservable10'
```

mbostock:
Oh, yeah, that’s probably a bug.

mbostock:
Well, hold on, that’s always true of D3’s package.json. So I think that’s just an esm.sh issue.

Fil:
This is a bit over my head… 🤷🏼
I still don't see how one can make sure when upgrading  Plot that d3-scale-chromatic@3.1 is here (I mean, for non-yarn users, who are upgrading d3… probably not a best practice).
Manually: `yarn upgrade d3` isn't enough if the requirement for d3-scale-chromatic is already satisfied by a 3.0.0; I needed `yarn upgrade d3 d3-scale-chromatic` to ensure the new symbol was there.

Fil:
importing `Plot = import("https://esm.sh/gh/observablehq/plot#mbostock/d3-7.9.0")` works, so I'm even more confused :)

mbostock:
Relevant history is https://github.com/d3/d3/issues/3256.

Fil:
interesting! I think the answer to my question is https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it

```
  "resolutions": {
    "d3-scale-chromatic": "^3.1.0"
  },
```

(still only for yarn users, though)

juba:
I think I encountered the same problem. I updated `Plot` and `d3` to 0.6.14 and 7.9.0 for my `pyobsplot` package, but `d3-scale-chromatic` stayed at 3.0.0, provoking the same error:

```
SyntaxError: The requested module 'd3' does not provide an export named 'schemeObservable10'
```

I fixed it by (temporarily) adding a direct dependency to `"d3-scale-chromatic": "^3.1.0"` in my `package.json`.

PR #2026, opened 2024-03-20
upgrade actions; add publish workflow

mbostock:
This will allow us to publish Plot from GitHub Actions, as we do for Framework.


Fil (approved):
ref. https://github.com/observablehq/framework/blob/main/.github/workflows/publish.yml (exact copy, so I guess it should work the same :) )

PR #2031, opened 2024-03-22
lazily materialize the groups

Fil:
Makes lazy calls to `take` (which is expensive on arrow tables). The total # of calls to `take` in the tests goes from 34068 to 11778 (differenceY and differenceYVariable are the main consumers of this function). This should give better performance to the group transform on large datasets.

Related: #2030

### DIFF
--- a/src/transforms/group.js
+++ b/src/transforms/group.js
@@ -362,7 +362,8 @@ function reduceMaybeTemporalAccessor(f) {
 
 export const reduceIdentity = {
   reduceIndex(I, X) {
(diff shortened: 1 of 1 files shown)


mbostock (commented):
I don’t like relying on Proxy to do this — it’s too magical. Can we instead change the group transform to not compute arrays of data unless explicitly requested?

Fil:
More experimentation: I've pushed another strategy, even though I'm not convinced it's better. Maybe it will give us new ideas.

PR #2040, opened 2024-03-27
waffle mark 🧇

### LINKED ISSUE #2029: Waffle mark
It could be like rect, but renders using an SVG pattern to produce multiple squares. It should automatically figure out how many units wide it is to produce a square grid.

Ref. https://observablehq.com/@fil/waffle-chart
Ref. https://observablehq.com/@hrbrmstr/observable-plot-waffle-wafflemark

mbostock:
Fixes #2029.

- [x] waffleX
- [x] waffleY
- [x] Rounded corners (rx, ry)
- [x] Automatic row/column size determination
- [x] Fractional cell rendering
- [x] Configurable units
- [x] Configurable gap
- [x] Fix patterns in test snapshots
- [x] Fix the **r** option; don’t allow asymmetric corner rounding
- [x] Fix orientation of partial rows with negative values
- [x] Fix rendering stroke glitch with partial cells
- [x] Handle constant **fill** and **stroke** better
- [x] Consolidate and optimize waffle rendering
- [x] Documentation
- [x] Tests


mbostock:
Current status:

Screenshot 2024-08-01 at 11 43 54 AM

I’m now computing the number of waffle cell columns correctly, and rendering them as a pattern. But I need to replace the rect elements generated by BarY with a path element so that the top & bottom cells aren’t truncated (unless there’s a fractional value).

mbostock:
Now with fractional cell rendering!

Screenshot 2024-08-01 at 1 15 25 PM

test/output/waffleShorthand.svg
mbostock, line 1: Hmm, the snapshots are all busted. The pattern id is `#plot-clip-1`, but the fill is `url(#plot-clip-333)`. I’ll fix…

mbostock:
Whew! 😅 That was a lot of work but I’m pretty happy with how this turned out.

mbostock:
localhost_5173_plot_marks_waffle

mbostock:
localhost_5173_plot_marks_waffle

Fil:
I was wondering how to use this for ISOTYPE charts, and it appears that it's possible to use the render transform to edit the patterns’ contents:

~~~
render: (index, scales, values, dimensions, context, next) => {
  const g = next(index, scales, values, dimensions, context);
  const pattern = d3.select(g).selectAll("pattern");
  pattern.selectAll("rect").remove();
  pattern.append("text").attr("x", 10).attr("y", 15).text("🚀");
  return g;
}
~~~

Capture d’écran 2024-08-04 à 12 18 39

This gives two ideas for future enhancements:
- extend the mark to accept a **symbol** option, or
- extend the mark to accept a **pattern** option

The rendered pattern might also be easier to manipulate if it was centered on the origin with a viewBox rather than extending from it (in that case, the values for x and y above would essentially be zero).


~~~diff
@@ -62,10 +62,11 @@ function waffleRender(y) {
     const basePattern = document.createElementNS(namespaces.svg, "pattern");
     basePattern.setAttribute("width", cellsize);
     basePattern.setAttribute("height", cellsize);
+    basePattern.setAttribute("viewBox", [-cellsize / 2, -cellsize / 2, cellsize, cellsize]);
     basePattern.setAttribute("patternUnits", "userSpaceOnUse");
     const basePatternRect = basePattern.appendChild(document.createElementNS(namespaces.svg, "rect"));
-    basePatternRect.setAttribute("x", gap / 2);
-    basePatternRect.setAttribute("y", gap / 2);
+    basePatternRect.setAttribute("x", (gap - cellsize) / 2);
+    basePatternRect.setAttribute("y", (gap - cellsize) / 2);
     basePatternRect.setAttribute("width", cellsize - gap);
     basePatternRect.setAttribute("height", cellsize - gap);
     if (rx != null) basePatternRect.setAttribute("rx", rx);
~~~

src/marks/waffle.js
```diff
@@ -0,0 +1,202 @@
...
+    const barx = this[y === "y" ? "_x" : "_y"](scales, values, dimensions);
+
+    // The length of a unit along y in pixels.
+    const scale = unit * scaleof(scales.scales[y]);
+
+    // The number of cells on each row of the waffle.
+    const columns = Math.max(1, Math.floor(Math.sqrt(barwidth / scale)));
+
+    // The outer size of each square cell, in pixels, including the gap.
+    const cellsize = scale * columns;
```
Fil, line 52:
```suggestion
    const cellsize = Math.min(barwidth, scale * columns);
```

Clamping `cellsize` to `barwidth` fixes the situation where waffles overlap visually, typically when you only have low values: `Plot.waffleY([1, 1, 1, 1]).plot()`. It cheats a bit with the *y* axis but I think it makes a more readable chart.

| before | after |
| - | - |
| before | after |

Fil:
It makes total sense that the number of rows/columns is completely automatic and not configurable, but it might be useful to call this out explicitly in the documentation?

Fil:
It might be nice in the future to allow a configurable orientation (e.g. filling rows from the left instead of from the right). Not a blocker at all, I'm not even sure I would use it.

src/marks/waffle.js
mbostock: Unfortunately, you can’t satisfy both constraints: you can’t have both square waffle cells and avoid overlap. I think it’s more than “it cheats a bit” — it’s grossly misleading. For example here the value 10 looks like ~4:

Screenshot 2024-08-04 at 3 01 09 PM

```js
Plot.waffleX([10]).plot()
```

If we wanted to allow non-square waffle cells, then we could avoid overlap in this case. But it’s pretty easy to achieve effectively the same result with the **clip** option:

Screenshot 2024-08-04 at 3 02 15 PM

```js
Plot.waffleX([10]).plot({clip: true})
```

I think we could file a feature request to support non-square waffle cells, and I can take another pass at it now maybe. But it was just a pain because the code is written in such a way to not have duplicate implementations for waffleX and waffleY, so it’s weird keeping track of which dimension is which. And I don’t think it’ll be very common to want a waffle chart when you only have a few points (but it could occur if the data is dynamic).

mbostock:
> It makes total sense that the number of rows/columns is completely automatic and not configurable, but it might be useful to call this out explicitly in the documentation?

That’s already called out here:

> The waffle mark automatically determines the appropriate number of cells per row or per column (depending on orientation) such that the cells are square, don’t overlap, and are consistent with position scales.

It would also be very easy to expose this as an option. In fact I had it as an option in the past. But it was a footgun: if you hard-code the value it’s easy to produce overlapping waffles. In practice, what I want is to guarantee that the rows/columns is either a factor or a multiple of the corresponding position scale. The waffle mark doesn’t have that information easily available at render time, though (since the axis could be rendered after the waffle, or there could be multiple axes)… maybe an option to limit it to a power of ten times 1, 2, 5? Anyway.

The number of rows/columns is also heavily affected by the values, so I’m not sure I have much useful hints. Maybe just, if you have narrower bands, you have fewer cells per row/column? Setting **padding** on the ordinal scale for example.

mbostock:
> It might be nice in the future to allow a configurable orientation (e.g. filling rows from the left instead of from the right).

Sure… though I’m skeptical of adding such an obscure feature unless someone can explain why it’s worth adding. You could always do it with a custom mark or render transform.

mbostock:
The ISOTYPE chart idea is interesting, but I’m not sure how well the waffle would work for that, since often the glyphs are non-square, and they can even vary across series. Which means that there’s no meaningful position encoding, and you’re just responsible for counting glyphs to perceive the encoded value.

image

But for sure there are more extensions we could do here than square/rounded squares for the pattern! I don’t think we should implement these now, with the possible exception of non-square aspect ratios as previously mentioned.

src/marks/waffle.js
Fil: Yes the use case I have in mind is when you want to animate / build up a chart by adding units one by one.

mbostock: Okay, added support for squished waffles.

Fil: Thanks for the counter example, I hadn't thought of the thin + long case.

A solution for my chart `Plot.waffleY([1, 1, 1, 1]).plot()` is to add an invisible mark such as `Plot.waffleY([4], {fill: []})`.

Fil (approved):
🧇 🧇 🧇

PR #2043, opened 2024-04-03
Patch analytics pageLoad event for now

visnup:
This actually isn't correct either, but restores the previous behavior that it seems to rely on. The current code watches router.route, but after an async import which means it actually misses the first actual page load and doesn't fire anything. A subsequent navigation counts as the first pageLoad.

I'll have a better fix soon, but wanted to restore the previous behavior in the meantime.


tophtucker (approved):
👍

PR #2046, opened 2024-04-08
Copy initial value from svg to figure

### LINKED ISSUE #2045: The initial value of a plot with a title and a pointer transform is undefined
It should be null, just like when there is no title.

Note that this is not merely an inconsistency, because it prevents downstream variables to resolve.

reported by @zachbogart

Fil:
closes #2045


mbostock (commented):
Seems correct but I’m curious how the svg.value gets set erroneously in the first place, and what prevents it from subsequently getting set again in the wrong place? (I can investigate this myself by looking at the code, but if you’re willing it would be nice to summarize more of what the underlying cause is here.)

Fil:
The pointer transform initializes svg.value:
https://github.com/observablehq/plot/blob/00610bbc1b32f6689fa71559f3db1f572688acfb/src/interactions/pointer.js#L182

it dispatches here before we actually have a figure:
https://github.com/observablehq/plot/blob/00610bbc1b32f6689fa71559f3db1f572688acfb/src/interactions/pointer.js#L129

mbostock (approved):
Thanks!

PR #2051, opened 2024-04-11
remove unused exports?

Fil:
I obtained the list with:
`>  npx ts-unused-exports tsconfig.json  src/**/*.js src/**/*ts test/**/*.js test/**/*.ts`

Not sure we need to remove all of this? Some things seem to be here for consistency.


src/format.js
```diff
@@ -14,7 +14,7 @@ const weekdayFormat = memoize1((locale, weekday) => {
   return new Intl.DateTimeFormat(locale, {timeZone: "UTC", ...(weekday && {weekday})});
 });
 
-export function formatNumber(locale = "en-US") {
+function formatNumber(locale) {
```
Fil, line 17: here I learned that we don't need to set the default locale (cf. #384) in this function, since it's only called from formatAuto.

src/options.js
```diff
@@ -50,12 +50,10 @@ export const singleton = [null]; // for data-less decoration marks, e.g. frame
 export const field = (name) => (d) => d[name];
 export const indexOf = {transform: range};
 export const identity = {transform: (d) => d};
-export const zero = () => 0;
 export const one = () => 1;
 export const yes = () => true;
 export const string = (x) => (x == null ? x : `${x}`);
 export const number = (x) => (x == null ? x : +x);
-export const boolean = (x) => (x == null ? x : !!x);
```
Fil, line 58: this seems to be here for consistency only

mbostock (commented):
I like removing the unused code, but I think we should keep the exports in most cases for consistency — for things that are conceptually “internally public” (_i.e._, intended to be used in other modules, even if they aren’t yet).

PR #2055, opened 2024-04-13
Fix Axis Tick Vector to Axis Tick Line

Stroked:
Axis uses Tick and Tick uses Line.

In the docs for Axis, there is repeated use of the term vector. Since, vector is also a mark type, it is kinda confusing.

### DIFF
--- a/docs/marks/axis.md
+++ b/docs/marks/axis.md
@@ -64,7 +64,7 @@ If you don’t declare an axis mark for a position scale, Plot will implicitly a
 
 Plot’s axis mark is a composite mark comprised of:
 
-* a [vector](./vector.md) for ticks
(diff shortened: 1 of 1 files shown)


mbostock:
The axis mark indeed does use the [vector mark](https://observablehq.com/plot/marks/vector) (not the line mark) to render ticks, as you can see from the implementation:

https://github.com/observablehq/plot/blob/943526511c163ef9ad6afd2031b073b97fdacc6e/src/marks/axis.js#L282

So these proposed edits are not correct.

Stroked:
ah, I see. 

How does the vector turn off the pointer? 

For the arrow mark, there is headLength. For the vector, is the draw function the only way?

PR #2069, opened 2024-06-02
lift side effect up to index.js

mbostock:
Ref. https://github.com/rollup/rollup/issues/5538


src/index.js
```diff
@@ -1,3 +1,11 @@
+import {Mark} from "./mark.js";
+import {plot} from "./plot.js";
+
+// Note: this side effect avoids a circular dependency.
+Mark.prototype.plot = function plotThis({marks = [], ...options} = {}) {
```
Fil, line 5:
```suggestion
Mark.prototype.plot = function ({marks = [], ...options} = {}) {
```

I don't think we need to name this function?

Fil (approved):
Thanks for the detailed bug report & repro repo.

I've disabled auto-merging since I have a suggestion, but it LGTM in any case.

PR #2074, opened 2024-06-08
tip title format

### LINKED ISSUE #2053: The tip mark’s format option should not be ignored when a title channel is present
Especially confusing when the title channel is implicit…

```js
Plot.tip(numbers, {x: Plot.index, y: Plot.identity, anchor: "bottom"})
```

I guess in this case the **format** option could be specified as a function, rather than an object?

mbostock:
Fixes #2053.

### DIFF
--- a/src/marks/tip.d.ts
+++ b/src/marks/tip.d.ts
@@ -74,7 +74,7 @@ export interface TipOptions extends MarkOptions, TextStyles {
    * is interpreted as a (UTC) time format for temporal channels, and otherwise
    * a number format.
    */
-  format?: {[name in ChannelName]?: boolean | string | ((d: any, i: number) => string)};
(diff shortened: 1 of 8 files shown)


src/marks/tip.d.ts
```diff
@@ -86,6 +86,18 @@ export interface TipOptions extends MarkOptions, TextStyles {
   textPadding?: number;
 }
 
+/**
+ * How to format channel values; one of:
+ *
+ * - a [d3-format][1] string for numeric scales
+ * - a [d3-time-format][2] string for temporal scales
+ * - a function passed a tick *value* and *index*, returning a string
```
Fil, line 94: is this _tick_ (or _tip_ or _title_)?

mbostock: It should be *channel* value here.

Fil (approved):

PR #2078, opened 2024-06-10
formatNumber; filter log tickFormat function

### LINKED ISSUE #2077: For a log scale, the tickFormat function should only be considered for ticks that should be visible
Log scales are a bit unusual in that all the ticks are drawn independent of the **ticks** option; instead, the **ticks** option only controls which ticks are labeled.

untitled - 2024-06-09T164644 913

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 7}})
```

untitled - 2024-06-09T164647 801

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 10}})
```

untitled - 2024-06-09T164649 511

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 20}})
```

This means that if you set the **tickFormat** option with a log scale, it (for the most part) ignores the **ticks** option. These three specifications produce equivalent output:

untitled - 2024-06-09T164820 342

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 7, tickFormat: String}})
```

untitled - 2024-06-09T164820 342

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 10, tickFormat: String}})
```

untitled - 2024-06-09T164820 342

```js
Plot.plot({x: {type: "log", domain: [1, 4200], ticks: 20, tickFormat: String}})
```

It‘d be cleaner if the **ticks** option filtered the tick labels, composing with the given **tickFormat**, rather than baking the filtering into the tick format. That way you could override the tick format without affecting the number of displayed ticks.

mbostock:
Fixes #2077. Also exposes Plot.formatNumber(*locale*).

### DIFF
--- a/src/marks/axis.js
+++ b/src/marks/axis.js
@@ -652,7 +652,7 @@ function inferTextChannel(scale, data, ticks, tickFormat, anchor) {
 // possible, or the default ISO format (2014-01-26). TODO We need a better way
 // to infer whether the ordinal scale is UTC or local time.
 export function inferTickFormat(scale, data, ticks, tickFormat, anchor) {
-  return typeof tickFormat === "function"
+  return typeof tickFormat === "function" && !(scale.type === "log" && scale.tickFormat)
     ? tickFormat
     : tickFormat === undefined && data && isTemporal(data)
     ? inferTimeFormat(scale.type, data, anchor) ?? formatDefault
--- a/test/output/logTickFormatFunctionSv.svg
+++ b/test/output/logTickFormatFunctionSv.svg
@@ -1,0 +1,82 @@
+<svg class="plot" fill="currentColor" font-family="system-ui, sans-serif" font-size="10" text-anchor="middle" width="640" height="60" viewBox="0 0 640 60" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <style>
+    :where(.plot) {
+      --plot-background: white;
+      display: block;
+      height: auto;
+      height: intrinsic;
+      max-width: 100%;
+    }
+
+    :where(.plot text),
+    :where(.plot tspan) {
+      white-space: pre;
+    }
+  </style>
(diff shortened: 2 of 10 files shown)


src/marks/axis.js
```diff
@@ -652,7 +652,7 @@ function inferTextChannel(scale, data, ticks, tickFormat, anchor) {
 // possible, or the default ISO format (2014-01-26). TODO We need a better way
 // to infer whether the ordinal scale is UTC or local time.
 export function inferTickFormat(scale, data, ticks, tickFormat, anchor) {
-  return typeof tickFormat === "function"
+  return typeof tickFormat === "function" && !(scale.type === "log" && scale.tickFormat)
```
Fil, line 655:
```suggestion
  return typeof tickFormat === "function" && scale.type !== "log"
```

Nit: I'm not sure about the case for the second part of the test. When scale.type is "log", the scale is an instance of d3.scaleLog() and it has a tickFormat method, which the user can't nullify.

test/output/logTickFormatFunctionSv.svg
Fil, line 1: Error may be mine, but this "invalid" message comes from the format in many European languages returning `2&nbsp;000` which is not accepted in SVG.

Fil (commented):
I like the feature — I hope we'll find a way to solve #384 too!

Two small comments/questions.

test/output/logTickFormatFunctionSv.svg
mbostock: The problem comes from running `beautify.html` on the generated SVG, meaning that the SVG is assumed to be in HTML. We’ll have to find a way to disable that for tests.

src/marks/axis.js
mbostock: I didn’t do this because I didn’t want to assume that log scales necessarily implement `scale.tickFormat`. That is currently true, but if this assumption doesn’t hold in the future, this could would break.

Fil (approved):

PR #2083, opened 2024-06-13
zoom option

### LINKED ISSUE #1590: Panning and zooming
It would be nice to support pan-and-zoom interaction (à la d3-zoom) in dense scatterplots with Plot. Presumably this would work by altering the scale domains and then re-rendering the plot.

mbostock:
This adds a top-level **zoom** option which can be set to true to enable panning and zooming (via [d3-zoom](https://d3js.org/d3-zoom)). The option can also be set to *xy* to zoom both the *x* and *y* scales, *x* to only zoom the *x* scale, *y* to only zoom the *y* scale, or false (the default) or null to disable zooming.

Setting the **zoom** option is functionally equivalent to setting the **domain**s of the *x* and *y* scales interactively; all marks that use position scales are affected, as well as any transforms or initializers that may depend on these scales. (See [Julien Colot’s notebook](https://observablehq.com/d/223f34bca10702a3) for a userland example of this technique.)

The goal of this initial implementation is not optimal performance, but rather functional correctness. I make some effort at performance — for example, tracking which channels need to be rescaled when the *x* or *y* domains change, and only re-rendering the affected marks — but there is room for future improvement, especially if mark implementations can assist by updating the DOM directly rather than re-rendering and replacing.

I am also interested in exploring usability improvements to panning and zooming. I’d like keyboard shortcuts (when the plot is focused), such as + to zoom in, - to zoom out, and arrow keys ↑↓←→ to pan. I also want a reset button to zoom to the home view, but I don’t know yet where that button should live. We might also support one-dimensional zooming by clicking in the axes.

I’d also like programmatic zooming with zoom transitions (a _plot_.zoom method), and a configurable duration for zoom transitions (such as double-clicking to zoom in). Another possibility is a dedicated method for updating a scale’s domain and recomputing everything downstream, though at least for now I’m just interested in position scales (_x_ and _y_).

TODO

- [x] Basic panning and zooming
- [ ] Handle _x_ or _y_ scales not being defined
- [ ] Handle _x_ or _y_ scales not being zoomable
- [ ] Keyboard shortcuts
- [ ] Reset button
- [ ] Constrained zoom (either translate or scale)
- [ ] Handle mark initializers
- [ ] Handle mark transforms
- [ ] Handle faceted marks
- [ ] Handle conditional marks

Fixes #1590.

Previously #1738 #1964.

### DIFF
--- a/src/plot.d.ts
+++ b/src/plot.d.ts
@@ -298,6 +298,9 @@ export interface PlotOptions extends ScaleDefaults {
    */
   projection?: ProjectionOptions | ProjectionName | ProjectionFactory | ProjectionImplementation | null;
 
+  /** TODO */
+  zoom?: boolean;
+
   /**
    * Options for the horizontal facet position *fx* scale. If present, the *fx*
    * scale is always a *band* scale.
--- a/src/plot.js
+++ b/src/plot.js
@@ -1,4 +1,4 @@
-import {creator, select} from "d3";
+import {creator, select, zoom as Zoom} from "d3";
 import {createChannel, inferChannelScale} from "./channel.js";
 import {createContext} from "./context.js";
 import {createDimensions} from "./dimensions.js";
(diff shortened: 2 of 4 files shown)


mbostock:
Work in progress…

https://github.com/observablehq/plot/assets/230541/12b7cbc4-8772-4969-a425-1ebbcf161db0

PBI-David:
How about using the escape key to reset the zoom? That way, everything can be controlled via keyboard shortcuts.

PR #2086, opened 2024-06-14
fix centroid/geocentroid with tips

### LINKED ISSUE #2085: centroid (and geoCentroid) sometimes fail when using tips
see https://talk.observablehq.com/t/faceting-a-choropleth/9387

Fil:
closes #2085

cc: @jwoLondon ~~(and can you give us provenance information for the dataset?)~~

~~An imperfection is that x and y show up in the tip in the londonCarAccessGeoCentroid example.~~


jwoLondon:
I derived the car access data from the UK Census (all covered by [Open Government Licence v3.0](http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/). Specifically the local authority level spatial aggregations for:

- 2021 : https://www.ons.gov.uk/datasets/TS045/editions/2021/versions/4
- 2011 : https://www.nomisweb.co.uk/census/2011/qs416ew
- 2001: https://www.nomisweb.co.uk/census/2001/cs060

(proportions calculated by me by comparing with total households available in the same tables).

Spatial London borough data derived from https://geoportal.statistics.gov.uk/datasets/caaf81b26a45453fa4849dbce31a3293_0/explore by selecting, simplifying and reprojecting in MapShaper.

Fil:
This is ready to review.

src/channel.js
```diff
@@ -5,15 +5,19 @@ import {registry} from "./scales/index.js";
 import {isSymbol, maybeSymbol} from "./symbol.js";
 import {maybeReduce} from "./transforms/group.js";
 
-export function createChannel(data, {scale, type, value, filter, hint, label = labelof(value)}, name) {
-  if (hint === undefined && typeof value?.transform === "function") hint = value.hint;
+export function createChannel(data, {scale, type, value, filter, hint, source, label = labelof(value)}, name) {
+  if (typeof value?.transform === "function") {
+    if (hint === undefined) hint = value.hint;
+    if (source === undefined) source = value.source;
```
mbostock, line 11: Why is this necessary? Should we do this in the `column` helper in `options.js` instead (per other comment)?

mbostock (changes requested):
I would love more comments to accompany code changes, either in the PR description, in self-review comments, or in the code itself. Without them I have to guess why the changes are necessary; this is a good exercise in active thinking, but it would be more efficient if you gave me some hints. Thank you!

src/marks/geo.js
```diff
@@ -23,6 +23,8 @@ export class Geo extends Mark {
       data,
       {
         geometry: {value: options.geometry, scale: "projection"},
+        x: {value: options.x, scale: "x", optional: true},
+        y: {value: options.y, scale: "y", optional: true},
```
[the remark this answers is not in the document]
Fil: They are used by the tip mark, which _derives_ the x and y channels from the geo mark. That's visible when interacting with the londonCarAccessGeoCentroid example (but unfortunately not caught by the automated test).

src/channel.js
Fil: This is to prevent the tip in londonCarAccessGeoCentroid to display x and y, which are meaningless pixel values. I'll try and use column.

src/marks/geo.js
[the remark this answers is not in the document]
mbostock: If I’m understanding correctly, you’re suggesting that the geo mark should be responsible for preparing the _x_ and _y_ channels for the **tip** option (and its derived tip mark) even though the geo mark doesn’t need these channels for itself.

This doesn’t feel like the right place to fix this problem. Either all marks should implicitly support _x_ and _y_ channels if declared under the expectation that maybe you’ll use the **tip** option and the derived tip mark will need them, or the geoCentroid transform should be responsible for declaring _x_ and _y_ as custom channels rather than simply passing them through as options.

Another possibility is that the geo mark should just support the **tip** option directly by implicitly invoking the geoCentroid transform (somehow?), or that the `derive` function for deriving a tip from an existing mark should similarly promote the _geometry_ channel into _x_ and _y_ channels, or that the tip mark and pointer transform should directly support reading the _geometry_ channel as an alternative to _x_ and _y_ channels.

mbostock:
A related question is should the **tip** option work directly with the geo mark (without requiring the geoCentroid transform explicitly)? It sure seems like that would be convenient. But using centroids for pointing at polygons maybe isn’t great, so I wonder if that’s an acceptable implementation, or if we should be thinking about other pointing mechanisms (_e.g._, SDF).

Fil:
(tracking the secondary issue of default tips in https://github.com/observablehq/plot/issues/2087)

test/plots/london-maps.ts
```diff
@@ -0,0 +1,72 @@
...
+        access,
+        Plot.centroid({
+          fx: "year",
+          geometry: (d) => boroughs.get(d.borough),
+          fill: "access",
+          stroke: "var(--plot-background)",
+          strokeWidth: 0.75,
+          tip: true
+        })
+      )
```
mbostock, line 40: I guess I’m not sure we should/need to support this use case now. A more explicit way to write this is:

```suggestion
      Plot.geo(access, {
        fx: "year",
        geometry: (d) => boroughs.get(d.borough),
        fill: "access",
        stroke: "var(--plot-background)",
        strokeWidth: 0.75
      }),
      Plot.tip(
        access,
        Plot.pointer(Plot.centroid({fx: "year", geometry: (d) => boroughs.get(d.borough), fill: "access"}))
      )
```

I don’t think it’s right to use the centroid/geoCentroid transform with the geo mark if the geo mark doesn’t need the _x_ and _y_ channels — if they’re just intended to be used with the **tip** option. It feels like we should just make the **tip** option work directly with the geo mark rather than going halfway.

mbostock (changes requested):

test/plots/london-maps.ts
mbostock, line 1: As I’ve said before, we should try to name tests more about the features they are testing than the datasets they are using. E.g., this should be a test of geo tips.

mbostock:
I’ve made a sketch at my alternate take in #2088. I think it could be further extended (for example maybe there’s an option to use the geoCentroid transform instead of the centroid transform, or maybe there’s an option to tell the pointer transform to use the nearest polygon instead of the nearest centroid — though in that case, we’d still need to pick an _xy_ point to place the tip) but maybe it’s an acceptable interim solution.

Fil:
superseded by #2088

Issue #1587, opened 2023-05-18
A better “label centroid” initializer?


Fil:
The centroid of a geometry is sometimes outside of it, for example (but not exclusively) with a MultiPolygon such as France+Guyane. This can result in labels or interactive tips that are quite far off from the shape they identify.

A great solution is to use @mapbox/polylabel, which guarantees that the center is inside —and furthermore at an optimal position for a centered label. The dependency is less that 1kB over the wire, including tinyqueue. An alternative approach could be to add an option to Plot.centroid, to make it choose the centroid of the largest polygon when given a multipolygon. Subpar solution, but cheaper.

fgregg:
the centroid of the largest polygon could also lead to the point being outside the polygon. this would be the case for a crescent-moon shaped polygon.

polylabel sounds great!

Fil:
We should probably call it labelCentroid, with an ellipseRatio option that defaults to 2 (?). And maybe add an _anchor_ option so that we can align text on the left (with textAnchor: start) or on the right (with textAnchor: end). Since polylabel returns the radius, it's not expensive to add.

fgregg:
this approach could also make it easier to scale labels so they fit inside of polygons, since you'll know the ellipse dimensions anyway.

Fil:
Related: 
* [ST_MaximumInscribedCircle](https://postgis.net/docs/ST_MaximumInscribedCircle.html): equivalent of the pole of inaccessibility.
* [ST_PointOnSurface](https://postgis.net/docs/ST_PointOnSurface.html): no particular specification re: the quality of the point. It just is on the surface.

PR #2088, opened 2024-06-14
promote geometry to x and y for tip

### LINKED ISSUE #1743: Plot.geo with tip: true should use Plot.centroid

### LINKED ISSUE #2085: centroid (and geoCentroid) sometimes fail when using tips

### LINKED ISSUE #2087: Add default tips for the geo mark
If the geo mark has no x and y channels, it would be nice if the **tip** option worked. It could create x and y with a centroid initializer. However, a system that tracks the closest geometry, relying on distance to the shape rather than to its centroid, would arguably be better. The difference in strategy is striking if you consider France, whose centroid is in the Atlantic, because Guyane. But centroids show the small features better.
 

(Note: this is a duplicate of #1743)

mbostock:
Fixes #1743. Fixes #2085. Fixes #2087. This makes the **tip** option “just work” with the geo mark. In the future, it’d be nice to have more control — say to use the geoCentroid transform instead of the centroid transform, or to compute the closest polygon instead of the closest centroid. But maybe this is okay for now?


Fil:
This is much easier to use and helps with #1743 (and maybe closes it? a secondary issue is the default tip contents, which I imagined could be based on the feature’s properties, if the geometry is a feature).

I still find it a bit troubling (in principle) that these two codes behave differently, with Plot.centroid “eating” the geometries if it is passed a non-default geometry option:

```
Plot.geo(access.map((d) => boroughs.get(d.borough)), Plot.centroid()) // displays the geometries
Plot.geo(access, Plot.centroid({geometry: (d) => boroughs.get(d.borough)})) // nothing is displayed
```

This can lead to surprising combinations: the code below draws the original shapes, and displays a tooltip on the geoCentroid.

```
Plot.geo(
  access.map((d) => boroughs.get(d.borough)),
  Plot.centroid({
    title: d => d.id,
    tip: true,
    geometry: (d) => ({type: "Point", coordinates: d3.geoCentroid(d)})
  })
)
```

But  if you remove Plot.centroid, then the Point geometries are drawn instead of the shapes:

```
Plot.geo(
  access.map((d) => boroughs.get(d.borough)),
  {geometry: (d) => ({type: "Point", coordinates: d3.geoCentroid(d)}), tip: true}
)
```

Lastly, I also want to explore a better default placement based on the pole of inaccessibility (#1587) — but this doesn't have to be blocking and can be done as a follow-up.

mbostock:
From my perspective, it never makes sense to use the centroid or geoCentroid transform with the geo mark. The centroid and geoCentroid transform turn {geometry} into {x, y}. The geo mark accepts {geometry} while most other marks accept {x, y}. In other words, the centroid transforms turn a square peg into a round peg, while the geo mark is a square hole.

Having the centroid and geoCentroid transforms also pass through the _geometry_ channel could work (instead of consuming them), but then we get back to the problem of the geo mark or the geoCentroid transform being responsible for declaring _x_ and _y_ channels even if it doesn’t need them directly. But it could do that only if the **tip** option were truthy, under the principle that the **tip** option is something all marks should support.

I can look into this approach tonight… it has the nice property that if you use the geoCentroid transform explicitly, it would disable the default centroid transform used by the **tip** option.

mbostock:
Okay, some big changes in the latest commit. This enables you to use the centroid or geoCentroid transform, or even to specify the _x_ and _y_ channels explicitly, in conjunction with the **tip** option. If you don’t specify _x_ or _y_ explicitly, then the geo mark will implicitly apply the centroid transform when the **tip** option is specified. I think this gives the most flexibility and the most convenience, and it doesn’t require any special treatment in `derive`.

In addition, the centroid and geoCentroid transforms now propagate the (memoized) _geometry_ channel instead of consuming it, just in case the downstream mark still wants access to it. I also (necessarily) made the geoCentroid transform more robust; it shouldn’t assume that the _x_ channel is computed before the _y_ channel (and before the _geometry_ channel); in some cases only a subset of channels will be computed.

Lastly I optimized `memoize1` for the single-argument case, and fixed the equality test which should use [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).

Fil (approved):

PR #2091, opened 2024-06-15
tip ignores literal colors

### LINKED ISSUE #2090: When a color channel is specified in literal color values, it should be ignored by the tip mark
For example, the **fill** channel appearing in the tip here isn’t very useful:

untitled - 2024-06-14T174016 066

```js
Plot.plot({
  grid: true,
  marks: [
    Plot.dot(penguins, {
      x: "culmen_length_mm",
      y: "culmen_depth_mm",
      fill: (d) => d.species === "Adelie" ? "red" : "black",
      tip: true
    })
  ]
})
```

mbostock:
Fixes #2090. Uses a new `channel.defaultScale` field to record what the `channel.scale` _would_ have been, if it weren’t for the specific values provided. This allows the tip mark to detect when a channel would have been associated with the _color_ scale, but isn’t, and hence contains only literal color values that should not be shown (at least by default) in the tip.

Screenshot 2024-06-14 at 6 08 37 PM

### DIFF
--- a/src/channel.js
+++ b/src/channel.js
@@ -48,11 +48,13 @@ export function inferChannelScale(name, channel) {
       case "stroke":
       case "color":
         channel.scale = scale !== true && isEvery(value, isColor) ? null : "color";
+        channel.defaultScale = "color";
         break;
       case "fillOpacity":
       case "strokeOpacity":
       case "opacity":
         channel.scale = scale !== true && isEvery(value, isOpacity) ? null : "opacity";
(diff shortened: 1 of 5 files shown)


Fil (approved):
The example in this PR would have displayed `color: orange` in the tips, which was not very informative.

There might be cases where you want the tip to display the actual color (for example if you are doing a visualization of colors) — this is possible by using the tip’s **format** option (_e.g._ `tip: {format: {fill: true}}`).

PR #2092, opened 2024-06-16
more GeoJSON awareness

### LINKED ISSUE #1172: Shorthand for GeoJSON properties
It’d be nice if Plot.geo supported shorthand for GeoJSON properties, e.g., 

```js
Plot.geo(counties, {fill: "rate"})
```

would be shorthand for

```js
Plot.geo(counties, {fill: d => d.properties.rate})
```

I’m not sure how to do it, though, because in theory the properties could be on the features directly, and we’d still want it to work in that case.

### LINKED ISSUE #1745: Make Plot more geojson-aware
Only the geo mark knows that data can be a FeatureCollection of a GeometryCollection. This makes pivoting from Plot.geo to Plot.dot a 3-changes operation:

> Plot.geo(countries) => Plot.**dot**(countries **.features**, **Plot.centroid()**)

In the spirit of #1743/#1744 we could have all the marks:
1. accept geo collections as data, and add a geometry: d => d option
2. apply the centroid initializer if they consume x, y and there is a geometry (and no x, y or frameAnchor); similarly, apply a _bounds_ initializer if they consume x1, y1, x2, y2 and …

mbostock:
Fixes #1745. Fixes #1172.

First, this changes the `arrayify` internal such that if `values` is a GeoJSON object, the corresponding array of features or geometries is returned. This makes GeoJSON objects usable as mark data for any mark, rather than being limited to the geo mark.

Second, this changes the `field` internal such that if the requested field is undefined on the given object `d`, it falls back to `d.properties` if `d` is a GeoJSON Feature object (`d.type === "Feature"`). This may have a slight performance overhead for such a core routine, but I expect the cost is negligible and I don’t think there is an alternative if we want GeoJSON to be more universally supported.


Fil (approved):
combine this with #2088 and it's total map bliss!

PR #2095, opened 2024-06-18
checkpoint markers

mbostock:
Not quite working, but trying to illustrate what I mean in https://github.com/observablehq/plot/pull/2094#pullrequestreview-2124039917. Could probably still be improved…

### DIFF
--- a/src/marker.js
+++ b/src/marker.js
@@ -1,5 +1,6 @@
-import {select, selectAll, group} from "d3";
 import {create} from "./context.js";
+import {unset} from "./memoize.js";
+import {keyof} from "./options.js";
 
 export function markers(mark, {marker, markerStart = marker, markerMid = marker, markerEnd = marker} = {}) {
   mark.markerStart = maybeMarker(markerStart);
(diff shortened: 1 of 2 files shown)


Fil:
Why do you say it's not quite working? It looks good to me :) (And the implementation is much nicer.)

src/marker.js
```diff
@@ -122,30 +162,12 @@ function applyMarkersColor(path, {markerStart, markerMid, markerEnd, stroke}, st
...
-      }
-      if (end) select(g.at(-1)).attr("marker-end", end);
-    }
-  } else {
-    if (start) path.attr("marker-start", start);
-    if (mid) path.attr("marker-mid", mid);
-    if (end) path.attr("marker-end", end);
-  }
+  if (markerStart) path.each(applyMarker("marker-start", markerStart, orient && ((i) => orient(i) & START)));
+  if (markerMid && orient) path.each(applyMarker("marker-start", markerMid, (i) => orient(i) ^ START));
```
Fil, line 170:
```suggestion
  if (markerMid && orient) path.each(applyMarker("marker-start", markerMid, (i) => !(orient(i) & START)));
```

(to not evaluate to END when a segment is START+END = 3)

PR #2096, opened 2024-06-21
fix date typing in arrow tables

Fil:
uses the same duck type test as https://github.com/observablehq/inputs/pull/263

addresses https://github.com/observablehq/framework/issues/1376 (for the Plot part)

Note: it's not as good as full Arrow support (where the data would not get transformed into an array of Proxy objects, and valueof would use _table_.getChild, see #2030).

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@types/node": "^20.5.0",
     "@typescript-eslint/eslint-plugin": "^7.2.0",
     "@typescript-eslint/parser": "^7.2.0",
+    "apache-arrow": "^16.1.0",
     "c8": "^9.1.0",
     "canvas": "^2.0.0",
     "d3-geo-projection": "^4.0.0",
@@ -87,7 +88,6 @@
     ]
   },
   "dependencies": {
-    "apache-arrow": "^16.1.0",
     "d3": "^7.9.0",
     "interval-tree-1d": "^1.0.0",
     "isoformat": "^0.2.0"
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     ]
   },
   "dependencies": {
+    "apache-arrow": "^16.1.0",
     "d3": "^7.9.0",
     "interval-tree-1d": "^1.0.0",
     "isoformat": "^0.2.0"
--- a/src/options.js
+++ b/src/options.js
@@ -148,6 +148,20 @@ export function arrayify(values) {
     case "Sphere":
       return [values];
   }
+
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     ]
   },
   "dependencies": {
+    "apache-arrow": "^16.1.0",
```
mbostock, line 90: This should be in devDependencies.

src/options.js
```diff
@@ -148,6 +148,20 @@ export function arrayify(values) {
...
+
+  // Duck type Arrow tables to retype date fields to Dates. Note that we only
+  // need the first non-nullish value to be typed correctly for isTemporal to
+  // return true.
+  const fields = values?.schema?.fields;
+  if (Array.isArray(fields)) {
+    values = Array.from(values);
+    for (const f of fields) {
+      if (String(f).endsWith("<MILLISECOND>"))
+        values.some((d, i) => d[f.name] != null && (values[i] = {...values[i], [f.name]: new Date(values[i][f.name])}));
```
mbostock, line 160: We already have Date coercion if the scale type is set to temporal. Can we use that instead here? I.e., change the isTemporal test instead of changing arrayify? I think we want to minimize the logic here.

mbostock (changes requested):

src/options.js
Fil: I don't know if we could do that. As soon as we apply Array.from to the arrow table, the schema information that we need to distinguish between a column of numbers and a column of dates is lost. The data becomes an Array of Proxy objects, with getters that return numbers (even on date-typed columns, per https://github.com/apache/arrow/pull/40892 — note that this change discusses a future improvement to make the Proxy return Date objects again, but I don't think it has happened yet).

But in fact, applying Array.from is already a losing proposition performance-wise, and we should use the columnar format directly (#2030). In a sense, this PR is only a stop-gap measure. I've ported the necessary changes to #2030 too, if we want to take that leap instead.

mbostock:
I incorporated this into #2115:

https://github.com/observablehq/plot/blob/4faebbc00e73c317823da89a94fac5c7eef0af4e/src/options.js#L67-L73

I think it would be slightly better in the future to avoid materializing the values as `Date` instances, and instead somehow retain a hint on the column that the numbers represent dates (like Apache Arrow does with `vector.type`). But that’s “just” an optimization that we can consider in the future.

Fil:
superseded by #2030

PR #2102, opened 2024-07-11
Determine the default height when the projection domain is set

### LINKED ISSUE #2063: Determine the default height when the projection domain is set
We currently determine the default height for whole-world projections automatically (for named projections):

https://github.com/observablehq/plot/blob/4bbad40906e851fe5ece146f34ecc854342b08da/src/projection.js#L237-L253

It’d be nice to do that if the **domain** is specified, too. For example:

```js
Plot.plot({
  projection: {
    type: "conic-conformal",
    parallels: [34.3333, 36.1667],
    rotate: [79, 0],
    domain: ncstate
  },
  marks: [
    Plot.geo(ncstate),
    Plot.geo(nccounties, {strokeOpacity: 0.1})
  ]
})
```

Fil:
(I don't think this feature is much documented, it "just works".)

closes #2063


src/projection.js
```diff
@@ -167,8 +167,10 @@ function scaleProjection(createProjection, kx, ky) {
       if (precision != null) projection.precision?.(precision);
       if (rotate != null) projection.rotate?.(rotate);
       if (typeof clip === "number") projection.clipAngle?.(clip);
-      projection.scale(Math.min(width / kx, height / ky));
-      projection.translate([width / 2, height / 2]);
+      if (width && height) {
```
mbostock, line 170: Should this check for nullish instead of zero?

Fil: good catch! I was thinking aspectRatio, but this part returns the actual projection, so it should respect a collapsed (degenerate) projection.

I fixed it and added a test case (without the fix, the test case displays a map in the default scale, instead of a 0-height map…).

Note that the degenerate case is still slightly problematic, as its path’s `d` property is an invalid string `"M NaN,NaN…"` instead of being null. (A projection with a strictly negative scale returns null, so the "issue" is only for scale 0.)

mbostock (approved):

PR #2114, opened 2024-07-26
Export inferTimeFormat function

mkfreeman:
This PR simply exports the `inferTimeFormat` function. 

The `inferTimeFormat` function is quite useful, and also depends on a number of other methods and constants (making it difficult to use without defining functions such as `formatTimeInterval` and `getTimeTemplate`).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -57,4 +57,4 @@ export {formatIsoDate, formatNumber, formatWeekday, formatMonth} from "./format.
 export {scale} from "./scales.js";
 export {legend} from "./legends.js";
 export {numberInterval} from "./options.js";
-export {timeInterval, utcInterval, inferTimeFormat} from "./time.js";
+export {inferTimeFormat, timeInterval, utcInterval} from "./time.js";
--- a/src/index.js
+++ b/src/index.js
@@ -57,4 +57,4 @@ export {formatIsoDate, formatNumber, formatWeekday, formatMonth} from "./format.
 export {scale} from "./scales.js";
 export {legend} from "./legends.js";
 export {numberInterval} from "./options.js";
-export {timeInterval, utcInterval} from "./time.js";
+export {timeInterval, utcInterval, inferTimeFormat} from "./time.js";


mbostock:
It’s unlikely we’d expose this internal method without a better understanding of the desired use case and with further thought to the interface. I saw the [discussion in the forum](https://talk.observablehq.com/t/plot-get-formatted-axis-tick-values/9550), but this change seems specific to time formats, so, I’m guessing you want it for other reasons? This might be better as an issue to discuss what you’re trying to do.

mkfreeman:
Thanks for your reply, @mbostock - what I'm really after is the formatted axis tick values so that I can compute their width/height and adjust the margins to leave sufficient space for the axis tick labels. Getting that behavior into Plot would be awesome, but I recognize that doing so efficiently (and exposing an API to control it) would be a significant undertaking. 

In case you're interested in more details, I'm generating Plots in Node, and re-rendering them to leave enough margin for the formatted ticks. For example, here I'm using [opentype](https://github.com/opentypejs/opentype.js) to measure the (advance) width of the text,  after which I'll regenerate the Plot:

```js
const somePlot = Plot.plot({...})
const yTicks = getTicks("y", somePlot.scale("y"));
// Get the margin left to determine the full width for the x labels
let maxYWidth = 0;
yTicks.forEach((tick) => {
  const computedWidth = font.getAdvanceWidth(tick, 10);
  maxYWidth = Math.max(maxYWidth, computedWidth + 10);
});
// Call Plot.plot() again with a new marginLeft that accounts for the maxYWidth.....
```

In the `getTicks` function,  the `inferTimeFormat` method really comes in handy (the other formatted values are pretty easy to replicate):

```js
function getTicks(direction, { type, domain, range }) {
  if (type === "band") return domain;
  else if ((type === "linear") | (type === "utc")) {
    const scaleType = type === "linear" ? "scaleLinear" : "scaleUtc";
    const scale = d3[scaleType]().domain(domain).range(range);
    const axis = direction === "x" ? d3.axisBottom(scale) : d3.axisLeft(scale);
    // https://github.com/observablehq/plot/blob/203bd4c9f7f41270ad2bc31d24efffb3787bd0e5/src/marks/axis.js#L547
    const tickSpacing = direction === "x" ? 80 : 35;
    const tickCount = inferTickCount(scale, tickSpacing);
    const allTicks = axis.scale().ticks(tickCount);
    const formatter =
      type === "utc"
        ? Plot.inferTimeFormat(type, allTicks, "bottom")
        :  scale.tickFormat();
    return axis
      .scale()
      .ticks(tickCount)
      .map((d, i) => formatter(d, i, allTicks));
  } else {
    return "error";
  }
}
```

I can obviously recreate that method, but it comes in handy here. I could also imagine cases where someone wants to use the formatted value of an axis tick in a piece of text that accompanies the chart. 

So, in short, exposing the formatted tick values would be great, but exporting `inferTimeFormat` helps users recreate the formatting. I understand if this doesn't fit into a patter of which methods you expose for external use.

mbostock:
It seems like exposing `inferTimeFormat` would only address part of the tick logic duplication problem. If you’re already rendering the plot twice, why not just pull the ticks out of the generated SVG?

mkfreeman:
Haha, that's where I started (which worked in the browser), but I wasn't able to select the elements in Node using `JSDOM`. I guess I could parse the SVG string but that seems pretty error prone. Is there a known/simpler way to extract those values? Previously I was using:

```js
node.querySelectorAll('[aria-label="x-axis tick label"] text')
```

mbostock:
Closing since I’d rather not expose this internal method. I would be open to exposing a public method for generating Plot’s default time formats, but we should open an issue first to discuss the potential design and use cases.

mkfreeman:
Yea, that makes sense - thanks for thinking it through. Agreed it would be good to think about which methods to expose for users (without overcommitting to their maintenance, of course).

PR #2115, opened 2024-07-28
support Apache Arrow as a normalized data representation

mbostock:
This allows an Apache Arrow Table to be used directly as the normalized representation of mark data (computed in `mark.initialize`), rather than requiring conversion to an array. This requires patching a few places to support both types, and then finally in `valueof` we materialize the columns as arrays using `table.getChild` and `vector.toArray`, which should be as efficient as possible.

It’s a bit tricky without type checking to make sure we haven’t missed things, but it doesn’t seem too bad, and we should be able to find the places with tests.

This also uncovered a number of preexisting problems with BigInt coercion (since Arrow represents dates as BigInt).

Fixes https://github.com/observablehq/plot/issues/191.
Fixes https://github.com/observablehq/framework/issues/1376.
Supersedes #2030.
Supersedes #2096.


src/mark.d.ts
```diff
@@ -32,6 +32,7 @@ export type TipPointer = "x" | "y" | "xy";
  *
  * - an array, typed array, or other iterable
  * - an object with a length property and indexed values
+ * - an Apache Arrow Table
  */
 export type Data = Iterable<any> | ArrayLike<any>;
```
mbostock, line 37: TODO: Maybe add `ApacheArrowTable` as another allowed type here, using a duck interface that we define.

src/options.js
```diff
@@ -236,20 +263,15 @@ export function maybeZ({z, fill, stroke} = {}) {
...
+  const n = isArray(data) ? data.length : data.numRows;
   const r = new Uint32Array(n);
   for (let i = 0; i < n; ++i) r[i] = i;
   return r;
 }
 
-// Returns a filtered range of data given the test function.
-export function where(data, test) {
-  return range(data).filter((i) => test(data[i], i, data));
-}
```
mbostock, line 248: This appears unused.

src/options.js
```diff
@@ -70,7 +92,7 @@ export function percentile(reduce) {
 
 // If the values are specified as a typed array, no coercion is required.
 export function coerceNumbers(values) {
-  return values instanceof TypedArray ? values : map(values, coerceNumber, Float64Array);
+  return isNumberArray(values) ? values : map(values, coerceNumber, Float64Array);
```
mbostock, line 95: (This fixes `coerceNumbers` for BigInt arrays.)

src/options.js
```diff
@@ -29,21 +51,25 @@ function maybeTake(values, index) {
 }
 
 function maybeTypedMap(data, f, type) {
-  return map(data, type?.prototype instanceof TypedArray ? floater(f) : f, type);
+  return map(data, isNumberType(type) ? (d, i) => coerceNumber(f(d, i)) : f, type); // allow conversion from BigInt
```
mbostock, line 54: This still isn’t quite right.

The common case we were addressing with `floater` is when `f` returns a `BigInt`, and `type` is a common typed array type such as `Float64Array`. But this leaves a few other cases we aren’t handling correctly: `type` could be `BigInt64Array` or `BigUint64Array`, in which case we need to coerce using `BigInt` rather than `Number`; `type` could be an integer type such as `Int32Array`, in which case the `NaN` handling in `coerceNumber` is superfluous and unnecessary.

I’ll work on this now…

mbostock: I guess we don’t really need to support the case where `type` is `BigInt64Array` or `BigUint64Array`; or in other words we could require you to coerce to `BigInt` explicitly yourself. And the second case of `coerceNumber` doing superfluous coercion when `type` is `Int32Array` isn’t really a problem; it’s just slightly slower than it needs to be.

src/options.js
mbostock: Also proposed for removal in #2051. 😅

Fil:
I think this works! I will close #2030 because the description and discussion here is much better.

Fil:
If we convert on the fly all the eligible `data` to Arrow tables, and run the tests, we see a few things:
- most tests pass
- a few tests crash when they try to call Date methods directly on the values (which are not dates anymore but numbers)
- a few tests result in different charts — mostly because invalid values are converted to 0 by tableFromJSON
- textOverflowClip &c crash because they use a custom transform which calls data.flatMap.

Nothing to see there.

More interestingly, we see crashes with:
- differenceY &c, and findArrow: because [Plot.find in group.js](https://github.com/observablehq/plot/blob/c911b1a7a60e3e88b9e8a3b16af9ae54b24c7044/src/transforms/group.js#L447) reads data[i] directly
- musicRevenueCustomOrder crashes because its custom order function gets passed undefined values (error coming from [orderComparator in stack.js](https://github.com/observablehq/plot/blob/c911b1a7a60e3e88b9e8a3b16af9ae54b24c7044/src/transforms/stack.js#L330) reading data[i] directly)

… and that is all!

So I guess these are two places that need an additional arrayify.

Fil:
We'll have to review all the remaining occurrences of `data[`:
- pointer.js
- sortData
- exclusiveFacets
- tree.js

domoritz:
> a few tests crash when they try to call Date methods directly on the values (which are not dates anymore but numbers)

It would probably be a good to have some kind of compatibility mode for arrow that returns dates instead of numbers. Jeff suggested this before.

src/transforms/group.js
```diff
@@ -444,6 +445,7 @@ export function find(test) {
   if (typeof test !== "function") throw new Error(`invalid test function: ${test}`);
   return {
     reduceIndex(I, V, {data}) {
+      data = arrayify(data);
```
mbostock, line 448: This will repeatedly convert the Arrow Table to an array of objects, so we should do something different that’s more efficient. Either we need to move this check outside the loop (somehow, if possible), or we can use data.at in the Arrow case to avoid converting the entire table each time.

src/transforms/stack.js
```diff
@@ -327,7 +327,10 @@ function orderAccessor(f) {
 }
 
 function orderComparator(f) {
-  return (data) => (i, j) => f(data[i], data[j]);
+  return (data) => {
+    data = arrayify(data);
```
mbostock, line 331: This isn’t as bad but we could use table.at here too, I expect.

mbostock:
Great job testing and finding these remaining incompatibilities, @Fil! Thank you.

mbostock:
I reviewed the other places you identified and optimized the handling of Arrow tables. PTAL!

Fil:
I've added a (convoluted) test for the code in tree.js, and a fix. I've also added tests for some of the other places we identified. Note that the comparator sort was not tested at all (with arrays nor arrow) until now!

mbostock:
Looks good to me! Please approve if you agree.

Fil (approved):

PR #2121, opened 2024-08-05
waffle multiple option

mbostock:
Now that we support non-square waffle cells, it feels safe and convenient to support a **multiple** option to affect the number of cells per row (or column, depending on the orientation). I’m not sure about the name of this option, but it seemed desirable to have an orientation-independent name so that it’s easy to switch between waffleX and waffleY.

### DIFF
--- a/docs/marks/waffle.md
+++ b/docs/marks/waffle.md
@@ -124,12 +124,21 @@ Plot.waffleX([apples], {y: ["apples"]}).plot({height: 240})
 The number of rows in the waffle above is guaranteed to be an integer, but it might not be a multiple or factor of the *x*-axis tick interval. For example, the waffle might have 15 rows while the *x*-axis shows ticks every 100 units.
 :::
 :::tip
-While you can’t control the number of rows (or columns) directly, you can affect it via the **padding** option on the corresponding band scale. Padding defaults to 0.1; a higher value may produce more rows, while a lower (or zero) value may produce fewer rows.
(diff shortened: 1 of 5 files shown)


docs/marks/waffle.md
```diff
@@ -124,13 +124,22 @@ Plot.waffleX([apples], {y: ["apples"]}).plot({height: 240})
...
 :::
 
 ## Waffle options
 
 For required channels, see the [bar mark](./bar.md). The waffle mark supports the [standard mark options](../features/marks.md), including [insets](../features/marks.md#insets) and [rounded corners](../features/marks.md#rounded-corners). The **stroke** defaults to *none*. The **fill** defaults to *currentColor* if the stroke is *none*, and to *none* otherwise.
 
+The waffle mark supports a few additional options to control the rendering of cells:
+
+* **unit** - the quantity each cell represents; defaults to 1
+* **multiple** - the number of cells per row (or column); defaults to undefine
```
Fil, line 137:
```suggestion
* **multiple** - the number of cells per row (or column); defaults to undefined
```

Fil, line 137:
```suggestion
* **multiple** - the number of cells per row (or column); defaults to undefined
```

Fil (commented):
multiple is fine; we could also use pods, echelons, places, tracks…

Fil (approved):

Issue #2006, opened 2024-02-28
Plot.boxX default fill should be responsive with dark mode


Fil:
The default fill for boxX is `#ccc`, which is way too light in dark mode (when `--plot-background` is dark and `currentColor` is light).

Should we address this by "hardcoding" a new color in box.js, say `color-mix(in oklab, var(--plot-background), currentColor 80%)`, or by introducing an intermediate `--color80`? 
The only other "hardcoded" colors are
* `stroke = "#555"` in treeLink (which would be `--color33` in my scheme)
* `positiveFill = "#3ca951", negativeFill = "#4269d0",` in differenceY (but that one is ok I think)

PR #2133, opened 2024-08-12
area line option

### LINKED ISSUE #1866: Add a line option to the area mark

### LINKED ISSUE #2111: The ridgeline plot example doesn’t handle overlapping correctly
Since 0.6.7, facets are no longer drawn as separate layers, so the lines are drawn atop all the areas.

0.6.7:
chart (11)

0.6.15:
chart (12)

We could use a custom mark to fix this that draws the line within the area mark, so that overlapping facets behave as desired. But that might be a fair amount of work?

Fil:
Adding a **line** option to the area mark makes it quite a bit more convenient to use than the areaY + lineY combo.

closes #1866 (line option on the area mark)
closes #2111  (The ridgeline plot example doesn’t handle overlapping correctly)

will need to update https://observablehq.com/@observablehq/plot-ridgeline

needs (and rebased on)  #2390 (stack invalid values as NaN)

### DIFF
--- a/test/plots/availability.ts
+++ b/test/plots/availability.ts
@@ -7,15 +7,7 @@ test(async function availability() {
   return Plot.plot({
     height: 180,
     marks: [
-      Plot.areaY(data, {
-        x: "date",
-        y: "value",
-        interval: "day",
-        curve: "step",
-        stroke: "red",
-        fill: "blue",
-        line: true
-      }),
+      Plot.areaY(data, {x: "date", y: "value", interval: "day", curve: "step", fill: "#f2f2fe", line: true}),
       Plot.ruleY([0])
     ]
   });


docs/marks/area.md
```diff
@@ -292,6 +296,8 @@ Points along the baseline and topline are connected in input order. Likewise, if
 
 The area mark supports [curve options](../features/curves.md) to control interpolation between points. If any of the **x1**, **y1**, **x2**, or **y2** values are invalid (undefined, null, or NaN), the baseline and topline will be interrupted, resulting in a break that divides the area shape into multiple segments. (See [d3-shape’s *area*.defined](https://d3js.org/d3-shape/area#area_defined) for more.) If an area segment consists of only a single point, it may appear invisible unless rendered with rounded or square line caps. In addition, some curves such as *cardinal-open* only render a visible segment if it contains multiple points.
 
+The **line** option <VersionBadge pr="TK" /> (boolean, defaults to false) indicates whether the mark should draw a line connecting the points with coordinates **x2** and **y2** (the “top of the area”). In that case, the **stroke** attribute defaults to *currentColor* and is applied to the line only, as well as the stroke opacity. The line uses the same **curve** as the area.
```
Fil, line 299:
```suggestion
The **line** option <VersionBadge pr="2133" /> (boolean, defaults to false) indicates whether the mark should draw a line connecting the points with coordinates **x2** and **y2** (the “top of the area”). In that case, the **stroke** attribute defaults to *currentColor* and is applied to the line only, as well as the stroke opacity. The line uses the same **curve** as the area.
```

test/plots/ridgeline.ts
```diff
@@ -0,0 +1,27 @@
...
+    y: {axis: null, range: [2.5 * 17 - 2, (2.5 - overlap) * 17 - 2]},
+    fy: {label: null, domain: traffic.map((d) => d.location)}, // preserve input order
+    marks: [
+      Plot.areaY(traffic, {
+        x: "date",
+        y: "vehicles",
+        fy: "location",
+        curve: "basis",
+        sort: "date",
+        fill: "color-mix(in oklab, var(--plot-background), currentColor 20%)",
```
Fil, line 22: see https://github.com/observablehq/plot/issues/2006

mbostock:
Superseded by #2407.

PR #2139, opened 2024-08-19
Documentation font changes

CobusT:
As per design's guidelines.
- Spline Sans Mono for the h1 headers
- Inter for the body text
- Font weight 500 for both

Also made a small fix to the footer (removed an old link)

### DIFF
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -24,6 +24,7 @@
   --vp-c-brand-soft: var(--vp-c-purple-soft);
   --vp-font-family-base: Inter, -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue",
     ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
+  --vp-font-family-brand-heading: "Spline Sans Mono", monospace;
 }
 
 .vp-doc {
@@ -31,9 +32,8 @@
 }
 
 .vp-doc h1 {
-  font-family: "Spline-Sans Mono", monospace;
+  font-family: var(--vp-font-family-brand-heading);
   font-weight: 500;
-  margin-bottom: 8px;
 }
 
 .vp-doc a {
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -22,13 +22,14 @@
   --vp-c-brand-2: var(--vp-c-purple-2);
   --vp-c-brand-3: var(--vp-c-purple-3);
   --vp-c-brand-soft: var(--vp-c-purple-soft);
-  --vp-font-family-base: -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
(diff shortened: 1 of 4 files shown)


Fil (changes requested):
I've changed a few things. I have just one pending question about the additional margin-bottom

docs/.vitepress/theme/custom.css
```diff
@@ -21,7 +22,22 @@
   --vp-c-brand-2: var(--vp-c-purple-2);
   --vp-c-brand-3: var(--vp-c-purple-3);
   --vp-c-brand-soft: var(--vp-c-purple-soft);
-  --vp-font-family-base: -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
+  --vp-font-family-base: Inter, -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue",
```
CobusT, line 25: I didn't intend the change the base font. Only the content font. This changes the left nav and the ToC also, which was not the intent of this PR.

docs/.vitepress/theme/custom.css
```diff
@@ -24,6 +27,20 @@
   --vp-font-family-base: -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
 }
 
+.vp-doc {
+  font-family: "Inter", sans-serif;
+  font-weight: 500;
+}
+.vp-doc h1 {
+  font-family: "Spline-Sans Mono", monospace;
+  font-weight: 500;
```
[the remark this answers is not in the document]
CobusT: There was actually a bug in the font name and it was showing monospace instead of Spline Sans Mono. Fixed.

docs/.vitepress/theme/custom.css
Fil: OK I'll revert that. At this size the fonts are almost identical, I can only differentiate them with the inspector :)

docs/.vitepress/theme/custom.css
```diff
@@ -24,6 +27,20 @@
...
+  margin-bottom: 8px;
```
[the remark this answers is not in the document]
CobusT: I am removing it.

Fil (approved):

docs/.vitepress/theme/custom.css
mbostock: I don’t think it looks good to have a mix of Inter for the body and San Francisco for the sidebar. We should use the same sans-serif font everywhere.

CobusT: I will run it by design.

mbostock: Thanks @CobusT. I asked Kyna and Alex earlier this evening, and also about consolidating Framework’s accent colors (since we’re also using a blue and a red in addition to phosphate there). I put up https://github.com/observablehq/framework/pull/1591 for Framework and if there’s agreement there I expect we can propagate some of those changes to Plot. Like maybe the Plot should use phosphate for its accent color instead of purple, too…

PR #2141, opened 2024-08-21
design tweaks

Fil:
Following https://github.com/observablehq/framework/pull/1591 we:
* use the same (sans serif) font on the body
* restore normal weight copy
* adopt phosphate as the brand color
* change the colors in the hero / home page to match

Notes:
* the "soft" color is used for tip backgrounds, it must be very close to the white (or black) background but with the same chroma as phosphate.
* the chroma of the pink colors chosen as alternative is from the Plot logo.
* "Inter" is the default from VitePress

Capture d’écran 2024-08-21 à 11 24 51
Capture d’écran 2024-08-21 à 11 25 59
Capture d’écran 2024-08-21 à 11 26 31
Capture d’écran 2024-08-21 à 11 26 43

### DIFF
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -14,6 +14,7 @@
   --vp-c-brand-soft: var(--theme-phosphate-soft);
   --mono-heading: "Spline Sans Mono", monospace;
   --vp-font-family-mono: var(--mono-heading);
+  --vp-font-family-base: Inter, -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
   --vp-code-font-size: 14px;
   --vp-code-line-height: 1.5;
 }
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -10,6 +10,9 @@
   --vp-c-brand-3: rgb(243, 139, 233);  /* home page alt color */
   --vp-c-brand-soft: var(--theme-phosphate-soft);
   --mono-heading: "Spline Sans Mono", monospace;
+  --vp-font-family-mono: var(--mono-heading);
+  --vp-code-font-size: 14px;
+  --vp-code-line-height: 1.5;
 }
 
 .dark {
@@ -23,6 +26,10 @@
 .vp-doc h1 {
   font-family: var(--mono-heading);
   font-weight: 500;
+}
+
+.vp-doc p {
+  line-height: 1.5;
 }
 
 .vp-doc a {
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -1,16 +1,14 @@
 :root {
-  --vp-c-purple-1: #7135be;
-  --vp-c-purple-2: #7f42cd;
-  --vp-c-purple-3: #9555e2;
-  --vp-c-purple-soft: rgba(155, 91, 233, 0.14);
   --vp-c-red: #f43f5e;
   --vp-c-green: #10b981;
   --vp-c-blue: #0092ff;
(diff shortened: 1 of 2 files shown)


CobusT:
Created a [PR](https://github.com/observablehq/plot/pull/2142) to this branch with some further tweaks.

Fil:
I've merged @CobusT 's addition, the screenshots now look crisper and the line-height is better: 
cobus

docs/.vitepress/theme/custom.css
```diff
@@ -25,6 +28,10 @@
   font-weight: 500;
 }
 
+.vp-doc p {
+  line-height: 1.5;
```
mbostock, line 32: This should probably be on .vp-doc rather than just p? Otherwise things like lists will have a different line-height which will look weird.

Fil: For me this is correct.

The default line-height on the body is set in `node_modules/vitepress/dist/client/theme-default/styles/base.css` as `24px`, but it's overloaded in
`node_modules/vitepress/dist/client/theme-default/styles/components/vp-doc.css` with the selector `.vp-doc p` (also `.vp-doc h1`, etc.). Resetting this on `.vp-doc` would not have enough specificity, and `.vp-doc p` is the only one we want to override.

mbostock: That seems like a bizarre out-of-the-box design by VitePress, but indeed you are right, the line-height is 24px for ul elements and 28px for p elements. It seems that VitePress avoids this problem on their own site by using p elements _within_ the li elements, restoring the 28px line-height. So, perhaps the intent is that you only use the 24px line-height for lists when you know that the list items all fit on a single line. 🤔 Anyway it means this is “correct” according to the weirdness of the VitePress default theme I guess.

docs/.vitepress/theme/custom.css
```diff
@@ -1,40 +1,36 @@
...
-  --vp-c-brand-3: var(--vp-c-purple-3);
-  --vp-c-brand-soft: var(--vp-c-purple-soft);
-  --vp-font-family-base: -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu,
-    roboto, noto, "segoe ui", arial, sans-serif;
-  --sans-text: Inter, var(--vp-font-family-base);
+  --theme-phosphate: #148576;
+  --theme-phosphate-soft: #d7fbf7;
+  --vp-c-brand-1: var(--theme-phosphate); /* link and brand color */
+  --vp-c-brand-2: var(--theme-phosphate);
+  --vp-c-brand-3: rgb(243, 139, 233); /* home page alt color */
```
mbostock, line 10: This doesn’t seem right: these three colors should have the same hue, but varying lightness.

Look at how the current button varies in color as you hover and click:

https://github.com/user-attachments/assets/f36bd40b-cf99-4f86-82d2-6dd07c880baf

Compare that to this branch, where the color changes from pink to green on hover, and there’s no variation in color on click:

https://github.com/user-attachments/assets/8eaf7830-7938-4e87-8646-fd5ae9280a29

mbostock (changes requested):

docs/.vitepress/theme/custom.css
Fil: Agree it's more "correct" than it is correct :-)

docs/.vitepress/theme/custom.css
Fil: Oops! that's right. Since I couldn't find documentation, I went back to the defaults. As you're saying brand 1, 2, 3 are defined as colors of increasing lightness in light mode, and increasing darkness in dark mode.

brand-light

brand-dark

I've done the same by deriving from phosphate in both modes, and re-adding the contrasting ("pink") color under another name.

mbostock (approved):
Big improvement! Thank you. 👏

Issue #33, opened 2020-11-05
Band chart (area with y1 and y2).


mbostock:
Related area #21, but this is interesting because it’s the first case where multiple abstract dimensions share the same y-scale, so we now need to support a many-to-one mapping from abstract dimensions to scales.

PR #2143, opened 2024-08-21
raster colorSpace

### LINKED ISSUE #2138: Support for modern color specifications in the raster mark
Enhancement request: _Allow the raster mark to use CSS4 colour specifications and the P3 colour space._

Marks that generate SVG currently work with [modern colour specifications](https://www.w3.org/TR/css-color-4/), meaning they can specify colours using new colour spaces such as [Oklch](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch). For example, this works correctly:

```javascript
Plot.cellX([1], { fill: "oklch(52.6% 0.24 29.5deg)" }).plot()
```

Not only does this give access to improved perceptual colour spaces, but also allows deeply saturated colours outside  the sRGB gamut but in the P3 gamut to be rendered correctly on displays that support P3 (with a graceful fallback for those which don't).

However, the raster mark, which relies on Canvas, does not currently support either modern colour specification, nor the P3 colour space. The following renders a raster as black rather than the intended saturated red:

```javascript
Plot.raster([1], {
  width: 1,
  height: 1,
  fill: "oklch(52.6% 0.24 29.5deg)"
}).plot()
```

It would be great to have support for modern colour specs in Canvas.

As far as I can see, the [current implementation](https://github.com/observablehq/plot/blob/1d01e254282a9cc3b8352f0b362ae3e2d35ad6f0/src/marks/raster.js#L129-L132) creates an sRGB canvas rather than a p3-friendly one. I think that would just require

```javascript
getContext("2d", { colorSpace: "display-p3" });
```

But handling CSS 4 colour specs looks a little more tricky as the implementation seems to rely on [d3.color](https://github.com/d3/d3-color) which currently does not handle the modern comma-free colour format. I note [d3 issue #87](https://github.com/d3/d3-color/issues/87) though.

mbostock:
Fixes #2138. Probably needs a test… I didn’t actually verify that it works.

### DIFF
--- a/docs/marks/raster.md
+++ b/docs/marks/raster.md
@@ -286,11 +286,12 @@ If **width** is specified, **x1** defaults to 0 and **x2** defaults to **width**
 
 The following raster-specific constant options are supported:
 
+* **colorSpace** - the [color space](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/colorSpace) <VersionBadge pr="2143" />
 * **interpolate** - the [spatial interpolation method](#spatial-interpolators)
 * **imageRendering** - the [image-rendering attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/image-rendering); defaults to *auto* (bilinear)
 * **blur** - a non-negative pixel radius for smoothing; defaults to 0
 
-The **imageRendering** option may be set to *pixelated* for a sharper image. The **interpolate** option is ignored when **fill** or **fillOpacity** is a function of *x* and *y*.
+The **colorSpace** may be set to *display-p3* for the Display P3 color space. The **imageRendering** option may be set to *pixelated* for a sharper image. The **interpolate** option is ignored when **fill** or **fillOpacity** is a function of *x* and *y*.
 
 ## raster(*data*, *options*) {#raster}
 
--- a/src/marks/raster.d.ts
+++ b/src/marks/raster.d.ts
@@ -146,6 +146,14 @@ export interface RasterOptions extends Omit<MarkOptions, "fill" | "fillOpacity">
   imageRendering?: string;
 
   /**
+   * The [color space][1] of the backing canvas. Defaults to *srgb*; set to
+   * *display-p3* for the Display P3 color space.
+   *
+   * [1]: https://developer.mozilla.org/en-US/docs/Web/API/ImageData/colorSpace
+   */
+  colorSpace?: "srgb" | "display-p3" | string;
+
+  /**
    * The fill, typically bound to the *color* scale. Can be specified as a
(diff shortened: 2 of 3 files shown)


Fil (changes requested):
Tested with success. This shows that it can work — technically speaking the serialization of the canvas does not suppress the color-space attribute. Using `colorSpace: "display-p3"` with no further change allows to make the image more vibrant. However, on its own this is incorrect for data visualization.

For example, if a color scale maps a value x to the `"red"` color, it is displayed as `"red"` or `"#ff0000"` in a dot mark or in a legend. In a traditional (`sRGB`) raster mark, it is rendered as the `(#ff, #00, #00)` byte triplet. However, when the raster is using the `display-p3` color space, the same color **must** be represented by its `xyz` coordinates `(#ea, #33, #23)`, not by its `srgb` values. If we don't do apply any conversion, it results in a different color (which we could name "vivid red").

In other words, when using `display-p3`, we have to "tone down" the byte values of colors that belong to the sRGB gamut, with an sRGB→P3 mapping [1]. This is necessary to ensure that colors are "true" and consistent, at least on p3 capable displays.

(As an aside, the developer might need to set `colorSpace` only if they detect that the display is capable of representing those colors — on my setup it depends  if I'm using the laptop screen or the external monitor, which further complicates the matter. But I think this is out of scope for this PR, and belongs to documentation.)

In a second step, we'll want to enable more colors — such as our "vivid red" — to be encoded too. But we need a way to specify them statically (like `color(display-p3 1 0 0)`) and in color interpolators. I think this means we need to create d3.xyz, d3.oklch as the equivalent of d3.rgb and d3.hcl, add CSS4 parsing to d3.color, and maybe also create a `"xyz"` color interpolator (d3.interpolateXyz, equivalent of d3.interpolateHsl).

In a later stage, we'll want to provide "p3-enabled" or "vivid" color schemes (continuous, ordinal, etc.). A cheap way of doing this might be to add an option such as `color: {theme: "oranges", vivid: true}` to "boost" the scheme by mapping r,g,b directly to p3. (We might also want to create new schemes, but that's more work.)


[1] references https://www.russellcottrell.com/photo/matrixCalculator.htm and https://observablehq.com/@jrus/srgb

PS: it's not clear to me if/how we should apply the gamma correction in this context.

jwoLondon:
Helpful comments @Fil 

On a possible `color: {theme: "oranges", vivid: true}`, and more generally with non-sRGB colour spaces, it is probably worth keeping the door open for `BT.2020` ('Rec2020") and other future enhanced colour spaces. There would be some ambiguity for `vivid:true`: Would that mean highest spec colour space or P3 or some other rule?

Fil (approved):

PR #2147, opened 2024-08-23
horizontal↔︎ and vertical︎↕︎

### LINKED ISSUE #2146: vertical︎↕︎ is displayed with the emoji font-variant
In #2141 I see an issue with Inter, the new font, which displays horizontal↔︎ as text and vertical︎↕︎ as emoji. (I don't know why it doesn't happen in main.)

Capture d’écran 2024-08-23 à 11 23 49

This should be solved in the far future when browsers support [font-variant-emoji](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-emoji), but in the meantime we need to fix it somehow. I tried the variation selector `U+FE0E`, but it's filtered out for some reason, whether I write as `&#xFE0E;` or `&#65038;`. I think I'll just do a `span::after`.

Fil:
closes #2146

(and it's also probably better for screen readers?)

| before | after |
| - | - |
| before | after |

### DIFF
--- a/docs/.vitepress/theme/custom.css
+++ b/docs/.vitepress/theme/custom.css
@@ -18,6 +18,16 @@
   --vp-code-line-height: 1.5;
 }
 
+.emoji-horizontal::after {
+  content: "↔︎";
+  font-family: var(--vp-font-family-mono);
+}
+
+.emoji-vertical::after {
+  content: "↕︎";
+  font-family: var(--vp-font-family-mono);
+}
+
 .dark {
   --theme-phosphate: #37d5be;
   --theme-phosphate-2: #28b39e;
--- a/docs/features/facets.md
+++ b/docs/features/facets.md
@@ -23,9 +23,9 @@ onMounted(() => {
 
 # Facets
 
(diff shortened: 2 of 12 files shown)


Fil:
Alternative way of fixing this issue is to set `--vp-font-family-base`:

```
:root: {
  --vp-font-family-base: Inter, -apple-system, BlinkMacSystemFont, "avenir next", avenir, helvetica, "helvetica neue", ubuntu, roboto, noto, "segoe ui", arial, sans-serif;
}
```

instead of the defaults from VitePress:
```
'Inter', ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'
```

or maybe we do both?

mbostock:
I’m inclined to change the definition of `--vp-font-family-base` to minimize churn and for better consistency with the Framework docs. Also it’s more internally consistent as we also have many instances of right→, ←left, up↑, and down↓. Not sure why these don’t also render as right➡️, left⬅️, up⬆️, down⬇️… thinking🤔 shrug🤷.

Fil:
ok 🤷🤷

PR #2155, opened 2024-08-28
add BigInt support to the normalize transform

### LINKED ISSUE #2154: Error when normalizing a BigInt variable
When data contains a `BigInt` variable, trying to apply `normalize` on this variable throws an error: `TypeError: Cannot convert a BigInt value to a number`.

I've created a small notebook to reproduce the issue:

https://observablehq.com/@juba/plot-normalize-bigint-error

It seems that the error may come from this line, but I'm far from an expert on this domain so I may be mistaken:

https://github.com/observablehq/plot/blob/0032c11160e6340ce44307675d93bed382d10446/src/transforms/normalize.js#L46C1-L47C1

Fil:
Note: it's a bit of a bummer in that we're adding overhead for all Arrays (by making a copy of the channel to a Float64Array). Maybe we could just do it if the first non-nullish value is BigInt?

cc: @juba

closes #2154

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -54,7 +54,7 @@ function maybeTypedMap(data, f, type) {
   return map(data, isNumberType(type) ? (d, i) => coerceNumber(f(d, i)) : f, type); // allow conversion from BigInt
 }
 
-function maybeTypedArrayify(data, type) {
+export function maybeTypedArrayify(data, type) {
   return type === undefined
     ? arrayify(data) // preserve undefined type
     : isArrowVector(data)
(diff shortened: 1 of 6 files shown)


mbostock (commented):
Should we fix this earlier, perhaps by having `maybeTypedArrowify` always coerce array of BigInt back to Number? Or is it Plot’s responsibility to undo BigInt’s strict coercion wherever BigInt may occur?

Fil:
It should definitely be earlier — in my code we coerce the whole channel as many times as there are series. But where? Maybe as high as in the map transform, where we do
https://github.com/observablehq/plot/blob/0032c11160e6340ce44307675d93bed382d10446/src/transforms/map.js#L36

In that case the mapper should possibly have a `type` key, like so:

```
  {
    mapIndex(I, S, T) {
      const b = basis(I, S);
      for (const i of I) {
        T[i] = S[i] === null ? NaN : S[i] / b;
      }
    },
    type: Float64Array
```

that would be passed to valueof?

Fil:
This is very close to https://github.com/observablehq/plot/issues/473

PR #2164, opened 2024-09-09
coerce Arrow BigInt to Number

mbostock:
Ref. https://github.com/observablehq/plot/pull/2155#pullrequestreview-2266406998

Note that in the case of DuckDB-Wasm, we default `config.query.castBigIntToDouble` to true, so we shouldn’t see BigInt in practice with DuckDB, but we might still see it when loading Arrow or Parquet files.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -69,6 +69,8 @@ function maybeTypedArrowify(vector, type) {
     ? vector
     : (type === undefined || type === Array) && isArrowDateType(vector.type)
     ? coerceDates(vector.toArray())
+    : (type === undefined || type === Array) && isArrowBigIntType(vector.type)
+    ? (type ?? Float64Array).from(vector.toArray(), Number)
     : maybeTypedArrayify(vector.toArray(), type);
 }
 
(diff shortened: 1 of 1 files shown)


Fil:
Sounds good. I'll work on a unit test.

Fil:
I've added a unit test, however it reveals some other issues. I'll comment inline.

test/plots/bigint.ts
```diff
@@ -25,3 +26,19 @@ export async function bigintOrdinal() {
 export async function bigintStack() {
   return Plot.barY(integers, {x: (d, i) => i % 5, y: "big1"}).plot();
 }
+
+export async function bigintNormalize() {
+  const table = await Arrow.tableFromIPC(fetch("data/height_frequency.arrow"));
+  return Plot.plot({
+    height: 500,
+    x: {percent: true, grid: true},
+    y: {domain: [1.3, 2.21]},
```
Fil, line 35: If you remove the y.domain here, the chart works in preview but crashes in node tests. I don't understand what the difference might be.

Fil: Capture d’écran 2024-09-09 à 19 39 39

This is how the chart looks like in preview. But it crashes in node. Is apache-arrow behaving differently in the browser?

Fil: OK the test is now fixed — I just don't use a file and generate the arrow structure on the fly.

PR #2169, opened 2024-09-11
use the context’s document to create the input Event

### LINKED ISSUE #2159: Use `context.document.defaultView.Event` instead of `Event`
Plot currently uses the `Event` global to construct events, but it should respect the top-level **document** option which can provide an alternative DOM implementation.

Ref. https://github.com/observablehq/plot/discussions/2158#discussioncomment-10512970

Fil:
closes #2159

cc: @StromFLIX

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -175,7 +175,8 @@ export function plot(options = {}) {
   context.dispatchValue = (value) => {
     if (figure.value === value) return;
     figure.value = value;
-    figure.dispatchEvent(new Event("input", {bubbles: true}));
+    const Event = context.document.defaultView?.Event;
+    if (Event) figure.dispatchEvent(new Event("input", {bubbles: true}));
   };
 
   // Reinitialize; for deriving channels dependent on other channels.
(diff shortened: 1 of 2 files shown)


src/plot.js
```diff
@@ -175,7 +175,8 @@ export function plot(options = {}) {
   context.dispatchValue = (value) => {
     if (figure.value === value) return;
     figure.value = value;
-    figure.dispatchEvent(new Event("input", {bubbles: true}));
+    const Event = context.document.defaultView?.Event;
+    if (Event) figure.dispatchEvent(new Event("input", {bubbles: true}));
```
mbostock, line 179: Why make it optional? Feels like that would be surprising: Plot is not able to function without a complete implementation, and the user wouldn’t notice when it fails silently.

Alternative:

```suggestion
    figure.dispatchEvent(new context.document.defaultView.Event("input", {bubbles: true}));
```

Fil: I was thinking of https://github.com/observablehq/plot/blob/09edfec1f26da02eeeb451f008dd2eb9fb8d8ffc/docs/components/PlotRender.js — but we wouldn't use this with tip: true

mbostock (approved):

src/plot.js
Fil: well… we do
fixed in #2185

PR #2175, opened 2024-09-24
include LegendOptions (e.g., className, fontSize, r…) in the type for scales

Fil:


mbostock (changes requested):
This seems right, but it also seems a little dangerous to mix the legend options into the scale options. Maybe we could do it more like the **tip** mark option, where instead of `legend: true` you could pass `legend: options`?

```js
Plot.plot({
  color: {
    type: "ordinal",
    scheme: "rainbow",
    legend: {
      className: "legend",
      style: {fontSize: "16px"}
    }
  },
  style: {fontSize: "7px"},
  marks: [Plot.cell("ABCDEFGHIJ", {x: Plot.identity, fill: Plot.identity})]
})
```

We would then change `ScaleDefaults` from:

```ts
legend?: LegendOptions["legend"] | boolean | null;
```

to:

```ts
legend?: LegendOptions | LegendOptions["legend"] | boolean | null;
```

It’d probably be safe to just make this change, but we could also be conservative and continue supporting the undocumented way of mixing legend options into scale options if desired.

mbostock (changes requested):
This was also superseded by #2249, which included type declarations for legend options within scale options. Note that we only want legend options for scales that can have legends, which is currently the _color_, _opacity_, and _symbol_ scales; if other scales support legends in the future (such as _r_ #236) then we can add those types as appropriate then.

There does seem to be a bug in the current implementation in that `ColorLegendOptions` only includes the options that are specific to color legends and does not include the generic `LegendOptions`. That’s because `LegendOptions extends ColorLegendOptions` rather than the other way around.

Fil:
superseded by #2421

PR #2180, opened 2024-09-30
filterFacets is a private method of context

Fil:
I believe this is not usable outside of the axis mark, and we won't document it as being part of the API (#1811). So let's reinforce that with the customary `_` prefix.


mbostock (commented):
I put it in the same category as getMarkState. If we need it for the axis mark, it might be need for other marks in the future…

Fil:
OK, I'll try harder to document it :)

Fil:
#2182

PR #2182, opened 2024-10-01
lifecycle

Fil:
A more ambitious documentation for #1711, describing the whole lifecycle of a chart.

alternative to #1811


Fil:
Two issues with Vitepress that need to be fixed:
- [ ] there are errors on build related to `Event`
- [ ] the three charts update when I click on any of the "replay" buttons

PR #2191, opened 2024-10-08
Add the Made by Observable button to the top nav

CobusT:
This PR adds the Made by Observable component to the Plot site.


docs/.vitepress/theme/CustomLayout.vue
```diff
@@ -35,4 +42,20 @@ const {Layout} = DefaultTheme;
   background-color: rgba(37, 37, 41, 0.5);
 }
 
+/* Hide the dark mode toggle */
+.VPNavBarAppearance,
+.VPNavBarExtra,
+.VPNavBarHamburger {
+  display: none !important;
+}
```
mbostock, line 50: This shouldn’t be necessary with `appearance: "force-auto"`?

mbostock: Or maybe the comment only applied to `.VPNavBarAppearance`, whereas we still also want to hide `.VPNavBarExtra` and `.VPNavBarHamburger`? Although it makes me wonder what functionality we are losing by hiding those latter two things?

CobusT: Yup, `.VPNavBarAppearance` is not needed anymore (removed it). In terms of what we are missing from the othhers, they are actually empty with our implementation.

mbostock (approved):
Looks great!

PR #2195, opened 2024-10-10
fix handling of null values in arrow vectors

### LINKED ISSUE #2194: Mishandling of Arrow's null values
In #2115 we did not consider that the arrow vector.toArray() can return *anything* when a value is null. (See https://github.com/observablehq/framework/pull/1738 for a concrete consequence).

After calling vector.toArray we have to clear out the null values (that can be skipped if there are no nulls).

Fil:
Call vector.toJSON instead of vector.toArray when there are null values.

closes #2194


src/options.js
```diff
@@ -68,8 +68,17 @@ function maybeTypedArrowify(vector, type) {
...
-    : maybeTypedArrayify(vector.toArray(), type);
+    ? coerceDates(vectorToArray(vector))
+    : maybeTypedArrayify(vectorToArray(vector), type);
+}
+
+function vectorToArray(vector) {
+  const values = vector.toArray();
+  if (values.nullCount) {
+    const NULL = vector.type?.typeId === 5 /* Utf8 */ ? null : undefined;
+    for (let i = 0; i < values.length; ++i) if (!values.isValid(i)) values[i] = NULL;
```
mbostock, line 79: We can’t use assignment (mutation) here because this might not be a copy. According to [the docs](https://arrow.apache.org/docs/js/classes/Arrow_dom.Vector.html#toArray):

> If this Vector contains a single Data chunk and the Vector's type is a primitive numeric type corresponding to one of the JavaScript TypedArrays, this method returns a zero-copy slice of the underlying TypedArray values. If there's more than one chunk, the resulting TypedArray will be a copy of the data from each chunk's underlying TypedArray values.

mbostock (changes requested):
To me this suggests that we can’t/shouldn’t use `vector.toArray` at all, and instead we should use the “slow” version `vector.toJSON` to let Arrow handle the conversion into usable JavaScript values. Maybe we could optimize it to allow `vector.toArray` if `vector.nullCount` is zero, though. But the type-specific handling of invalid values here seems likely to be wrong and not something that we want to replicate outside of Arrow.

Fil:
ok that's simpler 😓

mbostock (approved):

PR #2203, opened 2024-10-19
waffle href

### LINKED ISSUE #2202: The Waffle mark does not support href
The `A` element is (wrongly) created on the `pattern`’s `rect` instead of on the `path`.

Fil:
closes #2202


src/marks/waffle.js
```diff
@@ -104,6 +104,7 @@ function waffleRender(y) {
           )
           .attr("fill", (i) => `url(#${patternId}-${i})`)
           .attr("stroke", this.stroke == null ? null : (i) => `url(#${patternId}-${i})`)
+          .call(applyChannelStyles, this, {href})
```
mbostock, line 107: I think ariaLabel and title should travel with href, too. We want the visual channels to apply to the pattern, but the non-visual (semantic?) channels should apply to the path.

mbostock (approved):

Fil:
good call! thanks

src/marks/waffle.js
```diff
@@ -35,6 +35,7 @@ export class WaffleY extends BarY {
 
 function waffleRender(y) {
   return function (index, scales, values, dimensions, context) {
+    const {ariaLabel, href, title, ...visualValues} = values;
```
Fil, line 38: If this needs to be done somewhere else in the future, we should probably instead split `applyChannelStyles` into two functions (`applyChannelStyles` and `applyChannelSemantics`).

PR #2204, opened 2024-10-19
waffle stroke

### LINKED ISSUE #2186: Waffle pattern does not account for strokeWidth
When a Waffle mark has a stroke applied, the pattern size does not account for the added stroke width which causes neighboring strokes to bleed through at the edges.

Example:
```js
Plot.waffleX([6], {gap: 6, stroke: "black", strokeWidth: 5}).plot({height: 200})
```

Fil:
closes  #2186

(to apply after #2203)


test/output/waffleHref.svg
Fil, line 1: no visual difference here

test/output/waffleStroke.svg
Fil, line 1: this test and the other ones below have a small visible difference, but for the best

test/output/waffleStrokeWidth.svg
Fil, line 1: this would be unchanged

test/output/waffleStrokeWidthConst.svg
Fil, line 1: this test is the one exhibiting the bug (see description)

mbostock (commented):
A thing I noticed is that these waffles have invisible slivers that extend the full width of the band. Related to #2132 I guess. Need more time to work on this but we probably want to simplify the polygon we’re using to render the waffle cells.

image

Fil:
Yes, it explains the gradient we see at the top of the "before" image in the description. It would be nice to clean this up too. 

(Related: https://github.com/observablehq/plot/pull/2132)

Fil:
integrated in #2215

PR #2211, opened 2024-10-28
no ramp legend for continuous scales with a degenerate domain

### LINKED ISSUE #2210: `TypeError: o[n] is not a function` for empty color domain
I'm setting up a plot where the data set used for the plot may be empty. In such cases, I end up passing `color: { legend: true, domain }` where `domain` is an empty array but this leads to an error. Here's a demo that produces the error due to an empty data set and empty domain https://observablehq.com/d/f1142c360da5114b.

Locally I see:

```
TypeError: I[i2] is not a function
    piecewise piecewise.js:9
    legendRamp ramp.js:95
    legendColor legends.js:53
    createLegends legends.js:76
    plot plot.js:325
    render_chart main.ts:110
    do_query main.ts:278
    onreadystatechange utils.ts:6
    http_get utils.ts:3
    do_query main.ts:263
    query_keystroke main.ts:66
    timeoutId utils.ts:23
    setTimeout handler*debounce/< utils.ts:22
    App main.ts:33
    <anonymous> main.ts:288
```

Fil:
closes #2210

I considered an alternative where we would generate a ramp “from x to x” if there is 1 element, and “from NaN to NaN” if there is none. This would show the interpolator colors (like below).

However the bug report shows that we might be in a case where we normally expect an ordinal scale, but since the empty domain makes it impossible to automatically determine if we're ordinal or linear, the ramp is just a consequence of having defaulted to linear.

Capture d’écran 2024-10-28 à 07 02 39
Capture d’écran 2024-10-28 à 07 02 26

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -50,6 +50,7 @@ function legendColor(color, {legend = true, ...options}) {
     case "swatches":
       return legendSwatches(color, options);
     case "ramp":
+      if (color.domain.length < 2) return;
       return legendRamp(color, options);
     default:
       throw new Error(`unknown legend type: ${legend}`);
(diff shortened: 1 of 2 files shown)


Fil:
superseded by #2212

PR #2212, opened 2024-10-28
handle degenerate domains

### LINKED ISSUE #2210: `TypeError: o[n] is not a function` for empty color domain

mbostock:
Fixes #2210. Alternative to #2211.

### DIFF
--- a/src/scales.js
+++ b/src/scales.js
@@ -425,10 +425,11 @@ function inferScaleType(key, channels, {type, domain, range, scheme, pivot, proj
   if (kind === opacity || kind === length) return "linear";
   if (kind === symbol) return "ordinal";
 
-  // If the domain or range has more than two values, assume it’s ordinal. You
-  // can still use a “piecewise” (or “polylinear”) scale, but you must set the
-  // type explicitly.
-  if ((domain || range || []).length > 2) return asOrdinalType(kind);
+  // If a domain or range is explicitly specified and doesn’t have two values,
+  // assume it’s ordinal. You can still use a “piecewise” (or “polylinear”)
+  // scale, but you must set the type explicitly.
+  const n = (domain ?? range)?.length;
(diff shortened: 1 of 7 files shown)


mbostock:
Not sure why the test snapshots changed… maybe I’m running a new version of Node? 🤔

Fil:
Yes that's the reason. Same for me aaa8324f790293b84a8b377d6a4092d7ee2c3549 passes all tests with v21.6.0, but not v22.10.0. Probably the maths have changed :)

Means we have to round a bit before we compare numbers… it's going to be fun.

Fil:
can we go back to https://github.com/observablehq/plot/commit/aaa8324f790293b84a8b377d6a4092d7ee2c3549 and leave the tests aside for now? The fix lgtm!

mbostock:
Yeah we’re stuck on Node 20 for development until we either drop node-canvas or they release 3.0.0. #2234 So we’ll need to generate the snapshots on Node 20.

Fil (approved):
thanks!

PR #2215, opened 2024-10-30
waffle tips + waffle stroke + waffle unit: auto

### LINKED ISSUE #2214: waffle unit: auto
When a waffle mark depicts large values, it might show up invisible because the area for each "dot" becomes infinitesimal. This suggests that the default **unit** should be "auto" instead of 1, and adapt to the available area.

Image

Fil:
- [x] unit: auto (closes #2214)
- [x] #2132 (merge commit)
- [x] #2204 (merge commit)


mbostock:
I’d prefer to understand these three contributions separately rather than reviewing it all in this PR. I’m going to continue the reviews of the tip & stroke PRs before looking at this.

PR #2232, opened 2024-11-14
fix tickFormat type inference for empty domain

### LINKED ISSUE #2231: Temporal tickFormat for facet reported as invalid if dataset is empty
If a plot defines a temporal `tickFormat` like `%W` for a facet, and the dataset is empty, then rendering the Plot will fail with the error:
> Error: invalid format: %W

Example:
```js
Plot.barY([], {fy: d => d}).plot({fy: {tickFormat: "%W"}})
```
For comparison:
```js
Plot.barY([new Date()], {fy: d => d}).plot({fy: {tickFormat: "%W"}})
```

mbostock:
Fixes #2231. If an (ordinal) domain is empty, ignore the **tickFormat** since we can’t infer whether the data is temporal or quantitative, and instead use the default format. This shouldn’t affect the axis as we assume the tick format won’t be used anyway because the domain is empty and hence there can’t be any ticks.

### DIFF
--- a/src/marks/axis.js
+++ b/src/marks/axis.js
@@ -672,10 +672,10 @@ export function inferTickFormat(scale, data, ticks, tickFormat, anchor) {
     ? inferTimeFormat(scale.type, data, anchor) ?? formatDefault
     : scale.tickFormat
     ? scale.tickFormat(typeof ticks === "number" ? ticks : null, tickFormat)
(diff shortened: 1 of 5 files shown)


mbostock:
The `domain || range || []` is wrong… need to check what the inferred ordinal domain would be based on the channels.

Fil (approved):

PR #2234, opened 2024-11-15
limit numerical precision of test snapshots; fix node 22

mbostock:
This limits numerical precision to three fractional digits, similar to what we already do via d3-shape and d3-path (ref. #1193 #1143). Reduces the test snapshots from 58M to 51M. It looks like one test (projectionClipAngle) is still broken, I would guess to some other numerical instability that will be harder to fix. So, I may just not worry about having the test snapshots work on both Node 20 and Node 22, and only test against 22.

Related, it would be nice if Plot limited the precision of numeric attributes and numbers interpolated into templates by default, since we already do this for path data.


mbostock:
This is now blocked by https://github.com/Automattic/node-canvas/issues/2448.

I made some pretty good progress switching to [skia-canvas](https://github.com/samizdatco/skia-canvas) instead, and using module-alias to point JSDOM at skia-canvas instead of canvas. But it looks like image rendering (raster mark) still isn’t working, so that hit a dead end.

Then I looked at upgrading to [canvas 3.0.0-rc2](https://github.com/Automattic/node-canvas/releases/tag/v3.0.0-rc2), which supports Node 22, but that also changes the API in a way that breaks JSDOM https://github.com/Automattic/node-canvas/issues/2390. In fact they fixed it in https://github.com/Automattic/node-canvas/issues/2446, but they haven’t released a new release candidate yet, so there’s not currently a way to install a version that’s compatible with both Node 22 and JSDOM. 😓

PR #2239, opened 2024-11-19
A channel of explicit symbols does not show in the tip nor has a legend

### LINKED ISSUE #2237: Using a channel of named symbols doesn't work with tips
Image

```
Plot.dot(
  penguins,
  Plot.dodgeY({
    x: "culmen_length_mm",
    symbol: (d) => (d.sex === "FEMALE" ? "square" : "star"),
    tip: true
  })
).plot({ height: 220 })
```

The [Object object] is the `{draw: function}`  that corresponds to the symbol. Maybe it should just have a `toString`?

Fil:
closes #2237


src/marks/tip.js
```diff
@@ -347,8 +347,8 @@ function getSourceChannels(channels, scales) {
     if ((key === "x" || key === "y") && channels.geometry) continue; // ignore x & y on geo
     const source = getSource(channels, key);
     if (source) {
-      // Ignore color channels if the values are all literal colors.
-      if (source.scale == null && source.defaultScale === "color") continue;
+      // Ignore color/symbol channels if the values are all literal colors/symbols.
+      if (source.scale == null && ["color", "symbol"].includes(source.defaultScale)) continue;
```
mbostock, line 351: Seeing this test made me wonder whether we have an existing encapsulation of whether a scale allows literal values. The closest seems to be `inferChannelScale`, which suggests we might also want to treat the _opacity_ scale this way (when the domain is entirely values in [0, 1]). In fact, we could simplify this test as:

```suggestion
      if (source.scale == null && source.defaultScale) continue;
```

Since `channel.defaultScale` is only used when a channel can be defined literally.

Fil: 👍

PR #2242, opened 2024-11-21
Replace Svelte example with runes (v5.0.0+) and rm deprecated onmousemove

vincvdwal:


Fil:
Thanks! I think we also need to link to a different playground url.

(I pasted your changes and saved them to https://svelte.dev/playground/e65b5c87ae7e44239cef41ec3df28f52?version=5.2.7)

vincvdwal:
Np! I was wondering, maybe it needs 2 versions, since v5.0.0 with runes is kind of new? I don't know if a lot of people still use Svelte v4.x.x or prior. But I was kind of struggling getting it working. I also thought about removing the Typescript stuff since it's actually optional in Svelte

Fil:
We don't want to make this documentation about svelte versions. Being up-to-date is good if you're starting a new project.

PR #2243, opened 2024-11-21
clip: geojson

### LINKED ISSUE #1109: Arbitrary clip paths

Fil:
Note that [one case](http://localhost:8008/?test=usStateClipVoronoi) escapes the memoization: the pointer transform creates a new context, and marks that use it get a second clip path.

closes #1109


src/style.js
```diff
@@ -356,15 +352,27 @@ const getFrameClip = memoizeClip((clipPath, context, dimensions) => {
...
+    if (!geoClips.has(geo)) geoClips.set(geo, clip(geo, scales));
+    return geoClips.get(geo);
+  };
+}
+
+const getGeoClip = memoizeGeo((geo, scales) =>
+  memoizeClip((clipPath, context) => {
+    const {projection} = context;
+    if (!projection && geo.type === "Sphere") throw new Error(`clipping to the sphere requires a projection`);
+    clipPath.append("path").attr("d", geoPath(projection ?? xyProjection(scales))(geo));
```
mbostock, line 367: That’s a lot of plumbing of `scales` so we can call `xyProjection` here. I wonder if instead we should have a (bound) method or getter on the `context` so that we only need to provide the `context` here. Or maybe we _always_ have a `context.projection` and it’s initialized from the _x_ and _y_ scales if needed (but I think there are other places where we test if there’s a projection, so I doubt that will work).

test/output/contourVaporClip.svg
mbostock, line 1: This is lovely!

src/style.js
Fil: maybe be a context.geopath getter?

src/options.js
```diff
@@ -602,12 +602,14 @@ export function maybeNamed(things) {
...
 
-// TODO Accept other types of clips (paths, urls, x, y, other marks…)?
-// https://github.com/observablehq/plot/issues/181
 export function maybeClip(clip) {
   if (clip === true) clip = "frame";
   else if (clip === false) clip = null;
-  else if (clip != null) clip = keyword(clip, "clip", ["frame", "sphere"]);
+  else if (
+    clip != null &&
+    !["FeatureCollection", "Feature", "Polygon", "GeometryCollection", "MultiPolygon"].includes(clip.type)
```
mbostock, line 610: I don’t think we should restrict this to specific GeoJSON types. For example, you could have a FeatureCollection of a single Point feature, so why allow FeatureCollection but not Point? Also, a Sphere geometry should be treated the same as _sphere_.

Fil: You can't clip to a "Point" but I see your point. I've changed it to a duck-typing test of any GeoJSON object.

src/style.js
Fil: context.path it is

src/options.js
```diff
@@ -602,12 +616,12 @@ export function maybeNamed(things) {
...
 }
 
-// TODO Accept other types of clips (paths, urls, x, y, other marks…)?
-// https://github.com/observablehq/plot/issues/181
 export function maybeClip(clip) {
   if (clip === true) clip = "frame";
   else if (clip === false) clip = null;
-  else if (clip != null) clip = keyword(clip, "clip", ["frame", "sphere"]);
+  else if (isGeoJSON(clip)) {
+    if (clip.type === "Sphere") clip = "sphere";
```
Fil, line 623: note that other geojson pass through

docs/features/marks.md
```diff
@@ -493,7 +493,7 @@ All marks support the following style options:
 * **clip** - whether and how to clip the mark
 * **tip** - whether to generate an implicit [pointer](../interactions/pointer.md) [tip](../marks/tip.md) <VersionBadge version="0.6.7" />
 
-If the **clip** option is *frame* (or equivalently true), the mark is clipped to the frame’s dimensions; if the **clip** option is null (or equivalently false), the mark is not clipped. If the **clip** option is *sphere*, then a [geographic projection](./projections.md) is required and the mark will be clipped to the projected sphere (_e.g._, the front hemisphere when using the orthographic projection).
+If the **clip** option is *frame* (or equivalently true), the mark is clipped to the frame’s dimensions; if the **clip** option is null (or equivalently false), the mark is not clipped. If the **clip** option is *sphere* or a GeoJSON object with a polygonal geometry <VersionBadge pr="2243" />, then the mark will be clipped to the projected geometry (_e.g._, the front hemisphere when using the orthographic projection). A [geographic projection](./projections.md) is required for the sphere.
```
mbostock, line 496:
```suggestion
If the **clip** option is *frame* (or equivalently true), the mark is clipped to the frame’s dimensions; if the **clip** option is null (or equivalently false), the mark is not clipped. If the **clip** option is *sphere*, then the mark will be clipped to the projected sphere (_e.g._, the front hemisphere when using the orthographic projection); a [geographic projection](./projections.md) is required in this case. Lastly if the **clip** option is a GeoJSON object <VersionBadge pr="2243" />, then the mark will be clipped to the projected geometry. 
```

src/context.d.ts
```diff
@@ -18,6 +18,9 @@ export interface Context {
   /** The current projection, if any. */
   projection?: GeoStreamWrapper;
 
+  /** A function to draw GeoJSON with the current projection, if any, otherwise with the x and y scales. */
+  path?: () => GeoPath;
```
mbostock, line 22:
```suggestion
  /** A function to draw GeoJSON with the current projection, if any, otherwise with the x and y scales. */
  path: () => GeoPath;
```

src/mark.d.ts
```diff
@@ -296,11 +297,12 @@ export interface MarkOptions {
    *
    * - *frame* or true - clip to the plot’s frame (inner area)
    * - *sphere* - clip to the projected sphere (*e.g.*, front hemisphere)
+   * - geojson - a GeoJSON object with a (multi-) polygon geometry
```
mbostock, line 300:
```suggestion
   * - geojson - a GeoJSON object, typically with polygonal geometry
```

src/options.js
```diff
@@ -169,24 +169,38 @@ export function dataify(data) {
...
+      "GeometryCollection",
+      "Feature",
+      "LineString",
+      "MultiLineString",
+      "MultiPoint",
+      "MultiPolygon",
+      "Point",
+      "Polygon",
+      "Sphere"
+    ].includes(x.type)
```
mbostock, line 200: I believe using array.includes is quite a bit slower than using a switch statement since the array must be materialized each time the test is applied, causing more garbage collection. I’d prefer to use a switch statement.

src/options.js
mbostock: Let’s promote `clip = "sphere"` to `{type: "Sphere"}` here so we have one fewer code path later?

src/options.js
```diff
@@ -616,12 +616,18 @@ export function maybeNamed(things) {
   return isIterable(things) ? named(things) : things;
 }
 
+// A shared Sphere object coalesces all sphere clips.
+const sphere = {type: "Sphere"};
```
mbostock, line 620: This is for better memoization later?

src/style.js
```diff
@@ -306,24 +306,18 @@ export function* groupIndex(I, position, mark, channels) {
...
+    // Wrap the G element with another (untransformed) G element, applying the
+    // clip to the parent G element so that the clip path is not affected by
+    // the mark’s transform. To simplify the adoption of this fix, mutate the
+    // passed-in selection.node to return the parent G element.
+    selection = create("svg:g", context).each(function () {
+      this.appendChild(selection.node());
+      selection.node = () => this; // Note: mutation!
+    });
+    clipUrl = getFrameClip(context, dimensions);
+  } else if (clip?.type) clipUrl = getGeoClip(clip)(context);
```
mbostock, line 319:
```suggestion
  } else if (clip) clipUrl = getGeoClip(clip)(context);
```

src/style.js
```diff
@@ -356,11 +350,17 @@ const getFrameClip = memoizeClip((clipPath, context, dimensions) => {
...
+  const geoClips = new WeakMap();
+  return (geo, scales) => {
+    if (!geoClips.has(geo)) geoClips.set(geo, clip(geo, scales));
+    return geoClips.get(geo);
+  };
+}
+
+const getGeoClip = memoizeGeo((geo) =>
+  memoizeClip((clipPath, context) => clipPath.append("path").attr("d", context.path()(geo)))
+);
```
mbostock, line 363: I feel like there’s a way to simplify this memoization logic… but I’m just polishing. I’ll look at it if I’m inspired.

mbostock (approved):
Nice!

src/options.js
Fil: yes — we don't want each `"sphere"` and `{type:"Sphere"}` to generate a new clip-path

src/style.js
Fil: I agree that it feels convoluted, like a factory of factory… I didn't find a simpler way yet, but I'm sure there must be one.

Fil: done!

src/plot.js
```diff
@@ -236,6 +236,11 @@ export function plot(options = {}) {
     facetTranslate = facetTranslator(fx, fy, dimensions);
   }
 
+  // A path generator for marks that want to draw GeoJSON.
+  context.path = function () {
+    return geoPath(this.projection ?? xyProjection(scales));
+  };
+
```
mbostock, line 243: I didn’t catch this in the review, but I think this should be assigned quite a bit earlier, especially before we pass the *context* to mark initializers.

Fil: follow-up at https://github.com/observablehq/plot/pull/2252

PR #2244, opened 2024-11-21
tip: {…} can contain PointerOptions

### LINKED ISSUE #2134: Erroneous type definition for the tip {maxRadius} option
If you say `tip: {maxRadius: 100}`, the **maxRadius** option is correctly passed through to the pointer, affecting the interaction as expected; however our typescript defs reject it.

Capture d’écran 2024-08-14 à 18 50 15

Fil:
closes #2134


mbostock:
Oops, already fixed this in #2132 28bdcafe410adfedb4f209b37c9c7654a31f1bef. We don’t want to add this to the `TipOptions` (i.e., the options to the tip mark, since the tip mark doesn’t necessitate the use of the pointer transform); we instead want to allow it in the **tip** mark option.

PR #2249, opened 2024-11-23
top-level legend option, and better types

mbostock:
This introduces a new top-level **legend** option which is inherited by scales that support legends (_color_, _symbol_, and _opacity_). This makes it slightly more concise to enable legends, and takes us a step closer to enabling them automatically if desired. For example:

Screenshot 2024-11-23 at 9 31 27 AM

```js
Plot.plot({
  grid: true,
  inset: 12,
  legend: true,
  marks: [Plot.dot(decathlon, {x: "Long Jump", y: "100 Meters", symbol: "Country", stroke: "Country"})]
})
```

Note above that Plot is smart enough to realize that the _symbol_ legend subsumes the _color_ legend, and hence the _color_ legend does not need to be displayed separately. Plot now suppresses the _color_ legend automatically when the _symbol_ legend also encodes _color_. You can force separate legends by using the **stroke** _symbol_ legend option:

Screenshot 2024-11-23 at 9 35 04 AM

```js
Plot.plot({
  grid: true,
  inset: 12,
  legend: true,
  symbol: {stroke: "currentColor"},
  marks: [Plot.dot(decathlon, {x: "Long Jump", y: "100 Meters", symbol: "Country", stroke: "Country"})]
})
```

I considered the alternative below, which we could have display a colorless _symbol_ legend and no _color_ legend, but it would require more finagling to allow the _symbol_ legend to read the _color_ legend options; let me know if you prefer it:

```js
Plot.plot({
  grid: true,
  inset: 12,
  legend: true,
  color: {legend: false},
  marks: [Plot.dot(decathlon, {x: "Long Jump", y: "100 Meters", symbol: "Country", stroke: "Country"})]
})
```

I also wasn’t sure if we wanted a way to force the display of separate _symbol_ and _color_ legends, as this currently displays only a joint legend (which I think makes sense, since it’s equivalent to inheriting the top-level option):

```js
Plot.plot({
  grid: true,
  inset: 12,
  color: {legend: true},
  symbol: {legend: true},
  marks: [Plot.dot(decathlon, {x: "Long Jump", y: "100 Meters", symbol: "Country", stroke: "Country"})]
})
```

Lastly, this fixes a types bug where legend options were not allowed on scale options. For example, setting the **stroke** _symbol_ legend option previously errored with `Object literal may only specify known properties, and 'stroke' does not exist in type 'ScaleOptions'.` This also introduces more specific types for legend options for additional safety.

### DIFF
--- a/docs/features/legends.md
+++ b/docs/features/legends.md
@@ -76,7 +76,7 @@ Plot does not yet generate legends for the *r* (radius) scale or the *length* sc
 
 ## Legend options
 
-If the **legend** [scale option](./scales.md#scale-options) is true, the default legend will be produced for the scale; otherwise, the meaning of the **legend** option depends on the scale: for quantitative color scales, it defaults to *ramp* but may be set to *swatches* for a discrete scale (most commonly for *threshold* color scales); for *ordinal* *color* scales and *symbol* scales, only the *swatches* value is supported.
+If the **legend** [scale option](./scales.md#scale-options) is true, the default legend will be produced for the scale; otherwise, the meaning of the **legend** option depends on the scale: for quantitative color scales, it defaults to *ramp* but may be set to *swatches* for a discrete scale (most commonly for *threshold* color scales); for *ordinal* *color* scales and *symbol* scales, only the *swatches* value is supported. If the **legend* scale option is undefined, it will be inherited from the top-level **legend** plot option. <VersionBadge pr="2247" />
 
 <!-- TODO Describe the color and opacity options. -->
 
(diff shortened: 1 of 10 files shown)


Fil:
I love it. I don't see the need to have an option to force the separate legends when it's a double encoding. (Given that a user who really wants two legends can always create them with *plot*.legend.)

One caveat is that when we introduce radius legends (hopefully one day 😊 ), this will change the output (for the better…)?

mbostock:
Right, if you set the top-level **legend** option, you’re opting-in to potential future legends. I think that’s okay.

Fil (approved):

PR #2251, opened 2024-11-24
_plot_.rescale

### LINKED ISSUE #1590: Panning and zooming

mbostock:
This is a first cut at an alternate take to zooming, focusing on the programmatic API first rather than the interaction. This expose a _plot_.rescale method that takes scale definitions, allowing scales to be modified after a plot is rendered; the plot is re-rendered in place with the new scale definitions.

For zooming, we would use _plot_.rescale to redefine the _x_ and _y_ domains (or, in the case of an ordinal position scale, we might redefine the range instead). But the _plot_.rescale method could have other uses as well, such as adjusting the _r_ scale as you zoom in or out, switching from [a _linear_ to a _log_ scale](https://observablehq.com/@d3/new-zealand-tourists-1921-2018), redefining the _color_ scale to highlight certain elements, and so on.

Additionally, it should be simpler to sketch this if we focus on the programmatic API first. We should be able to easily implement interactive zooming and [zoom transitions](https://observablehq.com/@d3/scatterplot-tour) on top of this API.

The implementation so far is as minimal as I could make it. The plot is re-rendered, and then the contents of the old plot are replaced with the contents of the new plot. I don’t think this approach will work as-is (for example, I think it breaks the tip mark); but I think we can make a smarter, faster implementation that has the same API.

We also need to figure out what scale options are valid to pass to _plot_.rescale. Since these options are merged not with the original options, but with the materialized scales, the results may not be what you expect; for example, passing the **scheme** option for a _color_ scale typically has no effect because it is lower priority than the materialized **range** or **interpolate** option.

https://github.com/user-attachments/assets/9d8aae3a-c275-45c8-b267-2ff4da2bb223

Fixes #1590.

Previously #2083 #1964 #1738.

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -139,7 +139,7 @@ export function plot(options = {}) {
     stateByMark.set(mark, {data, facets, channels});
   }
 
-  // Initalize the scales and dimensions.
+  // Initialize the scales and dimensions.
   const scaleDescriptors = createScales(addScaleChannels(channelsByScale, stateByMark, options), options);
   const dimensions = createDimensions(scaleDescriptors, marks, options);
 
@@ -158,6 +158,11 @@ export function plot(options = {}) {
   context.ownerSVGElement = svg;
   context.className = className;
   context.projection = createProjection(options, subdimensions);
+
+  // A path generator for marks that want to draw GeoJSON.
+  context.path = function () {
+    return geoPath(this.projection ?? xyProjection(scales));
+  };
 
   // Allows e.g. the axis mark to determine faceting lazily.
   context.filterFacets = (data, channels) => {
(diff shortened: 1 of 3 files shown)


src/plot.js
```diff
@@ -159,6 +159,11 @@ export function plot(options = {}) {
   context.className = className;
   context.projection = createProjection(options, subdimensions);
 
+  // A path generator for marks that want to draw GeoJSON.
+  context.path = function () {
+    return geoPath(this.projection ?? xyProjection(scales));
+  };
+
```
Fil, line 166: see https://github.com/observablehq/plot/pull/2252

PR #2252, opened 2024-11-24
set context.path early

Fil:
Re: https://github.com/observablehq/plot/pull/2243#discussion_r1855336516

My thinking initially was that we needed the scales to be ready, but in fact the _position_ scales are ready when we create the context (initializers can set color, r, etc., but not x or y).

The centroid initializer now uses context.path. (The line mark render, and the link mark render, could already have used it before but I hadn't seen it; now they do, too.)

Note: currently legends create a context without passing the scales. They don't call context.path so it's not a problem, but if we wanted a legend to draw a geojson some time in the future, it would just have to pass scales.


Fil:
Marking as draft as I have to double-check that we're not applying scales twice.

mbostock (changes requested):
I think this belongs as an assignment in the `plot` function, rather than in `createContext`, because we likewise assign `context.projection` in the `plot` function, too.

src/plot.js
```diff
@@ -159,6 +159,11 @@ export function plot(options = {}) {
   context.className = className;
   context.projection = createProjection(options, subdimensions);
 
+  // A path generator for marks that want to draw GeoJSON.
+  context.path = function () {
+    return geoPath(this.projection ?? xyProjection(scales));
+  };
```
Fil, line 165:
```suggestion
  const projection = context.projection = createProjection(options, subdimensions);

  // A path generator for marks that want to draw GeoJSON.
  context.path = () => geoPath(projection ?? xyProjection(scales));
```

this is maybe more consistent?

PR #2254, opened 2024-11-25
temporal contours

### LINKED ISSUE #2250: Plot.contour does not support date values

Fil:
there were two issues:
1. blur2 does not support dates
2. the contour ticks should be derived as utc ticks not numerical ticks

closes #2250


src/marks/contour.js
```diff
@@ -187,10 +188,11 @@ function contourGeometry({thresholds, interval, ...options}) {
 // thresholds across facets. When an interval is used, note that the lowest
 // threshold should be below (or equal) to the lowest value, or else some data
 // will be missing.
-function maybeTicks(thresholds, V, min, max) {
+function maybeTicks(thresholds, V, min, max, temporal) {
   if (typeof thresholds?.range === "function") return thresholds.range(thresholds.floor(min), max);
   if (typeof thresholds === "function") thresholds = thresholds(V, min, max);
   if (typeof thresholds !== "number") return arrayify(thresholds);
+  if (temporal) return scaleUtc().domain([min, max]).nice(thresholds).ticks(thresholds);
```
mbostock, line 195: I think we could use `utcTickInterval` here instead?

```js
if (temporal) {
  thresholds = utcTickInterval(min, max, thresholds);
  return thresholds.range(thresholds.floor(min), max);
}
```

src/marks/contour.js
```diff
@@ -115,7 +115,8 @@ function contourGeometry({thresholds, interval, ...options}) {
     const {pixelSize: k, width: w = Math.round(Math.abs(dx) / k), height: h = Math.round(Math.abs(dy) / k)} = this;
     const kx = w / dx;
     const ky = h / dy;
-    const V = channels.value.value;
+    const temporal = isTemporal(channels.value.value);
+    const V = temporal && this.blur > 0 ? Float64Array.from(channels.value.value) : channels.value.value;
```
mbostock, line 119: Do we want to use `coerceNumbers` here whenever `this.blur > 0`, not just in the `temporal` case? I also wonder if we could provide a hint earlier when the _value_ channel is declared to say that it must include numbers (at least in the `this.blur > 0` case).

Fil: I'd love to make contours work with categorical data, but I feel it would necessitate a completely different approach (we'd have to compute the contours for each of the categories on a boolean raster saying whether a pixel belongs to that category… something like that… with blurring to accomodate the "random-walk" interpolator).

Fil:
rebased

src/marks/contour.js
Fil: this is much better, thanks; the test snapshot gets more "contrast" (since the domain is not artificially niced)

PR #2267, opened 2025-01-03
let the pointer's pointerdown event propagate

### LINKED ISSUE #2052: Pointer interaction interferes with external event handlers
The `pointerdown` handler for the pointer interaction calls `event.stopImmediatePropagation()`:
https://github.com/observablehq/plot/blob/main/src/interactions/pointer.js#L165

As a result, any event handlers added to the SVG after the fact are starved of expected `pointerdown` events. For example, in Mosaic vgplot this breaks discrete point selection if a Plot pointer or tip are active (https://github.com/uwdata/mosaic/issues/361).

For greater interoperability, might it be possible to remove this line, and instead handle any necessary event suppression in a Plot-internal fashion? Thanks!

Fil:
Controlling this event is necessary because (internally) we only want to process it once, even when it's tied to several marks. This typically happens on the crosshair mark which uses the same shared pointer transform on its submarks, resulting in an even number of flag toggling.

However for some applications the upstream app also wants to see that event.

In this PR, instead of preventing propagation, we ignore internal duplicate handling.

closes #2052

cc: @jheer

### DIFF
--- a/src/interactions/pointer.js
+++ b/src/interactions/pointer.js
@@ -4,6 +4,7 @@ import {isArray} from "../options.js";
 import {applyFrameAnchor} from "../style.js";
 
 const states = new WeakMap();
+const processedEvents = new WeakSet();
 
 function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...options} = {}) {
   maxRadius = +maxRadius;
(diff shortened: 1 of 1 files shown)


jheer:
Thanks @Fil! Would love to see this reviewed and released. We would make use of this immediately.

derekperkins:
Is there anything specific holding this back?

PR #2272, opened 2025-01-19
the window transform should respect *z* series

Fil:
Currently *z* is specified as `stroke: {value: "series"}`, it is ignored and we get the incorrect chart.

| correct | incorrect |
|--|--|
| Image | Image |

### DIFF
--- a/test/plots/index.ts
+++ b/test/plots/index.ts
@@ -356,4 +356,5 @@ export * from "./yearly-requests-dot.js";
 export * from "./yearly-requests-line.js";
 export * from "./yearly-requests.js";
 export * from "./young-adults.js";
+export * from "./window.js";
 export * from "./zero.js";
--- a/test/plots/window.ts
+++ b/test/plots/window.ts
@@ -1,0 +1,40 @@
+import * as Plot from "@observablehq/plot";
+import * as d3 from "d3";
+
+export async function windowZ() {
+  const random = d3.randomLcg(42);
+  const data = d3.sort(
+    ["a", "b", "c"].flatMap((type, i) =>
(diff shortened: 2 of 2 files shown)


Fil:
The discrepancy comes from the fact that the specification `{value: "type"}` is analyzed in mark.js to create the marks object’s this.channels; but in the map transform we don't make that same analysis and pass `{value: "type"}`directly to valueof.

The fix I pushed above fixes the situation for the map transform, but the same problem might appear in other places, and it might be better to fix it at the level of valueof. The patch in that case would be:

```diff
--- a/src/options.js
+++ b/src/options.js
@@ -32,6 +32,7 @@ function isBigIntType(type) {
 export const reindex = Symbol("reindex");

 export function valueof(data, value, type) {
+  if (isObject(value) && value.value !== undefined) ({value} = value);
   const valueType = typeof value;
```

Both patches pass all tests, but it only means we don't have a test for other places where z might be computed separately from channels (probably in the stack transform at least … this needs a bit more work to make sure we get the whole picture and add the necessary tests).

In fact, I'm wondering why we compute *z* separately as a valueof, and not take it from the mark's _channels_.

src/options.js
```diff
@@ -280,6 +280,7 @@ export function maybeTuple(x, y) {
 export function maybeZ({z, fill, stroke} = {}) {
   if (z === undefined) [z] = maybeColorChannel(fill);
   if (z === undefined) [z] = maybeColorChannel(stroke);
+  if (isOptions(z)) z = z.value;
```
Fil, line 283: Note: other properties on the options _z_ could be:
- scale — which scale to use
- label — a label for axes/legends
- hint — internal hints (e.g., {length: true} for stack)
- filter — a filter function for the channel
- optional — whether the channel can be null

none of these matter to determine how to group series.

PR #2285, opened 2025-02-20
ordinal opacity

### LINKED ISSUE #2284: Ordinal opacity scales
We support ordinal color scales:

Image

```js
Plot.legend({color: {type: "ordinal", domain: d3.range(10)}})
```

But trying to do the same with an opacity scale results in an error:

```
Error: ordinal opacity scales are not supported
```
```js
Plot.legend({opacity: {type: "ordinal", domain: d3.range(10)}})
```

I see no reason we couldn’t support ordinal opacity scales by dividing the range interval [0, 1] equally based on the cardinality of the domain.

Image

mbostock:
Fixes #2284.

### DIFF
--- a/src/legends.js
+++ b/src/legends.js
@@ -52,20 +52,21 @@ function legendColor(color, {legend = true, ...options}) {
     case "ramp":
       return legendRamp(color, options);
     default:
-      throw new Error(`unknown legend type: ${legend}`);
+      throw new Error(`unknown color legend type: ${legend}`);
   }
 }
 
-function legendOpacity({type, interpolate, ...scale}, {legend = true, color = rgb(0, 0, 0), ...options}) {
-  if (!interpolate) throw new Error(`${type} opacity scales are not supported`);
-  if (legend === true) legend = "ramp";
-  if (`${legend}`.toLowerCase() !== "ramp") throw new Error(`${legend} opacity legends are not supported`);
-  return legendColor({type, ...scale, interpolate: interpolateOpacity(color)}, {legend, ...options});
-}
-
-function interpolateOpacity(color) {
+function legendOpacity(opacity, {legend = true, color = "black", ...options}) {
+  if (legend === true) legend = opacity.type === "ordinal" ? "swatches" : "ramp";
   const {r, g, b} = rgb(color) || rgb(0, 0, 0); // treat invalid color as black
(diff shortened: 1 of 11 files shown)


mbostock:
Hmm, this still crashes if you try to have a threshold opacity scale, which is a separate but related bug:

```js
Plot.cellX(d3.range(10), {fill: "red", opacity: Plot.identity}).plot({opacity: {type: "threshold", legend: true, domain: [1, 2, 3], range: [0.2, 0.4, 0.6, 0.8]}})
```

But also it crashes if you try to display an ordinal opacity scale with a _ramp_ legend, which should be supported (because we can display ordinal color scales with a _ramp_ legend, too):

```js
Plot.cellX(d3.range(10), {fill: "red", opacity: Plot.identity}).plot({opacity: {type: "ordinal", legend: "ramp"}})
```

src/legends/ramp.js
```diff
@@ -70,18 +70,27 @@ export function legendRamp(color, options) {
     const interpolator =
       range === undefined
         ? interpolate
+        : type === "threshold"
+        ? scaleThreshold(
+            Array.from(domain, (_, i) => (i + 1) / (domain.length + 1)),
+            Array.from(range, interpolate)
+          )
```
mbostock: This is dead code; a threshold scale will never have an `interpolate` function, so we don’t have to worry about this case.

mbostock: Hmm, I stand corrected, because it _does_ apply in the edge case of the ramp legend for a threshold opacity scale, which pretends to be a color scale and has a defined interpolate even though it’s a threshold scale… I need to think more. 🤔

mbostock: Also I guess this isn’t testable…

mbostock:
Unfortunately it looks like the `color-mix(…)` technique of #2419 doesn’t actually work (at least here) in part because the canvas element the ramp legend is using is detached, and hence `currentColor` isn’t available. (And more generally it won’t be reliable to have a color scale that references `currentColor` if we’re baking the colors into an image for the legend…)

mbostock:
Resurrected the SVG filter technique from #2416. 👍

mbostock:
I decided it’s easier for the ramp & swatches legends to understand opacity scales, rather than trying to have the opacity scale masquerade as a color scale.

PR #2295, opened 2025-03-10
dataset mark option

Fil:


mbostock (commented):
Hmm, what is the motivation for this PR? Why is it useful to set data attributes in this context? Is it more than “advanced” uses? Is this desirable enough to merit a first-class feature, rather than say a render transform to set attributes?

And how often is it that you want to set a whole set of data attributes, derived from an object, rather than being more specific about what attributes you choose? It feels unlike Plot to have a channel whose values are objects, and allow different attributes if the objects are not homogeneous. And likewise there’s the complexity of dealing with primitive values (and the arbitrariness of `data-value`, and pulling out the channel label, _etc._).

I’m wondering about a more precise design where you can set individual data attributes, say like so for `data-name`:

```js
Plot.dot(athletes, {x: "weight", y: "height", dataset: {name: "name"}})
```

Or like so, which I think I prefer:

```js
Plot.dot(athletes, {x: "weight", y: "height", "data-name": "name"})
```

The latter would require extracting the `data-*` options out of the mark options, but that seems doable. But in any case, I’d like to hear the rationale to justify adding such a feature.

mbostock:
Would D3’s approach of setting __data__ work better (or some other property)? We could avoid serialization if it’s just for programmatic access.

PR #2322, opened 2025-05-13
support textOverflow in tips with key/value pairs

### LINKED ISSUE #2321: Plot.tip should support textOverflow
Currently, textOverflow is ignored in tips that use key/value pairs formatting. (It's appied only in the case where the tip is specified with the title channel).

Image

```js
Plot.tip(
  [
    {
      what: "A long long long long long long long long long long long long way home",
      textOverflow: "ellipsis-middle"
    }
  ],
  { x: "what" }
).plot()
```

Fil:
closes #2321

### DIFF
--- a/src/marks/tip.js
+++ b/src/marks/tip.js
@@ -45,7 +45,7 @@ export class Tip extends Mark {
       frameAnchor,
       format,
       textAnchor = "start",
-      textOverflow,
+      textOverflow = "ellipsis",
       textPadding = 8,
       title,
       pointerSize = 12,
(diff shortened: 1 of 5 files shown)


src/marks/tip.js
```diff
@@ -45,7 +45,7 @@ export class Tip extends Mark {
       frameAnchor,
       format,
       textAnchor = "start",
-      textOverflow,
+      textOverflow = "ellipsis",
```
mbostock: Changing the default causes a significant regression in the case where the tip is used to display the **title** channel (which wraps). Here’s before from tipDotFacets:

Image

Here’s after:

Image

So I don’t think we can simply change the default **textOverflow** to _ellipsis_. And we don’t know at the time the mark is constructed (or do we? probably not with an initializer?) whether we will end up using the _title_ channel and want wrapping, or key-value pairs for all channels and not want wrapping.

I’m not sure yet how we want to handle this.

Fil: Suggestion in [e0d4768](https://github.com/observablehq/plot/pull/2322/commits/e0d476874f3b05a42ffc575bb00c8d93e2f64e21)

PR #2329, opened 2025-06-06
Support interval+reduce in the area mark

### LINKED ISSUE #2328: The dense interval option on the area mark seems to require the **y** option
I would expect this to work:

Image

```js
Plot.areaY(data, {x: "date", reduce: "count", interval: "day"}).plot()
```

After all, it works for a line:

Image

```js
Plot.lineY(data, {x: "date", reduce: "count", interval: "day"}).plot()
```

Instead it seems you have to set the **y** option to anything:

Image

```js
Plot.areaY(data, {x: "date", y: true, reduce: "count", interval: "day"}).plot()
```

Fil:
Defaults for `x`, `y`, `indexOf`, and `identity` are now set inside `maybeDenseIntervalX` / `maybeDenseIntervalY`, instead of being scattered across the mark constructors. This fixes a bug where `areaY` (and `areaX`) with a dense interval did not apply the reducer to the y (or x) channel, because the default was set too late — after the dense interval transform had already been configured.

For example, this now works as expected:

```js
Plot.areaY(data, {x: "date", interval: "day", reduce: "count"})
```

Previously, the dense interval would not reduce `y` because `y` was still undefined when `maybeDenseIntervalX` ran. The `identity` default was only applied later by `maybeIdentityY`.

Closes #2328

### DIFF
--- a/src/marks/area.js
+++ b/src/marks/area.js
@@ -3,13 +3,8 @@ import {create} from "../context.js";
 import {maybeCurve} from "../curve.js";
 import {Mark} from "../mark.js";
 import {first, maybeZ, second} from "../options.js";
-import {
-  applyDirectStyles,
-  applyIndirectStyles,
-  applyTransform,
-  applyGroupedChannelStyles,
-  groupIndex
-} from "../style.js";
+import {applyDirectStyles, applyIndirectStyles, applyTransform, applyGroupedChannelStyles} from "../style.js";
+import {groupIndex} from "../style.js";
 import {maybeDenseIntervalX, maybeDenseIntervalY} from "../transforms/bin.js";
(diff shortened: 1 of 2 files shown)


mbostock (commented):
More description about how this fixes the problem would be helpful in the PR. (And it would be more relevant than your experience with AI, though I appreciate you sharing!) I guess this changes the behavior exactly when y is undefined, but preserves the current behavior when y is null. Will anything else be affected by this fix?

Fil:
Merging main was a bit complicated by 3542325cfbdc4529644b68c9b798d625af9bc445; then I saw that lineX could be simplified to follow the same pattern as we had in areaX (we were setting the defaults `x = indexOf, y = identity` in two places). PTAL?

PR #2332, opened 2025-06-19
implicit zero

### LINKED ISSUE #2331: Include zero in the domain of default positional scales if it's "almost" there?
This can help “ground” line charts when their minimum is somewhere between 0% and 7% of the maximum:

For example, on this multi-line chart it would be nice to include y=0 automagically. Or maybe it's too magic?

before

mbostock:
Fixes #2331. Breaks a bunch of tests, of course. And should probably only apply to positional scales, not to (_e.g._) color? And having an implicit rule would also be nice. Is this too opinionated?

### DIFF
--- a/src/scales/quantitative.js
+++ b/src/scales/quantitative.js
@@ -115,6 +115,14 @@ export function createScaleQ(
     scale.interpolate(interpolate);
   }
 
+  // If the zero option is not specified, then implicitly extend the domain to
+  // include zero if the minimum is less than 7% of the spread (and similarly
+  // for negative domains).
+  if (zero === undefined && type === "linear") {
+    const [min, max] = extent(domain);
+    zero = min > 0 ? min < 0.07 * (max - min) : max < 0 ? max > 0.07 * (min - max) : false;
+  }
+
   // If a zero option is specified, we assume that the domain is numeric, and we
   // want to ensure that the domain crosses zero. However, note that the domain
   // may be reversed (descending) so we shouldn’t assume that the first value is


Fil (approved):
When the `domain` is explicit and `zero` is implicit, we should not modify the domain.

For example:

```js
Plot.plot({x: { domain: [1, 16] }})
```

should result in a domain of [1, 16], not [0, 16]. However we want [0, 16] when we ask for

```js
Plot.plot({x: { domain: [1, 16], zero: true }})
```

When the two statements are present, `zero` has priority. It's useful in some cases.

However when `zero` is implicit, the `domain` statement should be respected, if present, and the implicit zero should only be applied if the domain is derived from the data.

I've pushed new code to that effect (and double-checked all tests).

src/scales/quantitative.js
```diff
@@ -282,13 +282,21 @@ export function inferDomain(channels, f = finite) {
...
+
+  if (zero === undefined && min * max > 0) {
+    // Default to zero for radius, opacity, and length scales
+    const scale = registry.get(key);
+    if (scale === radius || scale === opacity || scale === length) zero = true;
+    // If the zero option is not specified, then implicitly extend the domain to
+    // include zero if the minimum is less than 7% of the spread (and similarly
+    // for negative domains).
+    else if (type === "linear") zero = min > 0 ? min < 0.07 * (max - min) : max < 0 ? max > 0.07 * (min - max) : false;
+    if (zero) return [Math.min(0, min), Math.max(0, max)];
```
mbostock: This is a little confusing because now there are two places where `zero` is applied; one is here but only if `zero` is undefined, and other is in `createScaleQ` when `zero` is truthy. Hmm.

Fil: I've refactored the domain inference logic, PTAL

src/scales/quantitative.js
```diff
@@ -202,12 +183,13 @@ export function createScaleQuantize(
...
     n = range === undefined ? 5 : (range = [...range]).length,
     scheme = "rdylbu",
-    domain = inferAutoDomain(key, channels),
+    domain,
     unknown,
     interpolate,
     reverse
   }
 ) {
+  domain = deriveDomain(key, channels, domain);
```
Fil, line 192: Note : this is equivalent to what was before (because quantile scales do not support the `zero` option), but it feels better to use the same pattern as we do for linear scales. If that makes sense we could also decide to support the zero option for quantize scales, but this PR is not about adding this as a new feature.

Issue #1445, opened 2023-04-13
chore: refresh all snapshot tests after #1442


Fil:
#1442 special cases the mocha tests to avoid churn; in a quieter moment, remove the special case and change all snapshots

PR #2270, opened 2025-01-10
0.6.17

Fil:
TODO:
- [x] complete the CHANGELOG
  - [x] examples and explanations 
  - [x] make separate reference notebooks/pages? ([current notebook](https://observablehq.com/d/796358f1c340c7f1))
  - [x] world airports https://observablehq.com/@observablehq/plot-world-airports
  - [x] vapor https://observablehq.com/@observablehq/plot-us-water-vapor
  - [x] atan2 https://observablehq.com/@observablehq/plot-color-angle
- [x] fix typescript error (happens when we `yarn upgrade`)
- [x] fix typescript-eslint compat
- [x] upgrade all dependencies, including node 22
- [x] fix numerical instabilities in projectionClipAngle test (by rotating 89.9° instead of 90°)

After this is done:
- [ ] review all links in /releases that should now point to CHANGELOG-2024.md instead of CHANGELOG.md.

---

The typescript-eslint issue happens if we `yarn upgrade`, then we get this warning when we run `yarn test:tsc`:

```
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.7.3

Please only submit bug reports when using the officially supported version.
```

due to https://github.com/typescript-eslint/typescript-eslint/blob/81dc579e01c6e95bd347725eab4d9a26a5fc724d/package.json#L123

(A way of dealing with this is to avoid doing `yarn upgrade` for the moment.)

Fixes #2234.

PR #2365, opened 2026-02-18
refresh snapshots

Fil:
This defaults the snapshots (mocha) to use offset=0 (as a follow-up to #1442).

And we refresh all snapshots, which also benefit from the reduced precision in numbers (https://github.com/observablehq/plot/pull/2270, https://github.com/observablehq/plot/pull/2234)

closes https://github.com/observablehq/plot/issues/1445

Additionally, working on this revealed a bug, where the image generated when offset=0.5 was using a different origin for the hexes. As a result, the image produced was different when the DPI offset changed; two viewers using different screens would not see the same binning. The image now is the same independently of the DPI offset (but its coordinates are shifted by -0.5px and the path is translated by +0.5px). Note that the images in the test snapshots have changed, since offset=0 is now our baseline, but that's OK.

The bug is now fixed, but this reveals another issue: I've tried to create a new snapshot test that would show why we need [ox=0.5](https://github.com/observablehq/plot/blob/11a4751ac28bbbb061cddfb652f84370be9ae984/src/transforms/hexbin.js#L12) in `hexbin.js`, but I can't understand the logic any more, since what we gained on the right-hand-side by having the outer hexes be "inside" the frame, we lost on the left-hand side.

| ox=0 | ox=0.5 |
|-|-|
| ox=0 | ox= 5 |
| bad on the rhs | bad on the lhs |

(A better solution would be to rescale around the center by a factor 0.999999999…? which I'm doing here #2367)

### DIFF
--- a/test/output/arcCollatzUp.svg
+++ b/test/output/arcCollatzUp.svg
@@ -56,12 +56,12 @@
   </g>
   <defs>
     <linearGradient id="gradient0" x1="0%" y1="0%" x2="100%" y2="0%">
-      <stop offset="5%" stop-opacity="0.3"></stop>
-      <stop offset="95%" stop-opacity="1"></stop>
+      <stop offset="5%" stop-color="currentColor" stop-opacity="0.3"></stop>
+      <stop offset="95%" stop-color="currentColor" stop-opacity="1"></stop>
     </linearGradient>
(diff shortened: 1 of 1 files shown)


mbostock:
Duplicate of #2375 now? Or obsolete after #2373? We could use `setOffset(0)` in our tests but I think it’s nice to test against 0.5 because otherwise we won’t really have any tests for the offset being implemented correctly…

PR #2369, opened 2026-02-20
cross-facet brushing

Fil:
`Plot.brush({sync: true})` synchronizes the brush selection across all facets — when the user brushes in one facet, the same rectangle appears in all panes and reactive marks update across all facets.

The `filter` function's facet arguments are now optional: omitting `fx` or `fy` (or passing `undefined`) skips the facet check, selecting points across all values of that facet.

This resolves @allisonhorst 's suggestions https://github.com/observablehq/plot/pull/2364#issuecomment-3928139794 and https://github.com/observablehq/plot/pull/2364#issuecomment-3928143576.

I'm keeping this as a separate PR for now (against #2360) to make the logic more apparent. Ideally if we merge this I would migrate the tests and examples to brushX on parallel time series, rather a 2D brush on penguins.

Disclaimer: I used @claude to generate the tests.

### DIFF
--- a/docs/interactions/brush.md
+++ b/docs/interactions/brush.md
@@ -269,18 +269,8 @@ The brush value dispatched on [_input_ events](#input-events). When the brush is
 - **fx** - the *fx* facet value, if applicable
 - **fy** - the *fy* facet value, if applicable
 - **contains** - a method to test whether a point is inside the selection; see [input events](#input-events)
-- **pending** - `true` during interaction; absent when committed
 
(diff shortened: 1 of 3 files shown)


allisonhorst:
Not specific to the cross-facet brushing: one thing I noticed is that the opaque brush border is a bit distracting with the line mark. I think it'd be rare that a person would want to brush in the x and y direction on a line chart (instead of just brushing over time in the x direction, for example), but sharing in any case.

https://github.com/user-attachments/assets/7e2d1b17-813e-42c9-b4d0-6bbfee07c528

Fil:
> what happens here if we add an option for free x/y scales in faceted charts

Yes, it will be tricky. Most probably it just doesn't make sense to support syncing in that case?

Ref. #8

PR #2370, opened 2026-02-20
brush merge

Fil:
This branch merges all the branches that depend on `fil/brush-dataless` (#2360), fixes the conflicts, and changes the tests and examples of cross-facet brushing to a timeline

https://github.com/user-attachments/assets/86fd4345-e844-473a-95ef-833a2aa9df4f

(This branch combines everything, so while it may be better for testing, it's probably not as good for reviewing.)


allisonhorst:
I just went through all of the tests, looking great! The only thing I noted was the sticky tooltip behavior that I mentioned in the [tip+brush PR](https://github.com/observablehq/plot/pull/2361#issuecomment-3970614143).

PR #2371, opened 2026-02-24
pixelRound

Fil:
Given a continuous scale, returns a rounding function that snaps values to the coarsest precision that still distinguishes neighboring pixels. For temporal scales, this means utc or time intervals; for numeric scales, “nice” values = 10^n &times; {1, 2, 5}. For non-linear scales (log, pow, symlog, etc.) we compute precision locally.

This is a separate branch needed by both the dataless brush and the dataless crosshair. We wouldn't necessarily merge it directly since it's not doing anything on its own, but it can be easier to review independently.

### DIFF
--- a/src/precision.d.ts
+++ b/src/precision.d.ts
@@ -1,0 +1,19 @@
+import type {ScaleType} from "./scales.js";
+
+/** Internal d3 scale with type, as produced by createScaleFunctions. */
+export interface MaterializedScale {
+  (value: any): number;
+  type: ScaleType;
+  domain(): any[];
+  range(): number[];
+  invert(value: number): any;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


src/precision.d.ts
```diff
@@ -0,0 +1,19 @@
+import type {ScaleType} from "./scales.js";
+
+/** Internal d3 scale with type, as produced by createScaleFunctions. */
+export interface MaterializedScale {
```
mbostock: `InternalScale` or `InternalScaleFunction` is probably a better name here, since the existing `Scale` interface is already a “materialized” scale’s public representation, and what is unique about this interface is that it is the private internal representation (that we don’t want to expose publicly and that ideally we’d get rid of).

PR #2375, opened 2026-03-06
fix axis offset in snapshots

Fil:
Context:

The diff for https://github.com/observablehq/plot/pull/2373 was hard to read initially because a lot of changes stem from the fact that the axis offsets are removed.

To get deep in the woods: our snapshots in main, still using mocha, are in fact inconsistent. The  D3 axes have an offset of 0 (`window` is not defined when we load the module), whereas other elements in Plot use an offset of 0.5 (the fake `global.window` has been defined).

This branch adds devicePixelRatio = 2 to `global.window`, and hard-codes `offset = 0` so that the snapshots are now consistently using an offset 0 both in d3-axis and in Plot 😅 

Lots of diff, but easy to grep … the only changes (ideally) are in the Plot axes translate property, for instance:

```diff
-      <g class="tick" opacity="1" transform="translate(0.5,0)">
+      <g class="tick" opacity="1" transform="translate(0,0)">
```

However! when we actually regenerate the snapshots, we also get the updated "3 digits after decimal dot" precision change which we had never pushed, so we'll have to put a bit more effort into verifying the diff.

And also! By changing the offset, we modify some scales, and the hexbin/hexgrid coordinates change too, resulting in sometimes strikingly different images. So we have to fix that issue (which in my book is a bug: the fact that we render on a low DPI screen should not change the binning), and make hexbin independent of the offset.

So overall the changes to the snapshots fall into several categories. It's a lot to review manually, but a script can make this easier:

<details>

```js
// Usage: node verify.js

import {spawn} from "node:child_process";
import {max} from "d3";

function git(...args) {
  return new Promise((resolve, reject) => {
    const chunks = [];
    const child = spawn("git", args);
    child.stdout.on("data", (chunk) => chunks.push(chunk));
    child.on("close", (code) => code === 0 ? resolve(Buffer.concat(chunks).toString()) : reject(new Error(`git ${args[0]} failed`)));
    child.on("error", reject);
  });
}

const from = "origin/main";
const to = "origin/fil/fix-axis-offset";
const [diffOut, treeOut] = await Promise.all([
  git("diff", `${from}..${to}`, "--name-only", "--", "test/output/"),
  git("ls-tree", "-r", "--name-only", from, "test/output/")
]);
const files = diffOut.trim().split("\n").filter(Boolean);
const allFiles = treeOut.trim().split("\n").filter(Boolean);

const numRe = /-?\d+(\.\d+)?/g;

function removeTranslates(s) {
  return s.replace(/ transform="translate\([^)]*\)"/g, "");
}

// Classify numbers after stripping translates.
function classifyNumbers(old, neu) {
  const oldNums = Array.from(old.matchAll(numRe), (m) => +m[0]);
  const newNums = Array.from(neu.matchAll(numRe), (m) => +m[0]);
  if (oldNums.length !== newNums.length) return "needs-review";
  const maxDiff = max(oldNums, (a, i) => Math.abs(a - newNums[i]));
  if (maxDiff === 0) return "translate-only";
  if (maxDiff < 0.001) return "precision-only";
  if (maxDiff < 0.501) return "offset-shift";
  if (maxDiff < 1.001) return "integer-rounding";
  return "needs-review";
}

// Classify each file and collect into categories.
const categories = new Map([
  ["unchanged", []],
  ["translate-only", []],
  ["precision-only", []],
  ["offset-shift", []],
  ["integer-rounding", []],
  ["needs-review", []]
]);

const changedSet = new Set(files);

for (const file of allFiles) {
  if (!changedSet.has(file)) {
    categories.get("unchanged").push(file);
  }
}

await Promise.all(files.map(async (file) => {
  const [old, neu] = await Promise.all([git("show", `${from}:${file}`), git("show", `${to}:${file}`)]);
  const category = classifyNumbers(removeTranslates(old), removeTranslates(neu));
  categories.get(category).push(file);
}));

for (const [key, items] of categories) {
  console.log(`\n${key}: ${items.length} files`);
  for (const f of items) console.log(`  ${f}`);
}

```

Results:

```text
unchanged: 27 files
  ...
translate-only: 258 files
  ...
precision-only: 290 files
  ...
offset-shift: 206 files
  ...
integer-rounding: 6 files
  ...
needs-review: 11 files
  test/output/penguinVoronoi1D.svg
  test/output/faithfulDensity1d.svg
  test/output/geoTipXY.svg
  test/output/geoTipGeoCentroid.svg
  test/output/geoTipCentroid.svg
  test/output/geoTip.svg
  test/output/projectionBleedEdges2.svg
  test/output/projectionFitBertin1953.svg
  test/output/walmarts.html
  test/output/walmartsDensity.svg
  test/output/walmartsAdditions.svg
```

</details>

### DIFF
--- a/src/legends/ramp.js
+++ b/src/legends/ramp.js
@@ -3,7 +3,7 @@ import {inferFontVariant} from "../axes.js";
 import {createContext, create} from "../context.js";
 import {map, maybeNumberChannel} from "../options.js";
 import {interpolatePiecewise} from "../scales/quantitative.js";
-import {applyInlineStyles, impliedString, maybeClassName} from "../style.js";
+import {applyInlineStyles, impliedString, maybeClassName, offset} from "../style.js";
 
 export function legendRamp(color, options) {
   let {
@@ -164,6 +164,7 @@ export function legendRamp(color, options) {
         .tickFormat(typeof tickFormat === "function" ? tickFormat : undefined)
         .tickSize(tickSize)
         .tickValues(Array.isArray(ticks) ? ticks : null)
+        .offset(offset)
     )
     .attr("font-size", null)
     .attr("font-family", null)


src/style.js
```diff
@@ -6,7 +6,7 @@ import {isNone, isNoneish, isRound, maybeColorChannel, maybeNumberChannel} from
 import {keyof, number, string} from "./options.js";
 import {warn} from "./warnings.js";
 
-export const offset = (typeof window !== "undefined" ? window.devicePixelRatio > 1 : typeof it === "undefined") ? 0 : 0.5; // prettier-ignore
+export const offset = 0;
```
mbostock: This breaks a feature of Plot, though? This offset is used to have crisp rendering for lo-dpi screens. I’m okay with fixing the behavior in tests, but it seems wrong to remove this feature from production usage of Plot.

Fil:
Obsolete after #2373

PR #2381, opened 2026-03-13
no strokeDasharray on markers

### LINKED ISSUE #2131: strokeDashArray affects markers
## Current State
Certain settings on `strokeDasharray` property may break the markers. Here for instance instead of an arrow on the line we have a tick:
image

## Repro
```
Plot.plot({
  marks: [
    Plot.lineY([[0,5],[5,2],[10,0]], { 
      x: (d) => d[0], 
      y : (d) => d[1],
      strokeDasharray: "1,10",
      strokeWidth:3,
      markerMid: "arrow"
    })
  ]
})
```

## Expected State
`strokeDasharray` should not affect markers, or there should be a separate property to control it:
image

Fil:
closes #2131

### DIFF
--- a/src/marker.js
+++ b/src/marker.js
@@ -1,6 +1,6 @@
 import {create} from "./context.js";
 import {unset} from "./memoize.js";
-import {isNoneish, keyof} from "./options.js";
+import {keyof} from "./options.js";
 
 export function markers(mark, {marker, markerStart = marker, markerMid = marker, markerEnd = marker} = {}) {
   mark.markerStart = maybeMarker(markerStart);


Fil:
a) I believe `stroke-opacity` is useful to inherit. `fill` doesn't really make sense in any of the supported marks.

b) I researched the idea of using an external `<defs>` group, but I wouldn't want to add that group to the top-level `svg` (it makes it harder to control multiple renderings with, say, a pointer transform: the number of defs might grow unbounded). We could nest the current `<g>` under a wrapper `<g>` that also includes `<defs>` as a sibling of the styled `<g>`, but it feels "too much" given point a) above.

PR #2382, opened 2026-03-17
pointer pool option

### LINKED ISSUE #1835: Tips are chaotic on compound marks

Fil:
Adds a `pool` option to the pointer transform, that defaults to true on the tip mark (including marks derived with the `tip: true` mark option). Marks that pool their pointer transform show only the closest point among all marks, preventing collisions. Non-pooled marks (like crosshair sub-marks) are unaffected. This also handles facet deduplication.

closes #1835

https://github.com/user-attachments/assets/d4b63f63-51ea-469a-89c1-75a22823149b

You can opt out by setting `pool: false` (but why would you?). In the future this could be extended to different named pools (but again, why?).

### DIFF
--- a/src/mark.js
+++ b/src/mark.js
@@ -26,7 +26,7 @@ export class Mark {
       clip = defaults?.clip,
       channels: extraChannels,
       tip,
-      pool = defaults.pool,
+      pool = defaults?.pool,
       render
     } = options;
     this.data = data;


Fil:
rebased

src/interactions/pointer.d.ts
```diff
@@ -17,6 +17,13 @@ export interface PointerOptions {
   /** The vertical target position channel, typically bound to the *y* scale. */
   py?: ChannelValue;
 
+  /**
+   * Whether this mark participates in the pointer pool, which ensures that
+   * only the closest point is shown when multiple pointer marks are present.
+   * Defaults to true for the tip mark.
+   */
+  pool?: boolean;
```
mbostock: This isn’t the right place to declare this option. The pointer transform ignores `options.pool`; it only looks at `this.pool` (a.k.a. `mark.pool`), and `mark.pool` is only currently set by the tip mark. So, trying to use this option on any other mark won’t have any effect.

Either we should make this option work for every mark (either reading it from `options.pool` instead of `this.pool`, or by having the base mark constructor populate `this.pool` for all marks) or we should move the pool option to `TipOptions` where it will only work for the tip mark.

Moving it to `PointerOptions` (`options.pool`) feels the cleanest, but I’m not sure then how the default should be applied. According to this language, it should default to `this instanceof Tip`, but that seems funky (there’s no way for a custom mark to provide its own default for pool)… And there’s no way that the tip mark can reach into the pointer transform options to populate the default since the pointer transform is constructed prior to the tip mark. So I guess that means it belongs on `MarkOptions`, even though it’s only used by the pointer transform?

Why is it that only the tip mark wants pool to default to true?

mbostock: To answer my own question, I guess the tip mark wants pool to default to true because the tip mark is generally “exclusive” as in typically you only want to see one tip at a time. (Though we do support rendering tips statically as annotations, in which case you _do_ want multiple tips at the same time.)

src/interactions/pointer.js
```diff
@@ -72,32 +77,23 @@ function pointerK(kx, ky, {x, y, px, py, maxRadius = 40, channels, render, ...op
...
-            });
-            return;
-          }
-        }
-        render(ii);
+        if (!pool) return void render(ii);
+        if (ii == null) render(ii);
+        pool.set(renderIndex, {ii, ri, render});
+        if (pool.frame !== undefined) cancelAnimationFrame(pool.frame);
+        pool.frame = requestAnimationFrame(() => {
```
mbostock, line 91: Decorating onto `pool.frame` feels a bit dirty; can we use `state.poolFrame` instead? Or I guess the problem is when `state.pool` is undefined and we’re falling back to `facetPool`? So really `state.pool` needs to have a `state.pool.map` and a `state.pool.frame` to track the pool-specific state (and similarly `facetPool.map` and `facetPool.frame`)?

docs/marks/tip.md
```diff
@@ -201,7 +201,7 @@ Plot.plot({
 :::
 
 :::tip
-When multiple tips are visible simultaneously, some may collide; consider using the pointer interaction to show only the one closest to the pointer, or use multiple tip marks and adjust the **anchor** option for each to minimize occlusion.
+The tip mark defaults the [**pool**](../interactions/pointer.md#pointer-options) option to true, ensuring multiple marks with the *tip* option don’t collide.
```
mbostock: This is losing a useful tip relevant to the example above, which is rendering multiple tips statically!

mbostock:
Please merge if you approve. 🙏

Fil:
An alternative name could be `exclusivePointer`; would that be any better?

PR #2386, opened 2026-03-18
The interval option for line and area should round down

### LINKED ISSUE #1931: The interval option for line and area perhaps should round down rather than center
Say you have an `activity` table that has active users by `date` and `login`. You might want to draw stacked dots, and then draw a line between active users. But if you use the **interval** option, you might be surprised that the line is shifted to the right to the middle of the day (noon) rather than the start of the day (midnight).

It’s the difference between this

```js
Plot.lineY(activity, Plot.stackY({x: "date", interval: "day", z: "login"}))
```

and this

```js
Plot.lineY(activity, Plot.binX({x: "x1", y: "first", filter: null}, Plot.stackY({x: "date", interval: "day", z: "login"})))
```

Fil:
Capture d’écran 2026-03-18 à 11 25 47

closes #1931

### DIFF
--- a/src/transforms/bin.js
+++ b/src/transforms/bin.js
@@ -67,24 +67,22 @@ export function bin(outputs = {fill: "count"}, options = {}) {
   return binn(x, y, null, null, outputs, maybeInsetX(maybeInsetY(options)));
 }
 
-function maybeDenseInterval(x, options) {
-  options = withTip(options, x);
-  if (options.interval == null) return options;
+function maybeDenseInterval(bin, x, y, options) {
+  if (options?.interval == null) return options;
   const {reduce = reduceFirst} = options;
   const outputs = {filter: null, [x]: `${x}1`};
-  const y = x === "x" ? "y" : "x";
   if (options[y] != null) outputs[y] = reduce;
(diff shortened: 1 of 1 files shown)


mbostock (commented):
Hmm, this seems wrong to me. Here is my test case:

```js
test(async function aaplInterval() {
  const aapl = (await d3.csv<any>("data/aapl.csv", d3.autoType)).slice(0, 81);
  return Plot.plot({
    x: {grid: true, ticks: "week"},
    marks: [
      Plot.dotY(aapl, {x: "Date", y: "High", interval: "week"}),
      Plot.lineY(aapl, {x: "Date", y: "High", interval: "week", reduce: "max"}),
      Plot.lineY(aapl, {x: "Date", y: "High", interval: "week", reduce: "min"})
    ]
  });
});
```

In main, the dot and line align because both round to the middle of the interval:

Image

But in this branch, the dot and line don’t align because the line rounds down:

Image

Maybe #1931 is just ill conceived? Or should we get rid of `maybeIntervalMidX` and `maybeIntervalMidY`, and have the dot and text marks round down, too? That would look like this:

Image

PR #2388, opened 2026-03-18
fit tip

### LINKED ISSUE #2011: the default tip sometimes goes in the wrong direction
This example, given by @mootari, where the right-most tip should have been chosen to anchor right:

Capture d’écran 2024-03-06 à 16 38 34

```
Plot.rectX([1, 1, 1, 1, 1], {
  x: Plot.identity,
  fill: Plot.indexOf,
  title: () =>
    "Lorem ipsum lorem ipsum lorem ipsum Lorem ipsum lorem ipsum lorem ipsum Lorem ipsum lorem ipsum lorem ipsum",
  tip: true
}).plot({ height: 100, marginTop: 20 })
```

Fil:
fixes #2011

~~saving as draft as I don't think we have enough test coverage~~

### DIFF
--- a/test/output/tipAnchors.svg
+++ b/test/output/tipAnchors.svg
@@ -1,0 +1,98 @@
+<svg class="plot" fill="currentColor" font-family="system-ui, sans-serif" font-size="10" text-anchor="middle" width="640" height="160" viewBox="0 0 640 160" style="overflow: visible;" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <style>
+    :where(.plot) {
+      --plot-background: white;
+      display: block;
+      height: auto;
+      height: intrinsic;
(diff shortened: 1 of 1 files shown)


mbostock:
I made a notebook to explore the logic: https://observablehq.com/d/248a1e32c96c45db

The `fitLeft` check doesn’t strictly mean “the tip would fit using the left anchor” because the tip could still be outside the vertical bounds. It tells you when any of the left-anchored tips cross the right edge of the chart area.

Also I think the `pointerSize` is effectively doubled when using a corner anchor, which doesn’t seem right? (#2420)

Fil:
Rebased, and added a vertical test (and similar bug fix). It's a pity we don't have get `getBoundingClientRect` in JSDOM since it means the tests must be checked visually. Not sure it's worth to use playwright tho…

PR #2179, opened 2024-09-27
don’t stack with indexOf

mbostock:
Fixes #2178.

Issue #2178, opened 2024-09-26
`areaY` plot does not correctly render discontinuities


plmrry:
The `Plot.areaY` function does not seem to respect `null` values the same way `Plot.area` does:

https://observablehq.com/@plmrry/area-y-bug

<img width="467" alt="Screenshot 2024-09-26 at 6 37 06 PM" src="https://github.com/user-attachments/assets/80c01c3a-a27b-4641-a585-b03e633cdd62">

mbostock:
The stack transform is likely coercing null to zero. This might be intended… I’ll need to think about this more.

mbostock:
Although this isn’t explicitly documented, this is functioning (at least partly) as intended: the stack transform, like d3-array methods such as d3.sum, ignores null, NaN, and undefined values when computing the extent of each stacked column. The code can be seen here:

https://github.com/observablehq/plot/blob/e0bf09ea4dd537188c1fe11eca169d46bafb4fa8/src/transforms/stack.js#L107-L109

You probably didn’t intend to use the stack transform in this case, and the areaY shorthand probably shouldn’t use the stack transform since it doesn’t make sense in this context. You can fix your code by setting the **y2** channel to opt out of the stack transform:

```js
Plot.plot({
  width,
  height: 200,
  marks: [
    Plot.line(data, { x: (d, i) => i, y: (d) => d }),
    Plot.areaY(data, { y2: Plot.identity, opacity: 0.1 })
  ]
})
```

I’ll make the areaY shorthand do this by default.

PR #2390, opened 2026-03-19
stack invalid values as NaN

Fil:
We have originally stacked invalid values as 0 height segments (where `X1[i] == X2[i]` is the current height of the stack). But I think they should be ignored instead (`X1[i] = X2[i] = NaN`). This feels more principled.

Pros:
- cleaner information (an invalid value is kept throughout)
- allows to draw interrupted stacked lines

Con:
- gaps appear when an area "suddenly disappears" instead of vanishing in a continuous way (but this is "wrong", and can be solved by mapping NaNs to zeroes if the user wants, so it's not removing any feature).

| before | after |
|-|-|
| before | after |

blocks #2133

### DIFF
--- a/src/transforms/stack.js
+++ b/src/transforms/stack.js
@@ -92,8 +92,8 @@ function stack(x, y = one, kx, ky, {offset, order, reverse}, options) {
       const Z = valueof(data, z);
       const compare = order && order(data, X, Y, Z);
       const n = lengthof(data);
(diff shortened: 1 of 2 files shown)


mbostock:
Previously: https://github.com/observablehq/plot/issues/2178 https://github.com/observablehq/plot/pull/2179.

PR #2395, opened 2026-03-24
year format heuristic

### LINKED ISSUE #768: Smarter formatting for year channels?

Fil:
Format year-like numbers (integers in `[1000, 3000]`) without thousands separators on axes and tips.

Closes #768

This introduces two new helpers:
- `isYearDomain(domain)` — returns `true` if all finite numeric values are integers in `[1000, 3000]`; `false` for non-number types, empty domains; ignores NaN and Infinity
- `formatYear(value)` — formats as plain integer `YYYY` if the value is a number in `[0, 9999]` (or possibly `yyyy.xyz` if applied to a non-integer), returns "NaN" for NaN, else falls back to `formatNumber(value)` (once we decide to format without a comma, we do it over [0…9999] so that if 7000 unexpectedly appears, it is formatted consistently just like 2000).

When deriving an ordinal or linear scale, the data values are checked with `isYearDomain`, and a hint is added to the scale; in axes, the `tickFormat` then defaults to `formatYear`. For tips, when the channel is associated to the scale, we also check the hint; if there is no scale, we call `isYearDomain` too.

I chose to test `[1000, 3000]` for the heuristic (instead of, say, `[0, 9999]`), because the latter was generating many changes in snapshots. (The point is to have a heuristic that catches the most common cases while still being easy to remember; in that regard `[1000, 3000]` is slightly more complicated than`[0, 9999]`, but it seems more useful that way — or maybe it is that I'm too accustomed to it?)

I wonder if we should:
* accept `tickFormat: "year"` as an option (DONE)
* export `formatYear`?

### DIFF
--- a/test/plots/year-format.ts
+++ b/test/plots/year-format.ts
@@ -19,6 +19,15 @@ export async function yearFormat() {
 export async function yearFormatExplicit() {
   return Plot.plot({
     x: {tickFormat: "year"},
-    marks: [Plot.dot([{x: 4000, y: 1}, {x: 5000, y: 2}, {x: 6000, y: 3}], {x: "x", y: "y", tip: true})]
+    marks: [
+      Plot.dot(
+        [
+          {x: 4000, y: 1},
+          {x: 5000, y: 2},
+          {x: 6000, y: 3}
+        ],
+        {x: "x", y: "y", tip: true}
+      )
+    ]
   });
 }


docs/features/scales.md
```diff
@@ -177,6 +177,8 @@ Plot.plot({x: {type: "log", domain: [1e0, 1e5], grid: true}})
 
 If you prefer conventional notation, you can specify the **tickFormat** option to change the behavior of the axis. The **tickFormat** option can either be a [d3.format](https://d3js.org/d3-format) string or a function that takes a tick value and returns the corresponding string. Note, however, that this may result in overlapping text.
 
+For linear and ordinal scales, if the data values are all integers between 1,000 and 3,000, Plot assumes they represent years and defaults to a **tickFormat** without thousands separators — for example, `2026` instead of `2,026`. You can opt into this behavior explicitly with `tickFormat: "year"`. <VersionBadge pr="2395" />
```
mbostock: This documentation is separating the text describing the example from the example… I’ll find a better place for this in #2403.

PR #2397, opened 2026-03-27
plot.scale("projection")

### LINKED ISSUE #1191: expose the projection to make it reusable
maybe just as _plot_.scale("projection")? If the goal is to make it reusable, it doesn't matter if it's an opaque {stream}, and the code is ultra-simple. If we extend the scope and want to get the name back, it's much more difficult.

Fil:
`plot.scale("projection")` exposes the materialized projection from a plot

The returned object exposes the resolved projection options, reflecting the actual values used to construct the projection, making it possible to create a new plot with the same projection. The projection object also exposes an **apply** method that projects a [*longitude*, *latitude*] point to [*x*, *y*] pixel coordinates, as well as an **invert** method.

closes #1191
supersedes #2038

### DIFF
--- a/src/context.d.ts
+++ b/src/context.d.ts
@@ -1,6 +1,6 @@
 import type {GeoPath} from "d3";
 import type {MarkOptions} from "./mark.js";
-import type {ProjectionImplementation} from "./projection.js";
+import type {Projection} from "./projection.js";
 
 /** Additional rendering context provided to marks and initializers. */
 export interface Context {
@@ -17,7 +17,7 @@ export interface Context {
   className: string;
 
   /** The current projection, if any. */
-  projection?: ProjectionImplementation;
+  projection?: Projection;
 
   /** A function to draw GeoJSON with the current projection, if any, otherwise with the x and y scales. */
   path: () => GeoPath;


src/plot.js
```diff
@@ -340,6 +340,10 @@ export function plot(options = {}) {
     if ("value" in svg) (figure.value = svg.value), delete svg.value;
   }
 
+  if (context.projection) {
+    const {stream: _, ...projection} = context.projection;
+    scales.scales.projection = projection;
```
mbostock: No need to also put the projection here.

mbostock (changes requested):
Okay, I have pared back this PR by avoiding the introduction of a new “materialized projection” interface (_i.e._, `Projection`), instead exposing the existing `ProjectionImplementation` interface which was already exposed as `context.projection`.

This avoids the challenge of trying to normalize a provided `ProjectionImplementation` into a `Projection` (noted above). More importantly, it ensures when you reuse a projection across a plot, the projections behave identically rather than rescaling the projection to fit the dimensions of the other plot. This mirrors the current behavior of *plot*.scale (with respect to scales), and I believe is what we want. If you want to share the same projection type (or some other partial specification of a projection), then you should simply splat your projection options into multiple plots; you don’t need *plot*.scale to help with that.

All that said, I’d still like to expose a *projection*.apply and *projection*.invert convenience method, since that seems useful for implementing interaction. And in the case that the provided *projection* implementation is a function and exposes *projection*.invert, we can simply use that as-is. I’ll work on that next…

PR #2405, opened 2026-04-04
cleaner area line option (WIP)

Fil:
~~This still needs updates to the tests (to actually test that the line and fill options are going to the right place).~~

for #2133.


mbostock:
Superseded by #2407.

PR #2408, opened 2026-04-05
handle null lines

### LINKED ISSUE #2274: Uncaught TypeError when passing `null` to a mark's title

mbostock:
Supersedes #2275. Fixes #2274.

### DIFF
--- a/test/plot-snapshot.js
+++ b/test/plot-snapshot.js
@@ -10,6 +10,7 @@ export function test(plot) {
   const name = plot.name;
   it(name, async () => {
     const root = await (name.startsWith("warn") ? assert.warnsAsync : assert.doesNotWarnAsync)(plot);
+    if ("ready" in root) await root.ready;
     const ext = root.tagName === "svg" ? "svg" : "html";
     for (const svg of root.tagName === "svg" ? [root] : root.querySelectorAll("svg")) {
       svg.setAttributeNS("http://www.w3.org/2000/xmlns/", "xmlns", "http://www.w3.org/2000/svg");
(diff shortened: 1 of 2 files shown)


test/output/tipDispatch.svg
mbostock: This looks bad because JSDOM doesn’t implement `getBBox` (and our implementation returns 0×0), but it’s correct from a test perspective.

matthewCmatt:
Thanks for following up and fixing this!

Issue #2137, opened 2024-08-15
Seasonal transform


fgregg:
I find myself making a lot of cycle / season line charts like https://observablehq.com/@fgregg/chicago-citykeys-issued

Right now, i have to do some data munging to do them, and it would be really great if it was a supported transform. 

perhaps relates to #133

Fil:
Yes it would be useful to have as a `cycle: "1 year"` scale option, maybe?
Related experiment: https://observablehq.com/@observablehq/plot-cyclical-time-scale

fgregg:
wow, that's such a nice solution @Fil!

fgregg:
a `cycle` option would be great, maybe with a `cycle_start: 2024-06-01` option as well?

Issue #1888, opened 2023-10-02
Periodic scales?


Fil:
When dealing with time series, we often want to overlay lines to facilitate year on year comparisons. Or, when we facet by year, we'd like the time axis (within facets) to spread from start to end of year.

A way to do it is to change *x* though a scale transform, _e.g._:
- `d => d3.utcDay.count(d3.utcYear(d.Date), d.Date)` (number of days since start of the year), or
- `d => +d.Date - d3.utcYear(d)` (number of milliseconds since start of the year), or even
- `d => new Date(+new Date(Date.UTC(1972, 0, 1) + +d - d3.utcYear(d))` (a date between 1st Jan. 1972 and 31 Dec 1972 — a leap year)

This is a bit complex to get right (UTC, DST and leap years can make it hard to adjust), and it can cause issues elsewhere — for example in the tips we want the true data, not a shifted value, and on the date axis we want to suppress the year.

I wonder if we could have a "periodic: year" syntax in the scale definition which took care of all this for us. As it is orthogonal with intervals, we would accept the same DSL — "periodic: 6 months"; "periodic: 1" would plot the fractional value.

Here's a notebook that shows how what I'd like to simplify:

Fil:
See also https://github.com/observablehq/plot/issues/2137

PR #2413, opened 2026-04-07
horizon mark

mbostock:
This sketches a horizon mark, a variant of the area mark for making horizon charts. A horizon chart can now be as simple as:

```js
Plot.horizonY(industries, {x: "date", y: "unemployed", fy: "industry"}).plot()
```

The awkward part of the implementation is how the horizon mark affects the default _y_ scale domain. I used a channel hint which is consumed in `inferDomain`. However, this is made difficult by the stack transform applying its own “lengthy” hint, and needing to propagating the input channel hint (from _y_ to _y1_ and _y2_). Also, when the hint is constructed we don’t yet know the computed _y_-values, so I needed to use `valueOf` — gross! If I move the min/max hint to the value array instead of the channel object, I could perhaps avoid both problems.


Fil:
This feels related (or half related?) to the notion of periodic / cyclical scales:
- https://github.com/observablehq/plot/issues/2137
- #1888
- https://observablehq.com/@observablehq/plot-cyclical-time-scale
- https://observablehq.com/@observablehq/periodic-scale-1888

mbostock:
I don’t think horizon charts can be represented with a periodic scale because the area shape has to be drawn multiple times.

PR #2416, opened 2026-04-10
opacity legend 2

Fil:
Upgrade the color legend, to
* support for display-p3 colors
* ~~have a checkerboard~~
* support dark mode

_a test showing the ~~3~~ 2 new features:_

a test complete with display-p3 color, checkerboard, and dark mode


Fil:
a bit of work to merge…

test/plots/legend-opacity.ts
mbostock, line 1: I don’t think the tests should manually tests both light and dark; if we want that, we should do that universally; too much boilerplate, makes this PR harder to review since snapshots have moved, and GitHub doesn’t provide a nice way of reviewing HTML changes so we prefer to use SVG snapshots whenever possible. During development, you can manually switch your color scheme to test light and dark.

mbostock (changes requested):
I find the checkerboard ugly, but more importantly, it doesn’t match what you see in the chart itself (since charts don’t have a checkerboard background). The use of a checkerboard pattern therefore feels like a technical cue; the reader must apply their knowledge of the checkerboard convention from other applications to say “a ha, this is an opacity encoding.” Often I don’t think the user cares about the distinction between opacity and color, since in practice due to compositing an opacity encoding _is_ a color encoding.

Of course I like that the opacity legend supports CSS4 and dark mode, so we should do that, but I’m going to remove the rest.

mbostock (approved):

PR #2418, opened 2026-04-11
test raster mark support for mixed opacity

Fil:
adopt [globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation) in the raster color converter

test the color converter cache for opacity

this does not appear to be faster, clearRect is probably super fast; but it conveys the intent better and avoids a looped operation


test/plots/raster-precision.ts
```diff
@@ -39,3 +39,17 @@ test(async function rasterFacet() {
     ]
   });
 });
+
+test(async function rasterOpacityCache() {
+  // Test the opacity cache in the color converter
```
mbostock: In what way is this testing caching? I don’t think caching behavior is observable (or should be observable) externally.

mbostock (changes requested):
I don’t think we should do this. In my experience, setting globalCompositeOperation typically makes the canvas much slower by disabling hardware acceleration. Also browsers haven’t always implemented compositing consistently. I don’t see any performance cost here, but I also don’t a benefit in terms of logical clarity, so I’d rather stick with the common path of clearRect.

test/plots/raster-precision.ts
Fil: it's a test that mixes colors with opacities — it would fail if you forgot to set globalCompositeOperation or clear the canvas. None of the existing tests fail in that case.

mbostock: Okay, then we should call it “mixed opacity” in the sense that it alternates between opaque and partially opaque colors. There’s nothing in the test that is specific to caching.

mbostock (approved):
I removed the use of globalCompositeOperation and renamed the test (and made it slightly prettier).

PR #2419, opened 2026-04-11
fix ramp legend colorSpace

mbostock:
This replaces the SVG filter added in #2416 with a simpler `color-mix(…)` which is available since the ramp legend directly sets context.fillStyle (and hence we don’t have to worry, as we did with the raster mark, about converting a color string into RGBA values). It also allows the caller to set the **colorSpace** when constructing an opacity legend, so that P3 colors can be used as intended.

I’m not sure why the snapshots aren’t appearing correctly here; the tests seem to be correct locally.

### DIFF
--- a/src/legends/ramp.js
+++ b/src/legends/ramp.js
@@ -21,6 +21,7 @@ export function legendRamp(color, options) {
     fontVariant = inferFontVariant(color),
     round = true,
     opacity,
+    colorSpace,
     className
   } = options;
   const context = createContext(options);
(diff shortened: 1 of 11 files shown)


src/legends/ramp.js
```diff
@@ -21,6 +21,7 @@ export function legendRamp(color, options) {
     fontVariant = inferFontVariant(color),
     round = true,
     opacity,
+    colorSpace,
```
mbostock: Arguably we should default this to _display-p3_ so that it will render correctly in case the caller uses P3 colors (as sRGB colors will not be harmed by using _display-p3_). Maybe even it shouldn’t be an option?

mbostock:
I think the snapshots don’t render correctly because node-canvas doesn’t support [CSS Color Module Level 5](https://drafts.csswg.org/css-color-5/#color-mix) `color-mix(…)`… But I think this is still a preferable implementation and it should make #2285 easier.

src/legends/ramp.js
Fil: Sure, that will be much simpler

Fil:
I'm not seeing why it's preferable or simpler than the basic greyscale ramp + svg filter to add the color. (But as long as it works I'm fine with it.)

mbostock:
It’s simpler because you don’t need the filter — as the diff shows.

PR #2426, opened 2026-04-13
0.6.18

Fil:
* changelog
* version

(it's unfortunate that the diff is polluted as it mixes the rename operation of changelog.md to changelog-2025.md and the new write-up. See individual commits instead: [rename](https://github.com/observablehq/plot/pull/2426/commits/987dbbfa98fad2921cf285897d96736541b1ea75) & [new changelog](https://github.com/observablehq/plot/pull/2426/changes/82d7955639ac8bec2abe79743cbb095229b85d94))


CHANGELOG.md
```diff
@@ -0,0 +1,103 @@
...
+      fill: "island",
+      interpolate: "random-walk",
+      blur: 8,
+      colorSpace: "display-p3"
+    })
+  ]
+})
+```
+
+[Pointer tips](https://observablehq.com/plot/interactions/pointer) are now exclusive by default: when multiple pointer-driven marks overlap, only the closest point is shown. This avoids the common issue of multiple marks creating overlapping tips. This behavior is controlled by the new **pool** [mark option](https://observablehq.com/plot/features/marks#mark-options), which defaults to tru for tip marks; set `pool: false` to restore the previous behavior.
```
yurivish: should `tru` say `true`?

PR #2427, opened 2026-04-13
Add support for projections to the standalone Plot.scale function

Fil:
_e.g._ Plot.scale({projection: {type: "mercator", width, domain: …}}).

When merged, we must add this to the changelog PR #2426:

~~~md
The standalone [Plot.scale](https://observablehq.com/plot/features/scales#scale) function now also supports projections.

```js
const projection = Plot.scale({projection: {type: "mercator"}});
projection.apply([-1.55, 47.22]) // [316.7, 224.2]

const plot = Plot.plot({projection, marks: [Plot.sphere()]});
```
~~~

### DIFF
--- a/docs/features/scales.md
+++ b/docs/features/scales.md
@@ -1065,3 +1065,16 @@ As another example, below are two plots with different options where the second 
 const plot1 = Plot.plot({...options1});
 const plot2 = Plot.plot({...options2, color: plot1.scale("color")});
 ```
+
+Plot.scale also supports projections. <VersionBadge version="0.6.18" /> The returned projection object exposes *apply* and *invert* methods for converting between geographic and pixel coordinates, and can be passed as the **projection** option of another plot.
+
+```js
+const projection = Plot.scale({projection: {type: "mercator"}});
+projection.apply([-1.55, 47.22]) // [316.7, 224.2]
+```
+
+The projection's **width** defaults to 640, and its **height** defaults to the width times the projection's natural aspect ratio. You can override these with the **width** and **height** options, and inset the projection with the **margin** and **inset** options.
+
+```js
+const projection = Plot.scale({projection: {type: "albers-usa", domain, width: 960, height: 600}});
+```
--- a/docs/features/scales.md
+++ b/docs/features/scales.md
@@ -1066,7 +1066,7 @@ const plot1 = Plot.plot({...options1});
 const plot2 = Plot.plot({...options2, color: plot1.scale("color")});
 ```
 
-Plot.scale also supports projections. <VersionBadge version="0.6.18" /> The returned projection object exposes *apply* and *invert* methods for converting between geographic and pixel coordinates, and can be passed as the **projection** option of another plot.
+Plot.scale also supports projections. <VersionBadge pr="2427" /> The returned projection object exposes *apply* and *invert* methods for converting between geographic and pixel coordinates, and can be passed as the **projection** option of another plot.
 
 ```js
 const projection = Plot.scale({projection: {type: "mercator"}});
--- a/src/dimensions.d.ts
+++ b/src/dimensions.d.ts
@@ -1,3 +1,21 @@
+/** Options for specifying the dimensions of a plot or standalone projection. */
+export interface DimensionOptions {
+  /** The outer width in pixels, including margins. Defaults to 640. */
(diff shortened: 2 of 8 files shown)


docs/features/scales.md
```diff
@@ -1065,3 +1065,16 @@ As another example, below are two plots with different options where the second
 const plot1 = Plot.plot({...options1});
 const plot2 = Plot.plot({...options2, color: plot1.scale("color")});
 ```
+
+Plot.scale also supports projections. <VersionBadge version="0.6.18" /> The returned projection object exposes *apply* and *invert* methods for converting between geographic and pixel coordinates, and can be passed as the **projection** option of another plot.
```
mbostock:
```suggestion
Plot.scale also supports projections. <VersionBadge pr="2427" /> The returned projection object exposes *apply* and *invert* methods for converting between geographic and pixel coordinates, and can be passed as the **projection** option of another plot.
```

Let’s not get ahead of ourselves. 🙂

src/scales.d.ts
```diff
@@ -673,3 +674,4 @@ export interface Scale extends ScaleOptions {
  * ```
  */
 export function scale(options?: {[name in ScaleName]?: ScaleOptions}): Scale;
```
mbostock: Is `options` really optional here?

```suggestion
export function scale(options: {[name in ScaleName]?: ScaleOptions}): Scale;
```

src/scales.js
```diff
@@ -526,12 +527,28 @@ export function scale(options = {}) {
...
     if (!isScaleOptions(options[key])) continue; // e.g., ignore {color: "red"}
     if (scale !== undefined) throw new Error("ambiguous scale definition; multiple scales found");
-    scale = exposeScale(normalizeScale(key, options[key]));
+    scale = key === "projection" ? scaleProjection(options[key]) : exposeScale(normalizeScale(key, options[key]));
   }
   if (scale === undefined) throw new Error("invalid scale definition; no scale found");
   return scale;
 }
 
+function scaleProjection({
```
mbostock: Should this be called `exposeProjection` for symmetry?

src/scales.js
```diff
@@ -526,12 +527,28 @@ export function scale(options = {}) {
...
+  width = 640,
+  height,
+  margin = 0,
+  marginTop = margin,
+  marginRight = margin,
+  marginBottom = margin,
+  marginLeft = margin,
+  ...projection
+}) {
+  if (height === undefined) height = width * projectionAspectRatio(projection);
```
mbostock: This doesn’t look like what we do in `autoHeight`…

mbostock (commented):
As far as testing goes, I’m hoping that we can test this new implementation against the existing (more wasteful) one. This:

```js
Plot.plot({projection: {type: "mercator"}}).scale("projection")
```

Should be the same as this:

```js
Plot.scale({projection: {type: "mercator"}})
```

src/scales.d.ts
```diff
@@ -673,3 +674,4 @@ export interface Scale extends ScaleOptions {
...
+export function scale(options: {projection: ProjectionOptions & {width?: number; height?: number}}): Projection;
```
mbostock: This looks like it needs margin options, too. And probably aspectRatio? I guess those should be pulled out of PlotOptions and into something we can reuse here, maybe call it DimensionOptions?

src/scales.js
Fil: Yes. There is already an `exposeProjection `function but it could be renamed to `prepareProjection`.

src/scales.js
Fil: True. ~~I'll make a helper function to be used in both places.~~

Fil: This was addressed by reusing `createDimensions`.

src/dimensions.d.ts
```diff
@@ -1,3 +1,21 @@
+/** Options for specifying the dimensions of a plot or standalone projection. */
```
Fil, line 1: note that the terse definitions here are in fact only for `Plot.scale({projection})`, and get overridden by more detailed definitions in `Plot.plot()`, _i.e._ https://github.com/observablehq/plot/blob/1c6b239d1e42eb5c0d8c5b09f51b0c6ce51bea69/src/plot.d.ts#L11-L23

PR #2437, opened 2026-05-31
fix: preserve iterator data for mark facets

### LINKED ISSUE #2430: Invalid domain when facets are used with iterator
The following examples all produce the same (correct) output with a `y` domain of `["b"]`:
```js
Plot.dot(["ab"], {fy: "1"}).plot().scale("y").domain
```
```js
Plot.dot(new Set(["ab"]), {fy: "1"}).plot().scale("y").domain
```

However, when an iterator is passed in, no values are rendered and the `y` domain becomes `[undefined, undefined]`:
```js
Plot.dot(new Set(["ab"]).values(), {fy: "1"}).plot().scale("y").domain
```

(edit: simpler repro, but the original use case was `Map.values()`)

cat0825:
Fixes #2430.

This preserves iterator-backed mark data when mark-level facets are evaluated. The facet setup already materializes the iterator through `dataify`; without assigning that materialized array back to the mark, the mark later sees an exhausted iterator and derives an invalid y domain.

Added regression coverage for array, iterable, and iterator mark data with `fy`.

Verification:
- `pnpm run test:vitest run test/facet-test.js`
- `pnpm run test:tsc`
- `pnpm run test:lint`
- `pnpm run test:prettier`

### DIFF
--- a/src/plot.js
+++ b/src/plot.js
@@ -483,7 +483,9 @@ function maybeMarkFacet(mark, topFacetState, options) {
   // here with maybeTopFacet that we could reduce.
   const {fx, fy} = mark;
   if (fx != null || fy != null) {
-    const data = dataify(mark.data ?? fx ?? fy);
+    const value = mark.data ?? fx ?? fy;
+    const data = dataify(value);
+    if (mark.data === value && value?.[Symbol.iterator]?.() === value) mark.data = data;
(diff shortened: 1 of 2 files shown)


mbostock:
Thank you for the suggestion, but the data should be specified as an iterable so that it can be read repeatedly if necessary, and mutation of the mark should be avoided.

PR #2444, opened 2026-07-13
Allow tip format functions to return DOM nodes for rich text

### LINKED ISSUE #1767: feature request: to allow the title channel to contain rich text or markup

kirthi-b:
Closes the long-standing request for rich text and markup in tips (#1767).

This implements the "relatively easy" direction that mbostock outlined in the thread: a custom format function for the title channel, or any other channel, may now return a DOM node instead of a string. The returned node is appended directly to the tip rather than being wrapped in a text node, so you can style and hyperlink tip content, for example with an SVG tspan or anchor element built with Hypertext Literal:

```js
Plot.tip(data, {
  title: "name",
  format: {
    title: (name) => htl.svg`<tspan><tspan font-weight="bold">${name}</tspan> <a fill="blue" href="/details/${name}">details</a></tspan>`
  }
})
```

Safety: only a DOM node returned by a format function is treated as rich content. Plain string formats are unchanged and always rendered as text, so no HTML is ever parsed and this adds no injection surface. Because a returned node is appended as-is rather than measured, the lineWidth and textOverflow options do not apply to it.

Changes:
- src/marks/tip.js: append a returned node directly in renderLine, and route a title-channel node through renderLine; small isNode helper.
- src/marks/tip.d.ts: TipFormat may now return string | Node.
- docs/marks/tip.md: document the behavior, safety note, and an example.
- test/plots/tip-format.ts: snapshot tests for a title node and a name-value channel node.

I scoped this to the format-function approach only, and did not add the separate content channel or the href channel also floated in the thread, since those are larger API surfaces you may want to design yourselves. The VersionBadge version in the docs is a placeholder for you to set to whatever you plan to release.

Fixes #1767

### DIFF
--- a/docs/marks/tip.md
+++ b/docs/marks/tip.md
@@ -135,6 +135,19 @@ Plot.rectY(olympians, Plot.binX({y: "sum"}, {x: "weight", y: (d) => d.sex === "m
 
 The order and formatting of channels in the tip can be customized with the **format** option <VersionBadge version="0.6.11" pr="1823" />, which accepts a key-value object mapping channel names to formats. Each [format](../features/formats.md) can be a string (for number or time formats), a function that receives the value as input and returns a string, true to use the default format, and null or false to suppress. The order of channels in the tip follows their order in the format object followed by any additional channels. When using the **title** channel, the **format** option may be specified as a string or a function; the given format will then apply to the **title** channel. <VersionBadge version="0.6.15" pr="2074" />
 
+A format function may also return a DOM node instead of a string, such as an SVG tspan or anchor element (say generated with [Hypertext Literal](https://github.com/observablehq/htl)); the returned node is appended directly to the tip, allowing styled and hyperlinked rich text. <VersionBadge version="0.6.18" pr="1767" /> For example, to render the **title** channel in bold followed by a link:
+
+```js
+Plot.tip(data, {
+  title: "name",
+  format: {
+    title: (name) => htl.svg`<tspan><tspan font-weight="bold">${name}</tspan> <a fill="blue" href="/details/${name}">details</a></tspan>`
+  }
+})
+```
+
+Only DOM nodes returned by a format function are treated as rich content; plain strings are always rendered as text, so this does not introduce any HTML injection. Because a returned node is appended as-is rather than measured, the **lineWidth** and **textOverflow** options do not apply to it.
+
 A channel’s label can be specified alongside its value as a {value, label} object; if a channel label is not specified, the associated scale’s label is used, if any; if there is no associated scale, or if the scale has no label, the channel name is used instead.
 
 :::plot defer https://observablehq.com/@observablehq/plot-tip-format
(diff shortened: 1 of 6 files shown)


docs/marks/tip.md
```diff
@@ -135,6 +135,19 @@ Plot.rectY(olympians, Plot.binX({y: "sum"}, {x: "weight", y: (d) => d.sex === "m
 
 The order and formatting of channels in the tip can be customized with the **format** option <VersionBadge version="0.6.11" pr="1823" />, which accepts a key-value object mapping channel names to formats. Each [format](../features/formats.md) can be a string (for number or time formats), a function that receives the value as input and returns a string, true to use the default format, and null or false to suppress. The order of channels in the tip follows their order in the format object followed by any additional channels. When using the **title** channel, the **format** option may be specified as a string or a function; the given format will then apply to the **title** channel. <VersionBadge version="0.6.15" pr="2074" />
 
+A format function may also return a DOM node instead of a string, such as an SVG tspan or anchor element (say generated with [Hypertext Literal](https://github.com/observablehq/htl)); the returned node is appended directly to the tip, allowing styled and hyperlinked rich text. <VersionBadge version="0.6.18" pr="1767" /> For example, to render the **title** channel in bold followed by a link:
```
Fil:
```suggestion
A format function may also return a DOM node, such as an SVG tspan or anchor element (say generated with [Hypertext Literal](https://github.com/observablehq/htl)); the returned node is appended directly to the tip, allowing styled and hyperlinked rich text. <VersionBadge pr="2444" /> For example, to render the **title** channel in bold followed by a link:
```

docs/marks/tip.md
```diff
@@ -135,6 +135,19 @@ Plot.rectY(olympians, Plot.binX({y: "sum"}, {x: "weight", y: (d) => d.sex === "m
...
+```js
+Plot.tip(data, {
+  title: "name",
+  format: {
+    title: (name) => htl.svg`<tspan><tspan font-weight="bold">${name}</tspan> <a fill="blue" href="/details/${name}">details</a></tspan>`
+  }
+})
+```
+
+Only DOM nodes returned by a format function are treated as rich content; plain strings are always rendered as text, so this does not introduce any HTML injection. Because a returned node is appended as-is rather than measured, the **lineWidth** and **textOverflow** options do not apply to it.
```
Fil:
```suggestion
When returning nodes, the **lineWidth** and **textOverflow** options are ignored.
```

src/marks/tip.js
```diff
@@ -157,7 +157,11 @@ export class Tip extends Mark {
               this.setAttribute("stroke", "none");
               // iteratively render each channel value
               const lines = T[i];
-              if (typeof lines === "string") {
+              if (isNode(lines)) {
+                // A format function may return a DOM node (e.g., an SVG tspan
+                // or anchor) to display styled or hyperlinked rich text.
+                renderLine(that, {value: lines});
```
Fil:
```suggestion
              if (lines?.ownerDocument) {
                // A format function may return a DOM node (e.g., an SVG tspan
                // or anchor) to display styled or hyperlinked rich text.
                renderLine(that, {value: lines});
```

Precedent for this simpler test:
https://github.com/observablehq/plot/blob/356f579b1d947ee05a914420eddff0f29cee300a/src/plot.js#L364

Also I would like to avoid testing twice if this is a node.

Fil (changes requested):
Thank you for the PR. I'm sorry I haven't found enough time to fully review, but this looks promising. I noticed that the channel markup case is missing a space between the field name+colon and the contents (as you can see in tipFormatChannelMarkup.svg).

PR #2448, opened 2026-07-31
remove additional function execution

douglaskariuki022:
one of the functions on isTimeInterval function execution is not required on the module


douglaskariuki002:
see #2448

PR #2457, opened 2026-08-17
Rounded tooltips

pdebruic:
For a tip mark set a `radius` and this will make a rounded tooltip. 

Changed the tip paths to draw themselves in the clockwise direction. If a radius is set then add quadratic bezier curves in the appropriate corners based on the `frameAnchor`. 

I did not make curves from the flat edge to the "point".  Just the outer corners. 

image

Add tests


pdebruic:
Also added a `fontColor` field so the text is not always `currentColor`

